PR #282, opened 2015-12-08
Retrieve session data from session cookie

rschamp:
Assumes the session cookie is stored as JSON, which may or may not have been compressed via zlib (indicated by a leading `.`), which is then base64-encoded, and made URL-safe by replacing all `+` and `/` characters with `-` and `_` respectively.  The cookie may or may not be signed as well. The `Jar.unsign` method should work with URL-unsafe, uncompressed, unsigned values.

I had a lot of trouble using `browserify-zlib`, it never seemed to like the format I gave it. When I tried `pako` it just worked the first time, so I liked that feature.

This requires an update to scratchr2 to send the session cookie down in this format as non-HTTP-only. It also requires that we put what www expects into the session.

Halfway through doing this I had the thought that someone must have made a module for using Django session cookies in JS, but I couldn't find anything. It might be nice to split that out into its own package and use it that way — sessions aren't the only signed cookies you can use with Django, so it might be useful for others.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "lodash.range": "3.0.1",
     "minilog": "2.0.8",
     "node-sass": "3.3.3",
+    "pako": "0.2.8",
     "po2icu": "git://github.com/LLK/po2icu.git#develop",
     "react": "0.14.0",
     "react-addons-test-utils": "0.14.0",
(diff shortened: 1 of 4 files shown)


rschamp:
I have to close this because it's coming from my fork of the private repo. Reopening against my new public fork.

PR #344, opened 2016-01-29
Moving Microworlds pages to scratch-www & adding arts

morantsur:
Moved https://scratch.mit.edu/pathways/hiphop and fashion, and created a new microworld for arts. 
The final design is not done yet, but most of the code is already in place.


mewtaylor:
Re-opening from a forked branch

Jonathan50:
Whats microworlds?

PR #353, opened 2016-02-04
Moving Microworlds pages to scratch-www & adding arts

morantsur:
Moved https://scratch.mit.edu/pathways/hiphop and fashion, and created a new microworld for arts. 
The final design is not done yet, but most of the code is already in place.

Moved from #344


src/components/tipsslider/tipsslider.json
```diff
@@ -0,0 +1,50 @@
+[
```
mewtaylor: It might be helpful to have the json in the format that microworlds use as a sample – i.e.  carousel items chunked into sets.

src/components/tipsslider/tipsslider.jsx
```diff
@@ -0,0 +1,64 @@
...
+var React = require('react');
+var Slider = require('react-slick');
+
+var Thumbnail = require('../thumbnail/thumbnail.jsx');
+
+require('slick-carousel/slick/slick.scss');
+require('slick-carousel/slick/slick-theme.scss');
+require('./tipsslider.scss');
+
+var TipsSlider = React.createClass({
```
mewtaylor: small nitpick – it could be good to call it something like `NestedCarousel`, or in some way that describes a bit how it differs from the other Carousel component.

src/components/tipsslider/tipsslider.jsx
```diff
@@ -0,0 +1,64 @@
...
+            var items = this.props.items[i].tips;
+            var thumbnails = [];
+            for (var j=0; j < items.length; j++) {
+                    var item = items[j];
+                    thumbnails.push(<Thumbnail key={item.id}
+                               title={item.title}
+                               src={item.thumbnailUrl} />)
+            }
+            stages.push(
+            <div className="testing" key={"stage_" + i}>
```
mewtaylor: we can probably change this class name :)

src/components/tipsslider/tipsslider.scss
```diff
@@ -0,0 +1,75 @@
+@import "../../colors";
+
+.tipsslider {
+    $icon-size: 40px;
+    $button-offset: $icon-size + 5px;
+    $box-content-offset: 20px;
+
+    //padding: 12px $button-offset;
```
mewtaylor: can probably remove this

src/views/microworld_art/microworld_art.jsx
```diff
@@ -0,0 +1,4 @@
+var render = require('../../lib/render.jsx');
+var Microworld = require('../../components/microworld/microworld.jsx');
+
+render(<Microworld microworldData={require("./microworld_art.json")} />, document.getElementById('view'));
```
mewtaylor: I think the common convention for react is to require ahead of time to get the object, and then load it into the render method later on. @rschamp is that correct as far as you know?

src/components/microworld/microworld.jsx
```diff
@@ -0,0 +1,251 @@
...
+        var videoOpenArr = this.state.videoOpen;
+        videoOpenArr[key] = false;
+        this.setState({videoOpen: videoOpenArr});
+    },
+    getInitialState: function () {
+        return {
+            videoOpen: {},
+            featuredGlobal: {},
+            featuredLocal: {},
+            microworld_data: this.props.microworldData
```
mewtaylor: I think this could be moved to props instead of state if it's not going to be changed within the component itself

src/components/microworld/microworld.jsx
```diff
@@ -0,0 +1,251 @@
...
+                </div>
+            )
+        }
+    },
+    render: function () {
+        var classes = classNames(
+            'top-banner'
+        );
+
+        var microworldData = this.state.microworld_data;
```
mewtaylor: You could probably use `this.props.microworld_data.<...>` down below, which means you could get rid of this assigment

src/components/microworld/microworld.scss
```diff
@@ -0,0 +1,389 @@
...
+    .pathways .editor-and-tips {
+      padding: 15px 15px 0px 15px;
+      border-radius: 15px 15px 0 0;
+      box-shadow: 0px 0px 8px #cccccc inset, 0 2px 3px rgba(34, 25, 25, 0.3);
+      border-top: 1px solid #E0E0E0;
+      border-right: 1px solid #E0E0E0;
+      border-left: 1px solid #E0E0E0;
+      background-color: white;
+      margin-bottom: 50px;
+      /* offset-x | offset-y | blur-radius | color */
```
mewtaylor: could probably be deleted

mewtaylor:
Just a few comments, but this is awesome – I love how easy it makes it to prototype new pathways for people who are new to the code base! Besides the comments above, it might help to add some test cases to the PR description if there are any specific user behaviors that we want to check work correctly (i.e. videos showing/hiding, carousel item navigation, etc.)

Also, it looks like some tests are failing too, so it might make sense to run `make test` locally to see what might be failing.

thisandagain:
This is really wonderful @morantsur ! Excited to be able to prototype new Microworlds so easily.

/cc @ntlrsk @kaschm

morantsur:
Fixed all the comments, and removed some unneeded vars, ids and such.
I also changed the thumbnail component, to allow thumbnails without a link (used with the tips section).

I looked the the scss file for the tipsslider/nestedcarousel, and realized it was just a copy of the carousel scss, so I removed all the duplicate code - let me know if my solution makes sense. 

Where should I add the test cases? 

Thanks!

thisandagain:
@mewtaylor Could you follow-up on this with @morantsur ?

mewtaylor:
@thisandagain yes apologies for the delay!

mewtaylor:
@morantsur as discussed in person – if we can get the `nestedcarousel` to work with flex so that it's justified to the center, that would be great. Otherwise, looks great!

morantsur:
@mewtaylor 
I fixed the nested carousel issue, by just changing a few param and css of the slider, merged the code and added two changes I would like to get your OK for:
* Expanded that no-link option to include also the image itself (as it was an empty link before)
* I removed the routes to the microworlds, so for now they are detached. Once we decide it's good enough we can add them back. 

Thanks!

mewtaylor:
This looks awesome! :shipit:

rschamp:
@mewtaylor Talked with @thisandagain who thinks we should merge, but not set up the Fastly rules for this.  That will make it so this won't get stale, but also won't be released yet.

PR #355, opened 2016-02-05
Fix gh-354: 'Supported Organizations' in credits overflows

### LINKED ISSUE #354: 'Supporting Originizations' in credits overflows
This seems to be fixed by adding the CSS rule `word-wrap: break-word`.

Screenshot:
picture of bug

I'm on Debian GNU/Linux 7.0 "Jessie" with Iceweasel (rebranded Firefox) 38.6.0.
I don't get this issue with Safari 9.0.1 or Google Chrome on OS X.

TheGrits:
Should fix #354.
Can someone test this on Firefox? (I only have Chrome)

### DIFF
--- a/src/views/credits/credits.scss
+++ b/src/views/credits/credits.scss
@@ -3,10 +3,9 @@
 #view {
     p {
         line-height: 1.5rem;
-    }
-
-    a {
-       word-wrap: break-word; /* Overrides: https://github.com/LLK/scratch-www/blob/develop/src/main.scss#L43-L47 */
+        a {
+            word-wrap: break-word; /* Overrides: https://github.com/LLK/scratch-www/blob/develop/src/main.scss#L43-L47 */
+        }
     }
     
     ul {
--- a/src/views/credits/credits.scss
+++ b/src/views/credits/credits.scss
@@ -3,6 +3,7 @@
 #view {
     p {
         line-height: 1.5rem;
+        word-wrap: break-word;
     }
 
     ul {
--- a/src/views/credits/credits.scss
+++ b/src/views/credits/credits.scss
@@ -3,9 +3,12 @@
 #view {
     p {
         line-height: 1.5rem;
-        word-wrap: break-word;
     }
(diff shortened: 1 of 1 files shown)


mewtaylor:
@TheGrits this is actually related to a setting we currently have in `main.scss` for hrefs in text: https://github.com/LLK/scratch-www/blob/develop/src/main.scss#L43-L47. Could you make this override that specific setting for hrefs rather than having it be global for all paragraph text, and add a comment referencing what it's overriding?

Jonathan50:
Oh I see it's anchors only that overflow like that

TheGrits:
@mewtaylor Sure, sounds good. I'll try to push a commit to my patch sometime today.

mewtaylor:
@TheGrits thanks! @Jonathan50 – the idea behind nowrap on anchors in paragraph blocks is to ensure that anchors around multi-word text wrap together, rather than extending across multiple lines of text (i.e. on a screen in which one of the words in the anchor moves to the next line but the first word doesn't).

TheGrits:
@mewtaylor I have updated the style sheet, it should work now.

mewtaylor:
thanks @TheGrits ! I think it would be best if we could put the anchor formatting inside the paragraph formatting rather than next to it, like:
```SCSS
p {
    line-height: 1.5rem;
    a {
        word-wrap: break-word; /* Overrides: https://github.com/LLK/scratch-www/blob/develop/src/main.scss#L43-L47 */
    }
}
```
For uniformity, and to ensure the anchor formatting isn't applied to any elements that aren't text, like an img tag with a link, for instance. Could you make that change?

TheGrits:
@mewtaylor Sounds good, I'll make the change now.

TheGrits:
@mewtaylor Done!

mewtaylor:
@TheGrits awesome! Also... First PR alert!
confetti-horn-emoji

TheGrits:
:grinning:

Issue #357, opened 2016-02-06
Splash - Initial value of project count is arbitrary


TheGrits:
https://github.com/LLK/scratch-www/issues/161#issuecomment-171460369

The least that could be done is update the number, that should be an easy edit.
(The current initial value is 10569070 to be exact)

mewtaylor:
I think this might be covered under the scope of #161 – @thisandagain does covering this in the #161 enhancement make sense to you too?

TheGrits:
Hmm, I guess that does make sense.

thisandagain:
@mewtaylor Actually I kind of like splitting that meta-issue up. Let's keep this!

mewtaylor:
:+1: SGTM thanks @thisandagain @TheGrits !

PR #359, opened 2016-02-08
Fix gh-357: Splash - Initial value of project count is arbitrary

TheGrits:
Updated the number with a more recent number, I wasn't sure if using a string like "Loading..." would damage anything.

Temporary fix for #357

Also added a comment about the project count, but that doesn't change much. :stuck_out_tongue_winking_eye:

### DIFF
--- a/src/views/splash/splash.jsx
+++ b/src/views/splash/splash.jsx
@@ -27,7 +27,7 @@ var Splash = injectIntl(React.createClass({
     ],
     getInitialState: function () {
         return {
-            projectCount: 10569070,
+            projectCount: 12893209, // gets the shared project count
             activity: [], // recent social actions taken by users someone is following
             news: [], // gets news posts from the scratch Tumblr
             featuredCustom: {}, // custom homepage rows, such as "Projects by Scratchers I'm Following"
--- a/src/views/splash/splash.jsx
+++ b/src/views/splash/splash.jsx
@@ -27,7 +27,7 @@ var Splash = injectIntl(React.createClass({
     ],
     getInitialState: function () {
         return {
-            projectCount: 12893209, // gets the shared project count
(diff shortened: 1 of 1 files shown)


TheGrits:
If wanted, I could easily change the value to a string. Please let me know what you think.

mewtaylor:
@TheGrits thanks for this, and sorry for the delay – I personally would totally be interested to see if you had any ideas for the "blank" state. Do you have any interest in doing another PR that does something more along those lines?

Another idea that comes to mind, on top of the one you mentioned, would be displaying "millions of" while the string is being loaded, or "many".

TheGrits:
@mewtaylor My comment on Scratch was about my other PR, but this needs to be discussed too I guess. :stuck_out_tongue: 

I like the idea of using "millions of", something like "over 13 million" would work as well.

mewtaylor:
@TheGrits ah gotcha – I'll take a look at that one now too :). And yeah, I think that if we're going to fix #357 it might make sense to do something that's more "blank" state while we're at it, if you wanna change this PR to do something like that. 

@carljbowman @kaschm would be interested to hear what you think about the default state (i.e. before we retrieve the actual number of projects from the api). Ideas include:
* Display "**over 13 million** projects shared..."
* Display "**millions of** projects shared..."
* Display "Loading..." instead of the whole string

TheGrits:
@mewtaylor Ok, I changed it to "over 13 million". The only reason the checks are failing on the last PR update is because of the problem resolved by #361

mewtaylor:
@TheGrits awesome thanks! I'm going to wait to hear back from the others for confirmation that the text works before reviewing to make sure that's the text we want :).

mewtaylor:
@TheGrits looks good after the review of the text. Thanks!

Issue #300, opened 2016-01-01
[Homepage] Horizontal Scroll Bar Shows on Front Page


2jour:
OS: mac osx 10.10.5
Browser: chrome 46
Server: 2.2.4 production

![homepage](https://cloud.githubusercontent.com/assets/6530036/12071742/cf2d005a-b089-11e5-8d6e-c79efd92780b.jpg)

PR #360, opened 2016-02-08
Fix gh-300: Horizontal Scroll Bar Shows on Front Page

technoboy10:
Fixes issue #300.

To fix the horizontal scrollbar showing on the homepage (and other pages), I've set the horizontal padding on the view element to 0 in main.scss.

**Test Cases**
To test the fix, I checked that when scrolling vertically and attempting to scroll horizontally, no horizontal scrollbar appeared on the front page and credits page. My tests succeeded on Chrome 50, Firefox 44, and Safari 9.0.3 on OS X 10.11.3. 

~~**Possible problems**
I applied the fix to the main.scss file, which might cause problems if a horizontal scrollbar is desired on other pages that inherit from that file. I can't think of any page that would need a horizontal scrollbar, though.~~

### DIFF
--- a/src/main.scss
+++ b/src/main.scss
@@ -10,7 +10,6 @@ body {
     padding: 0;
     color: $type-gray;
     font-family: "Helvetica Neue", "Helvetica", Arial, sans-serif;
-    overflow-x: hidden;
 }
 
 /* Typography */
@@ -81,7 +80,7 @@ a:hover {
     /* NOTE: Margin should match height in navigation.scss */
     margin-top: 50px;
     background-color: $background-color;
-    padding: 20px 1px;
+    padding: 20px 0px;
     min-width: 100%;
     min-height: 768px;
 }
--- a/src/main.scss
+++ b/src/main.scss
@@ -80,7 +80,7 @@ a:hover {
     /* NOTE: Margin should match height in navigation.scss */
     margin-top: 50px;
     background-color: $background-color;
-    padding: 20px 0px;
+    padding: 20px 0;
     min-width: 100%;
     min-height: 768px;
 }
--- a/src/main.scss
+++ b/src/main.scss
@@ -10,6 +10,7 @@ body {
     padding: 0;
     color: $type-gray;
     font-family: "Helvetica Neue", "Helvetica", Arial, sans-serif;
+    overflow-x: hidden;
 }
 
 /* Typography */


technoboy10:
@mewtaylor

mewtaylor:
Hey @technoboy10  – I was just talking with @carljbowman a bit about this one. We were noticing that, while this PR would certainly fix the bug currently, it might have an adverse effect on ongoing efforts for responsive design – we probably want horizontal scrolling on mobile to be an option as we continue to get that view to work correctly. We have also set up our grid in such a way that was intended to prevent this from happening on desktop screens, so it'd be really helpful to figure out exactly where in our formatting that has broken down before applying a global standard like this.

Do you have some time to do a bit of investigation into that? One thing @carljbowman mentioned is that we might not be properly accounting for border sizing around elements within the grid system. He got that insight after seeing that elements like the Body, Nav, and Footer are just pixels larger than the window.

carljbowman:
@mewtaylor - Thanks for responding! This all looks great. Upon further review I am actually seeing issues with at `#view` level. Here are some contextualizing screens: 
MacPro - Chrome - 1280px

## Nav - Reporting 1280px 
screen shot 2016-02-09 at 9 39 07 am

## View - Reporting 1282px 
screen shot 2016-02-09 at 9 38 50 am

## Footer - Reporting 1280px 
screen shot 2016-02-09 at 9 39 20 am

Also interesting looking to the right of the Footer we have 2px of dark blue showing through from the `#view` element. 
screen shot 2016-02-09 at 9 40 38 am

technoboy10:
@mewtaylor Sure thing - I'll take a closer look. @carljbowman Thanks - I think I might know what's causing that.

technoboy10:
Looks like the min-width property on #view isn't checking for padding - @carljbowman is there a specific reason why there's a padding-right value of 1px on #view?

carljbowman:
@mewtaylor - Do you know of any reason that we could have 1px of padding left and right on view?

@technoboy10 - Good catch. That is the exact reason we are seeing an additional 2px.

technoboy10:
My test cases described in the initial PR also pass for the new fix.

mewtaylor:
Confirmed locally on Chrome and FF (OS X 10.11) that the horizontal bar does not appear, while it still does in both on production. Looks great thanks @technoboy10 !

Issue #358, opened 2016-02-08
Readme - Fix travis badge


thisandagain:
Our travis-ci badge is still pointing at our private repo from before `scratch-www` was made public.

/cc @mewtaylor

thisandagain:
[![Build Status](https://travis-ci.org/LLK/scratch-www.svg?branch=develop)](https://travis-ci.org/LLK/scratch-www)

@mewtaylor We should double-check that deployment is still working as well.

PR #361, opened 2016-02-12
Fix travis build fail

TheGrits:
Travis was complaining that link 188 had over 120 characters, this should fix it.
Here's where I saw it: https://github.com/LLK/scratch-www/issues/358#issuecomment-181396113


mewtaylor:
@TheGrits I'm not getting a lint failure on this line locally when I test the `develop` branch – is that the branch you were seeing it on?

Also in the future I think it'd be easier to open an issue before submitting a PR, just to make sure it can be reproduced first.

TheGrits:
@mewtaylor I saw it here: https://github.com/LLK/scratch-www/issues/358#issuecomment-181396113

mewtaylor:
Awesome thanks @TheGrits can repro now – would you mind updating the PR description to include the link to the comment?

TheGrits:
@mewtaylor Done.

mewtaylor:
@TheGrits looks great, just confirmed the fix locally. Thanks!

Issue #356, opened 2016-02-06
[Homepage] commas not showing in # of projects in Safari


ntlrsk:
Home page in Safari doesn't show commas in number of projects (when logged out).
Reproduced on two Mac laptop computers  - in Safari only (Mac OS X 10.10.5). 

![image](https://cloud.githubusercontent.com/assets/4721032/12868652/b7814392-ccd8-11e5-8ae2-4370f34f3cf0.png)

TheGrits:
Not happening to me. ChromeOS 7647.73.0, Chrome 48.0.2564.92, Flash 20.0 (release 0)

technoboy10:
Can also reproduce in Safari 9.0.3 on OS X 10.11.3.

Jonathan50:
Weird.
> Double-clicking the number on Safari brings up a FaceTime dialog - perhaps Safari is treating the string as a phone number?

Likely it.

thisandagain:
@technoboy10 I think that's exactly right. Good catch.

mewtaylor:
Turns out this is an issue with `toLocaleString`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString. It doesn't appear to have support in Safari.

This should be solved by moving to using react-intl for number formatting.

PR #367, opened 2016-02-26
Fix gh-366: Homepage - Failed PropType

### LINKED ISSUE #366: On homepage: "Failed propType: Invalid prop `projectCount` of type `string` supplied to `Intro`, expected `number`."
I get this when running locally.

TheGrits:
This should fix #366 
@mewtaylor @rschamp

### DIFF
--- a/src/components/intro/intro.jsx
+++ b/src/components/intro/intro.jsx
@@ -10,6 +10,9 @@ Modal.setAppElement(document.getElementById('view'));
 
 var Intro = React.createClass({
     type: 'Intro',
+    propTypes: {
+        projectCount: React.PropTypes.number
+    },
     getDefaultProps: function () {
         return {
             projectCount: 10569070,
--- a/src/components/intro/intro.jsx
+++ b/src/components/intro/intro.jsx
@@ -10,9 +10,6 @@ Modal.setAppElement(document.getElementById('view'));
 
 var Intro = React.createClass({
     type: 'Intro',
-    propTypes: {
-        projectCount: React.PropTypes.string
-    },
     getDefaultProps: function () {
         return {
             projectCount: 10569070,
--- a/src/components/intro/intro.jsx
+++ b/src/components/intro/intro.jsx
@@ -11,7 +11,7 @@ Modal.setAppElement(document.getElementById('view'));
 var Intro = React.createClass({
     type: 'Intro',
     propTypes: {
-        projectCount: React.PropTypes.number
+        projectCount: React.PropTypes.string
     },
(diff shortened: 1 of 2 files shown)


mewtaylor:
@TheGrits I'm getting an issue in the other direction with that change (when the number is retrieved it throws an error that it's not a string). I think the solution here might be to remove the `propTypes` validator for this value altogether

TheGrits:
@mewtaylor 
Makes sense, I have made the change. 
I don't seem to be getting any more errors regarding the change, so we should be good to go.

mewtaylor:
sweet! I'll take a look this afternoon!

rschamp:
I think needing to remove the validation hints at an issue with the original solution. We expect the count to be a number, since we call `toLocaleString` on the value — it doesn't work on strings.  Could we change it to instead display the "over X" text only if the value is unchanged from its default? Then we'll fix this bug as well as the localization issue at once.

The text that contains the count is also not localized, but that is a different issue.

mewtaylor:
@rschamp that sounds great to me! +1 on localization too, but I can open up a separate issue for that like you mentioned – there are few places, upon inspection, where we have default text that's not localized.

TheGrits:
@mewtaylor So what do you need me to do? I'm a little confused.

mewtaylor:
@rschamp correct me if I'm misunderstanding but:

@TheGrits I think the idea would be to leave the default as a number in `splash.jsx`, and rather than remove the `propType` check in `intro.jsx`, instead add a conditional check that would see if the value for `this.props.projectCount` is equal to `this.getDefaultProps().projectCount` in the `render()` method of `intro.jsx`. If they are equal (so we're using the default value), then we'd render `'over 13 million'` rather than `this.props.projectCount.toLocaleString()`. This could be done by assigning a variable in `render()` to `'over 13 million'` or `<number>.toLocaleString()` ahead of the `return` statement. Does that help?

rschamp:
@mewtaylor Pretty much what I was thinking. I think we might want to keep the two strings "A creative learning community with `X` projects shared" and "A creative learning community with over 13 million projects shared" separate for localization purposes, rather than just having `X` be a number or `over 13 million`.  The grammar isn't necessarily going to be the same for those two statements.

TheGrits:
@mewtaylor Makes a lot more sense, thanks. :+1: 
@rschamp Sounds good.

mewtaylor:
@TheGrits @rschamp #356 was fixed in a different way by https://github.com/LLK/scratch-www/pull/390, so I'm gonna close this.

PR #369, opened 2016-03-02
Migrate - Jobs Page

thisandagain:
- Migrates jobs view to `scratch-www` (@hlcolani)
- Updates listings to match https://www.media.mit.edu/about/opportunities
- Normalize styles

Resolves GH-306


mewtaylor:
:+1: looks great! Just one question – is it safe to assume we do not want/need the social media links that we have on the existing version of the page? (see below)
screen shot 2016-03-03 at 1 30 10 pm

thisandagain:
Yup! I dropped them purposefully. Thanks!

Issue #351, opened 2016-02-02
Upgrade to Node 4.2.X LTS


thisandagain:
We should move from Node 0.12 to the 4.2 LTS release now that it is stable.

- [x] Update `dependencies`
- [x] Update `devDependencies`
- [x] Test Node 4.2.X LTS locally
- [x] Add `4.2` to travis.yml
- [x] Land in staging for testing
- [ ] Update continuous deployment target
- [ ] Retire old environment

liam4:
Goodness me yes. Node 0.12 is terrible. :)

thisandagain:
> Node 0.12 is terrible. :)
Can you explain what you mean by "terrible"? Please try to be constructive with your comments.

liam4:
It's support for ES6 is much less complete than Node 4 and 5. NPM generally is more user friendly.

thisandagain:
"Less complete ES6 support" != "terrible". I'd appreciate if we could try to be constructive and explicit when working through technical issues.

liam4:
Okay, I'll try to be more constructive.

Jonathan50:
```
$ node --version
v5.4.1
$ nodejs --version
v0.10.29
$ 
```
this is confusing

mewtaylor:
@Jonathan50 I think it's safe to follow the `node` version (see the versions in the [changelog](https://github.com/nodejs/node/blob/v5.5.0/CHANGELOG.md#2016-01-12-version-541-stable-thealphanerd), which match up with it). If you have any additional questions/issues with downloading and installing node, it might also help to checkout their [repo](https://github.com/nodejs/node) – it's super well documented and might be helpful!

thisandagain:
`https://staging.scratch.mit.edu` is now running Node.js 4.3.2 LTS.

liam4:
:tada:

PR #370, opened 2016-03-02
Prepare for migration to Node 4.2 LTS

thisandagain:
- Updates `dependencies`
- Updates and removes redundant and otherwise unused modules from `devDependencies`. I attempted to keep this fairly minimal as to mitigate risk to our build process.
- Updates `travis.yml` to test against Node 4.2.X
- Add explicit support for `C++11` to resolve build issues in Travis CI `Precise` images

Resolves GH-351


package.json
```diff
@@ -52,10 +52,7 @@
     "minilog": "2.0.8",
     "node-sass": "3.3.3",
     "po2icu": "git://github.com/LLK/po2icu.git#develop",
-    "react": "0.14.0",
-    "react-addons-test-utils": "0.14.0",
-    "react-dom": "0.14.0",
-    "react-intl": "2.0.0-beta-1",
```
rschamp: I assume you removed these because we include them statically? Won't tests require them to run?

thisandagain: Correct. These aren't used in our current code as we include them in `./static/js/lib`.

thisandagain:
Failing due to native dependency of `iconv` within `po2icu`:
`scratch-www/node_modules/po2icu/node_modules/i18next-conv/node_modules/node-gettext/node_modules/iconv/build`

thisandagain:
Interestingly, I'm having trouble forcing Mac OS X to install `iconv` when running `npm install`. The `encoding` module *should* be using the non-native `iconv-lite` module by default.

thisandagain:
The fact that `iconv` is being installed is one problem. The default travis compiler not supporting C++11 is another. Addressing this issue in 834327a. 

Reference:
https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Node.js-v4-(or-io.js-v3)-compiler-requirements

PR #373, opened 2016-03-03
Fix gh-356: Homepage - Facetime Phone Numbers

### LINKED ISSUE #356: [Homepage] commas not showing in # of projects in Safari

TheGrits:
Should fix #356, but I only have Chrome. Can someone test it on Safari for me? Thanks.
(In case you were wondering, I did test on Chrome anyway.)
@mewtaylor @thisandagain

### DIFF
--- a/server/template.html
+++ b/server/template.html
@@ -11,6 +11,9 @@
 
         <title>Scratch - {{title}}</title>
 
+        <!-- Prevent Safari from making phone numbers -->
+        <meta name="format-detection" content="telephone=no">
+        
         <!-- Search & Open Graph-->
         <meta name="description" content="{{description}}" />
         <meta name="google-site-verification" content="m_3TAXDreGTFyoYnEmU9mcKB4Xtw5mw6yRkuJtXRKxM" />
--- a/src/components/intro/intro.scss
+++ b/src/components/intro/intro.scss
@@ -153,10 +153,12 @@
     }
 
     .project-count {
-        $project-count-color: hsl(318, 50%, 52%); 
+        text-decoration: none; /* should stop Safari from making it into a phone number */
+        $project-count-color: hsl(318, 50%, 52%);
         color: $project-count-color;
         font-size: 18px;
         font-weight: 700;
+        pointer-events: none; /* disables clicking */
     }
 
     .links {


mewtaylor:
OS X 10.10, Safari 9.0.3 – I'm still seeing spaces instead of commas. I wonder if there's a safari-specific meta tag of some kind we might need to look into here?

TheGrits:
@mewtaylor Yep, there is a meta tag we could use. [Documentation Here](https://developer.apple.com/library/safari/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html#//apple_ref/doc/uid/TP40008193-SW5)

TheGrits:
@mewtaylor Okay, I've added the meta tag I mentioned and tested locally on Chrome. Would you be willing to test it again? (Note: The meta tag, because I put it in the template.html file, should prevent Safari from making phone numbers not only on the project count, but also elsewhere on the site)

mewtaylor:
@TheGrits so odd – that seemed like it should do the trick based on the research you did, but oddly enough Safari was still doing the comma removal for me:
> screen shot 2016-03-07 at 8 55 03 am

Have you come across anything else that might help in the Safari docs?

TheGrits:
@mewtaylor That's the only meta tag I can find for the Apple documentation. The first version of the PR was the only other semi-clean solution, which leaves just one more solution that I found. 

Inserting a `<span></span>` in the middle of the number makes Safari not recognize it as a number that could be a phone number. It's real ugly though, I'm not sure that we would want to do this.

TheGrits:
@mewtaylor Ok, done some further research, and apparently what's done so far has solved half of the problem. Here's what's going on: The stuff that I have coded so far should prevent **mobile** Safari from making the number into a phone number. There is a different solution for desktop Safari that I'm going to look into now.

TheGrits:
@mewtaylor So the CSS that I added was supposed to fix the phone number part, which it should. I'm not sure how to prevent it from removing the commas though. (Besides the `<span>` solution)

mewtaylor:
@TheGrits thanks for doing that research! I wasn't able to find anything on the comma removal either. 

I think that, from here, maybe it makes the most sense to close this PR, since we seem to not be able to address the comma issue easily. The `meta` tag is probably useful for mobile on safari, but it might be best to open a separate PR with just that commit in it, since it's a separate issue. If you don't mind doing that, and linking to the ios docs you posted as part of that PR description, that would be awesome.

TheGrits:
@mewtaylor Sounds good.

PR #377, opened 2016-03-08
Fix gh-356: Facetime Phone Numbers

TheGrits:
Fixes mobile side of #356.

[Documentation of meta tag here](https://developer.apple.com/library/safari/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html#//apple_ref/doc/uid/TP40008193-SW5)

As per https://github.com/LLK/scratch-www/pull/373#issuecomment-193934786
@mewtaylor 
Tested locally on Chrome.

### DIFF
--- a/server/template.html
+++ b/server/template.html
@@ -10,7 +10,10 @@
         <meta name="viewport" content="width=942, initial-scale=1">
 
         <title>Scratch - {{title}}</title>
-
+        
+        <!-- Prevent mobile Safari from making phone numbers -->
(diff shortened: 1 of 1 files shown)


TheGrits:

mewtaylor:
@TheGrits looks great! Not sure exactly when we'll add this in yet – might wait on it until we do more mobile design updates, or for this next release.

TheGrits:
@mewtaylor Awesome! Ok, that's fine!

PR #379, opened 2016-03-10
[Master] – Wedo landing page

mewtaylor:
Updates applied based on #378. This also includes the initial pass on Global Typographic Standards.
- [x] Add URLs to all links on the page (still being generated)
- [x] Ater URLs have been added, update translation JSON to include messages that include links


mewtaylor:
@carljbowman links added, did some basic linter cleanup. I also moved the instructions section to the center on desktop – let me know if that's not the way it should be though!

Issue #372, opened 2016-03-03
Migrate - Cards


thisandagain:
https://scratch.mit.edu/info/cards/

Related #362

thisandagain:
@mewtaylor Could you take this and #362 ?

liam4:
On the topic of cards, why are images being used for "view card" and "get all the cards"?

mewtaylor:
@liam4 are you referring to the pdf icon to the left of the link?

liam4:
No, I was talking about the text..

Oh, haha, oops! Sorry, Firefox is weird in making the text act like an image.

Curious now though why the font "scratch-icons" renders as Times (I think)? That doesn't really fit with the style of the rest of the site..

mewtaylor:
@thisandagain @rschamp here's a proposal for how we might handle localizing links to static assets, at least as a short term solution while we work out which localization tool we use long term:
* use the `formattedMessage` system build into react-intl to handle the actual string to be placed in the anchor
* create a separate file in the views dir called `l10n-static.json` that contains the urls to be localized
* add a global config json file that contains, for each language, the localized versions of static urls for each language that has them. This would require being updated whenever we get new localized pdfs/assets, which is not ideal, but I think is fine for the short term since we don't get new pdfs/assets that often.
* use the `build-locales` script (during the `make translations` step) to add localized urls to the `{{view}}.intl.js` bundle for a given view. This will iterate over `l10n-static.json` files, and check in the global config file to see if a locale has the localized version of a file. If it doesn't, it'll default to english as it does currently

Your thoughts would be greatly appreciated!

rschamp:
That sounds like a good strategy to me.

mewtaylor:
@carljbowman I made some tiny tweaks to the design of the page – thoughts?
![cards](https://cloud.githubusercontent.com/assets/1815772/13791490/c73f3fe2-eac3-11e5-88c0-4def0b2338e1.png)

carljbowman:
@mewtaylor - This looks great! I do have two tiny details. 

1. If possible we should use the typography from our global type sheet: `_typorgraphy.scss`. That intro section gets a little lost with all the photos. There is a class paragraph style called `.intro` for these types of mastheads. 

[...]

Issue #362, opened 2016-02-12
update Scratch Cards page to include new cards


ntlrsk:
It would be great when possible to update the Scratch Cards page to include the thematic card sets: Dance, Name, and Hide-and-Seek (with others to come). 

http://scratch.mit.edu/dance/cards
http://scratch.mit.edu/name/cards
http://scratch.mit.edu/hide/cards

mewtaylor:
@ntlrsk do you have the images that would appear for these (i.e. the thumbnail for the cards page)? I do not have them anywhere yet. Just wanted to know if you do before I attempt to do any cropping or anything.

thisandagain:
@mewtaylor can this be closed?

mewtaylor:
Yes – sorry thought this was closed by the PR. Thanks!

PR #384, opened 2016-03-16
Implement GH-372: Add `/info/cards/` route

mewtaylor:
Also implements a method for handling translation of static files as described in https://github.com/LLK/scratch-www/issues/372#issuecomment-193350144. This also adds links as requested in #362, however we need to add thumbnails to our cdn in scratchr2 as well for that to work properly.


src/views/cards/cards.scss
```diff
@@ -0,0 +1,39 @@
+@import "../../colors";
+@import "../../typography";
+
+.intro {
```
rschamp: I think we already have an element with this class name — could you wrap all of the styles for the component in a class so there can't be any interaction? I know practically there won't be, but just for consistency.  E.g., add a `cards` class to the `div.inner`, and then wrap all these styles in `.cards`?

rschamp:
Looks good to me other than one nitpick.

src/views/cards/cards.scss
mewtaylor: :+1:

PR #386, opened 2016-03-18
Use Redux for session handling

rschamp:
This does a few things:

* Simplifies our `render` method by moving the `Navigation` and `Footer` components into a new `Page` component, used by each view.
* Removes the `Session` mixin and init step, handling it with redux instead.  The redux state tree just contains one object, `session`, which contains what used to be in `window._session`. The `refreshSession` global method is now a redux action by the same name.

There are a lot of ways we can organize the stuff for redux, but I kept overcomplicating things and confusing myself when I tried to impose a more complicated/modular structure, so I'm keeping it simple for now.  I am open to suggestions on that — https://github.com/erikras/ducks-modular-redux seems appealing and similar to the way we have other things organized.

/cc @thisandagain @tmickel


mewtaylor:
Mostly walked through this together (thanks @rschamp !). Still getting familiar with redux, but everything here looks good to me (especially the new `Page` wrapper)!

PR #387, opened 2016-03-18
Add "My Class" to student nav

rschamp:


src/components/navigation/navigation.jsx
```diff
@@ -292,6 +292,15 @@ var Navigation = React.createClass({
                                         </a>
                                     </li>
                                 ] : []}
+                                {this.state.session.permissions.student ? [
+                                    <li>
+                                        <a href={'/classes/' + this.state.session.user.classroomId + '/'}>
+                                            <FormattedMessage
+                                                id='general.myClass'
+                                                defaultMessage={'My Class'} />
```
mewtaylor: we're no longer putting default messages inline, and just having them be compiled in the `make translations` step.

mewtaylor:
Just one thing otherwise looks good!

PR #390, opened 2016-03-22
Fix GH-356: Use `formatNumber` to localize project count

### LINKED ISSUE #356: [Homepage] commas not showing in # of projects in Safari

### LINKED ISSUE #366: On homepage: "Failed propType: Invalid prop `projectCount` of type `string` supplied to `Intro`, expected `number`."

mewtaylor:
`toLocaleString()` is not supported in Safari, but react-intl has a polyfill, and so it is. Fixes #356. This also localized the intro description string, which wasn't previously. This also fixes #366 by doing as @rschamp suggested and checking if the count is the default count before setting the value.

Local browsers tested:
* Chrome 49
* Safari 9

### Test Cases ###
* Visit the homepage in any non-Safari browser, not logged in. There should not be a console error about propTypes.
* Visit the homepage in any non-Safari browser not logged in. The project count in the top right should appear as it does in production.
* Visit the homepage in safari not logged in. The project count in the top right should have commas in it.

### DIFF
--- a/src/components/intro/intro.jsx
+++ b/src/components/intro/intro.jsx
@@ -25,9 +25,11 @@ var Intro = React.createClass({
                 'intro.joinScratch': 'JOIN SCRATCH',
                 'intro.seeExamples': 'SEE EXAMPLES',
                 'intro.tagLine': 'Create stories, games, and animations<br /> Share with others around the world',
-                'intro.tryItOut': 'TRY IT OUT'
+                'intro.tryItOut': 'TRY IT OUT',
+                'intro.description': 'A creative learning community with <span class="project-count"> ' +
+                                     'over 13 million </span>projects shared'
             },
-            projectCount: 10569070,
+            projectCount: 13000000,
             session: {}
         };
     },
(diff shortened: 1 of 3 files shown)


src/components/intro/intro.jsx
```diff
@@ -25,9 +25,11 @@ var Intro = React.createClass({
                 'intro.joinScratch': 'JOIN SCRATCH',
                 'intro.seeExamples': 'SEE EXAMPLES',
                 'intro.tagLine': 'Create stories, games, and animations<br /> Share with others around the world',
-                'intro.tryItOut': 'TRY IT OUT'
+                'intro.tryItOut': 'TRY IT OUT',
+                'intro.description': 'A creative learning community with <span class="project-count"> ' +
+                                     'over 13 million </span>projects shared'
             },
-            projectCount: 10569070,
+            projectCount: 13000000,
```
rschamp: Do we need `projectCount` anymore?

mewtaylor: heh – great point :)

rschamp:
Looks great!

PR #391, opened 2016-03-22
Add method to retrieve token from session cookie and use it

rschamp:
Previously #282 and #365.

I got ahead of myself trying to make the whole session come from the cookie. All we need from the session cookie is the token, so this does just that. Later we can provide other methods to get the token.

Requires LLK/scratch-api#104


src/lib/jar.js
```diff
@@ -9,41 +10,66 @@ var xhr = require('xhr');
...
+            }
+
+            return callback(null, strData);
+        } catch (e) {
+            return callback(e);
+        }
+    },
+    get: function (name, callback) {
     // Get cookie by name
     var obj = cookie.parse(document.cookie) || {};
```
mewtaylor: I could be misreading the diff on github here, but should this be indented too?

rschamp: Whoops looks like it yeah.

src/lib/jar.js
```diff
@@ -9,41 +10,66 @@ var xhr = require('xhr');
...
+            var decompress = false;
+            if (b64Data[0] === '.') {
+                decompress = true;
+                b64Data = b64Data.substring(1);
+            }
 
-Jar.get = function (name, callback) {
-    // Get cookie by name
-    var obj = cookie.parse(document.cookie) || {};
+            // Django makes its base64 strings url safe by replacing + and / with - and _ respectively
```
mewtaylor: Wondering if we should drop in a link to Django's docs and/or code on this for future reference?

rschamp: Good idea

mewtaylor:
Thanks for adding in those notes! Looks great!

PR #392, opened 2016-03-22
Cleanup: catch more localization errors that might arise

mewtaylor:
This cleans up the localization stuff in a couple of ways:
1. Adds localization spot checks
  1. checks the localization for a few reference languages for the nav bar
  2. for our localized pages (splash, about, wedo2), it checks to see that there are locale strings, whether or not they're english (to make sure at least something is on the page).
2. Make `build-locales` throw errors it catches
  * The catch there before was meant to catch in the event that a file doesn't exist, and just continue (in which case the language will default to english). However, it was not specific enough, and was silencing other errors – now, it only catches the error if the file doesn't exist, and throws anything else, preventing a build.

### Test Cases ###
* travis-ci should pass
* translations should still show up in all environments

### DIFF
--- a/test/localization/spot_check_about_has_strings.js
+++ b/test/localization/spot_check_about_has_strings.js
@@ -7,7 +7,7 @@ var languages = require('../../languages.json');
 window = {};
 require('../../intl/about.intl.js');
 
-tap.test('spotCheckSplashStrings', function (t) {
+tap.test('spotCheckAboutStrings', function (t) {
     var isoCodes = Object.keys(languages);
     var keysToCheck = Object.keys(window._messages['en']).sort();
     for (var i in isoCodes) {
--- a/test/localization/spot_check_about_has_strings.js
+++ b/test/localization/spot_check_about_has_strings.js
@@ -1,0 +1,21 @@
+/*
+ * spot check that each language has values for the string id keys on About page
+ * that are contained in English (i.e. make sure strings will show up, not ids")
+ */
+var tap = require('tap');
+var languages = require('../../languages.json');
+window = {};
+require('../../intl/about.intl.js');
+
+tap.test('spotCheckSplashStrings', function (t) {
+    var isoCodes = Object.keys(languages);
+    var keysToCheck = Object.keys(window._messages['en']).sort();
+    for (var i in isoCodes) {
+        t.same(
+            Object.keys(window._messages[isoCodes[i]]).sort(),
+            keysToCheck,
+            'check About keys for language ' + isoCodes[i]
+        );
+    }
+    t.end();
+});
--- a/test/localization/spot_check_about_has_strings.js
+++ b/test/localization/spot_check_about_has_strings.js
@@ -2,17 +2,32 @@
  * spot check that each language has values for the string id keys on About page
  * that are contained in English (i.e. make sure strings will show up, not ids")
  */
+var merge = require('lodash.merge');
+var path = require('path');
 var tap = require('tap');
+
 var languages = require('../../languages.json');
-window = {};
-require('../../intl/about.intl.js');
+var localeCompare = require('../../bin/lib/locale-compare');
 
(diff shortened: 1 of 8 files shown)


test/localization/spot_check_about_has_strings.js
```diff
@@ -0,0 +1,21 @@
+/*
+ * spot check that each language has values for the string id keys on About page
+ * that are contained in English (i.e. make sure strings will show up, not ids")
+ */
+var tap = require('tap');
+var languages = require('../../languages.json');
+window = {};
+require('../../intl/about.intl.js');
+
+tap.test('spotCheckSplashStrings', function (t) {
```
rschamp: Copy-paste error

test/localization/spot_check_about_has_strings.js
```diff
@@ -0,0 +1,21 @@
...
+    var keysToCheck = Object.keys(window._messages['en']).sort();
+    for (var i in isoCodes) {
+        t.same(
+            Object.keys(window._messages[isoCodes[i]]).sort(),
+            keysToCheck,
+            'check About keys for language ' + isoCodes[i]
+        );
+    }
+    t.end();
+});
```
rschamp: Newline

test/localization/spot_check_wedo2_has_strings.js
```diff
@@ -0,0 +1,21 @@
+/*
+ * spot check that each language has values for the string id keys on Splash page
+ * that are contained in English (i.e. make sure strings will show up, not ids")
+ */
+var tap = require('tap');
+var languages = require('../../languages.json');
+window = {};
+require('../../intl/wedo2.intl.js');
+
+tap.test('spotCheckSplashStrings', function (t) {
```
rschamp: Copy-paste error

test/localization/spot_check_about_has_strings.js
```diff
@@ -0,0 +1,21 @@
+/*
+ * spot check that each language has values for the string id keys on About page
+ * that are contained in English (i.e. make sure strings will show up, not ids")
+ */
+var tap = require('tap');
+var languages = require('../../languages.json');
+window = {};
+require('../../intl/about.intl.js');
```
rschamp: For now I think this is fine but this is a little hacky.  It might be easier to test if you could test the build-locales script by individual language. Then you wouldn't need to build the whole thing to test specific languages, and you could assign the output to whatever variable you wanted. The full script could take the output and do what it currently does with `window._messages`.

rschamp:
Looks great overall. I think the awkwardness in testing `build-locales` is due to it being a monolith. Now might be the best time to modularize it that since we're working on it and it will make testing it more thoroughly easier in the future.

test/localization/spot_check_about_has_strings.js
mewtaylor: I agree that it is hacky, and am +1 on eventually having build-locales be able to return whatever variable you want rather than `window._messages`. One thing that came up for me during this, though, is that, being that this is meant to be a spot check of what will be used in the system, does it make sense to use what will be built by the `make` process?

rschamp: There definitely should be integration tests, but so long as unit tests catch the same errors as will break the system in the end, I don't think you need to test everything against the output of the build.  You could break this test up into a few different things — the build creates the file, it assigns to `window.messages`, and the keys for each language match those for `en`.  So long as the tests pass for each of these, then you're in the same place testing-wise even if you don't test the whole build.

mewtaylor: :+1:

mewtaylor:
@rschamp I modularized it, and updated the tests accordingly. They take a bit longer now, but are not dependent on another process to run.

rschamp:
Looks great, thanks!  There is room for improvement on `getIdsForView` because it mutates what you pass in and doesn't return anything.  But I don't think that is a big deal right now.

PR #393, opened 2016-03-23
Upgrade sass-lint, quiet various warnings

rschamp:
These warnings were annoying me because the noise makes it easy to miss real issues.

The `no-mergeable-selectors` rule is one that we do want to have, but right now it asks that you merge selectors in different `@media` blocks.  When the next release happens we should put that back.
https://github.com/sasstools/sass-lint/issues/307

Similarly, we want `force-element-nesting` but there is a problem with that because there's no easy way to have a nested selector in a list of selectors.
https://github.com/sasstools/sass-lint/issues/575

Finally, until they implement per-line overrides, we have to silence `class-name-format` because we don't have control over the `ReactModal` class names.  It's a useful rule to keep class names consistent though.  Per-line ignores should be coming soon.
https://github.com/sasstools/sass-lint/issues/70

I also fixed a few other warnings about child keys, and stopped configuring Raven in development.

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -1,6 +1,8 @@
 rules:
+  class-name-format: 0
   color-literals: 2
   final-newline: 2
+  force-element-nesting: 0
   hex-notation: 2
   indentation:
     - 2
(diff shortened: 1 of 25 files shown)


mewtaylor:
Awesome, I was just thinking about this too! Thank you so much for doing this!

rschamp:
I didn't check the output too closely, so I will just make sure to check staging closely in case I messed anything up.

PR #408, opened 2016-04-01
Add view for community blocks interview recruitment

sayamindu:
We are recruiting Scratchers for interviews on the Scratch Community
blocks feature; this is a part of my PhD disseration. This commit has
the PDF forms and the views that are needed for the recruitment.


thisandagain:
Looks good. Thanks @sayamindu .

@rschamp @mewtaylor Mind if we include this with the deploy on Monday?

PR #415, opened 2016-04-07
Add statistics page

kelseyc18:


src/views/statistics/queries.json
```diff
@@ -0,0 +1,40 @@
+{
```
thisandagain: This document can be removed as it is handled by the API, correct?

package.json
```diff
@@ -54,7 +54,11 @@
     "node-sass": "3.3.3",
     "pako": "0.2.8",
     "po2icu": "git://github.com/LLK/po2icu.git#develop",
+    "react": "0.14.0",
```
thisandagain: `react`, `react-dom`, and `react-intl` can be removed from the package as they are included as `externals` via Webpack.

src/views/statistics/statistics.jsx
```diff
@@ -0,0 +1,606 @@
...
+        rows.push(new_row);
+    });
+    return rows;
+}
+
+var MonthlyActivity = React.createClass({
+    mixins: [
+        Api
+    ],
+
```
thisandagain: Extra line break.

thisandagain:
Looks good overall though I think we could do some refactoring on the JSX to minimize boilerplate. I believe the Travis tests are failing due to the addition of `react` and `react-intl` in the package.

PR #416, opened 2016-04-07
[Master] Hotfix: Wrap wedo2 view in page component

mewtaylor:
Need to add the new `Page` component so that this page renders correctly. Currently shows up as a blue box.


rschamp:
LGTM — we should be able to check it out on staging after the tests run.

Issue #19, opened 2015-10-06
Migrate - Community Guidelines


thisandagain:
Existing: https://scratch.mit.edu/community_guidelines/
New Path: `/guidelines`

PR #421, opened 2016-04-13
Fix gh-19: Migrate Community Guidelines

TheGrits:
The raw HTML is tested locally on Chrome, but I don't think it is possible to run a Node.js server on ChromeOS.

Things that would be worth discussing:

**Bullets** 
Currently on Scratch there are cat faces for bullets, but on my screen they look quite pixel-ly. I currently have them as normal bullets, but it might be nice to keep the cat faces.

**Layout**
I put the header in a separate `div` for a different `background-color`, but I'm not sure if you don't want that.

If there is anything else I missed or if there is anything I did incorrectly, please do tell me. I would love to learn more about this type of stuff.

Scratch on!
The_Grits

**EDIT:**
Could the image be put into the site so that we can do something like this? https://github.com/LLK/scratch-www/blob/develop/src/views/jobs/jobs.jsx#L16
That would be ideal.


TheGrits:
FYI: The travis build is failing because of the url of the image at the bottom of the page.

src/views/guidelines/guidelines.jsx
```diff
@@ -0,0 +1,45 @@
...
+var Guidelines = React.createClass({
+    type: 'Guidelines',
+    render: function () {
+        return (
+            <div className="body">
+                <div className="top">
+                    <h1>Scratch Community Guidelines</h1>
+                </div>
+                <div className="bottom">
+                    <p>We need everyone's help to keep Scratch a friendly and creative
```
mewtaylor: The translation part of this needs the `FormattedMessage` component, rather than static text like this. See the [localization guide](https://github.com/LLK/scratch-www/wiki/Localization-Guide) for more info on what render looks like with localized strings, and let me know if you have any questions!

server/routes.json
```diff
@@ -43,5 +43,10 @@
         "pattern": "/wedo",
         "view": "wedo2",
         "title": "LEGO WeDo 2.0"
+    },
+    {
+        "pattern": "/guidelines",
```
mewtaylor: The url currently used is `/community_guidelines`, so we should continue with that at least for now, just while we're migrating. Otherwise, existing links in the django repo will break once this is in production.

src/views/guidelines/guidelines.scss
```diff
@@ -0,0 +1,40 @@
+@import "../../colors";
+
+#view {
+    .body {
+        #top, #bottom {
```
mewtaylor: These are declared as classes in the jsx file, but are referenced as ids here. I believe they should be declared as:
```(css)
.body {
    .top,
    .bottom {
.
.
.
}
```

src/views/guidelines/guidelines.jsx
```diff
@@ -0,0 +1,45 @@
...
+var Page = require('../../components/page/page.jsx');
+
+require('./guidelines.scss');
+
+var Guidelines = React.createClass({
+    type: 'Guidelines',
+    render: function () {
+        return (
+            <div className="body">
+                <div className="top">
```
mewtaylor: We have a component called [`Box`](https://github.com/LLK/scratch-www/blob/develop/src/components/box/box.jsx) which was created to have reusable styling for pages like this, where everything is contained within a bordered box (*see: "What's Happening" on the homepage, or the [`cards`](https://github.com/LLK/scratch-www/blob/develop/src/views/cards/cards.jsx) page*). I think we should re-use that component here.

mewtaylor:
Thanks for getting this going @TheGrits ! I left some notes for you above, let me know if you have any questions.

Additionally – we usually try to have PRs contain a limited amount of commits so that they are a bit easier to revert in case they ever need to be. If you could squash these commits, that'd be really helpful too!

TheGrits:
@mewtaylor The only reason I have so many commits is because I only use a ChromeOS text editor and the web version of Github. (Which also explains the CSS id thing. 😛) I'll try to find a way to squash down the number of commits.

I understand most of the stuff you said, the only thing I don't completely understand is the `Box` component. I'm sure with some tinkering around and looking at the other pages I'll figure it out... eventually. 😉

mewtaylor:
@TheGrits sounds good! definitely feel free to reach out if you get stuck on the Box component along the way.

TheGrits:
@mewtaylor What are your thoughts on the bullets?

TheGrits:
@mewtaylor Done! (Sweeet! I got the squash and merge to work! 🎉)

server/routes.json
```diff
@@ -43,5 +43,10 @@
         "pattern": "/wedo",
         "view": "wedo2",
         "title": "LEGO WeDo 2.0"
+    },
+    {
```
mewtaylor: Sorry to switch things up on you – it just so happens we merged in some changes to the way routes work, so it should read:
```(json)
{
    "name": "guidelines",
    "pattern": "^/community_guidelines$",
    "view": "guidelines/guidelines",
    "title": "Scratch Community Guidelines"
}
```

src/views/guidelines/guidelines.jsx
```diff
@@ -0,0 +1,51 @@
...
+var Page = require('../../components/page/page.jsx');
+var Box = require('../../components/box/box.jsx');
+
+require('./guidelines.scss');
+
+var Guidelines = React.createClass({
+    type: 'Guidelines',
+    render: function () {
+        return (
+            <Box title={<FormattedMessage id='guidelines.title' />}>
```
mewtaylor: The entire return should be wrapped inside of a 
```(html)
<div className="inner">
    <Box>
    .
    .
    .
    </Box>
</div>
```
That should handle centering the box on the page for ya.

src/views/guidelines/guidelines.jsx
```diff
@@ -0,0 +1,51 @@
...
+                <div className="bottom">
```
mewtaylor: I think that this div inside `Box` is not necessary. I think it'd be better to remove this, and make the header and unordered list direct children of the `Box` element:

```
<Box>
    <p>...</p>
    <ul>...</ul>
</Box>
```
Then you can remove the styling for `.bottom` from the scss file too

src/views/guidelines/guidelines.jsx
```diff
@@ -0,0 +1,51 @@
+var React = require('react');
+var FormattedMessage = require('react-intl').FormattedMessage;
+
+var render = require('../../lib/render.jsx');
+var Page = require('../../components/page/page.jsx');
```
mewtaylor: With the new changes that were just merged, this should now read:
```
var Page = require('../../components/page/www/page.jsx');
```

src/views/guidelines/guidelines.jsx
```diff
@@ -0,0 +1,51 @@
...
+                        <li>
+                            <b><FormattedMessage id='guidelines.honestyheader' /></b>
+                            <p><FormattedMessage id='guidelines.honestybody' /></p>
+                        </li>
+                        <li>
+                            <b><FormattedMessage id='guidelines.friendlyheader' /></b>
+                            <p><FormattedMessage id='guidelines.friendlybody' /></p>
+                        </li>
+                    </ul>
+                    <p><FormattedMessage id='guidelines.footer' /></p>
```
mewtaylor: I think this text, and the image below it should be centered. I'd recommend putting this into a div:
```
<div className="guidelines-footer">
    <p><FormattedMessage id='guidelines.footer' /></p>
    <img src="//cdn.scratch.mit.edu/scratchr2/static/images/help/spritesforcommunityguid.png"
         alt="sprites"/>
</div>
```
And then aligning its text in the center.

src/views/guidelines/guidelines.scss
```diff
@@ -0,0 +1,28 @@
+@import "../../colors";
```
mewtaylor: We're also trying to implement some new typography globals using the `_typography` file. I'd recommend adding here:
```(css)
@import "../../typography";
```
Then, you can remove the styles below for `p`, `h1`, and `h2`.

src/views/guidelines/guidelines.jsx
```diff
@@ -0,0 +1,51 @@
...
+var Guidelines = React.createClass({
+    type: 'Guidelines',
+    render: function () {
+        return (
+            <Box title={<FormattedMessage id='guidelines.title' />}>
+                <div className="bottom">
+                    <p><FormattedMessage id='guidelines.header' /></p>
+                    <ul>
+                        <li>
+                            <b><FormattedMessage id='guidelines.respectheader' /></b>
```
mewtaylor: @carljbowman thoughts on these being `h4` instead of `b`? that way it'd look like:
screencapture-localhost-8333-community_guidelines-1461350018840

src/views/guidelines/guidelines.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+require('./guidelines.scss');
+
+var Guidelines = React.createClass({
+    type: 'Guidelines',
+    render: function () {
+        return (
+            <Box title={<FormattedMessage id='guidelines.title' />}>
+                <div className="bottom">
+                    <p><FormattedMessage id='guidelines.header' /></p>
```
mewtaylor: I think it'd be good to add `className="intro"` to this `p` tag. That'll make the text slightly more pronounced than the text below.

mewtaylor:
@TheGrits thanks for the updates! We made some changes to the repo structure this iteration (wiki updates pending), which unfortunately means some other changes here – sorry about that! I'd recommend rebasing this on the `develop` branch, and then updating accordingly. You'll notice during the rebase that there is a conflict on the `server/routes.json` file – this is because that file has now moved to `src/routes.json`. In order to resolve the conflict, I'd just accept the "deleted" version of the file when you do the manual merge, and then add the route to `src/routes.json` as I described above.

Besides that, I left a couple of styling notes for you as well – small things, because otherwise it looks great! As for bullets – I think that bullets instead of scratch cats works for now, but @carljbowman it'd be good to get your thoughts on this if you have a moment (see the screen shot I left above).

@TheGrits let me know if you have any questions about the comments I left too – thanks!

TheGrits:
@mewtaylor No worries, I'll probably update this tomorrow.

mewtaylor:
@TheGrits thanks! Quick update: I just talked with @carljbowman, and we're thinking that instead of using a `<ul>` for this, we should instead use a `<dl>` tag for this. So, the list would look like:
```(html)
<p className="intro">...</p>
<dl>
    <dt><FormattedMessage id='guidelines.respectheader' /></dt>
    <dd><FormattedMessage id='guidelines.respectbody' /></dd>
    .
    .
    .
</dl>
<div>
```

The formatting for that will be handled by the `typography` file. However, while you're doing this, if you could also add a `margin` property to the typography file for `<dd>` tags, that would be appreciated – it should look like this:
```(css)
dl {
    .
    .
    .
    dt {
        font-weight: 700;
    }
    
    // add the lines below
    dd {
        margin: 0;
    }
}
```

Thanks so much!

TheGrits:
@mewtaylor While talking with @carljbowman did you get a chance to talk about the bullets?

TheGrits:
@mewtaylor Rebased and updated version is now here: #433

PR #422, opened 2016-04-13
Upgrade css-loader and autoprefixer-loader

rschamp:
These depended on postcss<5, which caused webpack to output the errors like

```shell
Container#eachAtRule is deprecated. Use Container#walkAtRules instead.
Container#eachRule is deprecated. Use Container#walkRules instead.
Container#eachDecl is deprecated. Use Container#walkDecls instead.
Node#style() is deprecated. Use Node#raw()
Node#_value was deprecated. Use Node#raws.value
Container#remove is deprecated. Use Container#removeChild
```

The autoprefixer-loader package is deprecated now and recommends using postcss-loader + autoprefixer itself.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,11 +31,11 @@
     "raven": "0.10.0"
   },
   "devDependencies": {
-    "autoprefixer-loader": "2.1.0",
+    "autoprefixer": "6.3.6",
     "classnames": "2.1.3",
     "cookie": "0.2.2",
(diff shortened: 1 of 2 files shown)


mewtaylor:
:shipit:

rschamp:
@Jonathan50 The :shipit: is the "ship it!" squirrel, so it's another way of signing off on the PR.

PR #420, opened 2016-04-09
Added support for redirect routes

sclements:
Added code for redirects and now redirecting the donation URL to the proper location. Fixes #412.

PR #425, opened 2016-04-16
Move app to S3 hosting

rschamp:
* Generates .html for each of the views with webpack
* Updates deployment from Elastic Beanstalk to S3
* Adds a script to update Fastly rules to make the S3 hosting work the same as the webserver:
  * Add rules to rewrite Express routes to the appropriate view HTML file
  * Generate the Pass condition from the app's routes and the directories and files in the static directory
  * Ensure the Pass condition is applied to the appropriate entries in Fastly

Known Issues:
* For deployment, it should probably activate the new Fastly config as soon as it's done, since the files on S3 will be updated. I know I can use async to know when everything is done but I kind of ran out of steam.
* It doesn't deal with removing deleted files on S3
* It doesn't deal with removing deleted views in the Fastly config
* It bases everything on the most recent version in Fastly, rather than a pre-determined starting point

Questions:
* Should `configure-fastly` construct the entire configuration, not just what's relevant to the app?
* When I got started, I just left routes.json and render.js in `server/`, but we don't need `server/` anymore... so where should those files live? (Mostly just asking because I'm tired atm)
* Is this going to work for all of our routes in the future? There is a max on the number of characters a condition statement can have.


bin/configure-fastly.js
```diff
@@ -0,0 +1,176 @@
...
+    action: 'set',
+    ignore_if_set: 0,
+    type: 'REQUEST',
+    dst: 'http.host',
+    src: '"' + s3Bucket + '"',
+    request_condition: NOT_PASS_REQUEST_CONDITION_NAME,
+    priority: 1
+};
+
+async.waterfall([
```
thisandagain: Not sure about the use of waterfall here. Wouldn't it make more sense for all of this to be in a single `async.auto`?

bin/configure-fastly.js
```diff
@@ -0,0 +1,176 @@
...
+                    async.series([
+                        async.apply(fastly.setCondition.bind(fastly), version.number, condition),
+                        async.apply(fastly.setFastlyHeader.bind(fastly), version.number, header)
+                    ], cb);
+                }, cb)
+            }
+        ], next);
+    }], function (err, responses) {
+        if (err) {
+            console.log(err);
```
thisandagain: If you want the process to exit with an error code (fatal exception), it might make more sense to throw an error so developers have a stack trace:
```js
throw new Error(err);
```

bin/lib/fastly-extended.js
```diff
@@ -0,0 +1,133 @@
+var Fastly = require('fastly');
```
thisandagain: Happy to merge all of this stuff into the upstream [fastly](https://github.com/thisandagain/fastly) module.

webpack.config.js
```diff
@@ -5,6 +5,36 @@ var webpack = require('webpack');
 
 var routes = require('./server/routes.json');
 
+
+function HtmlGeneratorPlugin (options) {
+    this.options = options;
+    return this;
+}
+
+HtmlGeneratorPlugin.prototype.apply = function (compiler) {
```
thisandagain: 👍

thisandagain:
This is great. A few fairly minor comments and some responses below. In general I think the only major question I have is around using Travis's S3 deployment system over simply including something like the [s3](https://www.npmjs.com/package/s3) module which could give us more control over the flow and allow us to deploy locally from our dev environment should a need arise.

#### Known Issues
> For deployment, it should probably activate the new Fastly config as soon as it's done, since the files on S3 will be updated. I know I can use async to know when everything is done but I kind of ran out of steam.

See comment below about baking more of the flow into the `Makefile`

> It doesn't deal with removing deleted files on S3

Could we use the `rsync` method rather than `cp` if we move the uploader into our own code?

> It doesn't deal with removing deleted views in the Fastly config

That is a little tricky. I think that is likely ok for now, but we should probably create an ops ticket for it.

> It bases everything on the most recent version in Fastly, rather than a predetermined starting point

That seems fine for now as well.

#### Questions
> Should configure-fastly construct the entire configuration, not just what's relevant to the app?

Seems like we should be doing more in the `Makefile` rather than adding more to the fastly configuration helper:
- build
- rsync to s3
- deploy fastly configuration

> When I got started, I just left routes.json and render.js in server/, but we don't need server/ anymore... so where should those files live? (Mostly just asking because I'm tired atm)

Do we even need `./server/` anymore? Seems like some of those files should be deleted and the root directory renamed / integrated with the rest of the build helper modules.

> Is this going to work for all of our routes in the future? There is a max on the number of characters a condition statement can have.

Hard to say, I think the routes I'm most interested in are `/projects/:project` and `/users/:user`. I can't find many routes longer than `/info/communityblocks-interviews` when looking at `scratchr2`.

bin/configure-fastly.js
rschamp: Thanks, `async.auto` is awesome.

bin/lib/fastly-extended.js
rschamp: Thanks. I think I'll make a separate PR for that, it's pretty specific to this task right now and could be generalized.

rschamp:
> Do we even need `./server/` anymore? Seems like some of those files should be deleted and the root directory renamed / integrated with the rest of the build helper modules.

I just tried to remove `./server`, and realized it's still useful as a dev server, since we still want it for proxying routes to scratchr2, as well as resolving the routes to the view HTML.  Did you have something else in mind for that?

rschamp:
@thisandagain This now uses the s3 module to sync between build and the S3 bucket, deleting unused files, and activates the configuration after deploying.

I also added the ability to configure redirect routes within `routes.json`, pulling from @sclements PR #420.  I'm not the happiest with the way that part of the script works, but wasn't really sure how to unravel it more.

Unfortunately the script provider isn't functional right now, so we can't deploy with make by branch using the `deploy` section of `.travis.yml`. I hacked together something similar in `after_success`.  It would be really nice if you could configure environmental variables by branch.

thisandagain:
This looks good. I think we could probably generalize / refactor a little more but this seems like a reasonable place to start. Regarding the dev server I wish we could use `webpack-dev-server` as it is so fast / works well with the watcher / handle live-reload but understand that is a tall order due to our pass-through system for `scratchr2`.

rschamp:
@thisandagain I spent some time trying to get webpack-dev-server's proxy system working for our pass through, but it doesn't really look like it's built for that. It's better if you have explicit urls you want to proxy, rather than proxying anything that isn't handled by webpack.  I think in the end having our own dev server using webpack-dev-middleware is probably as close as we can get, is much clearer to read for the proxy and url rewriting, and I believe can be used exactly the same way as webpack-dev-server itself as far as live reloading goes.

As far as generalizing and refactoring, what do you mean specifically?

I'm merging this now so that @mewtaylor can rebase on it, but am happy to work on it more later. I'm not super into the way it uses the routes, seems kind of overcomplicated.

thisandagain:
> As far as generalizing and refactoring, what do you mean specifically?

- Generalize the extensions to the `fastly` module and land them upstream
- Refactor the `bin/configure-fastly.js` file to make it testable (split out some of the code in the giant `async.auto` block) into utility methods
- Refactor the route / redirect handler

PR #429, opened 2016-04-21
Add initial site for Scratch Conference

mewtaylor:
Beyond adding the conference sub site, this PR does a few other things:
* Splits our components into two types: `presentation` and `container` components (/cc @rschamp @thisandagain )
* takes margins out of `typography` and makes them view-specific (/cc @carljbowman )
* changes how routes are rendered from `routes.json` (done to make sub-paths a bit easier):
   * `name` – how a bundle is named
   * `view` – a path to the view directory used to render the page

### Test Plan ###
* All pages should be viewable
* `Cards` and `HOC` pages should look as they did before, though they are using the `FlexBox` component now
* The conference site should have the following views:
  * `/conference` – landing page for the conference site  
  * `/conference/expect`
  * `/conference/plan`
* The conference site should be somewhat responsive
  * The nav should stack at the vertical tablet breakpoint
  * "Scratch Family" should disappear from the footer at the vertical tablet breakpoint
  * `FlexBox` components should generally move from horizontal to vertical at the vertical tablet breakpoint
  * h1 font-size should get smaller at the vertical tablet breakpoint


rschamp:
When we were talking about container vs presentational components, it was in the context of Redux which describes them here: http://redux.js.org/docs/basics/UsageWithReact.html#presentational-and-container-components (longer article here: https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0#.f8nzhsm21).  The gist is that container components just hook up the state and handle dispatching actions for presentation components.

The split you have now between `presentation/` and `container/` seems to be more about whether the component has content in it or not, which I think also makes sense, but I'm not sure if it's worth splitting them over. If we do have it split this way, then I think that `Page` should probably go into presentation since it holds no content.  I think it would probably be less confusing if they weren't split up though because I can see myself getting confused about what the right path to which component is without a very strong split between the types like it is right now.

rschamp:
For the change to `routes.json`, we need to update the other places that use the old sense of `route.view` to use `route.name`:
* https://github.com/LLK/scratch-www/blob/9b52097c381a271b08bb6e0934204d533ffe74da/bin/configure-fastly.js#L85
* https://github.com/LLK/scratch-www/blob/9b52097c381a271b08bb6e0934204d533ffe74da/bin/configure-fastly.js#L237
* https://github.com/LLK/scratch-www/blob/9b52097c381a271b08bb6e0934204d533ffe74da/mustache-renderer-webpack-plugin/index.js#L33

mewtaylor:
Changed as discussed offline – thanks @rschamp !

rschamp:
Looks good, thanks!

PR #431, opened 2016-04-22
Fix gh-426: Update Project Count

### LINKED ISSUE #426: Default project count in intro out of date again
A while ago we changed the default text on the homepage to read `Over 13 million` projects rather than an incorrect number (which it was before – see #357). We have very quickly moved into 14 million projects, though, making the current text out of date. Possible fixes:

* Continue updating this number every million we get
* Determine alternative text to display that captures the magnitude of projects created but does not need updating as often

TheGrits:
Update project count to (temporarily) fix #426  
I took the liberty to update _all_ references of 13 million projects to 14 million projects.


mewtaylor:
@TheGrits looks good thanks for this!

PR #433, opened 2016-04-23
Fix gh-19: Migrate Community Guidelines

TheGrits:
Rebased #421 and added all of the changes as per @mewtaylor's comments.

(#19 - so that #19 has a reference to this PR)

I already know that Travis will hate the image url, it would be great if a shorter url could be created.

### DIFF
--- a/src/views/guidelines/guidelines.jsx
+++ b/src/views/guidelines/guidelines.jsx
@@ -1,0 +1,41 @@
+var React = require('react');
+var FormattedMessage = require('react-intl').FormattedMessage;
+
+var render = require('../../lib/render.jsx');
+var Page = require('../../components/page/www/page.jsx');
+var Box = require('../../components/box/box.jsx');
+
+require('./guidelines.scss');
+
+var Guidelines = React.createClass({
+    type: 'Guidelines',
+    render: function () {
+        return (
+            <div className="inner">
+                <Box title={<FormattedMessage id='guidelines.title' />}>
+                    <p><FormattedMessage id='guidelines.header' className="intro" /></p>
+                    <dl>
+                        <dt><FormattedMessage id='guidelines.respectheader' /></dt>
+                        <dd><FormattedMessage id='guidelines.respectbody' /></dd>
+                        <dt><FormattedMessage id='guidelines.constructiveheader' /></dt>
+                        <dd><FormattedMessage id='guidelines.constructivebody' /></dd>
+                        <dt><FormattedMessage id='guidelines.shareheader' /></dt>
+                        <dd><FormattedMessage id='guidelines.sharebody' /></dd>
+                        <dt><FormattedMessage id='guidelines.privacyheader' /></dt>
+                        <dd><FormattedMessage id='guidelines.privacybody' /></dd>
+                        <dt><FormattedMessage id='guidelines.honestyheader' /></dt>
+                        <dd><FormattedMessage id='guidelines.honestybody' /></dd>
+                        <dt><FormattedMessage id='guidelines.friendlyheader' /></dt>
+                        <dd><FormattedMessage id='guidelines.friendlybody' /></dd>
+                    </dl>
+                    <div className="guidelines-footer">
+                      <p><FormattedMessage id='guidelines.footer' /></p>
+                      <img src="//cdn.scratch.mit.edu/scratchr2/static/images/help/spritesforcommunityguid.png" alt="sprites"/>
+                    </div>
+                </Box>
+            </div>
+        );
+    }
+});
+
+render(<Page><Guidelines /></Page>, document.getElementById('app'));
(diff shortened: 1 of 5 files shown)


TheGrits:

src/views/guidelines/guidelines.scss
```diff
@@ -0,0 +1,11 @@
+@import "../../colors";
+@import "../../typography";
+
+.view {
+    .guidelines-footer {
+        text-align: center;
```
mewtaylor: @TheGrits this isn't aligning center for me when I look at it in my development environment right now – is it for you?

TheGrits: @mewtaylor That's odd. Does removing the rule for the image help?

mewtaylor: I haven't tried it yet – have you tried looking at it in your browser?

TheGrits: I have tested removing the image rule, I'm using Chrome on ChromeOS.

TheGrits: This is what the HTML produces in my browser:
screenshot
It's kind of odd, images don't normally align with the `text-align` rule. But this is what my CSS for the footer is:
screenshot

mewtaylor: Ah I I think I see – `.view` should actually be `#view` in this case (it's an `id` in the template). Changing that works for me – and I think that the `text-align: center` is sufficient, so I think you could remove the `img` formatting below

src/views/guidelines/guidelines.scss
```diff
@@ -0,0 +1,11 @@
+@import "../../colors";
+@import "../../typography";
+
+.view {
+    .guidelines-footer {
```
mewtaylor: I think it might be good to add some margins between the `dt/dd` tags, so that it looks like this:
```
#view {
    dl {
        dt {
            margin-top: 1.5rem;
        }
    }

    .guidelines-footer {
        margin-top: 1.5rem;
        text-align: center;
    }
}
```
@TheGrits If you could make those slight changes, I think this is ready to go!

mewtaylor:
@TheGrits sorry for the delay from yesterday – just one last thing then I think this is ready to go :)

src/views/guidelines/guidelines.jsx
```diff
@@ -0,0 +1,41 @@
...
+                        <dt><FormattedMessage id='guidelines.privacyheader' /></dt>
+                        <dd><FormattedMessage id='guidelines.privacybody' /></dd>
+                        <dt><FormattedMessage id='guidelines.honestyheader' /></dt>
+                        <dd><FormattedMessage id='guidelines.honestybody' /></dd>
+                        <dt><FormattedMessage id='guidelines.friendlyheader' /></dt>
+                        <dd><FormattedMessage id='guidelines.friendlybody' /></dd>
+                    </dl>
+                    <div className="guidelines-footer">
+                      <p><FormattedMessage id='guidelines.footer' /></p>
+                      <img src="//cdn.scratch.mit.edu/scratchr2/static/images/help/spritesforcommunityguid.png" alt="sprites"/>
```
mewtaylor: I'm getting a lint error on this line being too long – @TheGrits could you move the `alt` attribute to the following line, like below?
```(html)
<img src="//cdn.scratch.mit.edu/scratchr2/static/images/help/spritesforcommunityguid.png"
     alt="sprites"/>
```

TheGrits:
@mewtaylor I have made the changes according to your comments! :)

src/views/guidelines/guidelines.scss
```diff
@@ -0,0 +1,16 @@
+@import "../../colors";
+@import "../../typography";
+
+#view {
+    .guidelines-footer {
+        text-align: center;
```
mewtaylor: I think this is a duplicate of the line below

TheGrits: @mewtaylor lol Fixed. (oops!)

mewtaylor:
@TheGrits thanks for the speedy turnaround! Just one duplicate line and then we're good :)

TheGrits:
@mewtaylor Wait, another duplicate line? (Or have you not seen the most recent commit?)

mewtaylor:
@TheGrits sorry – you were so fast to respond, that you fixed it before I could leave the comment! That was the same line, you got it. 

However the linter is now erroring about two other things – let me drop in some comments above to point them out

src/views/guidelines/guidelines.scss
mewtaylor: thanks :). I think the linter wants these to be switched though – `margin-top` above `text-align`

TheGrits: @mewtaylor Sounds good, I'll make the change now.

TheGrits:
@mewtaylor Huzzah! The changes are made and Travis gives a thumbs up!

TheGrits:
@mewtaylor lol Now you're the one that's faster than the other!

mewtaylor:
@TheGrits haha yeah I guess we both move fast! Thanks again for the back and forth on this one

TheGrits:
@mewtaylor No problem, I'm glad I could help. :)

Issue #310, opened 2016-01-07
Migrate - DMCA


kaschm:

TheGrits:
@mewtaylor Shouldn't this issue be closed by #435?

mewtaylor:
@TheGrits yup – thanks!

PR #424, opened 2016-04-15
Fix gh-310: Migrate DMCA

TheGrits:
After (I hope) figuring out the `Box` component, this one was a no-brainer. 

HTML tested locally on Chrome. (ChromeOS)

Sorry about those commits before the squash and merge, I couldn't figure out how to get rid of them. :\

@mewtaylor @thisandagain @rschamp 

EDIT:
Just so that #310 has a reference to this PR.

PR #435, opened 2016-04-25
Fix gh-310: Migrate DMCA

TheGrits:
Rebased and updated version of #424 

(Also, hardly any extra commits! 😁)

HTML tested locally on Chrome. (ChromeOS)

(Just so that #310 has a reference to this PR.)


src/views/dmca/dmca.jsx
```diff
@@ -0,0 +1,37 @@
...
+                    <p>Room E14-445A</p>
+                    <p>Cambridge, MA 02139</p>
+                    <p>Tel: (617) 253-9783</p>
+                    <br/>
+                    <p><FormattedMessage id='dmca.llkresponse' /></p>
+                    <br/>
+                    <p><FormattedMessage id='dmca.assessment' /></p>
+                    <br/>
+                    <p><FormattedMessage id='dmca.eyetoeye' /></p>
+                    <br/>
```
mewtaylor: paragraph tags have their own margins set, so I think the `<br/>` tags between paragraphs can be taken out.

src/views/dmca/dmca.jsx
```diff
@@ -0,0 +1,37 @@
...
+            <div className="inner">
+                <Box title={'DMCA'}>
+                    <p><FormattedMessage id='dmca.intro' /></p>
+                    <br/>
+                    <p>Copyright Agent / Mitchel Resnick</p>
+                    <p>MIT Media Laboratory</p>
+                    <p>77 Massachusetts Ave</p>
+                    <p>Room E14-445A</p>
+                    <p>Cambridge, MA 02139</p>
+                    <p>Tel: (617) 253-9783</p>
```
mewtaylor: The margin between lines in the address is quite large when using separate `<p>` tags for each line. I'd recommend putting the entire address into one `<p>` tag, and then using the `<br/>` tag to do line breaks within that paragraph.

src/views/dmca/dmca.scss
```diff
@@ -0,0 +1,2 @@
+@import "../../colors";
+@import "../../typography";
```
mewtaylor: I'm a bit torn on this – on one hand, doing the import here and then importing the `dmca.scss` file in `dmca.jsx` keeps to our conventions, however it seems that if all we're doing is importing in a view's scss file that we could just as easily do those imports on the jsx file directly, and delete the scss file.

mewtaylor:
@TheGrits looks good overall – left you a couple of suggestions, and one open ended question :).

src/views/dmca/dmca.scss
TheGrits: Although that is a good idea, I'm still leaning towards keeping the `dmca.scss` file in case changes are needed in the future.

TheGrits:
@mewtaylor OK, I have edited the `dmca.jsx` file according to your comments. 😀

src/views/dmca/dmca.scss
TheGrits: @mewtaylor Maybe we should see what thisandagain or carljbowman think?

mewtaylor: @TheGrits this is good for now, thanks! We might do a bit of restructuring with the typography file, but this keeps to current convention.

src/views/dmca/dmca.jsx
```diff
@@ -0,0 +1,32 @@
...
+        return (
+            <div className="inner">
+                <Box title={'DMCA'}>
+                    <p><FormattedMessage id='dmca.intro' /></p>
+                    <p>Copyright Agent / Mitchel Resnick<br/>
+                    MIT Media Laboratory<br/>
+                    77 Massachusetts Ave<br/>
+                    Room E14-445A<br/>
+                    Cambridge, MA 02139<br/>
+                    Tel: (617) 253-9783</p>
```
mewtaylor: Just one nitpick (sorry) – our convention for tabbing with multiline paragraphs is to indent the text on a new line, like this:
```
<p>
    Copyright Agent / Mitchel Resnick<br/>
    MIT Media Laboratory<br/>
    77 Massachusetts Ave<br/>
    Room E14-445A<br/>
    Cambridge, MA 02139<br/>
    Tel: (617) 253-9783
</p>
```
If you could change that, I think we're all set, because looking at it in the browser it looks great!

mewtaylor:
@TheGrits just left one small nitpick, otherwise this looks read to go! Thanks!

TheGrits:
@mewtaylor Done!

mewtaylor:
Awesome :)

technoboy10:
Looks good to me!

QA tests:
- [x] Text looks okay
- [x] Important info (address, fair use section) is correct
- [x] All links go to intended destination
- [x] Can still log in and out

PR #437, opened 2016-04-25
Bugfix: Temporarily remove "View Full Schedule" link

mewtaylor:
Since we don't have one yet.

### Test Cases ###
* Visit `/conference/expect` – there shouldn't be a link to "View Full Schedule" in the schedule section


carljbowman:
@mewtaylor - Good catch! Looks good to me.

PR #444, opened 2016-04-27
Make viewport width a variable value

mewtaylor:
So that we can allow some pages to be responsive (i.e. conference, which is responsive at the moment).

### Test Cases ###
* View splash page on a desktop – it should show up as usual
* View conference page on a desktop – it should show up as usual
* View splash page on a mobile device – it should be zoomed out to a desktop view
* View conference page on a mobile device – it should be responsive


rschamp:
Nice and simple!

PR #450, opened 2016-04-28
[Master] Mark version with webpack

rschamp:
Moves tagging git version into webpack so it's necessarily within build. On Travis, just use the environment variable rather than going to git.


.travis.yml
```diff
@@ -24,6 +24,7 @@ env:
   - S3_LOCAL_DIR=build
   - S3_REGION=us-east-1
   - SKIP_CLEANUP=true
+  - WWW_VERSION==${TRAVIS_COMMIT:0:5}
```
mewtaylor: extra =

PR #458, opened 2016-04-29
[Master] Use s3cmd for S3 deployment

rschamp:
The `s3` module is unreliable for syncing uploads, apparently because of the difficulty of using `aws-sdk`. Therefore, use the more reliable Python-based tool `s3cmd` for this task.

While there, I added a purge all step in the `sync-fastly` command.


thisandagain:
This looks good. Could you create a backlog bug to re-investigate the `s3` module again in the future once some of the dust settles?

PR #463, opened 2016-04-30
Explore and Search pages

rthorn:
This includes the redesigned Explore page and a Search page based off of that. There is also a new tabs component that is used for both pages. As of right now, studios are not currently integrated into either page.


src/components/carousel/carousel.jsx
```diff
@@ -34,7 +34,7 @@ var Carousel = React.createClass({
             slidesToScroll: 5,
             variableWidth: true
         });
-        var arrows = this.props.items.length > settings.slidesToShow;
+        var arrows = this.props.items.length > settings.slidesToShow && settings.slidesToScroll>0;
```
thisandagain: Nit: Add spaces between operator and operands.

src/components/carousel/carousel.jsx
```diff
@@ -43,13 +43,24 @@ var Carousel = React.createClass({
             <Slider className={classes} arrows={arrows} {... settings}>
                 {this.props.items.map(function (item) {
                     var href = '';
+                    var thumbnail_url = item.thumbnail_url;
+                    if (thumbnail_url==undefined && item.image!=undefined) {
```
thisandagain: Nit: Add spaces between operator and operands.

thisandagain: Nit: Always use strict equality.

src/views/explore/explore.jsx
```diff
@@ -0,0 +1,148 @@
...
+var render = require('../../lib/render.jsx');
+
+var Api = require('../../mixins/api.jsx');
+
+var Page = require('../../components/page/www/page.jsx');
+var Box = require('../../components/box/box.jsx');
+var Tabs = require('../../components/tabs/tabs.jsx');
+var SubNavigation = require('../../components/subnavigation/subnavigation.jsx');
+var Carousel = require('../../components/carousel/carousel.jsx');
+var offset = 0;
```
thisandagain: `offset`, `more`, and `tab` should be included in the React component's state (`getInitialState`) rather than in the global module.

src/views/explore/explore.jsx
```diff
@@ -0,0 +1,148 @@
...
+var more = [];
+var tab = 'all';
+var acceptableTabs = ['all','animations','art','games','music','stories'];
```
thisandagain: `acceptableTabs` should be a property of the React component.

src/views/explore/explore.jsx
```diff
@@ -0,0 +1,148 @@
...
+
+require('./explore.scss');
+
+var Explore = injectIntl(React.createClass({
+    type: 'Explore',
+    mixins: [
+        Api
+    ],
+    getInitialState: function () {
+        return {};
```
thisandagain: `exploreAll` should be defined in initial state as `null` or `[]`.

src/views/explore/explore.jsx
```diff
@@ -0,0 +1,148 @@
...
+            tabText = '&q='+tab;
+        }
+        offset+=16;
+        this.api({
+            uri: '/search/projects?limit=16&offset='+offset+tabText
+        }, function (err, body) {
+            if (!err) this.setState({exploreMore: body});
+        }.bind(this));
+    },
+    renderRows: function () {
```
thisandagain: This might be good to talk about tomorrow in-person.

src/views/search/search.jsx
```diff
@@ -0,0 +1,148 @@
...
+
+var Page = require('../../components/page/www/page.jsx');
+var Box = require('../../components/box/box.jsx');
+var SubNavigation = require('../../components/subnavigation/subnavigation.jsx');
+var Tabs = require('../../components/tabs/tabs.jsx');
+var Carousel = require('../../components/carousel/carousel.jsx');
+var offset = 0;
+var more = [];
+var tab = 'projects';
+var searchTerm = '';
```
thisandagain: Same here ... it seems like this should be handled by react's state machine.

src/views/explore/explore.jsx
```diff
@@ -0,0 +1,148 @@
...
+        }, function (err, body) {
+            if (!err) this.setState({exploreAll: body});
+        }.bind(this));
+    },
+    getExploreMore: function () {
+        var tabText = '';
+        if (tab!='all') {
+            tabText = '&q='+tab;
+        }
+        offset+=16;
```
thisandagain: `16` should probably be a property defined in default props.

thisandagain:
Looks good! There is a fair amount of boilerplate being shared by both the search and explore views. Let's come up with a strategy to refactor this a bit tomorrow.

src/views/search/search.jsx
```diff
@@ -0,0 +1,148 @@
...
+            searchTerm = searchTerm.substring(0,searchTerm.indexOf('/'));
+        }
+        while (searchTerm.indexOf('&')>-1) {
+            searchTerm = searchTerm.substring(0,searchTerm.indexOf('&'));
+        }
+        searchTerm = searchTerm.split('+').join(' ');
+        this.getSearchResults(0);
+    },
+    getSearchResults: function () {
+        var termText = '';
```
thisandagain: This could probably be refactored a bit.

src/views/search/search.jsx
```diff
@@ -0,0 +1,148 @@
...
+        }
+        offset+=16;
+        this.api({
+            uri: '/search/projects?limit=16&offset='+offset+termText
+        }, function (err, body) {
+            if (!err) this.setState({searchMore: body});
+        }.bind(this));
+    },
+    renderRows: function () {
+        var row2 = this.state.searchResults;
```
thisandagain: Remove `carousel` and replace with new component called `grid` that accepts an array of projects.

src/views/search/search.jsx
```diff
@@ -0,0 +1,148 @@
...
+
+var Search = injectIntl(React.createClass({
+    type: 'Search',
+    mixins: [
+        Api
+    ],
+    getInitialState: function () {
+        return {};
+    },
+    componentDidMount: function () {
```
thisandagain: Add a `@todo` comment here to make note of how this should be migrated to React-Router once available.

static/svgs/view/view_type-gray.svg
```diff
@@ -0,0 +1,118 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
rthorn: You may want to replace this with an svg matching the love and remix icons in file format.

static/svgs/favorite/favorite_type-gray.svg
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
rthorn: You may want to replace this with an svg matching the love and remix icons in file format.

src/components/grid/grid.jsx
```diff
@@ -0,0 +1,77 @@
...
+    type: 'Grid',
+    getDefaultProps: function () {
+        return {
+            items: require('./grid.json'),
+            itemType: 'projects',
+            showLoves: false,
+            showFavorites: false,
+            showRemixes: false,
+            showViews: false,
+            perRow: 4
```
thisandagain: The `row` concept could probably be removed now that you are using flexbox.

src/components/grid/grid.scss
```diff
@@ -0,0 +1,45 @@
...
+.grid {
+    display: flex;
+    padding: 12px;
+    flex-direction: row;
+    flex-wrap: wrap;
+    justify-content: space-around;
+    align-items: center;
+
+    .thumbnail {
+        padding: 12px;
```
thisandagain: You'll want to add `display: inline-block` to this for older IE support (no flexbox).

static/svgs/favorite/favorite_type-gray.svg
thisandagain: Could you run this through [svgo](https://github.com/svg/svgo)?

static/svgs/view/view_type-gray.svg
thisandagain: Same here.

thisandagain:
This is basically ready to go. @mewtaylor could I get a second set of eyes?

thisandagain:
👍 from my perspective. Assigning to @mewtaylor for final review.

src/routes.json
```diff
@@ -24,6 +24,18 @@
         "title": "Hour of Code"
     },
     {
+        "name": "explore",
+        "pattern": "^/explore/:projects/:all/?$",
```
mewtaylor: @rschamp I'm not we've yet figured out how to handle this between fastly vs express yet. Since these are now turned into varnish rules, should we be doing something like `^/explore/\d{1,15}/\w{1,30}/?$` instead (that is a rough guestimate of the rule)

src/views/explore/explore.jsx
```diff
@@ -0,0 +1,159 @@
...
+
+var Page = require('../../components/page/www/page.jsx');
+var Box = require('../../components/box/box.jsx');
+var Tabs = require('../../components/tabs/tabs.jsx');
+var SubNavigation = require('../../components/subnavigation/subnavigation.jsx');
+var Grid = require('../../components/grid/grid.jsx');
+
+require('./explore.scss');
+
+// @todo migrate to React-Router once available
```
mewtaylor: @rschamp @thisandagain I think we should chat briefly about this at some point. When I looked at it last week it seemed pretty straigthforward to do a multi-page app using router and webpack together w/ chunking the bundle, but I bet @rschamp might have some more insight after the webpack work you've done recently!

src/views/explore/explore.jsx
```diff
@@ -0,0 +1,159 @@
...
+                    </a>;
+        }
+        return allTab;
+    },
+    render: function () {
+        return (
+            <div>
+                <div className='outer'>
+                    <Box title={'Explore'}>
+                        <SubNavigation className='categories'>
```
mewtaylor: Are we getting rid of the custom tag search in this subnav in favor of the new search page?

src/views/explore/explore.jsx
```diff
@@ -0,0 +1,159 @@
...
+                break;
+            }
+        }
+        window.location = window.location.origin + '/explore/'+newType+'/'+this.props.tab;
+    },
+    getBubble: function (type) {
+        var allBubble = <a href={'/explore/'+this.props.itemType+'/'+type+'/'}>
+                        <li>
+                            <FormattedMessage
+                                id={'explore.' + type}
```
mewtaylor: I think we still need an `l10n.json` file for these, unless I'm missing it

src/views/search/search.jsx
```diff
@@ -0,0 +1,124 @@
...
+            var currentOffset = this.state.offset + this.props.loadNumber;
+            this.setState({offset: currentOffset});
+        }.bind(this));
+    },
+    getTab: function (type) {
+        var term = this.props.searchTerm.split(' ').join('+');
+        var allTab = <a href={'/search/'+type+'?q='+term+'/'}>
+                        <li>
+                            <FormattedMessage
+                                id={'explore.'+type}
```
mewtaylor: I think that, since this is a separate view, these might need to be in a separate `l10n.json` file than explore – unless the thought was to include all of the `explore.*` strings in the main `l10n.json` file?

src/components/tabs/tabs.jsx
```diff
@@ -0,0 +1,26 @@
...
+var SubNavigation = require('../../components/subnavigation/subnavigation.jsx');
+var React = require('react');
+
+require('./tabs.scss');
+
+/**
+ * Container for a custom, horizontal list of navigation elements
+ * that can be displayed within a view or component.
+ */
+var Tabs = React.createClass({
```
mewtaylor: This is awesome – I could definitely use a tabs component in conference, so this is great to have!

mewtaylor:
@rthorn this is awesome!

Sorry for my delayed response – I left a couple of comments, but I think the only thing we need to do before pulling this in is handle the localisation strings, and maybe the fastly rules thing too.

It would be great, when we discuss react-router, to also talk about how to bring redux into this. I think it could be helpful to use redux in a future version of this to separate out layout from functionality in the views (which might also help with setting up some unit tests).

src/routes.json
rschamp: Yeah, good point, I think we could do a replacement like the one below to cover these so they would all become e.g. `^/explore/.*?/.*?/?$`, since it would be tough to configure the translation of the format of each argument from express to fastly.
```javascript
pattern.replace(/(:[^/]*)\//gi, ".*?/")
```

src/routes.json
```diff
@@ -24,6 +24,18 @@
...
+        "view": "explore/explore",
+        "title": "Explore"
+    },
+    {
+        "name": "search",
+        "pattern": "^/search/:projects?$/?$",
```
rschamp: This regex looks a little screwy, is it supposed to be `^/search/:projects/?$`?

src/views/explore/explore.jsx
rschamp: Yes I tried out using `require.ensure` for dynamic script loading yesterday and it seemed to work pretty well.  Although I believe we wouldn't even need to use that in this case since all of the explore pages use the same javascript, so we could treat it as a self-contained SPA.

src/routes.json
mewtaylor: i think that makes sense – do we want to do something a bit more restrictive though? i.e. make it be alphanumeric, and make it be under a certain character limit? Or would that make the rule too large...

mewtaylor:
@rthorn @thisandagain moved this to #559, closing for now. Thanks again for all this amazing work, @rthorn !

PR #465, opened 2016-05-02
[Master] Hotfix: Add missing env variables needed for the build

rschamp:
Fixes the issue by adding `API_HOST`, `GA_TRACKER` and `SENTRY_DSN` so that the build can use them.

There are so many that depend on the branch, and are required by either the build or deploy script, so I consolidated all the variables to be set in the environment setup section. Unfortunately there's no easy way to override variables by branch, and this seemed the cleanest way to do it.


thisandagain:
👍  once Travis is ✅

Issue #457, opened 2016-04-29
Cosmetic issue when logging in - on production


jwzimmer:
If you sign in using the drop-down from the Home Page, the "Sign In" button and "Forgot Password" text are too close together.

Steps to repro:
- Go to scratch.mit.edu
- Log out if you are logged in
- Click on Sign In in the upper right
- A dropdown should come down with username/ password fields
- Note that the Sign In button and Forgot Password text encroach on each other uncomfortably

Browser: Chrome Version 50.0.2661.86 (64-bit)

This minor cosmetic issue is visible even when my browser is maximized/ full-screen.

Note there is no impact on functionality, the log in still works fine.

<img width="316" alt="screen shot 2016-04-29 at 1 03 02 pm" src="https://cloud.githubusercontent.com/assets/1320456/14923389/fa38fe50-0e0a-11e6-87e9-a6df18fc3899.png">

PR #470, opened 2016-05-09
Fix gh-457: Log In Margin

TheGrits:
Fix cosmetic issue reported in #457 

Tested locally on Chrome.

4.4px is the max margin to the right of the button without the "Forgot password?" text wrapping to the next line.

### DIFF
--- a/src/components/login/login.scss
+++ b/src/components/login/login.scss
@@ -7,7 +7,11 @@
         padding-top: 5px;
         font-weight: bold;
     }
-
+    
+    .right {
+        float: right;
+    }
+    
     .spinner {
         margin: 0 .8rem;
         width: 1rem;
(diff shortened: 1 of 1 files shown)


TheGrits:

rschamp:
Thanks @TheGrits! The issue with this solution is that while it will work for "Forgot password?" in English, other languages may use more or less space. It looks like somewhere along the line, the rule for `.right` went missing. Could you add a rule for `.right { float: right; }`, then it will always use the maximum space available until it wraps, and will work for translations of "Forgot password?" as well.

TheGrits:
@rschamp Ah, that makes sense. I'll fix it now.

TheGrits:
@rschamp Done!

rschamp:
Thanks!

TheGrits:
No problem!

PR #471, opened 2016-05-11
Fix gh-309: Migrate Privacy Policy

TheGrits:
Migrated the [Privacy Policy](https://scratch.mit.edu/privacy_policy/) page as per #309 

HTML tested locally on Chrome 50 (ChromeOS)

<sup>I'm learning a lot doing this stuff, it's awesome!</sup>


TheGrits:
I'm not sure where Travis is getting the unexpected `}` from. hmm...

src/views/privacypolicy/l10n.json
```diff
@@ -0,0 +1,34 @@
...
+  "privacypolicy.infoprotectionheader": "How does the Scratch Team protect my information?",
+  "privacypolicy.infoprotectionbody": "The Scratch Team has in place physical and electronic procedures to protect the information we collect on the Scratch website. We strictly limit individual access to the Scratch servers and the data we store on them. However, as effective as these measures are, no security system is impenetrable. We cannot completely guarantee the security of our database, nor can we guarantee that the information you supply will not be intercepted while being transmitted to us over the Internet.",
+  "privacypolicy.policychangesheader": "Notifications of Changes to the Privacy Policy",
+  "privacypolicy.policychangesbody": "We review our security measures and Privacy Policy on a periodic basis, and we may modify our policies as appropriate. We may also change or update our Privacy Policy if we add new services or features. If we make any changes to our privacy practices, we will amend this Privacy Policy accordingly and post the amended policy on the Scratch website. We encourage you to review our Privacy Policy on a regular basis.",
```
rschamp: I think the issue with travis is this trailing comma. In JSON, the final element may not end with a comma.

TheGrits: Good call, that makes sense with the error message.

rschamp:
@TheGrits This looks great, thank you! Sorry to always do this to you, but would you mind rebasing this on develop and removing `privacypolicy.scss`? There are no conflicts, but we just merged a commit that removes `_typography.scss`, so the page will not work when we merge it.

rschamp:
One other thing. For legal reasons we can't translate the text on legal notices like this privacy policy and our terms of service. So, would you please remove the localizations you added as well? Thank you for being thorough with that though!

TheGrits:
@rschamp Don't worry about it, I'll get to it now.

TheGrits:
Moved to #477

PR #473, opened 2016-05-12
Lint JSON and all of our .js and .scss

### LINKED ISSUE #472: Lint JSON in lint step
Invalid JSON can cause unexpected errors, so catch invalid JSON in the lint step rather than having what uses the JSON crash at runtime.

rschamp:
This adds a JSON linter to catch JSON errors before trying to build or run.

It also adds directory rules to eslint and sass-lint, so that when we add new directories or new depths of directories, we don't need to add them to the lint step.

When I added those rules, the linter found errors in un-linted files (including a bug!), which I fixed.

Fixes #472

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,4 @@
+node_modules/*
+static/*
+build/*
+intl/*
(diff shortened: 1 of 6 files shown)


rschamp:
Travis apparently has stopped installing devDependencies. I filed a support ticket.

PR #475, opened 2016-05-12
Fix GH-[441, 440]: Get rid of typography file, use `render` in conference

### LINKED ISSUE #440: Use lib.render for all views
From #438:

The conference views don't use lib.render because it doesn't use intl. This causes it to be built differently than the rest of the views since it needs to import `main.scss` itself. IMO, ideally all the views should all function the same way following the principle of least surprise.

If in `lib.render`, `window._messages` falls back to `{}`, pages without intl packages can still be rendered successfully.

Then the import of `main.scss` can be taken out of the conference `Page` component, and the conference views can use `lib.render` in order to use `main.scss` styles.

### LINKED ISSUE #441: Merge/import _typography.scss with main.scss
From #438 

Should the `_typography.scss` file be merged into `main.scss`? Right now some pages import `_typography.scss` to get styles that apply to `#view`, but some pages don't — this could become messy as more views are added. However there may be some unintended consequences that we'll need to deal with if we apply the styles in `_typography.scss` globally.

mewtaylor:
This PR covers a few ergonomics issues:
1. Merge typography into main (fixes #441)
2. Get rid of use of `#view` in scss 
3. Make `render` work with conference pages (fixes #440)

The consequences of this involve some slight typography changes in places on the homepage, and the about page (but it's now uniform with conference, and with hoc/credits).

### Test Cases ###
* Formatting should be essentially the same (outside of some slight type differences) on:
  * homepage
  * `/about`
  * `/info/credits`
  * `/info/cards`
  * `/hoc`
  * `/jobs`
  * `/dmca`
* Formatting for the conference pages should not be different at all


src/components/intro/intro.scss
```diff
@@ -1,5 +1,12 @@
 @import "../../colors";
 
+h1 {
+    line-height: 2.125rem;
+    color: $ui-orange;
+    font-size: 1.625rem;
+    font-weight: 400;
+}
+
```
rschamp: This should be within the `.intro` block so it won't affect `h1`'s outside of the component.

rschamp:
🙌  This is awesome. The only adjustment I see would be to make sure all components get a unique className, and then wrap all their styles in that, so it's scoped to the component. Otherwise whenever that component is present, the unscoped styles will override other styles on the page. I think this should apply to views as well for consistency, but also since it could potentially affect the navigation and footer styles if plain element selectors are in the view scss.

mewtaylor:
👍 on the unique class name – I did that in conference but forgot to add it to the other stuff here, that should be a quick add.

src/views/conference/expect/expect.scss
```diff
@@ -1,177 +1,174 @@
 @import "../../../colors";
 @import "../../../frameless";
-@import "../../../typography";
 
-#view {
```
rschamp: Missing the `.expect` scope here

src/views/conference/index/index.scss
```diff
@@ -1,74 +1,78 @@
...
 #view {
     background-color: $ui-light-gray;
     min-height: initial;
+}
 
-    .title-banner {
-        margin-bottom: 0;
-        background-image: url("/images/conference/index/title-banner.jpg");
-        padding: 48px 0;
+.title-banner {
```
rschamp: Should get wrapped within `.index` below.

src/views/conference/plan/plan.scss
```diff
@@ -1,113 +1,111 @@
 @import "../../../colors";
 @import "../../../frameless";
 
-#view {
```
rschamp: Needs `.plan` scope

src/views/splash/splash.jsx
```diff
@@ -4,7 +4,7 @@ var omit = require('lodash.omit');
 var React = require('react');
 var render = require('../../lib/render.jsx');
 
-var actions = require('../../redux/actions.js');
+var authActions = require('../../redux/auth.js');
```
rschamp: Doesn't exist, but you knew that.

rschamp:
Looking really good, almost there!

src/views/splash/splash.jsx
mewtaylor: Whoops, thought I removed that from the commit like the intro one. Thanks!

rschamp:
Nice, thanks so much for doing this!

rschamp:
@carljbowman Adding you to this so you can do a once-over of staging (after staging is working) and let us know of anything that should be fixed before release.

src/components/footer/www/footer.scss
```diff
@@ -18,12 +18,16 @@
         dt {
             display: block;
             margin-bottom: 8px;
-            font-weight: bold;
         }
 
         dd {
             display: block;
             margin: 5px 0;
+            line-height: 1.2rem;
```
carljbowman: We should update the `#footer .list .dl` font-size. Things look a bit large.

screen shot 2016-05-27 at 11 40 10 am

I propose `font-size: .8rem`
screen shot 2016-05-27 at 11 40 58 am

src/components/navigation/container/navigation.scss
```diff
@@ -72,7 +73,7 @@
     .link {
         > a {
             display: block;
-            padding: 17px 15px 0 15px;
+            padding: 15px 15px 2px 15px;
```
carljbowman: Text looks miss-aligned vertically. Can we update to `padding: 13px 15px 3px 15px`

PR #476, opened 2016-05-13
Improve performance of Travis builds

rschamp:
Use the newly published to npm version of po2icu, which uses i18next-conv 2.2.5 which uses iconv-lite rather than iconv.  This avoids the node-gyp rebuild step.  Because of that, we can remove the gcc add-on step.

Also cache the pip requirements so they aren't installed each time.


rschamp:
@mewtaylor I am going to merge this since Travis is broken right now without it.

PR #479, opened 2016-05-16
Move externals, polyfill and init to common chunk

rschamp:
This should allow us to define and update our dependencies more easily (via npm) as well as reducing the number of requests.

Additionally, when we add common packages this will avoid the issue after deploys where a user may get a cached HTML page, but new Javascript, leading to the dreaded blank page because the browser didn't download all dependencies.

Using mangle with UglifyJS also made the total file size for the common chunk bundle smaller than the current total for all of our external/polyfill scripts (1.18MB vs 1.55MB).

@thisandagain please explain ;) if this is a bad move, since you set up the method using externals and removed the packages from our dependencies.  I got a peerDependency error when I installed a package and realized there was no easy way to figure out what version we were on for the external packages — probably the most important versions to be aware of.


package.json
```diff
@@ -58,11 +58,15 @@
     "pako": "0.2.8",
     "po2icu": "0.0.2",
     "postcss-loader": "0.8.2",
-    "react-addons-test-utils": "0.14.7",
```
rschamp: Not used anymore

webpack.config.js
```diff
@@ -97,10 +105,15 @@ module.exports = {
             {from: 'intl', to: 'js'}
         ]),
         new webpack.optimize.UglifyJsPlugin({
+            mangle: true,
             compress: {
                 warnings: false
             }
         }),
+        new webpack.DefinePlugin({
+            'process.env.NODE_ENV': '"production"'
```
rschamp: This quiets Redux

webpack.config.js
```diff
@@ -97,10 +105,15 @@ module.exports = {
             {from: 'intl', to: 'js'}
         ]),
         new webpack.optimize.UglifyJsPlugin({
+            mangle: true,
             compress: {
                 warnings: false
             }
         }),
+        new webpack.DefinePlugin({
+            'process.env.NODE_ENV': '"' + process.env.NODE_ENV + '"'
```
rschamp: Switches Redux and React in/out of production mode.  Redux reports a warning if you run it minified but with NODE_ENV !== "production", which is expected since we run minified code in development.

webpack.config.js
```diff
@@ -97,10 +105,15 @@ module.exports = {
             {from: 'intl', to: 'js'}
         ]),
         new webpack.optimize.UglifyJsPlugin({
+            mangle: true,
```
mewtaylor: this is true by default – is this here to be explicit?

thisandagain:
This looks great. Thanks @rschamp. The only reason we used `externals` initially was because webpack support for code splitting was still pretty immature.

webpack.config.js
rschamp: Oh, thanks. I didn't realize it was on by default because when I added this it reduced the file size significantly.  Apparently I did something else at the same time, so I'll double check and remove it if it's unnecessary.

mewtaylor: oh interesting! Keeping it is fine with me either way – even though it's true by default I like that it's explicit, in case it leads to any issues in the future.

mewtaylor:
This looks great to me too! Just for reference, since we talked about IRL: The `common` bundle is now 1.18MB, as compared to 1.58MB worth of code when it was split up before, correct?

rschamp:
@mewtaylor It mysteriously changed to 1.23MB since I last checked without changing anything, but yes. I added up the files that currently constitute the common bundle and they are 1.58MB total.

src/lib/render.jsx
```diff
@@ -1,5 +1,7 @@
 var redux = require('redux');
 var thunk = require('redux-thunk').default;
+// JSX syntax transforms to React.createElement
+var React = require('react'); // eslint-disable-line
```
rschamp: When the JSX below (`<StoreProvider store={store}>...`) gets transpiled into regular JS, the JSX syntax  `<Component .../>` becomes `React.createElement(Component ...)`, so even though we don't explicitly use React in this file, it's required due to the JSX syntax. There's no way currently to tell eslint that JSX syntax == React requirement (see https://github.com/eslint/eslint/issues/1534), so I needed to disable linting on this line.  Previously since React was an external, we didn't actually need to ever require react to use it, except for the fact that eslint would complain.

PR #480, opened 2016-05-17
fix chinese translation

laiqun:
language selection box have zh-cn and zh-tw ,but in react-intl/lib/locale-data/ don't have zh-cn.js and zh-tw.js ,so change 'zh-cn' and 'zh-tw' to 'zh'

### DIFF
--- a/static/js/lib/react-intl-with-locales.js
+++ b/static/js/lib/react-intl-with-locales.js
@@ -2879,7 +2879,10 @@ var IntlProvider = (function (_Component) {
             config[name] = _this2.props[name];
             return config;
         }, {});
-
+        if(config.locale === 'zh-cn'||config.locale ==='zh-tw')
+        {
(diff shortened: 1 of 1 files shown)


mewtaylor:
Thanks for the pr @laiqun ! However, since `react-intl-with-locales.js` is an external dependency, it would be best to not make changes to that file ourselves. We currently solve this by loading in [custom language configs](https://github.com/LLK/scratch-www/blob/develop/custom-locales.json) for those languages [here](https://github.com/LLK/scratch-www/blob/develop/src/lib/intl.jsx), which follows react-intl's system of language registration too. This is similar to how `zh-Hans` is added in `intl-messageformat` itself in the [source](https://github.com/yahoo/intl-messageformat/blob/master/dist/locale-data/zh.js#L2).

laiqun:
Thanks,Honey

PR #481, opened 2016-05-17
Fix Spacing Issues on Privacy Policy

TheGrits:
screenshot

I noticed on the new [privacy policy page](https://staging.scratch.mit.edu/privacy_policy/) there were some spacing issues beside links. (Some links didn't have a space between a word before or after)

Also, I noticed that bolding doesn't seem to work on the staging site. The `h3`s aren't bolded (very well if at all), and neither are anything that I put in `<b>` tags. Is this just an issue on the staging site or will changes need to be made for the page?

### DIFF
--- a/src/views/privacypolicy/privacypolicy.jsx
+++ b/src/views/privacypolicy/privacypolicy.jsx
@@ -18,7 +18,7 @@ var Privacypolicy = React.createClass({
                          to explain what information we collect, how we use it,
                          and what we're doing to keep it safe. If you have any
                          questions regarding this privacy policy, you can
-                         <a href="/contact-us">contact us</a>.
+                         <a href="/contact-us"> contact us</a>.
                     </p>
                     <p><i>
                         Please do not share personal contact information, such as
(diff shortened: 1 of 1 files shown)


TheGrits:

rschamp:
Nice catch @TheGrits!

It looks like you figured this out but to be clear, this is due to the way React deals with tags within inline text. It wraps all text within `<span>` elements and strips whitespace between tags, so `my text <a>my link</a>` becomes `<span>my text</span><a>my link</a>`.

The idiomatic way to deal with this issue is to add `{' '}` between text and links, so would you update this with that method?

Not sure what's going on with the `<b>` tags, but will let you know what I find out about that.

TheGrits:
@rschamp I'm not exactly sure what you mean. Do you mean literally add `{' '}`? And where? (Within the span tag? After the closing span tag?)

rschamp:
Yes please literally add `{' '}` after the text immediately before the `<a...`, e.g., 
```diff
 -                          questions regarding this privacy policy, you can
 +                          questions regarding this privacy policy, you can{' '}
                           <a href="/contact-us">contact us</a>.
  ```

The span tags are invisible to the developer and added in the transformation from JSX to DOM elements.

TheGrits:
@rschamp Sounds good, thanks for clarifying!

rschamp:
Thanks!

TheGrits:
@rschamp No problem!

PR #488, opened 2016-05-17
[Privacy Policy] One space more

TheGrits:
I can't believe I missed this one earlier. \*facepalm\* Sorry about that.
screenshot

Also, I'm thinking about solutions for the non-bolding problem. Perhaps we should increase the font-weight for the `<h3>`s?

(And yes, I double-checked to make sure there were no more this time)


TheGrits:
@rschamp

rschamp:
Thanks!

Issue #484, opened 2016-05-17
Minor cosmetic issue (spacing) on new developer page


jwzimmer:
@rschamp These two words are missing a space, in Design for Tinkerability > "To support this style of interaction, we design our **interfacesto** encourage quick experimentation and rapid cycles of iteration."

https://scratch.mit.edu/developers

Chrome, 50.0.2661.94 (64-bit)
<img width="745" alt="screen shot 2016-05-17 at 12 25 49 pm" src="https://cloud.githubusercontent.com/assets/1320456/15330714/085abd28-1c2c-11e6-8a2b-fe8e495e3a0d.png">

TheGrits:
@jwzimmer @technoboy10 already made a PR that fixes this in #483

jwzimmer:
Reopening since the PR #483 has not actually been merged yet, will close after PR is merged. : )

PR #489, opened 2016-05-17
Hotfix: Resolve gh-484 (cosmetic issue in developers page)

technoboy10:
Changes "interfacesto" to "interfaces to".

@mewtaylor


mewtaylor:
Thanks @technoboy10 !

@rschamp looks like there's something here which is in master and not develop – or is it just the fact that there's a cherry pick?

rschamp:
Yes, that's a cherry pick for an issue that began when I updated .travis.yml to fix S3. It hadn't been installing devDependencies, but we didn't see that until I cleared caches on Travis, so that's expected.

This PR is only failing because #473 clashed with #476, so I'm going to merge it. It's one of those days :(

PR #491, opened 2016-05-18
Fix fallout from common chunk refactor

rschamp:
I didn't realize what I was walking into... Fixes these things:
 * Raven is not available globally so can't be accessed from the `<script>` in `template.html`, so require Raven in `init.js` and set it up there.
 * Polyfills shouldn't be bundled since they need to update the global environment. Put them back the way they were.
 * The npm installed version of react-intl needs locale data to be added on demand, so do that in `intl.jsx`.


mewtaylor:
This is awesome – well done tracking down those issues! They were not easy...

PR #496, opened 2016-05-18
Use `routes.json` to get localization paths

mewtaylor:
Before we were using glob, which was about to start failing on subdirectories in views (which we started using in `conference`). Instead of searching for `l10n.json`, it seemed more appropriate to instead look for localization by using the configured pages that need to be localized.

### Test Cases ###
* run `make translations && npm start`. Check the website in pt_BR – it should be localised.

### DIFF
--- a/bin/build-locales
+++ b/bin/build-locales
@@ -66,7 +66,8 @@ try {
 
 // get global locale strings first.
 var globalTemplateFile = path.resolve(__dirname, '../src/l10n.json');
-var ids = JSON.parse(fs.readFileSync(globalTemplateFile, 'utf8'));
+var ids = require(globalTemplateFile);
+// var ids = JSON.parse(fs.readFileSync(globalTemplateFile, 'utf8'));
 
 // message key with english string values (i.e. default values)
 var viewLocales = {
@@ -76,7 +77,6 @@ var viewLocales = {
 };
 var idsWithICU = localeCompare.idToICUMap('general', ids);
 var icuWithIds = localeCompare.icuToIdMap('general', ids);
-// localeCompare.getIdsForView('general', globalTemplateFile, viewLocales, idsWithICU, icuWithIds);
 
 var views = [];
 var localizedAssetUrls = {};
@@ -92,15 +92,17 @@ for (var v in routes) {
         var subdir = routes[v].view.split('/');
         subdir.pop();
         var l10n = path.resolve(__dirname, '../src/views/' + subdir.join('/') + '/l10n.json');
-        fs.accessSync(l10n, fs.F_OK);
-        ids = JSON.parse(fs.readFileSync(l10n, 'utf8'));
+        ids = require(l10n);
         viewLocales[routes[v].name] = {
             en: ids
         };
         idsWithICU = merge(idsWithICU, localeCompare.idToICUMap(routes[v].name, ids));
         icuWithIds = merge(icuWithIds, localeCompare.icuToIdMap(routes[v].name, ids));
     } catch (err) {
-        // check that routes' view path is not malformed, error if it is
+        if (err.code !== 'MODULE_NOT_FOUND') {
+            throw err;
+        }
+
         try {
             fs.accessSync(path.resolve(__dirname, '../src/views/' + routes[v].view + '.jsx'));
         } catch (err) {
@@ -111,11 +113,12 @@ for (var v in routes) {
 
     // get asset url translations
     try {
-        var l10nStatic = path.resolve(__dirname, '../src/views/' + routes[v].view + '/l10n-static.json');
-        fs.accessSync(l10nStatic, fs.F_OK);
+        subdir = routes[v].view.split('/');
+        subdir.pop();
+        var l10nStatic = path.resolve(__dirname, '../src/views/' + subdir.join('/') + '/l10n-static.json');
         localizedAssetUrls[routes[v].name] = {};
 
-        var assetUrls = JSON.parse(fs.readFileSync(l10nStatic, 'utf8'));
+        var assetUrls = require(l10nStatic);
         for (var lang in localizedUrls) {
             localizedAssetUrls[routes[v].name][lang] = {};
             for (var key in assetUrls) {
@@ -127,7 +130,16 @@ for (var v in routes) {
             }
         }
     } catch (err) {
-        // :)
+        if (err.code !== 'MODULE_NOT_FOUND') {
+            throw err;
+        }
+
+        try {
+            fs.accessSync(path.resolve(__dirname, '../src/views/' + routes[v].view + '.jsx'));
+        } catch (err) {
+            // the config for the view is not set up correctly, so throw the error.
+            throw err;
+        }
     }
 }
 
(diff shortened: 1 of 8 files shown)


bin/build-locales
```diff
@@ -64,66 +64,82 @@ try {
...
-// Sample structure: { 'general-general.blah': 'blah', 'about-about.blah': 'blahblah' }
-var idsWithICU = {};
-// reverse (i.e. english string with message key as the value) object for searching po files.
-// Sample structure: { 'blah': 'general-general.blah', 'blahblah': 'about-about.blah' }
-var icuWithIds = {};
-
 // get global locale strings first.
 var globalTemplateFile = path.resolve(__dirname, '../src/l10n.json');
-localeCompare.getIdsForView('general', globalTemplateFile, viewLocales, idsWithICU, icuWithIds);
+var ids = JSON.parse(fs.readFileSync(globalTemplateFile, 'utf8'));
```
rschamp: Can we just require this like routes?

bin/build-locales
```diff
@@ -64,66 +64,82 @@ try {
...
 
+// message key with english string values (i.e. default values)
+var viewLocales = {
+    general: {
+        en: ids
+    }
+};
+var idsWithICU = localeCompare.idToICUMap('general', ids);
+var icuWithIds = localeCompare.icuToIdMap('general', ids);
+// localeCompare.getIdsForView('general', globalTemplateFile, viewLocales, idsWithICU, icuWithIds);
```
rschamp: Remove commented out code

bin/build-locales
mewtaylor: 👍

bin/build-locales
mewtaylor: ah whoops – thanks!

bin/build-locales
```diff
@@ -64,66 +64,82 @@ try {
...
+            for (var key in assetUrls) {
+                if (localizedUrls[lang].hasOwnProperty(key)) {
+                    localizedAssetUrls[routes[v].name][lang][key] = localizedUrls[lang][key];
+                } else {
+                    localizedAssetUrls[routes[v].name][lang][key] = assetUrls[key];
+                }
             }
         }
+    } catch (err) {
+        // :)
```
rschamp: Why is this error silenced?

mewtaylor: Ah whoops – forgot to add to this what I did to the other `l10n` file above. Thanks!

bin/build-locales
```diff
@@ -64,66 +64,82 @@ try {
...
+        };
+        idsWithICU = merge(idsWithICU, localeCompare.idToICUMap(routes[v].name, ids));
+        icuWithIds = merge(icuWithIds, localeCompare.icuToIdMap(routes[v].name, ids));
+    } catch (err) {
+        // check that routes' view path is not malformed, error if it is
+        try {
+            fs.accessSync(path.resolve(__dirname, '../src/views/' + routes[v].view + '.jsx'));
+        } catch (err) {
+            // the config for the view is not set up correctly, so throw the error.
+            throw err;
```
rschamp: Is this handled this way to silence errors for nonexistent l10n.json files? If so, I think it'd be more appropriate to explicitly check if the file exists so that anything else that goes wrong will get thrown and we'll see it.

mewtaylor: I don't think I understand – isn't that what this does? It gets the l10n file if it exists, and if it doesnt, then checks to see if the file for the view exists. If the view doesn't, it knows that there's a problem – if it does, then the assumption is that the view is not localised (i.e. conference pages, privacy policy, etc.). Unless there's another case I'm not accounting for...

rschamp: It does determine if the l10n.json or view path exist, but the issue is that it's swallowing any error that isn't the one you're looking for. So if anything throws an error you don't expect — the errors you really want to know about — it silences them. That's a recipe for bugs that are hard to debug because the script will carry on like nothing happened but the results won't be what you expect.

So I think you should literally check if the file exists, and then decide to continue or not, not in a try...catch, so that any errors you don't expect get thrown and we'll know when something goes wrong.

mewtaylor: ah right gotcha. I think that the way `accessSync` works is by raising an exception, as does doing a `require`, and those are the two recommended ways for checking if a file exists in node. So, I think that I'm going to check the error code here for `ENOENT`, which wouldn't be thrown by anything else in this sequence.

rschamp: I think that makes sense, or I think you could also just limit the scope of your try{} block to just the accessSync line. Then skip over the rest of the code via an if statement or continue or something.

rschamp:
As far as the approach goes, I think building the file based on the views referenced in routes makes sense since we don't include l10n.json anywhere but the views.

LGTM other than the error handling.

bin/build-locales
```diff
@@ -64,66 +64,94 @@ try {
...
 // Sample structure: { 'sdfas43534sdfasdf': 'general-general.blah', 'lkjfasdf4t342asdfa': 'about-about.blah' }
 var md5WithIds = localeCompare.getMD5Map(icuWithIds);
 
 // Get ui localization strings first
 var isoCodes = Object.keys(languages);
-for (i in isoCodes) {
+for (var i in isoCodes) {
     var translations = localeCompare.getTranslationsForLanguage(isoCodes[i], idsWithICU, md5WithIds);
-    for (var key in translations) {
+    for (key in translations) {
```
rschamp: Is this to avoid a `no-redeclare` linter error? In that case I would keep the `var` and just call `key` something distinct and more descriptive. Same with `i` above, if these were called `isoCode` and `messageId`, I think it would help some of the complexity of this script if these names were more descriptive. This code is a little hard to follow if you don't know what it does already since it's transforming and consolidating a bunch of different types of objects. If the loops that operated on these described the data they were working with, I think it would help make it more understandable, and would keep you from needing to scrolling around to check the comments for what the structure of the objects are.

bin/build-locales
```diff
@@ -64,66 +64,94 @@ try {
...
+    } catch (err) {
+        if (err.code !== 'MODULE_NOT_FOUND') {
+            throw err;
+        }
+
+        try {
+            fs.accessSync(path.resolve(__dirname, '../src/views/' + routes[v].view + '.jsx'));
+        } catch (err) {
+            // the config for the view is not set up correctly, so throw the error.
+            throw err;
```
rschamp: This error would be more helpful if you threw a new one in this case that said that the config for the view isn't set up correctly like you say in the comment, like `throw new Error("View " + routes[v].view + " in route " + routes[v].name + " does not exist")`.

bin/build-locales
mewtaylor: 👍

bin/build-locales
mewtaylor: The error is going to say something pretty close to that, since it's an `access` error. It'll say something along the lines of `ENOENT`, and about the file not existing. My thinking behind rethrowing it rather than delcaring a new error was that it'd be more helpful to have the full path to file in the stack trace, rather than just the view name, and relative path from `./src/views`.

PR #497, opened 2016-05-19
Feature: conference schedule page

mewtaylor:
Adds the conference schedule and details pages. Uses similar logic to @rthorn 's for handling url patterns (which can be updated to use `react-router` if/when we decide to do that). This also re-arranges the redux subdirectory slightly.

@rschamp for the redux structure, I went with something I'd seen before where actions and reducers that are related live together in one file, and the reducers are all imported in the centralized `reducer.js` file. I also am aware that redux will often suggest separate subdirectories for `actions` and `reducers`, but this seemed to be sufficient for the amount of reducers we have currently. Let me know what you think, though!

I also added to the fastly config a conversion from expressJS patterns to fastly patterns – I think it's the right place, but am not sure.

### Test Cases ###
* Login should still work
* Splash pages should still detect if you're logged in
* The `/conference/schedule` page should include tabs for different days of the conference
* The `/conference/:id/details` page should show info on conference sessions


src/redux/reducer.js
```diff
@@ -1,43 +1,15 @@
...
+var conferenceDetailsReducers = require('./conference-details.js');
 
 var appReducer = combineReducers({
-    session: sessionReducer,
-    token: tokenReducer
+    session: authReducers.sessionReducer,
+    token: authReducers.tokenReducer,
+    day: conferenceScheduleReducers.dayReducer,
+    schedule: conferenceScheduleReducers.scheduleReducer,
+    details: conferenceDetailsReducers.detailsReducer
```
rschamp: I like the organization changes you made for redux.  I think we could benefit from following the [ducks](https://medium.com/@scbarrus/the-ducks-file-structure-for-redux-d63c41b7035c#.j91k3mtyl) method more, and isolate each module to one reducer per file.  That way when you combine reducers here, the state can just be the name of the reducer, so it's clear what part of the state comes from where.  Right now `day`, `schedule`, and `details` are ambiguous.

So I suggest breaking up authReducers into token and session, and conferenceScheduleReducers int conferenceDay and conferenceSchedule.  Since in ducks, they recommend you export the reducer, and put the actions as properties of the reducer, for example:
```javascript
var TYPES = keyMirror({
//...
});
var reducer = function (state, action) {
//...
}
reducer.fooAction = function () {/* ... */}
reducer.thunkifiedAction = function () {
    return function (dispatch) {
        /* ... */
        // This needs to dispatch another action
        return dispatch(reducer.fooAction());
    }
}
// ...
module.exports = reducer;
```

and then `combineReducers` would be a simple map:
```javascript
var appReducer = combineReducers({
    session: session,
    token: token,
    conferenceDay: conferenceDay,
    conferenceSchedule: conferenceSchedule,
    conferenceDetails: conferenceDetails
});

// or in ES6
var appReducer = combineReducers({
    token, session, conferenceDay, conferenceSchedule, conferenceDetails
});
```

I think this would keep things more organized since then the state is inherently namespaced, and the name of the state implies the name of the module, so that's where you would go to get actions, or to reference for how the reducer uses those actions.

...

Alternatively I would just suggest being more specific in your naming of `day`, `schedule`, and `details` here :)

src/views/conference/schedule/schedule.jsx
```diff
@@ -0,0 +1,122 @@
...
+            </div>
+        );
+    }
+});
+
+var mapStateToProps = function (state) {
+    return {
+        day: state.day,
+        schedule: state.schedule,
+        fetching: state.fetching,
```
rschamp: I think this was leftover from a previous version of the code. You never use `props.fetching` and `state.fetching` doesn't exist either.

src/views/conference/details/details.jsx
```diff
@@ -0,0 +1,88 @@
...
+            </div>
+        );
+    }
+});
+
+var mapStateToProps = function (state) {
+    return {
+        details: state.details,
+        fetching: state.fetching,
+        error: state.error
```
rschamp: Same here. I think both `fetching` and `error` are properties of `details`?

src/views/conference/schedule/schedule.jsx
mewtaylor: ah yes – thanks!

src/components/navigation/www/navigation.jsx
```diff
@@ -5,7 +5,7 @@ var ReactIntl = require('react-intl');
 var FormattedMessage = ReactIntl.FormattedMessage;
 var injectIntl = ReactIntl.injectIntl;
 
-var actions = require('../../../redux/actions.js');
+var actions = require('../../../redux/auth.js');
```
rschamp: For consistency elsewhere, `authActions` here? Also will make it clear which actions those are when they're used in the file, since we have different types of actions now.

mewtaylor: yeah – I think that was a merge conflict that I did not handle fully. thanks!

src/redux/reducer.js
mewtaylor: 1 reducer per file makes sense for now – I think that it could become unwieldly if we start to have a considerable number of actions that are unique to specific components, but I also guess the idea is that actions are not tightly coupled to components, and can be used across components, right?

I'll split it out further into the separate reducer files as part of this :)

src/redux/conference-details.js
```diff
@@ -0,0 +1,80 @@
...
+    if (typeof state === 'undefined') {
+        state = {};
+    }
+    switch (action.type) {
+    case Types.SET_DETAILS:
+        return action.details;
+    case Types.SET_DETAILS_FETCHING:
+        return {fetching: action.fetching};
+    case Types.SET_DETAILS_ERROR:
+        return {error: action.error};
```
rschamp: Won't this obliterate everything in the state, and replace it with `{error: action.error}`? E.g., if your state is `{... whateverMakesDetails, fetching: false, error: false}`, that will all just become `{error: action.error}` or `{fetching: action.fetching}`?

src/redux/reducer.js
rschamp: Yeah exactly. I would much rather have a bunch of focused files that are small and easy to understand and test(?!), than having it consolidated into huge files.  Having it more modular would also make it possible to only include the reducers a particular view needs in the app reducer.

src/redux/conference-details.js
mewtaylor: This will replace everything in `this.props.details` with the details, `fetching`, or `error` – which is intentional, since that prop should not contain multiple of those states at any point in time.

src/redux/conference-schedule.js
```diff
@@ -0,0 +1,159 @@
...
+                });
+                var schedule = [];
+                for (var i = 0; i < scheduleByChunk.info.length; i++) {
+                    schedule.push({
+                        info: scheduleByChunk.info[i],
+                        items: scheduleByChunk.chunks[scheduleByChunk.info[i].name]
+                    });
+                }
+                dispatch(module.exports.setDay(day));
+                dispatch(module.exports.setSchedule(schedule));
```
rschamp: Since these are so intertwined, I think you could actually leave this as a single reducer, and have day and the schedule  details be two parts of the same state.

src/redux/conference-schedule.js
```diff
@@ -0,0 +1,159 @@
...
+ *  @return {Object}     Schedule for the day, broken into chunks
+ */
+module.exports.getDaySchedule = function (day) {
+    return function (dispatch) {
+        api({
+            uri: '/conference/schedule/' + day
+        }, function (err, body) {
+            if (err) {
+                dispatch(module.exports.setDayError(err));
+                dispatch(module.exports.setScheduleError(err));
```
rschamp: And if you combine them into one state, then this can just be one error, since they are both always set at the same time.

src/redux/conference-details.js
```diff
@@ -0,0 +1,80 @@
...
+    return function (dispatch) {
+        api({
+            uri: '/conference/' + id + '/details'
+        }, function (err, body) {
+            if (err) {
+                dispatch(module.exports.setDetailsError(err));
+                return;
+            }
+
+            if (typeof body !== 'undefined') {
```
rschamp: What does it mean if body is undefined? Should that be a `Not Found` error as well?

mewtaylor: I was trying to copy the structure that we had set up in `refreshSession` for consistency – I can add in an error dispatcher for this though

rschamp: Oh I see. Yeah maybe update both so the pattern is `if (err || typeof body === 'undefined') { // set error to err || 'Empty response'}` and then we can continue knowing there's no error and the body is there?

mewtaylor: I put it at the end already since it seemed like it might be easier (so there's no additional conditional check for if it's `err` or not inside that `or`). That cool with you?

src/redux/conference-schedule.js
```diff
@@ -0,0 +1,136 @@
...
+                        }
+                    }
+                });
+                var schedule = [];
+                for (var i = 0; i < scheduleByChunk.info.length; i++) {
+                    schedule.push({
+                        info: scheduleByChunk.info[i],
+                        items: scheduleByChunk.chunks[scheduleByChunk.info[i].name]
+                    });
+                }
```
rschamp: I think this could do with some refactoring at some point. It's a lot of code so it's hard to tell what's going on here. Maybe split out the reduce and sorting functions? Also this for loop can become a map.

src/redux/conference-schedule.js
```diff
@@ -0,0 +1,136 @@
...
+        dispatch(module.exports.setScheduleFetching());
+        dispatch(module.exports.getDaySchedule(day));
+    };
+};
+
+/**
+ * Gets the schedule for the given day from the api
+ * @param  {String} day  Day of the conference (Thursday, Friday or Satrurday)
+ *
+ *  @return {Object}     Schedule for the day, broken into chunks
```
rschamp: What is a chunk?

src/redux/session.js
```diff
@@ -0,0 +1,58 @@
...
+            host: '',
+            uri: '/session/'
+        }, function (err, body) {
+            if (err) return dispatch(module.exports.setSessionError(err));
+
+            if (typeof body !== 'undefined') {
+                if (body.banned) {
+                    return window.location = body.url;
+                } else {
+                    dispatch(module.exports.getToken());
```
rschamp: This is in `token`

src/redux/conference-schedule.js
mewtaylor: chunk is a part of a day when parallel events are happening. Like a time slot. Now that I've said time slot I realize that's maybe a better naming convention.

src/redux/token.js
```diff
@@ -0,0 +1,55 @@
...
+module.exports.setToken = function (token) {
+    return {
+        type: Types.SET_TOKEN,
+        token: token
+    };
+};
+
+module.exports.setTokenError = function (error) {
+    return {
+        type: Types.SET_SESSION_ERROR,
```
rschamp: Should be `SET_TOKEN_ERROR`

src/redux/conference-schedule.js
mewtaylor: Ordering is important here. The reason it's somewhat involved is because of the formatting conversions we need to do between the tabular structure the data is stored in and the multiple levels of division that we need for the presentation (i.e. schedule for a day, for a time slot/chunk, for an actual event, etc.), since sometimes order matters and other times it doesn't. I'm happy to split out the sorting to a separate method that this calls, though

src/redux/conference-schedule.js
rschamp: Thanks for updating it, that's clearer. Don't forget to update this docstring too.

rschamp:
Nice work organizing redux! And using redux for the view state for that matter. I agree that react-router will help us out with things like tabs.

jwzimmer:
On staging, using @mewtaylor's test cases:

**Chrome Version 50.0.2661.102 (64-bit)**

- [X] Login should still work
- Sign in as mod jwzimmer1990; go to editor; still signed in; go to My Stuff; still signed in; can delete projects from My Stuff page; still signed in
- Sign in as not-a-mod a76fhd5ehjsak; go to editor; still signed in; drag block in script area; project saves; still signed in
- [ ] Splash pages should still detect if you're logged in
- [X] The /conference/schedule page should include tabs for different days of the conference
- Includes tabs for Thursday/ Friday/ Saturday
- Tabs are clickable and have different content
- [ ] The /conference/:id/details page should show info on conference sessions
- These are clickable and have content, e.g. https://staging.scratch.mit.edu/conference/16/details
- **However, there is not a details page for every session - is this intended design?**

mewtaylor:
Yes that is intended design, if I understand correctly - some items, like `breakfast` for instance, do not have any details associated with them, so they are not clickable.

jwzimmer:
Adding QA-OK since we closed #517 as having to do with fastly (not on our end).

rschamp:
Sorry @carljbowman I told you to put pr-needs work but I guess that doesn't make sense if it's already merged.  So let's use needs-qa for logging QA issues on a merged PR.

carljbowman:
Tested on MacOSX in Chrome Version 50.0.2661.102 (64-bit) & on iOS Safari 
Screens from Chrome

### Navigation Breaks on Mobile
'Schedule' should be on the same line as 'What to Expect' & 'Plan Your Visit'

screen shot 2016-05-27 at 2 48 38 pm

### Schedule Sub-Nav and Section Titles break

- 'Saturday' should be on the same line as 'Thursday' & 'Friday'. We can remove the icons if that makes it possible 
- We should make the Section titles smaller, and if they do wrap to a new line it should push content not over lay.

screen shot 2016-05-27 at 2 52 16 pm

### Schedule detail page 
Content should be no larger than 8 columns wide. We can go to 100% on breakpoints after desktop
Description should have styling of `<p>`. The line-height is too tight which affects readablity on screen

screen shot 2016-05-27 at 3 00 59 pm


### Footer extra spacing (not blocking)
Looks like the new type standards added a lot of addtional spacing on the footer, really elongating the page vertically. 

Family links are too large and too spaced and the line about 'Scratch is a project...' should use the class `.legal` which has a `font-size: .8rem`

jwzimmer:
I saw this on my iPhone 6 in Chrome:
After scrolling up and down a bit, there was a little gap between the header and the top of the screen, where I could see content that would normally be hidden by the header.

img_7288
img_7287

jwzimmer:
@mewtaylor & @carljbowman , I'm actually still seeing some weird things on my iPhone on staging. The Scratch logo and words in the header seem too close together and I still get the gap between header and top of screen. I'm not sure if @mewtaylor's latest changes are up on staging so please disregard if this stuff has been addressed already.

image
image

mewtaylor:
pushed some changes to help with some mobile issues. I'm still seeing the squished nav, but that is proving harder to fix than expected. It is pre-existing, though, so Im not sure its a blocker (though I will try to fix it when I get back next week).

jwzimmer:
@mewtaylor, I wouldn't think it's a blocker, personally.

jwzimmer:
@carljbowman and @mewtaylor, you might already know this, but it looks like the gap between the header and the top of the screen is not our fault, but a chrome mobile or iOS bug - I happened to run into it on another website on my phone. : )

jwzimmer:
@carljbowman can you look over this to verify it is acceptable, design-wise? The squished nav issue is non-blocking so I made a new case for it. As far as I know this PR is going into today's release. If it looks good to you, you can remove needs-QA label and add QA-OK label. Thank you.

carljbowman:
@jwzimmer - Awesome thanks for logging that responsive nav ('squishy nav') issue. All the core scaling and display is looks good to me. There are some addtional tweaks to mobile we can make later. Nothing blocking. 

👍

jwzimmer:
Added test cases on wiki: https://github.com/LLK/scratch-www/wiki/Testing-Scratch-www#All_Test_Cases_Conference_Site

PR #498, opened 2016-05-20
Always use regex form of route patterns

rschamp:
When talking to Fastly, don't ever use an express-style url pattern for consistency, and so we don't use : characters in condition names.

FYI @mewtaylor


mewtaylor:
Thanks @rschamp !

PR #501, opened 2016-05-20
Upload svgs with MIME type image/svg+xml

rschamp:
Unfortunately there's no way to map extensions to MIME type. The s3cmd tool guesses wrong and uploads svgs with text/html.  Upload everything but svgs first, then upload only svgs using the correct MIME type.

Fixes an issue where svg `` tags were appearing as not found.

FYI @mewtaylor

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,8 @@ webpack:
 	$(WEBPACK) --bail
 
 sync-s3:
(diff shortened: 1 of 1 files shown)


mewtaylor:
👍 thanks @rschamp !

Issue #500, opened 2016-05-20
[Staging] Footer link for Press not working


TheGrits:
Staging:
![screenshotstaging](https://cloud.githubusercontent.com/assets/12952370/15432644/23f28a18-1e7d-11e6-993e-b577b0dc6774.png)
Production:
![screenshotproduction](https://cloud.githubusercontent.com/assets/12952370/15432663/3dddc6ae-1e7d-11e6-897b-49436afcf41c.png)

mewtaylor:
Fixed by #503

PR #503, opened 2016-05-20
Fix gh-500: Footer press link

TheGrits:
The string associated to the link was (previously) non-existent. This should fix #500 

@mewtaylor

PR #504, opened 2016-05-20
Fix gh-500: Footer press link

TheGrits:
As per https://github.com/LLK/scratch-www/pull/503#issuecomment-220647078


TheGrits:
@mewtaylor

PR #510, opened 2016-05-23
Fix gh-507: Consistently change URI

### LINKED ISSUE #507: [Conference] Inconsistent location anchor on schedule
On the schedule page in the staging site, selecting different days does _not_ change the URI, but using the "back to full schedule" links on a session page _does_ change the URI. 

Either the full schedule links should link to the homepage without location anchors, or clicking on days on the main page should change the URI.

technoboy10:
Fixes #507 by adding a location hash after clicking on a different schedule day (conforms with the behavior of the 'back to full schedule' links on detail pages). 

I used window.history.replaceState to change the hash, so that moving back and forward in history drops the user into the place in the schedule where they left off, but that behavior is certainly up for discussion.

**Test Cases**
Clicking on days in the schedule should change the location hash in the address bar, and moving back and forward in browser history should load the schedule for the most recently accessed day.

### DIFF
--- a/src/views/conference/schedule/schedule.jsx
+++ b/src/views/conference/schedule/schedule.jsx
@@ -22,6 +22,7 @@ var ConferenceSchedule = React.createClass({
         this.handleScheduleChange(day);
     },
     handleScheduleChange: function (day) {
+        window.history.replaceState(history.state, '', "#" + day);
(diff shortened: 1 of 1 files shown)


carljbowman:
👍 Awesome! This is a great addition for navigating the Schedule. Thanks @technoboy10

Issue #29, opened 2015-10-07
Migrate - Terms of Use


kaschm:
https://scratch.mit.edu/terms_of_use/

Also need to add a new section. Per recent discussions with the Community and Moderator groups, please add an item 5 to the list of prohibited things in section 3.5:

5. Using alternate accounts or organizing voting groups to manipulate site statistics, such as purposely trying to get on the "What the Community is Loving/Remixing" rows of the front page, for example.

Moved from LLK/scratchr2#2736
/cc @Paddle2See

PR #512, opened 2016-05-24
Fix gh-29: Migrate Terms of Use

TheGrits:
Migrate [Terms of Use](https://scratch.mit.edu/terms_of_use) as per #29.

HTML tested locally on Chrome 50.0 on ChromeOS.

A couple things to note:
-  3.2 is nonexistent. 3.1 and 3.3 exist, but not 3.2.
- 7.2 should be updated as users can now delete their own users without emailing the ST.
- 16 says that the terms of use were last revised March 4, 2015, but at the bottom of the page it says that the last update was April 2016

A little side note, this will probably be my last PR for quite a while. I've got an internship this summer, so I'll be less active both on Scratch and here. I'll probably make a PR once and a while, but definitely not a frequently as I have been recently.

### DIFF
--- a/src/views/termsofuse/termsofuse.jsx
+++ b/src/views/termsofuse/termsofuse.jsx
@@ -37,7 +37,7 @@ var Termsofuse = React.createClass({
                     <p>
                         1.3 Scratch is open to children and adults of all ages, and
                          we ask that you keep this in mind when using the Scratch
-                         services. When you use Scratch, you agree to abide by the
+                         services. When you use Scratch, you agree to abide by the{' '}
                          <a href="/community_guidelines">Scratch Community Guidelines</a>.
                     </p>
                     <p>
--- a/src/views/termsofuse/termsofuse.jsx
+++ b/src/views/termsofuse/termsofuse.jsx
@@ -16,7 +16,7 @@ var Termsofuse = React.createClass({
                          <a href="https://scratch.mit.edu">scratch.mit.edu</a>{' '}
                          and all associated services, including but not limited
                          to the <a href="http://day.scratch.mit.edu">Scratch Day</a>{' '}
-                         and <a href="http://scratchx.org">ScratchX</a> websites 
+                         and <a href="http://scratchx.org">ScratchX</a> websites
                          (collectively "Scratch"). The Scratch Team is part of the
                          Lifelong Kindergarten Group in the Media Laboratory at the
                          Massachusetts Institute of Technology ("MIT"). Please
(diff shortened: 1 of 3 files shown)


TheGrits:

src/views/termsofuse/termsofuse.jsx
```diff
@@ -0,0 +1,514 @@
+var React = require('react');
+var render = require('../../lib/render.jsx');
+var Page = require('../../components/page/www/page.jsx');
+var Box = require('../../components/box/box.jsx');
+
+var Termsofuse = React.createClass({
```
technoboy10: Would it be better to call this 'Terms' rather than 'Termsofuse'? I'm thinking it might be a little more concise and easier to read.

src/views/termsofuse/termsofuse.jsx
```diff
@@ -0,0 +1,514 @@
...
+                         <a href="/privacy_policy">Privacy Policy</a>, which identifies
+                         how the Scratch Team uses, collects, and stores information
+                         it collects through the Services. By using Scratch, you
+                         additionally agree that you are comfortable with Scratch's
+                         Privacy Policy.
+                    </p>
+                    <p>
+                        1.3 Scratch is open to children and adults of all ages, and
+                         we ask that you keep this in mind when using the Scratch
+                         services. When you use Scratch, you agree to abide by the
```
technoboy10: There doesn't seem to be a space between 'the' and 'Scratch'; it might be better to use `{' '}` at the end of lines (see https://github.com/LLK/scratch-www/blob/develop/src/views/developers/developers.jsx).

technoboy10:
Looks good! I left you a couple of comments to look at when you get the chance. Thanks! :)

src/views/termsofuse/termsofuse.jsx
TheGrits: @technoboy10 I used that elsewhere, I just forgot it here. Thanks for pointing that out!

src/views/termsofuse/termsofuse.jsx
TheGrits: @technoboy10 I think that's a good idea. Would this require renaming the branch?

TheGrits: @technoboy10 I went ahead and did it anyway.

TheGrits:
@technoboy10 Okay, I have made changes according to your comments. Thanks! :)

technoboy10:
Thanks, merged!

PR #513, opened 2016-05-24
Fix pdf localization check

mewtaylor:
It was adding the `(English)` suffix no matter what before, when it should do so only if the cards are only available in english, and the language is not english. This fixes it by getting the locale from props, and also by making the conditional check a bit more safe by using `indexOf` instead of a complete equal check.

### Test Cases ###
* Visit `/info/cards` in English. No links should have the `(English)` suffix anymore
* Visit `/info/cards` in Spanish. The first link should not have the suffix, but the others should.

### DIFF
--- a/src/views/cards/cards.jsx
+++ b/src/views/cards/cards.jsx
@@ -11,6 +11,12 @@ require('./cards.scss');
 
 var Cards = injectIntl(React.createClass({
     type: 'Cards',
+    pdfLocaleMismatch: function (locale, pdf, englishPdf) {
+        if (pdf === englishPdf && locale.indexOf('en') !== 0) {
+            return true;
+        }
+        return false;
+    },
     render: function () {
         var locale = this.props.intl.locale || 'en';
         var formatMessage = this.props.intl.formatMessage;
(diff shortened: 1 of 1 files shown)


src/views/cards/cards.jsx
```diff
@@ -52,7 +52,7 @@ var Cards = injectIntl(React.createClass({
                                     <FormattedMessage id='cards.viewCard' />
                                     {(
                                         formattedLinks['cards.starterLink'] === englishLinks['cards.starterLink'] &&
-                                        locale !== 'en'
+                                        locale.indexOf('en') === -1
```
rschamp: Given how many times it's used, it would be good to refactor this logic out into something like a `localeIsEnglish` helper method.  It also seems more correct to check if `locale.indexOf('en') !== 0` since if it appears elsewhere, it's not English. The most correct might be `locale.split('_')[0] !== 'en'`.

mewtaylor: 👍 on generalizing. Just fyi, I switched to using `this.props.intl.locale` to avoid having to clean it like you suggest in `locale.split('_')[0] !== 'en'`, since we already to that in `render`. It doesn't appear that the value in `this.props.intl.locale` could ever be something that includes `en` otherwise. I guess switching does prevent against adding future locales that would have that though

rschamp: Yeah, was mostly thinking about it being robust for the future even though it works with the languages we have now.

mewtaylor: Cool! The changes are in, whenever you have a minute :).

rschamp: Thanks!

PR #514, opened 2016-05-25
Fix gh-307: Migrate FAQ page

technoboy10:
Migrates the FAQ page from scratchr2 to scratch-www, fixing #307.

**TODO**

- [x] Spin off the sidebar nav into its own component
- [x] Clean up scss
- [x] Potentially make new template/submodule for text-heavy pages based on this?
- [ ] Reopen against `develop` branch


src/routes.json
```diff
@@ -107,7 +113,7 @@
     },
     {
         "name": "guidelines",
-        "pattern": "^/community_guidelines$",
+        "pattern": "^/community_guidelines/$",
```
TheGrits: Whoops, that would be my mistake. Shouldn't there also be a `?` in there too though?

`^/community_guidelines/?$`
instead of
`^/community_guidelines/$`

src/views/faq/faq.jsx
```diff
@@ -0,0 +1,421 @@
...
+                                </li>
+                            </a>
+                        </SubNavigation>
+                    </div>
+                </TitleBanner>
+                <div className="inner">
+                    <section id="about-scratch">
+                        <span className="nav-spacer"></span>
+                        <h2><FormattedMessage id='faq.aboutTitle' /></h2>
+                        <FlexRow className="sidebar-row">
```
mewtaylor: I think that these might make more sense if they use the `dl/dd/dt` tag structure, since they have sort of "title" text (the question), and then "description" text (the answer). Also, since they're always going to be in columns, I think we could take out the use of `FlexRow`

mewtaylor:
@technoboy10 this looks amazing! I just left one suggestion, outside of the questions that you had for @carljbowman.

PR #477, opened 2016-05-15
Fix gh-309: Migrate Privacy Policy

TheGrits:
Rebased and updated version of #471, to resolve #309.

HTML tested locally on Chrome 50 (ChromeOS)

@rschamp @thisandagain @mewtaylor

PR #518, opened 2016-05-27
Fix Password Reset Link

TheGrits:
As per https://github.com/LLK/scratch-www/pull/477#issuecomment-222148473

Not sure how that happened as I copied and pasted most (if not all) the text.


TheGrits:
@rschamp

rschamp:
Ah, thanks @TheGrits! Please open against release/2.2.9, as this is a QA issue for the release. In the future you don't have to worry about fixing things that happen in QA, since this is part of our release process.

TheGrits:
@rschamp Sounds good, thanks.

Issue #396, opened 2016-03-25
Splash header needs a default state


mewtaylor:
Currently, once the browser loads a page, that page then makes a request to the api to see if the user who's loading the page is logged in or not. While the page is waiting for the api to make a response, the nav defaults to its signed out state – it shows the `intro` component, which is what is displayed when a user is logged out.

This is usually not noticeable, as often the api responds quickly with information about whether the user is logged in. However, this can still be confusing for users. It might briefly appear to them that they are logged out – prompting them to try and login – when in reality, the page is still trying to determine if they are or not.

We need the splash view to have a "default" state – a design for before it knows whether or not the user is logged in.

Radical-from-Scratch:
In my opinion, checking to see if logged in or not should be before even the first line of HTML. I always had a suspicion  that this would happen, and it also has a more notable difference on the site then you'd think.

thisandagain:
/cc @carljbowman 

I think there are a couple parts to this. First, I agree we need to design an explicit "empty" state prior to the navigation being rendered. In some ways, this is really a UI design issue. If we had a little more consistent UI in terms of scale / segmentation between logged-in and logged-out, the design of the empty state could look much more natural rather than adding visual complexity for a split second during page load (which would likely just make the problem worse). Without this, I think we likely have enough information about the user to "do the right thing" on subsequent page loads after we have already performed a session check. This could influence what the default state looks like as to help avoid the visual "jar" that happens today by optimizing for the 90%+ case of a user continuing to be logged in should a session cookie exist.

Radical-from-Scratch:
Ok with me

technoboy10:
Resolved by #539.

PR #539, opened 2016-06-01
Fix GH-395 and GH-396: Add initial state to splash header and navbar

### LINKED ISSUE #395: Nav bar needs an initial state
Currently, once the browser loads a page, that page then makes a request to the api to see if the user who's loading the page is logged in or not. While the page is waiting for the api to make a response, the nav defaults to its signed out state – it shows "Join Scratch", and "Sign in" components.

This is usually not noticeable, as often the api responds quickly with information about whether the user is logged in. However, this can still be confusing for users. It might briefly appear to them that they are logged out – prompting them to try and login – when in reality, the page is still trying to determine if they are or not.

We need to have the nav component display a "default" state – a design for before it knows whether or not the user is logged in.

technoboy10:
This PR adds a {'loaded':true} item to the session object when a session is successfully loaded. This allows the navbar and splash header to have different states for 1) logged in users, 2) logged out users, and 3) not fully loaded pages. 

Fixes #395 and #396.

For the design, I've replicated the scratchr2 navbar design (search field is farthest element on the right side), and omitted the welcome element or news elements from the splash page.

Labeled as needs-discussion in case the way I'm modifying the session object isn't Kosher.

### DIFF
--- a/src/components/navigation/www/navigation.jsx
+++ b/src/components/navigation/www/navigation.jsx
@@ -225,7 +225,7 @@ var Navigation = React.createClass({
                             <Input type="hidden" name="sort_by" value="datetime_shared" />
                         </form>
                     </li>
-                    {this.props.session.status == sessionActions.Status.FETCHED ? (
+                    {this.props.session.status === sessionActions.Status.FETCHED ? (
                         this.props.session.results.user ? [
                             <li className="link right messages" key="messages">
                                 <a
--- a/src/components/navigation/www/navigation.jsx
+++ b/src/components/navigation/www/navigation.jsx
@@ -44,19 +44,19 @@ var Navigation = React.createClass({
         };
     },
     componentDidMount: function () {
-        if (this.props.session.results.user) {
+        if (this.props.session.session.user) {
             this.getMessageCount();
             var intervalId = setInterval(this.getMessageCount, 120000); // check for new messages every 2 mins.
             this.setState({'messageCountIntervalId': intervalId});
         }
     },
     componentDidUpdate: function (prevProps) {
-        if (prevProps.session.results.user != this.props.session.results.user) {
+        if (prevProps.session.session.user != this.props.session.session.user) {
             this.setState({
                 'loginOpen': false,
                 'accountNavOpen': false
             });
-            if (this.props.session.results.user) {
+            if (this.props.session.session.user) {
                 this.getMessageCount();
                 var intervalId = setInterval(this.getMessageCount, 120000);
                 this.setState({'messageCountIntervalId': intervalId});
@@ -81,13 +81,13 @@ var Navigation = React.createClass({
         }
     },
     getProfileUrl: function () {
-        if (!this.props.session.results.user) return;
-        return '/users/' + this.props.session.results.user.username + '/';
+        if (!this.props.session.session.user) return;
+        return '/users/' + this.props.session.session.user.username + '/';
     },
     getMessageCount: function () {
         this.api({
             method: 'get',
-            uri: '/users/' + this.props.session.results.user.username + '/messages/count'
+            uri: '/users/' + this.props.session.session.user.username + '/messages/count'
         }, function (err, body) {
             if (err) return this.setState({'unreadMessageCount': 0});
             if (body) {
@@ -175,14 +175,14 @@ var Navigation = React.createClass({
     },
     render: function () {
         var classes = classNames({
-            'logged-in': this.props.session.results.user
+            'logged-in': this.props.session.session.user
         });
         var messageClasses = classNames({
             'message-count': true,
             'show': this.state.unreadMessageCount > 0
         });
         var formatMessage = this.props.intl.formatMessage;
-        var createLink = this.props.session.results.user ? '/projects/editor/' : '/projects/editor/?tip_bar=home';
+        var createLink = this.props.session.session.user ? '/projects/editor/' : '/projects/editor/?tip_bar=home';
         return (
             <NavigationBox className={classes}>
                 <ul>
@@ -226,7 +226,7 @@ var Navigation = React.createClass({
                         </form>
                     </li>
                     {this.props.session.status === sessionActions.Status.FETCHED ? (
-                        this.props.session.results.user ? [
+                        this.props.session.session.user ? [
                             <li className="link right messages" key="messages">
                                 <a
                                     href="/messages/"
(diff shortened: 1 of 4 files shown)


rschamp:
> Labeled as needs-discussion in case the way I'm modifying the session object isn't Kosher.

I think the ideal would be if we had a standard setup for the api calls we handle with redux, that looked like:
* Before any API call happens, the contents of the objects returned by the API call were like `{status: NOT_FETCHED, results: null}`
* When we make an API call, this becomes `{status: FETCHING, results: null}`
* Finally when results are returned `{status: FETCHED, results: [...] or {...}}`

This way we could tell the difference between empty results, and objects that aren't requested yet and gives us the ability to add more metadata about the session without inserting it into the object itself.  It also would allow us to be smarter about making API requests, because we could decide to use results we've already retrieved rather than making a new request every time a component is mounted.

So this would require some changes to the places that use the session object, since we'd want the session state to contain the fetching state in addition to the results, not just the session itself.

The other nice thing to have is a standard way of retrieving the initial state (e.g. `{status: NOT_FETCHED, results: null}` here), because we're about to add some functionality that will require persisting it locally across page loads, so having a framework for initial state that's easy to use would be very helpful.

technoboy10:
@rschamp Made the changes you suggested, but I had to use an empty object instead of null for the initial state (because existing code broke otherwise).

src/components/navigation/www/navigation.jsx
```diff
@@ -225,99 +225,101 @@ var Navigation = React.createClass({
                             <Input type="hidden" name="sort_by" value="datetime_shared" />
                         </form>
                     </li>
-                    {this.props.session.user ? [
-                        <li className="link right messages" key="messages">
-                            <a
-                                href="/messages/"
-                                title={formatMessage({id: 'general.messages'})}>
+                    {this.props.session.status == sessionActions.Status.FETCHED ? (
```
rschamp: Use `===` for comparison — it gives more predictable results when the values you're comparing are different types because it doesn't do implicit type conversion.  For example,

```
'' == '0'           // false
0 == ''             // true
0 == '0'            // true

false == 'false'    // false
false == '0'        // true

false == undefined  // false
false == null       // false
null == undefined   // true

' \t\r\n ' == 0     // true
```

technoboy10:
@rschamp Fixed.

src/redux/session.js
```diff
@@ -31,27 +41,29 @@ module.exports.setSessionError = function (error) {
...
     return {
         type: Types.SET_SESSION,
-        session: session
+        session: {'status': status,'results': results}
     };
 };
 
 module.exports.refreshSession = function () {
     return function (dispatch) {
+        dispatch(module.exports.setSession(module.exports.Status.NOT_FETCHED, {}));
```
rschamp: A few things here:
* The status at the beginning of the api call should be fetching to reflect that the api call has started but not completed. `NOT_FETCHED` should be the initial state before anything has touched the session
* I think we should split the two parts of the state and handle them separately so we can retain the session data until we have the refreshed content. E.g., at the beginning just update the status and don't change the session data, then when you've fetched it, dispatch two actions, one to update the status and one to update the session data.
* Splitting the actions to update the status and session data separately will require changing the reducer so it only updates either the session data or the status. When you do that you should not do what is most intuitive and change the state object passed into the reducer — the state object should be treated as immutable, so you should use `lodash.defaultsDeep` to return a new state object with the appropriate values. e.g. `case Types.SET_SESSION_STATUS: return defaultsDeep({}, {status: action.status}, state)`.  Sorry this is kind of overcomplicated for the task at hand, but it's part of the way redux works that isn't obvious if you're just looking at the existing code here.

technoboy10:
@rschamp Does this look better?

rschamp:
Looks awesome, thanks!

technoboy10:
**Test Cases**
1) When logged out, make sure that the "Join Scratch" and "Sign In" buttons appear on the navigation bar, as well as the "Create stories, games, and animations" header.
2) Log in and make sure that the full navigation bar shows up (messages, my stuff, profile dropdown). 
3) Reload page and ensure that the "Join Scratch" and "Sign In" buttons and "Create stories, games, and animations" header never appear while the page is loading.
4) Repeat steps 1-3 on a mobile device.

src/redux/session.js
```diff
@@ -1,21 +1,35 @@
...
 module.exports.sessionReducer = function (state, action) {
     // Reducer for handling changes to session state
     if (typeof state === 'undefined') {
-        state = {};
+        state = module.exports.getInitialState();
     }
     switch (action.type) {
     case Types.SET_SESSION:
-        return action.session;
+        return defaultsDeep({results: action.session}, state);
```
rschamp: Just noticed this — this should be updated to `{session: action.session}`, right?

technoboy10: @rschamp Absolutely - fixed now.

jwzimmer:
**Test Cases**
Chrome Version 51.0.2704.84 (64-bit)
- [x] When logged out, make sure that the "Join Scratch" and "Sign In" buttons appear on the navigation bar, as well as the "Create stories, games, and animations" header.
  - [x] Project rows should be unchanged
  - [x] Links at bottom of page + language menu should be unchanged
  - [x] Scratch video should play
  - [x] Content under "Create stories, games, and animations" should be unchanged (number of projects can be different on staging vs. production)
- [ ] Log in and make sure that the full navigation bar shows up (messages, my stuff, profile dropdown). 
- [ ] Reload page and ensure that the "Join Scratch" and "Sign In" buttons and "Create stories, games, and animations" header never appear while the page is loading.
- [ ] Repeat the above steps on a mobile device.

technoboy10:
@jwzimmer I checked on my phone (OnePlus One, same screen size as iPhone), and problems 1-3 also exist on the current Scratch site, so they probably aren't caused by this change. I couldn't reproduce problem 2, however.

jwzimmer:
@technoboy10 Okay, great (sorry, I should have remembered to check if it was the same on production).

jwzimmer:
I just want to check if there's any way I can check in the UI that these issues are fixed by this PR (like if I delete a session cookie or something):
- https://github.com/LLK/scratch-www/issues/395
- https://github.com/LLK/scratch-www/issues/396#issuecomment-209649679
Other than that, this issue has been tested. (Almost ready to QA-OK it.) : )

jwzimmer:
**Additional Test Cases**
Chrome Version 51.0.2704.84 (64-bit)

- [ ] Throttled connection using Chrome JavaScript Console/ Dev Tools > Network tab > DSL & GPRS & Good 2G & WiFi (the idea being that slowing the connection would expose the default state in the UI, since before we retrieve the user's signed in/ out status using a cookie, the default navbar and Splash header should be visible)
  - [ ] The default navbar should look like the regular navbar, but without the Sign In/ Join Scratch links (for signed out state) or the username & Messages & My Stuff icon (for signed in state), and the other elements should be spread out slightly to fill the space they would normally occupy
  - [ ] The default Splash header should not have the video or Scratch Cat & friends (for signed out state) or the What's Happening or Scratch News (for signed in state)
  - [ ] The default Splash header should have the project rows
- [ ] With the DSL throttled connection, I deleted the cookies for staging & production, and reloaded the page. The default state should still be the default, and then in the absence of a pre-existing cookie, the user should be in the signed out state.

PR #542, opened 2016-06-01
Fix GH-307: Restyle static pages

technoboy10:
This PR restyles the Terms of Use, FAQ, DMCA, and Privacy Policy pages, fixing #307.

Everything should be good to go for this, unless there's some stylistic preferences about the way I did the scss template for infopage.


technoboy10:
@rschamp Assigning you since you were assigned to the previous PR.

jwzimmer:
When this case is finished, we can reopen case https://github.com/LLK/scratch-www/issues/546 and close scratchr2 issue 3686. (Explantion: the case is just to update the FAQ with current support information, as in, what browsers and OS's we support - I opened the case in scratch-www, then closed it and opened scratchr2 3686  because I realized the FAQ are still on scratchr2... then @TheGrits helpfully pointed out that the FAQ is soon to be on www anyway, so when this case is finished I figure I will reopen #546 and close 3686.)

rschamp:
@technoboy10 Could you explain your decision to use a mixin in addition to the `InformationPage` component?

We're still trying to figure out the best way to organize our CSS, but I think right now for consistency it would be best if the `infopage` mixin was applied to the `InformationPage` component. Then  anything in an `InformationPage` should receive those styles rather than needing to apply the mixin manually to everything using the `InformationPage` component.

technoboy10:
@rschamp Sounds good! I'm honestly not sure why I didn't do that before, but oh well. :P

technoboy10:
@rschamp Moved all the styling into the InformationPage component.

src/components/informationpage/informationpage.scss
```diff
@@ -0,0 +1,98 @@
+@import "../../colors";
+@import "../../frameless";
+
+$ui-secondary: darken($ui-blue, 10%);
+
+#view {
+    padding: 0;
+}
+
+.title-banner {
```
rschamp: Please give the outer `<div>` a className like `information-page`, and then nest everything within this file (other than `#view`) within an `.information-page { ... }` block. This will keep the `InformationPage` styles specific to that component.

technoboy10:
Moved everything but the view styling into .information-page .

rschamp:
Looks great, thank you!

jwzimmer:
@technoboy10, is [this page](https://staging.scratch.mit.edu/scratch_1.4/) included under this case? (As in, will it be migrated to scratch-www as part of this PR?)

technoboy10:
@jwzimmer Not as part of this PR, no.

PR #549, opened 2016-06-03
[Develop] Hotfix: conference schedule bug fixes

mewtaylor:
1. Sort schedule using 24-hour clock to handle redundant times in am and pm
2. Sort time slots by type AND time, since some time slots have the same type.


mewtaylor:
@sleggss @rschamp just tested on mobile and desktop, and it is working as expected, so I'm going to hotfix this now.

PR #551, opened 2016-06-06
[Master] hotfix: mobile nav conference

### LINKED ISSUE #529: Squished/ Responsive nav: Conference page on mobile
Non-blocking issue found in https://github.com/LLK/scratch-www/pull/497

I saw this on my iPhone 6 in Chrome; space between icon and links should be larger:
img_7288

mewtaylor:
Also fix the spelling for Nichole Pinkard's name (apologies!). 

It appears that flexbox has an easier time formatting wrapped rows on mobile than columns. Fixes #529.

/cc @sleggss

### DIFF
--- a/src/components/navigation/conference/navigation.scss
+++ b/src/components/navigation/conference/navigation.scss
@@ -44,7 +44,7 @@
     @media only screen and (max-width: $mobile - 1) {
         .inner > ul {
             li {
-                &.right {
+                &.right {                                        
                     ul {
                         flex-flow: row nowrap;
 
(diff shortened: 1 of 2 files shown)


rschamp:
I'm confused about displaying with column vs row — wouldn't these do two drastically different things?  How are they interchangeable like this?

mewtaylor:
@rschamp `wrap` makes it so that the element goes on to the next row (i.e. display as a column) when the screen width is too small, which is how this works. The conference icon and the nav links are split into two separate elements which are displayed in flex, allowing all nav links to move to the second line when `wrap` is present.

I'm still not clear why `column` behaves differently when on desktop chrome versus mobile chrome, but this fix seems to be handling that particular edge case where `column` did not.

rschamp:
It makes sense how row works, I guess I am more confused how the column setting works in the first place.  But that's not really important at this point!

LGTM

PR #570, opened 2016-06-14
Bugfix: shorten fastly condition

mewtaylor:
* removes `/components` from routes.json, as it's a dev-only route now
* generalizes the matching pattern for the following using a new object key, `routeAlias`, to group together patterns into singular checks:
  1. match all `^/explore` to www
  2. match all `^/search` to www
  3. match all `^/conference` to www

This of course means all old search pages, explore pages and conference pages will now be inaccessible if they're not in www, including old conference pages for 2015 and 2014 ( @sleggss please let me know if this is an issue).

Currently, the output of this is now fewer than 400 characters:
`'req.url~"^/css/|^/favicon.ico|^/images/|^/js/|^/pdfs/|^/routes.json|^/svgs/|^/version.txt|^/?$|^/about/?$|^/developers/?$|^/hoc/?$|^/explore|^/search|^/info/credits/?$|^/info/faq/?$|^/info/cards/?$|^/info/communityblocks-interviews/?$|^/jobs/?$|^/wedo/?$|^/conference|^/info/donate/?|^/DMCA/?$|^/community_guidelines/?$|^/privacy_policy/?$|^/terms_of_use/?$|^/\\?|^/[^/]*.html$"'`

### DIFF
--- a/bin/configure-fastly.js
+++ b/bin/configure-fastly.js
@@ -44,9 +44,13 @@ var getStaticPaths = function (pathToStatic) {
  * the express route and a static view file associated with the route
  */
 var getViewPaths = function (routes) {
-    return routes.map(function (route) {
-        return route.pattern;
-    });
+    return routes.reduce(function (paths, route) {
+        var path = route.routeAlias || route.pattern;
+        if (paths.indexOf(path) === -1) {
+            paths.push(path);
+        }
+        return paths;
+    }, []);
 };
 
 /*
--- a/src/routes.json
+++ b/src/routes.json
@@ -26,12 +26,14 @@
     {
         "name": "explore",
         "pattern": "^/explore/:projects/:all/?$",
+        "routeAlias": "^/explore",
         "view": "explore/explore",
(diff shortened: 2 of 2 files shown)


rschamp:
If it comes to it we can save some characters down the line by grouping info under `^/info/(?:credits|faq|cards|communityblocks-interviews|donate)/?`

mewtaylor:
@rschamp that'll be a good one to keep in mind for the next round of updates :).

Also @rschamp @thisandagain I just added a negative lookahead to allow access to earlier conference page sites. Does that addition look ok?

thisandagain:
@mewtaylor We need to keep the 2014/2015 redirects?

mewtaylor:
@thisandagain yes – I chatted with @kaschm and @sleggss about that. Seems like people generally want to have records of what happened for at least a few years back. We should determine just how far back we'd be okay with our archives on the web going, though.

thisandagain:
Ok. Then I'm 👍  to land.

PR #589, opened 2016-06-21
[WIP] Restructure Splash page

technoboy10:
This PR refactors the splash page by
  - Moving all API calls to Redux actions
  - Putting splash-specific presentational components in a splash/components folder
  - Removing presentational and layout components from splash.jsx and putting them in components.

**Open Questions**
The src/redux folder is pretty large now; would it be a good idea to structure that folder further? (if there are very view-specific redux files, it might be good to put those into one folder inside src/redux)

The split between container and presentational components is a bit tricky as far as *also* having a view/component split - is the view a container component, or something different? (In the PR, I've treated the view as the most top-level container component)

In regards to presentational view-specific components, I split the globally featured, user featured, and shuffled rows into separate components since they each required slightly different logic in each. I'm not entirely sure this is the right decision, but it seems to work well at the moment.


technoboy10:
@rschamp @mewtaylor @thisandagain

technoboy10:
(The Travis error is caused by my usage of ES6 functional components, I think)

src/views/splash/components/admin.jsx
```diff
@@ -0,0 +1,44 @@
+var React = require('react');
+
+var AdminPanel = require('../../../components/adminpanel/adminpanel.jsx');
+var Button = require('../../../components/forms/button.jsx');
+
+require('./splash.scss');
+
+var SplashAdmin = (props) => {
```
rschamp: Let's avoid this syntax until we convert the entire project over to ES6.

src/views/splash/components/admin.jsx
```diff
@@ -0,0 +1,44 @@
+var React = require('react');
```
rschamp: In the case where you need React for the jsx syntax, disable the require line like so https://github.com/LLK/scratch-www/blob/633fee162aa81665e539f06cfcf46151cd2c22d6/src/lib/render.jsx#L4 — there's not a good way to let the linter know about jsx transpilation yet.

src/redux/activity.js
```diff
@@ -0,0 +1,75 @@
...
+    SET_ACTIVITY_ERROR: null
+});
+
+module.exports.Status = keyMirror({
+    FETCHED: null,
+    NOT_FETCHED: null,
+    FETCHING: null
+});
+
+module.exports.getInitialState = function () {
```
rschamp: Do we need to export this?

src/redux/activity.js
```diff
@@ -0,0 +1,75 @@
...
+    return function (dispatch) {
+        dispatch(module.exports.setStatus(module.exports.Status.FETCHING));
+        api({
+            uri: '/proxy/users/' + username + '/activity?limit=5'
+        }, function (err, body) {
+            if (err) return dispatch(module.exports.setActivityError(err));
+
+            if (typeof body !== 'undefined') {
+                dispatch(module.exports.setActivity(body));
+                dispatch(module.exports.setStatus(module.exports.Status.FETCHED));
```
rschamp: I think we use this style in other reducers, but I don't particularly like referencing `module.exports` in the module itself.  Could we update this so that the reducer is the only thing exported, and everything else is a property of the reducer?

src/views/splash/splash.jsx
```diff
@@ -359,10 +153,10 @@ var Splash = injectIntl(React.createClass({
             'teacherbanner.resourcesButton': formatMessage({id: 'teacherbanner.resourcesButton'}),
             'teacherbanner.faqButton': formatMessage({id: 'teacherbanner.faqButton'})
         };
-        if (this.state.projectCount === this.getInitialState().projectCount) {
+        if (this.props.projectCount.projectCount === countActions.getInitialState().projectCount) {
```
rschamp: Here I see why it's being exported... it's weird we were calling `getInitialState()` at all — I think it would be cleaner to let getInitialState be something that only the reducer uses, and if other things need access to values from `getInitialState`, those values should be constants used by `getInitialState` and elsewhere. E.g, add a constant to `countActions` called `DEFAULT_PROJECT_COUNT`, and use that here and in `getInitialState`.

src/views/splash/components/activity/activity.jsx
```diff
@@ -0,0 +1,81 @@
+var React = require('react');
```
rschamp: I don't think we want to move all these components so that they're linked to the view. That makes them seem owned by the view, so it will be confusing if we ever want to use them in a different view.  What was the reasoning to moving these all within the splash directory?

src/redux/handle-dismiss.js
```diff
@@ -0,0 +1,71 @@
+var keyMirror = require('keymirror');
```
rschamp: I think `handle-dismiss` is a little vague — this reducer might be better named after template-cue (until we have a better name for that, template-cue is a little vague too).

src/redux/reducer.js
```diff
@@ -5,10 +5,22 @@ var detailsReducer = require('./conference-details.js').detailsReducer;
...
 
 var appReducer = combineReducers({
     session: sessionReducer,
     token: tokenReducer,
+    rows: rowReducer,
+    activity: activityReducer,
+    news: newsReducer,
+    projectCount: countReducer,
+    handleDismiss: dismissReducer,
+    homepageCache: cacheReducer,
```
rschamp: Another case for having the reducer be the thing exported — we could put all the reducers in one directory and use webpack to glob them all together and combine them at once.

src/views/splash/components/custom.jsx
```diff
@@ -0,0 +1,111 @@
...
+        var rows = [];
+        var keys = Object.keys(this.props.featured);
+        for (var i=0; i < keys.length; i++){
+            if (this.props.featured[keys[i]]){
+                if (this.props.featured[keys[i]].length > 0) {
+                    rows.push(
+                        <Box title={
+                                formatMessage({
+                                    id: this.rowMap[keys[i]].id,
+                                    defaultMessage: this.rowMap[keys[i]].defaultMessage})}
```
rschamp: Seems like this could just be `formatMessage(this.rowMap[keys[i]])`?

src/views/splash/components/custom.jsx
```diff
@@ -0,0 +1,111 @@
...
+        custom_projects_loved_by_following: {
+            id: 'splash.projectsLovedByScratchersFollowing',
+            defaultMessage: 'Projects Loved by Scratchers I\'m Following'
+        },
+        custom_projects_in_studios_following: {
+            id: 'splash.projectsInStudiosFollowing',
+            defaultMessage: 'Projects in Studios I\'m Following'
+        }
+    },
+    componentWillReceiveProps: function (){
```
rschamp: What's going on in `componentWillReceiveProps`? This looks like it's the same as `render`?

src/views/splash/components/emailconfirmation.jsx
```diff
@@ -0,0 +1,36 @@
...
+require('./splash.scss');
+
+var ConfirmEmail = React.createClass({
+    type: 'ConfirmEmail',
+    render: function () {
+        return (
+            <div>
+                <DropdownBanner key="confirmedEmail"
+                        className="warning"
+                        onRequestDismiss={this.props.handleDismiss.bind(this.props.self, 'confirmed_email')}>
```
rschamp: What's `this.props.self`?

src/views/splash/splash.jsx
```diff
@@ -50,14 +46,12 @@ var Splash = injectIntl(React.createClass({
...
-                this.getActivity();
-                this.getFeaturedCustom();
-                this.getNews();
+                this.props.dispatch(activityActions.getActivity(this.props.session.session.user.username));
+                this.props.dispatch(rowActions.getCustom(this.props.session.session.user.id));
+                this.props.dispatch(newsActions.getNews());
             } else {
-                this.setState({featuredCustom: []});
-                this.setState({activity: []});
                 this.setState({news: []});
```
rschamp: In case this wasn't just a mistake — this can be deleted too, right?

src/views/splash/splash.jsx
```diff
@@ -1,42 +1,38 @@
...
+    type: 'View',
     getInitialState: function () {
         return {
-            projectCount: 14000000, // gets the shared project count
-            activity: [], // recent social actions taken by users someone is following
-            news: [], // gets news posts from the scratch Tumblr
-            featuredCustom: {}, // custom homepage rows, such as "Projects by Scratchers I'm Following"
-            featuredGlobal: {}, // global homepage rows, such as "Featured Projects"
             showEmailConfirmationModal: false, // flag that determines whether to show banner to request email conf.
             refreshCacheStatus: 'notrequested'
```
rschamp: This state is unused too.

src/views/splash/splash.jsx
```diff
@@ -448,10 +223,15 @@ var Splash = injectIntl(React.createClass({
 var mapStateToProps = function (state) {
     return {
         session: state.session,
-        permissions: state.permissions
+        permissions: state.permissions,
+        rows: state.rows,
+        activity: state.activity,
+        news: state.news,
+        projectCount: state.projectCount,
+        homepageCache: state.homepageCache
```
rschamp: Something to think about — do more work in `mapStateToProps` so that we only pass in the data we need rather than broad categories. E.g., instead of just `session`, pass in the user if it exists. If you need to know if the session is loaded, then set another prop, `sessionLoaded` for that purpose.  That will help get rid of the weird doubling `this.props.stateKey.stateKey` everywhere, and can help us be more explicit about what we're using, so we'll know better when something is unused and can get removed from the state.

src/views/components/components.jsx
```diff
@@ -1,7 +1,7 @@
 var React = require('react');
 var render = require('../../lib/render.jsx');
 
-var Activity = require('../../components/activity/activity.jsx');
+// var Activity = require('../../components/activity/activity.jsx');
```
mewtaylor: we can remove this

src/views/components/components.jsx
```diff
@@ -35,9 +35,9 @@ var Components = React.createClass({
                     <Carousel />
                 </Box>
                 <h1>{'What\'s Happening??'}</h1>
-                <Activity />
+                // <Activity />
```
mewtaylor: we can remove this

src/views/components/components.jsx
```diff
@@ -35,9 +35,9 @@ var Components = React.createClass({
...
                 <h1>{'Nothing!!!'}</h1>
-                <Activity items={[]} />
+                // <Activity items={[]} />
```
mewtaylor: we can remove this

src/views/splash/custom.jsx
```diff
@@ -0,0 +1,111 @@
...
+            id: 'splash.projectsLovedByScratchersFollowing',
+            defaultMessage: 'Projects Loved by Scratchers I\'m Following'
+        },
+        custom_projects_in_studios_following: {
+            id: 'splash.projectsInStudiosFollowing',
+            defaultMessage: 'Projects in Studios I\'m Following'
+        }
+    },
+    componentWillReceiveProps: function (){
+        var formatMessage = this.props.intl.formatMessage;
```
mewtaylor: this looks like the same functionality that's in `render`, so this can probably be removed too.

src/views/splash/custom.jsx
```diff
@@ -0,0 +1,111 @@
...
+                {rows}
+            </div>
+        );
+    }
+});
+
+module.exports = CustomRows;
+
+/*
+<Box title={
```
mewtaylor: we can remove this

src/views/splash/custom.jsx
```diff
@@ -0,0 +1,111 @@
...
+                        <Box title={
+                                formatMessage({
+                                    id: this.rowMap[keys[i]].id,
+                                    defaultMessage: this.rowMap[keys[i]].defaultMessage})}
+                             key={keys[i]}>
+
+                            <Carousel items={this.props.featured[keys[i]]} />
+                        </Box>
+                    );
+                    /*TODO: This seems really messy to me,
```
mewtaylor: I see what you're saying, but think it's fine. One other way to do it would be to do this as a `map` inside the return statement, but that might look equally (or more) messy too...

src/views/splash/splash.jsx
mewtaylor: I think that makes sense to me – wondering if we should have that be part of this, though, or its own issue/PR?

rschamp: I'd rather it be part of this PR since this PR adds a lot of redux state as props.

src/views/splash/custom.jsx
rschamp: I agree... it doesn't look like this does anything.  This kind of stuff shouldn't happen in `componentWillReceiveProps` either — seems like it was supposed to go into `render` but didn't end up there, and then it looks like it is `render` so it didn't get cleaned up? Maybe?

src/views/splash/splash.jsx
```diff
@@ -1,42 +1,38 @@
...
+var View = injectIntl(React.createClass({
+    type: 'View',
     getInitialState: function () {
         return {
-            projectCount: 14000000, // gets the shared project count
-            activity: [], // recent social actions taken by users someone is following
-            news: [], // gets news posts from the scratch Tumblr
-            featuredCustom: {}, // custom homepage rows, such as "Projects by Scratchers I'm Following"
-            featuredGlobal: {}, // global homepage rows, such as "Featured Projects"
             showEmailConfirmationModal: false, // flag that determines whether to show banner to request email conf.
```
rschamp: Looks like this one is too... but that looks like an existing mistake. The only local state in the view I see now is `emailConfirmationModalOpen` — I think that's what this key should be.

src/views/splash/custom.jsx
```diff
@@ -0,0 +1,111 @@
...
+                     * although not *technically* against the container/presentation split.
+                     */
```
rschamp: Could you expand on what you think is messy?

src/views/splash/emailconfirmation.jsx
```diff
@@ -0,0 +1,36 @@
+var React = require('react');
```
rschamp: It doesn't look like this file is used. Did I miss it?

mewtaylor:
@technoboy10 @rschamp closing this for now, and will bring this into the homepage redesign when we do that later this fall. Thanks again for establishing this methodology @technoboy10, it's going to be a great foundation off of which to build our redux system!

PR #595, opened 2016-06-22
Fix GH-572: HOC page formatting

### LINKED ISSUE #572: HOC page formatting issue
I've just noticed a formatting issue on the Hour of Code page (https://scratch.mit.edu/hoc). 

The activity cards and guides' names get all messed up whenever their size is too big for the column their in. Have a look at the images below. The problem seems to affect all languages.

In pt-br:
20160608-hoc_formatting_issue

In en:
20160608-hoc_formatting_issue2-en

Thanks!

technoboy10:
This PR fixes #572 using @Choco31415's suggestion.

Tested on Chrome 51, Firefox 46.0.1, and Safari 9.1.1.

**Testing Steps**
Making the width of the window small enough to cause the center box to resize should *not* cause the headers in "Activity Cards and Guides" section to collapse on themselves. 

I did have some issues in Safari with the headers extending past the box, but this behavior is the same on the existing Scratch website.

### DIFF
--- a/src/views/hoc/hoc.scss
+++ b/src/views/hoc/hoc.scss
@@ -9,7 +9,7 @@ $base-bg: $ui-white;
 
 .hoc {
     .title-banner {
-        &.wbb-bg {    
+        &.wbb-bg {
             background-image: url("/images/hoc2015/hide-bg.jpg");
         }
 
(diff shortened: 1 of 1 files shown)


thisandagain:
Thanks!

Issue #311, opened 2016-01-07
Educator Page Update


kaschm:
Update `https://scratch.mit.edu/educators/` with new content that highlights Scratch use in classroom and after-school settings.

Need:

- [x] Final content structure (@kaschm)
- [x] Design (@carljbowman)
- [x] Final links to Scratch Foundation blog when interviews are up (@kaschm)

thisandagain:
Blocked on content & design. Assigning to @kaschm for tracking.

kaschm:
Thanks @thisandagain !

kaschm:
Final content here: https://docs.google.com/document/d/16gJYgW-di4M0pFE0mCWmnRmRMTYnz-jo4Sw_OoVxUjE/edit

kaschm:
@carljbowman , screenshot for the page is here:
https://drive.google.com/a/media.mit.edu/file/d/0BwdnRUms3m2bMHZacWdoNWhsa1E/view?usp=sharing

carljbowman:
## Scratch For Educators 

**Assets**
[for-educators-assets.zip](https://github.com/LLK/scratch-www/files/303307/for-educators-assets.zip)


**Missing Content** 
@kaschm - For each of the 4 User Stories 
- [ ] Name of person and association (e.g. school)
- [ ] Photo 

### Desktop (12cols)
![r3_educators-landing](https://cloud.githubusercontent.com/assets/3409578/15869878/40292550-2cbc-11e6-9721-9ee395031797.png)

![r3_educators-landing 12cols](https://cloud.githubusercontent.com/assets/3409578/15869896/5977a70c-2cbc-11e6-8e2e-be9c68ef8356.png)

### Tablet – Desktop (8cols)
![r3_educators-landing 8cols](https://cloud.githubusercontent.com/assets/3409578/15869906/606813f8-2cbc-11e6-8ddb-8c2fe8a9bdcb.png)

### Mobile – Tablet (6cols)
![r3_educators-landing 6cols](https://cloud.githubusercontent.com/assets/3409578/15869925/6be89ec8-2cbc-11e6-8ebf-051b43198e0b.png)

### Mobile – (4cols)
![r3_educators-landing 4cols](https://cloud.githubusercontent.com/assets/3409578/15869938/77b52fc8-2cbc-11e6-9d55-1927e52d5226.png)

carljbowman:
## Scratch for Educators

[...]

PR #601, opened 2016-06-23
Fix GH-311: Educator landing page

technoboy10:
Probably needs some work, as sass-lint stopped working on my machine so I couldn't reliably test the .scss files.

This PR adds the Educator Landing page.


src/views/teachers/landing/landing.scss
```diff
@@ -0,0 +1,521 @@
...
+            padding-bottom: 0;
+
+            h1 {
+                margin: 0;
+                color: $ui-white;
+            }
+
+
+            .masthead-info {
+                display: flex;
```
mewtaylor: If this is on `FlexRow` it may not need this attribute

src/views/teachers/landing/landing.scss
```diff
@@ -0,0 +1,521 @@
...
+            max-width: $cols4;
+        }
+    }
+
+    section {
+        margin-bottom: 3rem;
+
+        /*p {
+            max-width: $cols8;
+        }*/
```
mewtaylor: commented out code – could be removed

src/views/teachers/landing/landing.scss
```diff
@@ -0,0 +1,521 @@
...
+
+                p {
+                    margin: 10px 0;
+                    font-size: .75rem;
+                }
+            }
+        }
+    }
+
+    #resources {
```
mewtaylor: I think it would be best, if possible, to switch these to class names rather than id's.

PR #612, opened 2016-06-24
Fix GH-607: Language chooser should not be 100% width

### LINKED ISSUE #607: Language dropdown is 100% width
screenshot 2016-06-24 09 54 08

technoboy10:
This PR fixes #607 by adding a stylesheet for the languagechooser component that sets the `select` element's width to 13.75 rem (3 columns).

### DIFF
--- a/src/components/languagechooser/languagechooser.jsx
+++ b/src/components/languagechooser/languagechooser.jsx
@@ -6,6 +6,8 @@ var languages = require('../../../languages.json');
 var Form = require('../forms/form.jsx');
 var Select = require('../forms/select.jsx');
 
+require('./languagechooser.scss');
+
 /**
  * Footer dropdown menu that allows one to change their language.
  */
--- a/src/components/languagechooser/languagechooser.scss
+++ b/src/components/languagechooser/languagechooser.scss
@@ -1,8 +1,10 @@
 @import "../../frameless";
 
 .language-chooser {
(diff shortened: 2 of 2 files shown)


rschamp:
@carljbowman @mewtaylor We've decided that all dropdowns should be 100% the width of their container by default? Seems backwards to me — they should have a fixed width, but in some situations (mobile) the dropdown should go to 100%. Is the 100% default for dropdowns consistent with text inputs?

mewtaylor:
@rschamp I don't think it's that they're set to 100% by default – the way I did it was to not provide a default width, and have that be determined by the implementing component/view. You think there should be a default width, though?

rschamp:
[This is what I mean by the default 100% width we currently have defined.](https://github.com/LLK/scratch-www/blob/develop/src/components/forms/select.scss#L17) That rule isn't present for [text inputs](https://github.com/LLK/scratch-www/blob/develop/src/components/forms/input.scss). I agree with this being set by the parent component where necessary, but this should be consistent across all the form components so they can be used together easily.  I don't think the 100% width makes sense for the default though, more often than not we will want a fixed width like the browser default.

mewtaylor:
@rschamp ah whoops I must've missed taking that one out when I took it out of the `input` file, sorry about that! I totally agree about not having it 100% by default.

PR #615, opened 2016-06-28
[Bugfix] Add permissions dispatch for `WillReceiveProps`

mewtaylor:
We still need `componentWillMount` to handle the initial render for a logged in teacher first visting the page. Fixes issue in which a teacher would not see the banner immediately after dynamic login.

### DIFF
--- a/src/components/navigation/www/navigation.jsx
+++ b/src/components/navigation/www/navigation.jsx
@@ -259,14 +259,14 @@ var Navigation = React.createClass({
                                             <FormattedMessage id="general.myStuff" />
                                         </a>
                                     </li>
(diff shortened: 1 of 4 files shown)


rschamp:
I think this will work. However I think we should still think about consolidating the way we check for the educator permission so that we get consistent behavior (and so future bugs will all be located on the same surface).

Related to that, I think this action may be more appropriate to dispatch with the session action.  You will be able to tell if the cookie has changed between the time the session is requested and returned — if it has changed, then refresh the storage using the cookie.  Then the storages based on the cookie will always be in sync with the session.

Issue #537, opened 2016-05-31
Gap above Log In dropdown on Homepage - on production


jwzimmer:
Per conversation with @carljbowman, spin off case for high-prio issue from #532.

**Issue on Homepage**
Chrome

Scratch login on homepage has a gap:
<img width="1016" alt="screen shot 2016-05-31 at 10 37 30 am" src="https://cloud.githubusercontent.com/assets/1320456/15678272/b50fc1ae-271b-11e6-890d-8a25c56215f7.png">

jwzimmer:
Note: when you're not logged in, the menu looks like a speech bubble.
<img width="326" alt="screen shot 2016-06-08 at 4 37 49 pm" src="https://cloud.githubusercontent.com/assets/1320456/15910090/5fdcc646-2d97-11e6-938f-1ca01f2a352b.png">

carljbowman:
@jwzimmer - Thanks for logging this.

The Logged out 'speech' bubble was not affected by the global typography, everything there looks fine on Logged out state. 

**The dropdown when logged in should be attached to the nav.**

![nav_desktop](https://cloud.githubusercontent.com/assets/3409578/15910837/c818b1f4-2d9a-11e6-9980-0b9656f5918c.png)

**Hover on List Item**

![nav_desktop_hover-list-item](https://cloud.githubusercontent.com/assets/3409578/15910849/d59560a2-2d9a-11e6-8420-dbfaeefdfc49.png)

PR #619, opened 2016-06-29
Darken Account Nav Link when Menu Open

technoboy10:
This PR adds an `open` class to the account nav link whenever the profile menu is open and darkens the link, as per Carl's sketch in #537.

### DIFF
--- a/src/components/navigation/www/navigation.jsx
+++ b/src/components/navigation/www/navigation.jsx
@@ -241,7 +241,8 @@ var Navigation = React.createClass({
                                 </a>
                             </li>,
                             <li className="link right account-nav" key="account-nav">
-                                <a className="user-info" href="#" onClick={this.handleAccountNavClick}>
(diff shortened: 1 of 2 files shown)


src/components/navigation/www/navigation.jsx
```diff
@@ -241,7 +241,8 @@ var Navigation = React.createClass({
                                 </a>
                             </li>,
                             <li className="link right account-nav" key="account-nav">
-                                <a className="user-info" href="#" onClick={this.handleAccountNavClick}>
+                                <a className={this.state.accountNavOpen ? 'user-info open' : 'user-info'}
```
mewtaylor: I think it might be more straightforward to do this before the return and use the `classes` library. That way, it doesn't need to be an either/or, but dependent on the presence of an extra class.

technoboy10:
@mewtaylor Fixed!

Issue #626, opened 2016-06-29
Teacher Banner loses content but remains (purple bar) after logging out of teacher account


jwzimmer:
/cc @mewtaylor 

Re https://github.com/LLK/scratch-www/pull/558
Chrome Version 51.0.2704.106 (64-bit)
- **The banner should not be present when I log out after logging in as a teacher**
  - The banner loses its content but still appears (it goes away after you refresh the page)
<img width="1019" alt="screen shot 2016-06-29 at 11 09 56 am" src="https://cloud.githubusercontent.com/assets/1320456/16457585/0f62897c-3dea-11e6-9899-c9036b20d2f6.png">

mewtaylor:
Fixed in https://github.com/LLK/scratch-www/issues/640

PR #640, opened 2016-06-30
Fix GH-626: Set null contents to null instead of error throw

mewtaylor:
Before, if a cookie didn't exist, it would throw an error, not causing permissions/tokens to change to empty values. This fixes that (and #626) by setting the value to undefined instead.

### Test Case ###
* see #626 for the test case


src/lib/jar.js
```diff
@@ -80,6 +81,8 @@ var Jar = {
             if (err) return callback(err);
             Jar.unsign(value, function (err, contents) {
                 if (err) return callback(err);
+                if (typeof contents === 'undefined') return callback(null, contents);
```
rschamp: We could also avoid this if we handled the empty value ahead of `Jar.unsign`. Not a big deal, but would simplify what happens a little bit. "get —> check undefined —> return" instead of "get —> unsign —> check undefined —> return —> check undefined —> return".

PR #656, opened 2016-06-30
Fix GH-651: Orangeify navbar on Staging

### LINKED ISSUE #651: Staging - Change navigation bar background color when in staging environment
We need some way to visually distinguish between production and staging. In `scratch.ly` we use a really terrible background that is distracting. I'm wondering if we might be able to just change the navbar background color or something else to make it easier to immediately identify which server you are on:
image

technoboy10:
This PR fixes #651 by adding a `staging` class to the navbar and the dropdown if an environment variable named `SCRATCH_ENV` is set to "staging". 

screenshot 2016-06-30 18 08 25

### DIFF
--- a/src/components/dropdown/dropdown.scss
+++ b/src/components/dropdown/dropdown.scss
@@ -15,6 +15,10 @@
     font-size: .8125rem;
     font-weight: normal;
 
+    &.staging {
+        background-color: $ui-orange;
+    }
+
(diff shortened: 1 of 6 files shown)


TheGrits:
@technoboy10 haha I love it!

thisandagain:
📙 🔶 🔸

PR #662, opened 2016-07-03
Use babel-loader for jsx transpilation

rschamp:
This is to remove the deprecation warning about React.__spread. It will give us other nice things too, though.


thisandagain:
👍  Thanks for putting this together.

PR #668, opened 2016-07-05
Fix GH-666: Make category options into key/value

### LINKED ISSUE #666: Explore - XHR request should include `q=*` not `q=all`
Currently, the explore page issues an XHR request with a query string of `q=all` which actually does a search for projects that include `all` in the metadata. This should be changed to either `q=*` or simply drop that query string parameter altogether.

mewtaylor:
so that we can store the query string with the tab name, since it's not the same in every instance. fixes #666, and makes a couple other style tweaks.

### DIFF
--- a/src/views/explore/explore.jsx
+++ b/src/views/explore/explore.jsx
@@ -1,3 +1,4 @@
+var classNames = require('classnames');
 var injectIntl = require('react-intl').injectIntl;
 var FormattedMessage = require('react-intl').FormattedMessage;
 var React = require('react');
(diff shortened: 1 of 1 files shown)


src/views/explore/explore.jsx
```diff
@@ -80,34 +86,28 @@ var Explore = injectIntl(React.createClass({
...
-                    </a>;
-        if (this.props.category === type) {
-            allBubble = <a href={'/explore/' + this.props.itemType+'/' + type + '/'}>
-                        <li className='active'>
-                            <FormattedMessage id={'general.' + type} />
-                        </li>
-                    </a>;
-        }
-        return allBubble;
+        let classes = classNames({
```
thisandagain: This will break, no?

thisandagain:
@mewtaylor @rschamp Worried about our use of `let` here until we have moved over to Babel (not this iteration). Should this be rolled back?

mewtaylor:
@thisandagain you're correct – see https://github.com/LLK/scratch-www/commit/9a037fda818c415097df26e9b6eed4f9c5078924

PR #677, opened 2016-07-06
Updated Project Counts

St19Galla:
Awhile ago at issue #426 a discussion was started about what fix should be used for the project count:

- Continue updating this number every million we get
- Determine alternative text to display that captures the magnitude of projects created but does not need updating as often

The topic was closed without a clear decision, so I went ahead and updated the project counts from 14 to 15 million as per https://scratch.mit.edu/statistics/


St19Galla:
@thisandagain @rschamp @mewtaylor

St19Galla:
It looks like this has been fixed with the new update :) Closing.

mewtaylor:
@St19Galla oh whoops sorry about that! I should've posted that on this PR. Thanks for following up :)

Issue #681, opened 2016-07-06
For Educators – 'Coming soon' button should not have a pointer cursor on hover


carljbowman:
<img width="1051" alt="screen shot 2016-07-06 at 11 49 40 am" src="https://cloud.githubusercontent.com/assets/3409578/16624711/c52fa48a-436f-11e6-8c10-3b827131fe6b.png">

PR #682, opened 2016-07-06
Fix GH-680 and GH-681: Educator landing style tweaks

### LINKED ISSUE #680: For Educators - Entire user story element should be a link
The entire 'card' should link to the article, not just the title. 

screen shot 2016-07-06 at 11 46 51 am

technoboy10:
This PR fixes #680 and #681 by changing the story divs to `a` elements and setting `cursor: default` on the Coming Soon button.


mewtaylor:
@technoboy10 looks good! One recommendation I have would be to make "Coming Soon" not a button, rather than do the `cursor: default`, though. It already has a fair amount of styling specified on it, and it seems like it's only static text with styling for now.

technoboy10:
@mewtaylor Done!

Issue #684, opened 2016-07-06
Teacher Registration: Confusing use of "characters" in username validation - excludes "?", "!" (non-alphanumeric characters)


jwzimmer:
We may want a narrower description of "characters" in the validation message for username... A lot of users may think ?,&,!, etc., would be considered "characters" when we're really trying to restrict to alphanumeric chars, I think. /cc @kaschm & @carljbowman 

<img width="668" alt="screen shot 2016-07-06 at 3 22 39 pm" src="https://cloud.githubusercontent.com/assets/1320456/16631587/dfeb9acc-438d-11e6-8341-847ea1c943cb.png">

kaschm:
Good catch. @rschamp, would it be correct to assume we really mean 

'Your username may only contain numbers, letters, and dashes' ? 

@jwzimmer @carljbowman , is this any clearer?

jwzimmer:
(@kaschm I thumbs-up'd your comment, but then thought you might not see that since it doesn't incur a notification - yes, I think the text you wrote is clearer. : ) )

rschamp:
@kaschm yes, correct.

St19Galla:
#685

PR #685, opened 2016-07-07
Fix gh-684 - Changing error message

St19Galla:
I edited the error message to specifically say `letters, numbers,` instead of the possible misleading `characters` as per #684

### DIFF
--- a/src/views/teacherregistration/l10n.json
+++ b/src/views/teacherregistration/l10n.json
@@ -1,7 +1,7 @@
 {
     "teacherRegistration.usernameStepDescription": "Fill in the following forms to request an account. The approval process may take up to 24 hours.",
     "teacherRegistration.usernameStepTitle": "Request a Teacher Account",
-    "teacherRegistration.validationUsernameRegexp": "Your username may only contain letters, numbers, and \"-\"",
+    "teacherRegistration.validationUsernameRegexp": "Your username may only contain letters, numbers, \"-\", and \"_\"",
     "teacherRegistration.validationUsernameMinLength": "Usernames must be at least 3 characters",
     "teacherRegistration.validationUsernameMaxLength": "Usernames must be at most 20 characters",
     "teacherRegistration.validationPasswordLength": "Passwords must be at least six characters",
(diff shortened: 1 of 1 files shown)


St19Galla:
@thisandagain @rschamp @mewtaylor

thisandagain:
Good catch. 👍

rschamp:
We should note it also allows underscores (`\w`).

mewtaylor:
@rschamp ah right thanks – @kaschm @jwzimmer would adding "numbers, letters, _underscores_ and..." be too long, or is that okay?

St19Galla:
Couldn't we just make it like the main website registration and have it say `Your username may only contain letters, numbers, \"-\", and \"_\"` ?

mewtaylor:
@thisandagain @carljbowman @St19Galla here's a couple of screenshots, for reference:
screen shot 2016-07-07 at 8 47 02 am
screen shot 2016-07-07 at 8 48 15 am

St19Galla:
@carljbowman @mewtaylor @thisandagain Using the word underscore looks a bit neater, and it seems to work on English as per the screenshots above, but will switching to languages with longer text break the layout?

carljbowman:
@St19Galla This tool tip is designed to grow vertically and continue to wrap the text. Yeah carrot orienting the tip with the associated form remains fixed. 

That said longer content should not be an issue.

St19Galla:
@carljbowman Great, thanks :) So, with that being said, do you guys want to going with `underscores` or `\"_\"`?

carljbowman:
@St19Galla - I am thinking we should go with the original content from the current Join Flow as you mentioned:

`Your username may only contain letters, numbers, \"-\", and \"_\"`

#### Current 
screen shot 2016-07-07 at 1 53 48 pm

#### Suggested
9f0dec0c-441f-11e6-8136-e9859bc79ff3

St19Galla:
@thisandagain @rschamp @mewtaylor @carljbowman done :)

thisandagain:
@mewtaylor R?

mewtaylor:
looks great @St19Galla thanks!

PR #687, opened 2016-07-07
Fix GH-686: Add language to search / explore api requests

### LINKED ISSUE #686: Search / Explore - Ensure language preference is attached to request
Currently (even if you language is set to something other than English), the explore and search pages will make an XHR request to:
`https://api.scratch.mit.edu/search/projects?limit=16&offset=0&q=*`

This request should reflect language preference. E.g.:
`https://api.scratch.mit.edu/search/projects?limit=16&offset=0&language=es&q=*`

mewtaylor:
Fixes #686, thanks @thisandagain !

### DIFF
--- a/src/views/explore/explore.jsx
+++ b/src/views/explore/explore.jsx
@@ -64,6 +64,7 @@ var Explore = injectIntl(React.createClass({
             uri: '/search/' + this.props.itemType +
                  '?limit=' + this.props.loadNumber +
                  '&offset=' + this.state.offset +
+                 '&language=' + this.props.intl.locale +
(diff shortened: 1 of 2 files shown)


thisandagain:
Is `this.props.intl.locale` always defined?

mewtaylor:
@thisandagain anywhere we use `render`, should be: https://github.com/LLK/scratch-www/blob/develop/src/lib/render.jsx#L38

thisandagain:
Ok. Just want to make sure we don't need to define a fallback here. Thanks.

Issue #694, opened 2016-07-07
Floating field validation for Country - Teacher Registration


jwzimmer:
The Country validation is off to the right instead of aligned like the other validations. It is also not outlined in orange.

Chrome, staging

<img width="846" alt="screen shot 2016-07-07 at 12 05 17 pm" src="https://cloud.githubusercontent.com/assets/1320456/16660156/375576ea-443b-11e6-971b-ab345b837116.png">

jwzimmer:
This alignment may not be ideal either:

<img width="650" alt="screen shot 2016-07-07 at 12 08 17 pm" src="https://cloud.githubusercontent.com/assets/1320456/16660249/8a9ceafe-443b-11e6-998b-ff2efcb501eb.png">

mewtaylor:
For the `Gender` field, were you thinking about alignment horizontally, or alignment vertically?

jwzimmer:
Both? Probably vertically though? I mean, as is, it's not too bad, but it kind of looks like it's pointing to a non-existent field, to me...

mewtaylor:
Ah ok – @jwzimmer the way it was aligned vertically was to be with the `Gender` title, since that's the "field" that it's referring to. But perhaps that would be clearer if it were differently aligned horiztonally?

jwzimmer:
@mewtaylor I don't know - if @carljbowman thinks it's fine as is, that's fine with me. : )

carljbowman:
@jwzimmer @mewtaylor 

RE Select Country tooltip: This is definitely a bug and should be aligned like 'Address Line 1'

RE Gender tooltip: It is not clear to me that the tooltip is pointing at the title 'Gender', it seems just slightly off and by that nature unintentional. This behavior is also inconsistent with how we center align vertically to fields. My leaning is towards center aligning vertically with the set of fields.

That said, we will have a similar issue on 'Type of Organization'. In these cases we may want to be more specific with the tooltip content: 'Please select a gender, this field is required'. For the case of the 'Type of Organization': 'Please check all the Types of Organizations that apply, this field is required'.

jwzimmer:
Also seeing this in IE 11:

<img width="620" alt="screen shot 2016-07-07 at 4 18 57 pm" src="https://cloud.githubusercontent.com/assets/1320456/16668438/f0cb527a-445e-11e6-92e2-a6aff51ef600.png">

jwzimmer:
[...]

Issue #693, opened 2016-07-07
Phone number field not required but it should be - Teacher Registration


jwzimmer:
Either phone number & checkmark should both be optional, or both should be required.

Right now, you can leave the phone number field blank but you are forced to check the box saying Scratch can call you, which doesn't make sense.

Chrome, staging

<img width="443" alt="screen shot 2016-07-07 at 12 00 00 pm" src="https://cloud.githubusercontent.com/assets/1320456/16660060/d9c9d5c0-443a-11e6-97a6-cab75745300c.png">
<img width="737" alt="screen shot 2016-07-07 at 12 10 02 pm" src="https://cloud.githubusercontent.com/assets/1320456/16660332/dc685120-443b-11e6-9a1a-4cd51bbed23b.png">

jwzimmer:
Possibly not putting in a phone number caused this?
<img width="666" alt="screen shot 2016-07-07 at 12 17 58 pm" src="https://cloud.githubusercontent.com/assets/1320456/16660575/f1d74326-443c-11e6-8bc8-f25ddab35303.png">

thisandagain:
This is prio-high but not for the beta test. Adding to July milestone.

thisandagain:
Nvmd! The connection to #696 makes this a blocker.

mewtaylor:
Fixed in https://github.com/LLK/scratch-www/pull/702

Issue #691, opened 2016-07-07
Entered too-long phone number during Teacher Registration


jwzimmer:
It seemed to accept this phone number even though it is too long, for some reason? I am not sure under exactly what conditions the validation fails - I thought at first that it might be thrown off by ending a phone number in 0, but it correctly rejects e.g. +1 (738) 982-73000, +1 (738) 982-73120, +1 (617) 253-88100, etc. It seems to be something about the rest of the phone number that fools it, since it accepts +1 (123) 456-78931 (and other similar numbers) as well - maybe the first two 1's? It rejects e.g. +1 (234) 567-89391.

Maybe this is actually a valid US phone number pattern? : \

No validations being triggered:
<img width="380" alt="screen shot 2016-07-07 at 11 06 13 am" src="https://cloud.githubusercontent.com/assets/1320456/16658056/de0e33fe-4432-11e6-847e-cc41a82b121c.png">
<img width="384" alt="screen shot 2016-07-07 at 11 03 15 am" src="https://cloud.githubusercontent.com/assets/1320456/16657974/89517614-4432-11e6-9c3e-a1d66e9222e3.png">
<img width="418" alt="screen shot 2016-07-07 at 11 03 03 am" src="https://cloud.githubusercontent.com/assets/1320456/16657973/89513f28-4432-11e6-9d37-e0545f239cea.png">

Validation:
<img width="613" alt="screen shot 2016-07-07 at 11 06 00 am" src="https://cloud.githubusercontent.com/assets/1320456/16658075/f5847822-4432-11e6-85fc-1d4349d5f8ff.png">

![out](https://cloud.githubusercontent.com/assets/1320456/16657835/e12ec644-4431-11e6-931e-f1612b7d55a6.gif)

mewtaylor:
Fixed in https://github.com/LLK/scratch-www/pull/702

jwzimmer:
I can still enter a longer phone number, on production, in Chrome:

Number is invalid:
<img width="639" alt="screen shot 2016-07-09 at 6 36 51 pm" src="https://cloud.githubusercontent.com/assets/1320456/16710654/4e97c0a4-4604-11e6-90f9-4f30e651fdd9.png">

Add a 1, and it's valid (but too long for a US number?):
<img width="506" alt="screen shot 2016-07-09 at 6 36 39 pm" src="https://cloud.githubusercontent.com/assets/1320456/16710653/4e9601c4-4604-11e6-869a-8ac54d064f47.png">

[...]

PR #702, opened 2016-07-07
Fix GH-[695, 694, 693, 691] – Add custom validation for organization and phone

### LINKED ISSUE #695: Type of Organization should be required - Teacher Registration
If you don't check anything for Type of Organization, you can still proceed to the next step (the top two fields are the only ones that seem to be required on this step). We meant to require type of organization... we're missing that validation. /cc @carljbowman & @kaschm 

screen shot 2016-07-07 at 12 10 17 pm
screen shot 2016-07-07 at 12 10 23 pm

mewtaylor:
It appears that the `required` field is not applied in the way we'd like it to be for the custom checkbox group and custom phone input fields, so do a `required` check on form submission. Additionally, this adds in a few style fixes. Fixes #695, #694, #693, #691.

/cc @rschamp @carljbowman

### DIFF
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -286,6 +286,14 @@ module.exports = {
                 waiting: false
             };
         },
+        onValidSubmit: function (formData, reset, invalidate) {
+            if (formData.phone.national_number.length !== formData.phone.country_code.format.length) {
+                return invalidate({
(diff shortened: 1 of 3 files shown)


jwzimmer:
I am no longer seeing the issues from #691, #693, #694, or #695, although there are issues of when the validations are triggered (https://github.com/LLK/scratch-www/issues/712, etc.).

PR #717, opened 2016-07-11
[Master] Hotfix: Allow urls with `ajax` to pass through

mewtaylor:
Trying #716 again due to CI failure.


mewtaylor:
@thisandagain appears it was a git configuration issue after all – passing now!

PR #720, opened 2016-07-11
[Develop] GH-710: only show username in search and explore

### LINKED ISSUE #710: Search / Explore - Replace stats with author username
In each project cell we should replace the "stats" information with the author's username (see: splash) as per discussion and user feedback.

mewtaylor:
fixes #710


thisandagain:
You'll need to apply this to `search.jsx` as well:
https://github.com/LLK/scratch-www/blob/develop/src/views/search/search.jsx

thisandagain:
:shipit:

PR #727, opened 2016-07-18
Fix GH-674: Waiting room cards should be centered in IE

### LINKED ISSUE #674: Cards need to be centered in teacher waiting room - in IE11
What @cwillisf is seeing in IE11:
screen shot 2016-07-05 at 6 01 32 pm

When @cwillisf shrinks the window a bit, the Scratch logo gets centered (I also see this in Chrome, presumably that behaviour is for mobile/ tablet view?) but the "Get Started with Resources" is still to the left:
screen shot 2016-07-05 at 6 02 33 pm

What I see in Chrome:
screen shot 2016-07-05 at 5 59 45 pm

technoboy10:
This PR fixes #674 by setting the horizontal margin of the waiting room card elements to auto instead of 0.

### DIFF
--- a/src/views/teacherregistration/teacherregistration.scss
+++ b/src/views/teacherregistration/teacherregistration.scss
@@ -149,7 +149,7 @@ body {
         }
 
         .card {
-            margin: 1rem 0;
+            margin: 1rem auto;
             padding: 1.5rem;
             width: initial;
 


technoboy10:
This change does not change how the cards look on other browsers (Chrome 51, Firefox 46 on OS X 10.11).

technoboy10:
For future reference, it appears that IE doesn't set the cards to the correct width so they appear off-center. For the time being, it seems that centering the cards is the best option.

PR #730, opened 2016-07-20
Add student registration flow

rschamp:
Resolves https://github.com/LLK/scratchr2/issues/3679
Requires PRs in both github.com/LLK/scratchr2 and github.com/LLK/scratch-api

Adds class registration page
Also refactors teacher registration styles so that the registration steps may be reused
I alphabetized routes.json by path since there was not much rhyme or reason to the old order. Redirects are at the bottom.

## Test cases
* Student registration should follow as described in https://github.com/LLK/scratchr2/issues/3679
* Teacher registration should still display correctly

PR #731, opened 2016-07-20
Fix GH-699: Fallback to mobile validation bubble on IE

### LINKED ISSUE #699: Misaligned validation for consent checkbox in IE 11 - Teacher Registration
The validation for the checkbox covers the text instead of being off to the right side as it is in Chrome:

screen shot 2016-07-07 at 3 22 15 pm
(phone number field)
screen shot 2016-07-07 at 3 22 03 pm

Staging, IE 11 (Win 7 VM)

technoboy10:
This PR fixes #699 by adding a media query to target IE 10 and 11 and changes the validation bubble styling to the mobile style for the bubbles. (I also added a few selectors to the media query which originally sets the transform property for smaller browsers, to be more consistent with how the page displays validation bubbles).

It also sets the -ms-reveal and -ms-clear pseudoelements on IE to `display: none` to better match the experience on other browsers.


mewtaylor:
@technoboy10 it looks like a couple things are moving around in #730 from the `teacherregistration.scss` file – would you be okay rebasing this branch off of that one when it does, and just moving these stylings into the new location (`steps.scss`, I believe) when that happens?

technoboy10:
@mewtaylor Sure thing!

technoboy10:
@mewtaylor Ok, this should be good to go!

PR #753, opened 2016-07-25
Fix GH-749: Fix formatting for a few form elements

### LINKED ISSUE #749: Validation messages are not aligned with inputs on teacher/student registration

mewtaylor:
and make use of the FRC `elementWrapperClassName`, since they way we did it before – custom `className` strings on the element – was not working. Fixes #749 .

### Elements to test ###
* Phone number input
* "Why use scratch" textarea
* Checkbox group

### DIFF
--- a/src/components/forms/checkbox.jsx
+++ b/src/components/forms/checkbox.jsx
@@ -15,7 +15,7 @@ var Checkbox = React.createClass({
             this.props.className
         );
         return (
-            <FRCCheckbox elementWrapperClassName={classes} {... this.props} />
+            <FRCCheckbox rowClassName={classes} {... this.props} />
         );
     }
 });
--- a/src/components/forms/checkbox.jsx
+++ b/src/components/forms/checkbox.jsx
@@ -15,9 +15,7 @@ var Checkbox = React.createClass({
             this.props.className
         );
         return (
-            <div className={classes}>
-                <FRCCheckbox {... this.props} />
-            </div>
+            <FRCCheckbox elementWrapperClassName={classes} {... this.props} />
         );
     }
 });
--- a/src/components/forms/phone-input.jsx
+++ b/src/components/forms/phone-input.jsx
@@ -53,9 +53,9 @@ var PhoneInput = React.createClass({
                                      label={null}
                                      disabled={this.isFormDisabled() || this.props.disabled}
                     />
+                    {this.renderHelp()}
                     {this.renderErrorMessage()}
                 </div>
-                {this.renderHelp()}
             </Row>
         );
     }
--- a/src/components/forms/phone-input.jsx
+++ b/src/components/forms/phone-input.jsx
@@ -42,7 +42,7 @@ var PhoneInput = React.createClass({
         return (
             <Row {... this.getRowProperties()}
                  htmlFor={this.getId()}
-                 className={classNames('phone-input', this.props.className)}
+                 elementWrapperClassName={classNames('phone-input', this.props.className)}
             >
                 <div className="input-group">
                     <ReactPhoneInput className="form-control"
--- a/src/components/forms/phone-input.jsx
+++ b/src/components/forms/phone-input.jsx
@@ -42,7 +42,7 @@ var PhoneInput = React.createClass({
         return (
             <Row {... this.getRowProperties()}
                  htmlFor={this.getId()}
-                 elementWrapperClassName={classNames('phone-input', this.props.className)}
+                 rowClassName={classNames('phone-input', this.props.className)}
             >
                 <div className="input-group">
                     <ReactPhoneInput className="form-control"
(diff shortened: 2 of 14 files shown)


src/components/forms/checkbox.jsx
```diff
@@ -15,9 +15,7 @@ var Checkbox = React.createClass({
             this.props.className
         );
         return (
-            <div className={classes}>
-                <FRCCheckbox {... this.props} />
-            </div>
+            <FRCCheckbox elementWrapperClassName={classes} {... this.props} />
```
rschamp: Why did you decide to move the class to an internal part of the component?  I would expect the `className` I give to a component to be applied to the outermost element of that component so that I can target anything within the component under that `className`, e.g.
```jsx
<Checkbox className="my-unique-checkbox" />
```
```scss
.my-unique-checkbox {
    // Styles for the outermost element
    .help-block { // Or some other sibling to the element wrapper
        // ...
    }
}
```

Most importantly doing it this way makes it so you don't have to know how the checkbox component works in order to add outermost block-level styles to it (`float`, etc.). Currently, wouldn't you need to add styles to the row classname, and then it wouldn't be namespaced to the className you gave to your component?

mewtaylor: I did it this way so that the stylings could be uniformly applied to elements inside of `.form-group` elements. The way FRC works, it builds a `Row` around everything, so I was seeing this more the outermost wrapper around the element itself, even if it wasn't outside of `Row`. It made more sense for me for it to be `.form-group > .[element]` rather than `.[element] > .form-group`.

The other reason is because this is the only place we do this – I added that wrapper in the initial commit of this feature because applying `className` directly onto `FRCCheckbox` wasn't working. We had this set up, applying `className` to FRC elements to provide custom formatting, everywhere. However, I discovered in this bug that FRC was basically ignoring all of those classnames, and not applying them, since that's not how it works. 

When I learned how `elementWrapperClassName` worked, it seemed to make more sense to me to use that, rather than adding another `div` wrapper around all FRC components to make this uniform.

rschamp: I don't agree that we want `.form-group > .[element]` (I think it's better to think `.form-group > .[component]`) since that's inverted in terms of scoping. It makes a particular component's `.form-group` inaccessible without affecting _all_ `.form-group`s. How would you put a border around one checkbox's row, but not all of them on the page?

I agree with getting rid of the `<div></div>` wrapper, but I think you want the `className` to become the `rowClassName` not the `elementWrapperClassName`.

src/components/forms/phone-input.jsx
```diff
@@ -53,9 +53,9 @@ var PhoneInput = React.createClass({
                                      label={null}
                                      disabled={this.isFormDisabled() || this.props.disabled}
                     />
+                    {this.renderErrorMessage()}
                 </div>
                 {this.renderHelp()}
```
rschamp: I was wrong about this being outside the row on the others: https://github.com/twisty/formsy-react-components/blob/master/src/input.js#L92 https://github.com/twisty/formsy-react-components/blob/master/src/textarea.js#L59

mewtaylor: Cool – I'll move it in!

src/components/forms/textarea.jsx
```diff
@@ -17,7 +17,7 @@ var TextArea = React.createClass({
         return (
             <FRCTextarea {... this.props}
                          className={classes}
-                         elementWrapperClassName="textarea-group" />
+                         rowClassName="textarea-row" />
```
rschamp: Shouldn't this get classes?

mewtaylor: not the way it was originally set up – `.textarea` is applied to style the `textarea` element itself, rather than wrappers, since it need to override defaults. We could switch it so that custom styling applies to the wrapping row, if we want to. I'd be fine with that.

rschamp: If you would, could you update the styles for all of the input components so it works in the way we want, where the `className` is applied to the top-level element?  It looks like even `Input` ignores `className` right now.

rschamp: Oh nvm, I was looking at something outdated

mewtaylor: I think you're still right about `Input` – must've missed that when I was doing the others, I'll do that now

rschamp: Thanks!

rschamp:
@mewtaylor While you're working on it, just found this one:
image

mewtaylor:
@rschamp already fixed

src/components/registration/steps.jsx
```diff
@@ -315,6 +315,7 @@ module.exports = {
                             <div className="gender-input">
                                 <Input name="user.genderOther"
                                        type="text"
+                                       className="no-label"
```
rschamp: Rather than needing to know about this class, could `Input` just give the class to itself if there is no `label` prop?

thisandagain:
@mewtaylor PR failing CI due to trailing spaces.

PR #761, opened 2016-07-26
Fix GH-757: Redirect to `/educators/classes` if a confirmed teacher

### LINKED ISSUE #757: "Empty" waiting room should redirect to "My Classes"
Staging, IE 11 on Windows 8.1

This one is a little hard to repro on staging, but the use case is pretty clear: imagine being a teacher who is attempting to create a Teacher Account. You go through the process and verify your email address. You then need to wait the X hours to have your account verified but you are antsy and keep the "Waiting Room" page open in a tab and keep hitting "refresh" from time to time to see if it has gone through yet.

## Repro
- Create new Teacher Account
- Complete email verification and return to "Waiting Room"
- Once waiting period ends, hit "refresh"

## Expected
- Redirected to "My Classes"

## Actual
- See "empty" waiting room with link to resources

## Reference
image

mewtaylor:
Fixes #757. Note: Because we are getting this info from the session, the redirect may not happen until the page is rendered first in cases where the connection is slow, or the db is slow.

### Test Cases ###
* With a confirmed teacher account, try to go directly to the `waiting` page. It should redirect you to "My Classes"

### DIFF
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -763,14 +763,8 @@ module.exports = {
             return {
                 email: null,
                 invited: false,
-                confirmed: false,
-                educator: true
+                confirmed: false
             };
-        },
-        componentWillReceiveProps: function (nextProps) {
-            if (nextProps.educator && nextProps.confirmed) {
-                window.location.href = '/educators/classes/';
-            }
         },
         render: function () {
             return (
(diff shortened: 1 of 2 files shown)


src/components/registration/steps.jsx
```diff
@@ -762,9 +762,16 @@ module.exports = {
...
                 email: null,
-                invited: false
+                invited: false,
+                confirmed: false,
+                educator: true
             };
         },
+        componentWillReceiveProps: function (nextProps) {
+            if (nextProps.educator && nextProps.confirmed) {
+                window.location.href = '/educators/classes/';
```
rschamp: This behavior belongs in the view rather than the component

Issue #217, opened 2015-11-06
Navigation - Make top navigation and footer responsive


thisandagain:
All of the work we are doing to integrate grid systems and flex box into our new views is limited by the fact that our navigation is not responsive which leads to some bugs (see: #214) as well as a less than ideal viewing experience on mobile devices.

- [x] Design responsive top navigation
- [x] Design (or eliminate) responsive footer
- [ ] Implement updated top navigation
- [ ] Implement updated footer

/cc @carljbowman @kaschm @mewtaylor @rschamp

thisandagain:
@carljbowman Can you share specs with @mewtaylor when you are ready? It would be nice to get this wrapped up this iteration.

carljbowman:
## Responsive Nav / Footer 
*omg it's happening* 🎉

**Overview Deck: [R3_Responsive Nav & Footer.pdf](https://github.com/LLK/scratch-www/files/354249/R3_Responsive.Nav.Footer.pdf)**

--

### Navigation Logged Out

- [ ] Change 'Join Scratch' to 'Sign up' verbiage across the site: navigation, sign up flow
- [ ] Change 'Sign in' to 'Log in' verbiage across the site: navigation, login dropdown, editor nav

#### 12 cols
![r3_responsive nav footer_nav - desktop - logged out](https://cloud.githubusercontent.com/assets/3409578/16688429/0b1bfc5a-44ec-11e6-9919-d6543122f10a.png)

#### 8 cols 
![r3_responsive nav footer_nav - tablet to desktop - logged out](https://cloud.githubusercontent.com/assets/3409578/16688439/1b56bc90-44ec-11e6-92c2-ba01887633df.png)

#### 6 cols
![r3_responsive nav footer_nav - mobile to tablet - logged out](https://cloud.githubusercontent.com/assets/3409578/16688461/2e1a998c-44ec-11e6-9294-07235300bbac.png)

#### 4 cols
![r3_responsive nav footer_nav - mobile - logged out](https://cloud.githubusercontent.com/assets/3409578/16688469/368d7e40-44ec-11e6-85b9-ae7dcad98db5.png)

-----

### Navigation Logged In

#### 12 cols
![r3_responsive nav footer_nav - desktop - logged in](https://cloud.githubusercontent.com/assets/3409578/16688498/50eacedc-44ec-11e6-98d4-ddf07dda485e.png)

#### 8 cols
[...]

PR #767, opened 2016-07-27
[WIP] Fix GH-217: Implement responsive nav and footer

technoboy10:
This PR is a first pass at implementing the responsive navigation bar and footer outlined in #217. It also modifies the carousel and splash styles to make the splash page usable on mobile devices.

The PR adds the 'react-responsive' package as a dependency. My reasoning behind adding the dependency was 
* The 4-6 column layout of the footer required a different layout in the JSX and could not be implemented using SCSS media queries
* Controlling what elements were hidden in the nav bar in pure SCSS decreased readability of the scss file significantly. 
* Using the MediaQuery component provided by react-responsive allows separating responsive layout vs responsive styling (it also removes hidden items from the DOM instead of just setting `display: none` on them, which might marginally improve performance)


technoboy10:
I'm not sure why the localization checks are failing... :/

mewtaylor:
@technoboy10 looks like it's a problem in `develop`, rather than your branch. Looking into it...

technoboy10:
Talked with @carljbowman and we decided to put the verbiage changes in a separate issue/PR since it's causing the tests for this PR to fail.

src/components/footer/www/footer.jsx
```diff
@@ -4,165 +4,206 @@ var FormattedMessage = require('react-intl').FormattedMessage;
...
+                                </a>
+                            </dd>
+                            <dd>
+                                <a href="http://www.scratchfoundation.org/">
+                                    <FormattedMessage id='general.scratchFoundation' />
+                                </a>
+                            </dd>
+                        </dl>
+                    </div>
+                </MediaQuery>
```
rschamp: There's a lot of duplication here, which will make this annoying to maintain — especially if this is the example the rest of the site would need to follow.

rschamp:
@technoboy10 I'm not sold on the `MediaQuery` component. Could you provide a comparison of the CSS-only and `MediaQuery` component methods for a specific case? Maybe the footer because I don't see why the component helps us more than a few `display: none`s within media queries there.

technoboy10:
@rschamp The main issue in the footer is that the layout in JSX for the footer links can't be turned into the layout in the spec by just hiding elements in CSS. The best I was able to get was this: 
screenshot 2016-08-01 13 42 13

(Those links are in two columns in the spec, but in JSX they're laid out in five columns)

technoboy10:
@rschamp The nav probably doesn't need the MediaQuery component, so maybe only using MediaQuery in the footer makes sense?

src/components/carousel/carousel.jsx
```diff
@@ -27,12 +27,29 @@ var Carousel = React.createClass({
...
+            centerMode: false,
             dots: false,
             infinite: false,
             lazyLoad: true,
             slidesToShow: 5,
             slidesToScroll: 5,
-            variableWidth: true
+            variableWidth: true,
+            responsive: [
+                {breakpoint: 480, settings: {
```
mewtaylor: Could we get these values from frameless too?

technoboy10: @mewtaylor Definitely!

src/components/footer/www/footer.jsx
mewtaylor: I'm not convinced that this will be an example the rest of the site has to follow. This feels more like a unique case than a standard case, given the way the content reduces.

The only other way this could feasibly work, from what i can see, is if we were to restructure the footer so that the column headers were their own `FlexRow`, and then all the rest of the content was in one giant `FlexRow` that was column oriented, and overflowed into the next column. Then you could potentially use `display: none`.

However, that might be really tricky to line up correctly with column headers.

src/components/page/www/page.jsx
```diff
@@ -13,13 +13,13 @@ var Page = React.createClass({
         return (
             <div className="page">
                 <div id="navigation" className={classes}>
-                    <Navigation />
+                    <Navigation/>
```
mewtaylor: I was under the impression the space in here was one of our style conventions. Thoughts @rschamp and @technoboy10 ?

src/routes.json
```diff
@@ -3,7 +3,8 @@
         "name": "splash",
         "pattern": "^/?$",
         "view": "splash/splash",
-        "title": "Imagine, Program, Share"
+        "title": "Imagine, Program, Share",
+        "viewportWidth": "device-width"
```
mewtaylor: yay :)

src/components/page/www/page.jsx
technoboy10: Oops, I meant to add the space back in (for some reason it didn't commit?).

mewtaylor: yeah that happens sometimes, certainly done that myself!

technoboy10:
I did some research about using flexbox or CSS3 columns to potentially get rid of the MediaQuery component in the footer, but 
- CSS3 columns and column-break are only supported (barely) in Chrome, so that's not an option
- Using flexbox appears to work, but trying to get the columns to break before the next section isn't really possible without being hacky (especially since the columns don't have equal numbers of items in them).
screenshot 2016-08-02 11 33 55

So it seems that using the MediaQuery component is the best option for the footer.

mewtaylor:
@technoboy10 thanks for the change on the nav bar! IMHO, I think that this looks good to go now, with using `MediaQuery` to display the footer. As soon as @rschamp is okay with this, I'd say it's good to go.

rschamp:
This looks good to me, thanks @technoboy10!

Thanks for avoiding `MediaQuery` where possible — I see why it's necessary now for the footer design. At some other time maybe we can work with @carljbowman on adjusting the design so that it can work with CSS alone.

thisandagain:
This looks great. Thanks for getting this done @technoboy10

PR #778, opened 2016-07-28
Set common y-origin for validation bubbles

### LINKED ISSUE #770: Wrong card used for register link
This context should use the old card. /cc @rschamp 

Repro:
- Go to https://staging.scratch.mit.edu/classes/80/register/50c71e56fafa49a783bcf93f4b3fde04

screen shot 2016-07-27 at 5 17 14 pm

From https://github.com/LLK/scratch-www/pull/765

rschamp:
I believe this will keep the misalignment from happening across browsers. Sets the top left corner to the top left corner of the input block.

Tested in Chrome 52 and confirmed this fixes the issue there, but Sauce was having network issues so I wasn't able to check this in other browsers yet. I think the absence of the `top` declaration was the source of the pain for this though.

Resolves #770

### DIFF
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -25,8 +25,9 @@
         $arrow-border-width: 1rem;
         display: block;
         position: absolute;
+        top: 0;
         left: 0;
-        transform: translate(20rem, -4rem);
+        transform: translate(16rem, 0);
         margin-left: $arrow-border-width;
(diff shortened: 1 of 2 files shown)


rschamp:
Merging for QA, @mewtaylor please review though

mewtaylor:
👍 thanks @rschamp – odd about that switch to chrome 52, but this definitely looks helpful!

Issue #773, opened 2016-07-27
Type of Organization includes "afterscool" typo - Teacher Registration


jwzimmer:
<img width="224" alt="screen shot 2016-07-27 at 6 25 26 pm" src="https://cloud.githubusercontent.com/assets/1320456/17194733/9830cc48-5427-11e6-9d1e-541966971d2c.png">
<img width="481" alt="screen shot 2016-07-27 at 6 25 48 pm" src="https://cloud.githubusercontent.com/assets/1320456/17194732/982a5f2a-5427-11e6-9c96-39b0e0c32876.png">

mewtaylor:
fixed in https://github.com/LLK/scratch-www/pull/779

PR #779, opened 2016-07-28
Fix GH-[772,773]: Fix “afterschool” typo, default country and margin for username

### LINKED ISSUE #772: Country dropdown defaults to United States for student & teacher registration
In the spec it defaults to "select country" everywhere, I think. (See https://github.com/LLK/scratchr2/issues/3679.)

In Address in Teacher Registration, it defaults to "select country", but in the demographic info for students & teachers, it defaulted to United States.
screen shot 2016-07-27 at 6 28 33 pm

/cc @carljbowman

mewtaylor:
This fixes #772 and #773.

### DIFF
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -124,12 +124,14 @@ module.exports = {
                     <Card>
                         <Form onValidSubmit={this.onValidSubmit}>
                             <div>
-                                <b>{formatMessage({id: 'registration.createUsername'})}</b>
-                                {this.props.usernameHelp ? (
-                                    <p className="help-text">{this.props.usernameHelp}</p>
-                                ):(
-                                    null
-                                )}
+                                <div className="username-label">
+                                    <b>{formatMessage({id: 'registration.createUsername'})}</b>
+                                    {this.props.usernameHelp ? (
+                                        <p className="help-text">{this.props.usernameHelp}</p>
+                                    ):(
+                                        null
+                                    )}
+                                </div>
                                 <Input className={this.state.validUsername}
                                        type="text"
                                        name="user.username"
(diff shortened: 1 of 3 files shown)


src/components/registration/steps.jsx
```diff
@@ -280,6 +282,17 @@ module.exports = {
...
+                label: formatMessage({id: 'teacherRegistration.selectCountry'}),
+                disabled: true,
+                selected: true
+            }).sort(function (a, b) {
+                if (a.disabled) return -1;
+                if (b.disabled) return 1;
+                if (a.value === this.props.defaultCountry) return -1;
+                if (b.value === this.props.defaultCountry) return 1;
+                return 0;
+            }.bind(this));
```
rschamp: Isn't this taken from elsewhere? Can it be put in a function and reused?

mewtaylor: I tried that – the issue I was running into was the need for `formatMessage`, which we inject at the class level. I was having trouble finding a way to abstract that away, do you have any thoughts on it?

rschamp: I believe it would work to just use `<FormattedMessage>`. Or you may be able to just import `formatMessage` from react-intl? Not sure if that function can be used standalone.

mewtaylor: right – I forgot that that component can work like that outside of render 👍

src/components/registration/steps.jsx
```diff
@@ -27,6 +27,17 @@ var Tooltip = require('../../components/tooltip/tooltip.jsx');
...
 
 var DEFAULT_COUNTRY = 'us';
+var COUNTRY_OPTIONS = countryData.countryOptions.concat({
+    label: <intl.FormattedMessage id="teacherRegistration.selectCountry" />,
+    disabled: true,
+    selected: true
+}).sort(function (a, b) {
+    if (a.disabled) return -1;
+    if (b.disabled) return 1;
+    if (a.value === DEFAULT_COUNTRY) return -1;
```
rschamp: This makes US always be at the top, but the previous behavior would put the user's last country selection at the top. If it were a function that took a default_country argument, you could pass it `this.props.default_country` and it would behave exactly as before.

mewtaylor: I didn't realize that's how it was working before – I didn't see that `teacherregistration` was doing that. Why don't we just have the value selected be the country from before, rather than sorting it to the top?

rschamp: I guess we can do that, but it's not how the spec was, and not how it was behaving before.

mewtaylor: ok – i'll just go do default then.

carljbowman: The desired behavior is to populate the Country field on Address with the previously selected Country from Personal Information

PR #786, opened 2016-07-28
Fix GH-775: Fix gender input error alignment in IE11

### LINKED ISSUE #775: Misaligned validations in student join flow in IE11
I put this in a separate issue from the Chrome validations, but we can combine them if that's preferred.

IE11, Win7 VM, Staging

For example:
- Repro by following steps at https://staging.scratch.mit.edu/classes/80/register/50c71e56fafa49a783bcf93f4b3fde04
screen shot 2016-07-27 at 6 55 27 pm

mewtaylor:
Also re-apply the arrow styling on the organization checkbox group, since it appears to be aligned correctly. Fixes #775.

/cc @technoboy10


src/components/registration/steps.scss
```diff
@@ -212,5 +212,57 @@
...
+
+                        border-bottom: 1px solid $active-gray;
+                        border-left: 1px solid $active-gray;
+                        border-radius: 5px;
+
+                        background-color: $ui-orange;
+                        width: $arrow-border-width;
+                        height: $arrow-border-width;
+
+                        content: "";
```
rschamp: This is pretty messy to have it all duplicated — maybe create a mixin for this?

rschamp: Alternately, be specific about which input types should receive the first override so you don't have to un-override here.

mewtaylor: I just realized it doesn't need to be all duplicated, actually – just the `&before` and a couple other things. I'd probably rather do that than a mixin, thoughts?

mewtaylor: Could be wrong about this – but if we're going with a general rule of it going underneath in IE11, wouldn't it be better to keep it general like this, and just override in the 1 (maybe 2) cases it needs to be?

rschamp: Ideally if it's a general rule, but if it's not then it's better to only override the ones you need to override so you don't have to reverse the rules immediately after.  If the arrow is ok in some cases, I wonder if it's truly impossible to fix in the others?

Also, it looks like `:not()` is supported all the way back to IE 9, so you could also try that for the exclusions: http://www.w3schools.com/cssref/sel_not.asp

rschamp: The reason why I think you shouldn't override then un-override is that the un-override is just a more complicated version of copy-pasting, so when things change you're even less likely to update the corresponding rule in the un-override.  Having browser-specific hacks is messy enough to not have to also deal with confusing browser-specific hacks.

mewtaylor: +1 to giving it a try by remove the browser-specific styling, now that we have a bit more time to look at it. I'll try for the `not` thing otherwise, thanks I didn't know about that!

rschamp:
Hurray for fixing things by accident!

jwzimmer:
@mewtaylor I'm still seeing the gender field misaligned in the x direction (Staging, Firefox 48 on Win 7 VM):
screen shot 2016-08-19 at 1 19 51 pm

mewtaylor:
@jwzimmer sorry the title here is not specific enough – this was to change the alignment from being "mobile" styled all the time, to appearing with an arrow next to it.

The issue you're referring to here is being tracked already, I believe, in https://github.com/LLK/scratch-www/issues/775#issuecomment-236697946

jwzimmer:
@mewtaylor Okay, so it fixes everything in #775 except that last part (the x alignment). Thanks.

mewtaylor:
Patch for the L/R misalignment in https://github.com/LLK/scratch-www/pull/858

PR #793, opened 2016-08-01
[Master] Add "Request Account" button to Educators landing page

rschamp:
To be deployed on August 3.

### DIFF
--- a/src/components/subnavigation/subnavigation.jsx
+++ b/src/components/subnavigation/subnavigation.jsx
@@ -9,10 +9,21 @@ require('./subnavigation.scss');
  */
 var SubNavigation = React.createClass({
     type: 'SubNavigation',
+    getDefaultProps: function () {
+        return {
+            align: 'middle'
+        };
+    },
     render: function () {
         var classes = classNames(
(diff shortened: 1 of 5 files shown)


src/components/subnavigation/subnavigation.jsx
```diff
@@ -9,10 +9,21 @@ require('./subnavigation.scss');
...
         var classes = classNames(
-            'sub-nav',
-            this.props.className
+            [
+                'sub-nav',
+                this.props.className
+            ],
+            {
+                'sub-nav-align-left': this.props.align === 'left',
+                'sub-nav-align-right': this.props.align === 'right'
```
mewtaylor: do we have a case where this would be `right`?

mewtaylor:
One question, and looks like a failing check on an empty definition, but once those are resolved 👍

src/components/subnavigation/subnavigation.jsx
rschamp: Not yet, but it would be weird if the align prop took "left" but not "right" as an argument.

PR #795, opened 2016-08-01
[Master] Hotfix: Create space between paragraph elements

### LINKED ISSUE #792: Add margins to schedule elements on Scratch Conference page
Mac, Chrome Version 52.0.2743.82 (64-bit)

The names are very close to the flag icon:
https://scratch.mit.edu/conference/schedule#thursday
screen shot 2016-08-01 at 1 59 21 pm

@carljbowman's comment about this:
> We should have a margin on those elements. If we are going to hotfix stuff to the conference site this week, we should update. If the hot fix happened then its not worth fixing.

mewtaylor:
Fixes #792


carljbowman:
👍 Looks good to me. Thanks for tweaking this.

Issue #797, opened 2016-08-01
Type of Organization > Other is not a required field but Type of Organization is - Teacher Registration


jwzimmer:
Chrome 52, Mac

If you choose the "Other" option for Type of Organization during Teacher Registration, you don't have to fill out the Other field. Since Type of Organization is required and Other is a field within it rather than just the option "Other" (text), I think this is a little bit inconsistent, but not the worst behaviour... Is it supposed to be required, @carljbowman & @kaschm?

<img width="441" alt="screen shot 2016-08-01 at 3 42 25 pm" src="https://cloud.githubusercontent.com/assets/1320456/17306632/0fa94e8c-57ff-11e6-826e-5814d1c3a951.png">

kaschm:
I think it should be required. @carljbowman ?

carljbowman:
I agree. If the user selects 'Other', then the field should be required and if they attempt to go to the next step having not entered anything a tooltip should prompt them to enter something. 

Long term, we should create a type style ( italic ), for form fields that are pre-populated with a action. For example: 'Select Country' or 'Other'. Something to distinguish the content is temporary.

rschamp:
This used to work — it's a regression.

mewtaylor:
Fixed in https://github.com/LLK/scratch-www/pull/803

Issue #712, opened 2016-07-09
Timing for validation for too-long phone number is confusing - Teacher Registration


jwzimmer:
I don't see the validation for a longer phone number, on production, in Chrome, until after I click "Next" (which is confusing since you do see the validations for other kinds of incorrect input in the phone field as soon as you enter it).

(This should go along with the other cases relating to when validations should be triggered, which should be considered, as @carljbowman [said](https://github.com/LLK/scratch-www/issues/683).)

Number is invalid:
<img width="639" alt="screen shot 2016-07-09 at 6 36 51 pm" src="https://cloud.githubusercontent.com/assets/1320456/16710654/4e97c0a4-4604-11e6-90f9-4f30e651fdd9.png">

Add a 1, and it appears to be valid:
<img width="506" alt="screen shot 2016-07-09 at 6 36 39 pm" src="https://cloud.githubusercontent.com/assets/1320456/16710653/4e9601c4-4604-11e6-869a-8ac54d064f47.png">

Update:
**No error on Next, I can proceed to next step**
![out](https://cloud.githubusercontent.com/assets/1320456/18144244/b32d873e-6f93-11e6-9963-8c659c170b4b.gif)

mewtaylor:
Fixed in https://github.com/LLK/scratch-www/pull/803

mewtaylor:
The fix from #803 ended up causing some issues with accessibility and international phone numbers (see #816). So, this needs a separate solution

mewtaylor:
We are treating this as no longer an issue, since we no longer have a custom length checker, and the length issue seems to be handled by our validator: https://github.com/LLK/scratch-www/issues/691#issuecomment-243889806

Issue #705, opened 2016-07-08
Type of Organization validation should be triggered at the same time as other validations on the same page - Teacher Registration


jwzimmer:
Chrome, ON PRODUCTION : )

It wasn't required at all before the fix to https://github.com/LLK/scratch-www/issues/695 but I could not get past that step in Chrome without submitting an answer to Type of Organization, so that is functionally fixed, but you don't see the validation if you click "Next" with other empty required fields... you see the validation if you click "Next" and it's the only empty required field...

Click "Next" without filling anything in this step out - you see validations for the top two fields but not for Type of Organization:
<img width="502" alt="screen shot 2016-07-08 at 10 15 13 am" src="https://cloud.githubusercontent.com/assets/1320456/16690330/590ab600-44f5-11e6-932c-6d2d52fcd33f.png">

Click "Next" after you have filled out the other two fields but haven't answered type of organization:
<img width="511" alt="screen shot 2016-07-08 at 10 15 27 am" src="https://cloud.githubusercontent.com/assets/1320456/16690329/5909e770-44f5-11e6-8b46-0d04000bdf17.png">

kaschm:
Related: https://github.com/LLK/scratch-www/issues/683, https://github.com/LLK/scratch-www/issues/712

mewtaylor:
Fixed in https://github.com/LLK/scratch-www/pull/803

PR #803, opened 2016-08-02
Fix GH-[683,705,712,797]: Updates to registration UX for teachers

### LINKED ISSUE #683: Teacher Registration - I don't see validations (e.g. for username already exists) until I click "Next"
Staging, Chrome

In teacher registration, I don't see the validation that I'm trying to sign up with an extant username until I click the Next button:
screen shot 2016-07-06 at 3 13 52 pm

In our other join flow, you see the verification that username exists right after you type it:
screen shot 2016-07-06 at 3 13 44 pm

Most of the validations for the fields in the teacher registration flow are not triggered until you click "next".

However, the phone number validation seems to be as soon as the system detects that the field has been typed in but the existing number isn't valid... e.g. if I type my real phone number, it comes up as "invalid" until the last digit... (which doesn't seem like obviously the best behaviour to me? Maybe it could wait until the user deactivates the field or otherwise signals they're "done" typing):
screen shot 2016-07-06 at 3 49 00 pm

Ideally, I think we would be consistent with when validations are triggered for fields across the platform. /cc @carljbowman

mewtaylor:
1. Phone too long validation occurs before submit now.
2. Organization “other” type is now being required
3. Organization type required now shows up at the same time as other organization validation errors
4. Username check now occurs earlier, on blur rather than on submit

Fixes #683, #705, #712, #797.

### Test Cases ###
* Try to register with a pre-existing username. It should show you an error after you focus out from the username field
* Try to type a phone number that's too long – it should show an error and prevent moving to the next step
* Try to submit without choosing an organization type or an organization name. The organization type error should appear
* Try to choose "Other" for organization without putting anything in the black input. An error should appear, and prevent you from moving to the next step.

### DIFF
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -82,37 +82,52 @@ module.exports = {
         onChangeShowPassword: function (field, value) {
             this.setState({showPassword: value});
         },
-        onValidSubmit: function (formData, reset, invalidate) {
-            this.setState({waiting: true});
+        validateUsername: function (username) {
             api({
                 host: '',
-                uri: '/accounts/check_username/' + formData.user.username + '/'
+                uri: '/accounts/check_username/' + username + '/'
             }, function (err, res) {
                 var formatMessage = this.props.intl.formatMessage;
-                this.setState({waiting: false});
-                if (err) return invalidate({all: err});
+                if (err) {
+                    this.refs.form.refs.formsy.updateInputsWithError({all: err});
+                    return false;
+                }
                 res = res[0];
                 switch (res.msg) {
                 case 'valid username':
                     this.setState({
                         validUsername: 'pass'
                     });
-                    return this.props.onNextStep(formData);
+                    return true;
                 case 'username exists':
-                    return invalidate({
+                    this.refs.form.refs.formsy.updateInputsWithError({
                         'user.username': formatMessage({id: 'registration.validationUsernameExists'})
                     });
+                    return false;
                 case 'bad username':
-                    return invalidate({
+                    this.refs.form.refs.formsy.updateInputsWithError({
                         'user.username': formatMessage({id: 'registration.validationUsernameVulgar'})
                     });
+                    return false;
                 case 'invalid username':
                 default:
-                    return invalidate({
+                    this.refs.form.refs.formsy.updateInputsWithError({
                         'user.username': formatMessage({id: 'registration.validationUsernameInvalid'})
                     });
+                    return false;
                 }
             }.bind(this));
+        },
+        onUsernameBlur: function (event) {
+            this.validateUsername(event.currentTarget.value);
+        },
+        onValidSubmit: function (formData) {
+            this.setState({waiting: true});
+            this.validateUsername(formData.user.username);
+            this.setState({waiting: false});
+            if (this.state.validUsername === 'pass') {
+                this.props.onNextStep(formData);
+            }
         },
         render: function () {
             var formatMessage = this.props.intl.formatMessage;
@@ -139,7 +154,7 @@ module.exports = {
                         )}
                     </p>
                     <Card>
-                        <Form onValidSubmit={this.onValidSubmit}>
+                        <Form onValidSubmit={this.onValidSubmit} ref="form">
                             <div>
                                 <div className="username-label">
                                     <b>{formatMessage({id: 'registration.createUsername'})}</b>
@@ -152,6 +167,7 @@ module.exports = {
                                 <Input className={this.state.validUsername}
                                        type="text"
                                        name="user.username"
+                                       onBlur={this.onUsernameBlur}
                                        validations={{
                                            matchRegexp: /^[\w-]*$/,
                                            minLength: 3,
@@ -473,8 +489,21 @@ module.exports = {
         onChooseOrganization: function (name, values) {
             this.setState({otherDisabled: values.indexOf(this.organizationL10nStems.indexOf('orgChoiceOther')) === -1});
         },
-        onValidSubmit: function (formData, reset, invalidate) {
+        validateOrganizationType: function (formData) {
             if (formData.organization.type.length < 1) {
+                return false;
+            }
+            return true;
+        },
+        onSubmit: function (formData, reset, invalidate) {
+            if (!this.validateOrganizationType(formData)) {
+                return invalidate({
+                    'organization.type': this.props.intl.formatMessage({id: 'teacherRegistration.validationRequired'})
+                });
+            }
+        },
+        onValidSubmit: function (formData, reset, invalidate) {
+            if (!this.validateOrganizationType(formData)) {
                 return invalidate({
                     'organization.type': this.props.intl.formatMessage({id: 'teacherRegistration.validationRequired'})
                 });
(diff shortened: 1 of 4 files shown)


src/components/registration/steps.jsx
```diff
@@ -473,8 +489,21 @@ module.exports = {
         onChooseOrganization: function (name, values) {
             this.setState({otherDisabled: values.indexOf(this.organizationL10nStems.indexOf('orgChoiceOther')) === -1});
         },
-        onValidSubmit: function (formData, reset, invalidate) {
+        validateOrganizationType: function (formData) {
```
rschamp: If this were a regular validation for the organization CheckboxGroup, then wouldn't it automatically be run `onSubmit` (and ahead of `onValidSubmit`)?

src/components/registration/steps.jsx
```diff
@@ -82,38 +82,53 @@ module.exports = {
...
                 }
             }.bind(this));
         },
+        onUsernameBlur: function (event) {
+            this.validateUsername(event.currentTarget.value);
+        },
+        onValidSubmit: function (formData) {
+            this.setState({waiting: true});
+            this.validateUsername(formData.user.username);
+            this.setState({waiting: false});
```
rschamp: Here `validateUsername` is treated like it's synchronous, but that method uses an asynchronous API call. The `waiting` state change and `onNextStep` calls should happen in a callback. If this does work locally, it's because `this.validateUsername(formData.user.username)` returns in time for the rest of the function to continue due to the speed of the local connection. It would surely break running remotely.

src/components/registration/steps.jsx
```diff
@@ -82,38 +82,53 @@ module.exports = {
...
                         'user.username': formatMessage({id: 'registration.validationUsernameVulgar'})
                     });
+                    return false;
                 case 'invalid username':
                 default:
-                    return invalidate({
+                    this.refs.form.refs.formsy.updateInputsWithError({
                         'user.username': formatMessage({id: 'registration.validationUsernameInvalid'})
                     });
+                    return false;
```
rschamp: Does the return of `true`/`false` have any effect, or could this be reduced to just return the statement above?

src/components/registration/steps.jsx
mewtaylor: whoops – meant to put a callback in there instead. 👍

src/components/registration/steps.jsx
mewtaylor: once I add the callback (which I forgot to do), `true/false` handles the difference between `onValidSubmit` and `onSubmit`. Since we don't want to call `onNextStep` for `onSubmit`, but do for `onValidSubmit`.

src/components/registration/steps.jsx
mewtaylor: IIRC the `required` prop does not work as we expect in this case for `CheckboxGroup`. I'm not sure if we could implement a validator in `validations.jsx` directly, but will take a look.

src/components/registration/steps.jsx
```diff
@@ -494,7 +515,7 @@ module.exports = {
                                  tipContent={formatMessage({id: 'registration.nameStepTooltip'})} />
                     </p>
                     <Card>
-                        <Form onValidSubmit={this.onValidSubmit}>
+                        <Form onValidSubmit={this.onValidSubmit} onSubmit={this.onSubmit}>
```
rschamp: Now this can be `onValidSubmit={this.props.onNextStep}` and the `onValidSubmit` method can go away.

src/components/registration/steps.jsx
```diff
@@ -82,38 +82,63 @@ module.exports = {
...
                 default:
-                    return invalidate({
+                    this.refs.form.refs.formsy.updateInputsWithError({
                         'user.username': formatMessage({id: 'registration.validationUsernameInvalid'})
                     });
+                    break;
+                }
+
+                if (typeof callback !== 'undefined') {
+                    callback(isValid);
```
rschamp: One last suggestion that I think is just stylistic — if you make `callback` optional on the first line of the function with `callback = callback || function () {};`, then you can always just `return callback(true/false)` without checking the type of `callback`.

If the reason you changed the style here was to avoid repeating the `if (typeof callback !== 'undefined') {...}` in every `case`, then this would allow you to just return within each `case` too. Personally I find the inline returns more readable than the fall-through style, since you know what happens as soon as you see a `return`.

jwzimmer:
Testing, Staging @mewtaylor 

**Chrome Version 52.0.2743.82 (64-bit), Mac OS X El Capitan**
- I'm still seeing the phone number issue from #712; although now I don't see the validation when I click "Next", so I can enter a too-long phone number and get to the next step.
out
- I noticed I can type in the Type of Organization > Other field when it's checked, then uncheck it, and whatever I typed stays. This seems not ideal to me. It didn't seem to cause an error at the end of registration, but I don't think it's the standard behaviour.
screen shot 2016-08-22 at 11 00 59 am
- The Type of Organization does occur at the same time as the other validations on that page now, so that's working.

jwzimmer:
Testing, Staging
@mewtaylor 

* Just from typing in a few known usernames to test out the validation (I didn't click next, just clicked in between username and password fields, I got this error:
screen shot 2016-08-22 at 5 58 21 pm
screen shot 2016-08-22 at 5 57 02 pm

jwzimmer:
In Safari on Mac (see comment above), I'm still seeing this error after clearing cache and history. The repro steps are: go to `educators/register`, click Next, put your cursor in username field, click Next again.
screen shot 2016-08-23 at 1 30 44 pm

mewtaylor:
@jwzimmer the issue appears to be that it attempts to check the empty username once you engage the username field, and is finding an empty string. I'm not 100% sure, but I *think* that this would've existed before this particular fix, since all this fix does is add a check on focus out.

Adding an additional check to make sure the username check has a value at all before sending it to the server seems reasonable, and could be done for the release, if desired.

jwzimmer:
@mewtaylor well, since the validation ("oops...") goes away if you fill out the fields, and it doesn't seem to cause an error at the end of registration, I'm fine with opening a separate case for it.

mewtaylor:
@jwzimmer sounds good to me!

jwzimmer:
@mewtaylor Just to clarify - it does seem like this error is more likely to be encountered by a user because of this PR, right? Since the username check happens on blur, rather than clicking "Next", the user is more likely to encounter this error (e.g. if they enter a username, then start typing a password, then clear the username because they see the "this username is already in use" validation, then finish the password...)?

mewtaylor:
@jwzimmer yeah that is true, it'll be more apparent in that case

jwzimmer:
@mewtaylor Okay, thanks. I think separate case is fine, but I think it would be good to get to https://github.com/LLK/scratch-www/issues/872 in the next milestone... what do you think?

mewtaylor:
@jwzimmer totally agree!

jwzimmer:
Just wanted to doublecheck the test cases one last time, Chrome 52 on Mac OS X El Capitan:
- [x] Try to register with a pre-existing username. It should show you an error after you focus out from the username field
- [x] Try to submit without choosing an organization type or an organization name. The organization type error should appear
- [x] Try to choose "Other" for organization without putting anything in the black input. An error should appear, and prevent you from moving to the next step.

PR #826, opened 2016-08-10
[Master] Hotfix: Remove `phoneLength` validation

### LINKED ISSUE #816: International phone number parsing
teacheraccountbug

As reported by @LiFaytheGoblin  the format above should be valid, no?

mewtaylor:
It appears that some valid phone numbers in some countries differ in length from what is specified in the util’s phone number format. Fixes #816.

### DIFF
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -397,6 +397,14 @@ module.exports = {
                 waiting: false
             };
         },
+        onValidSubmit: function (formData, reset, invalidate) {
+            if (!formData.phone || formData.phone.national_number === '+') {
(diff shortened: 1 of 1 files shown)


rschamp:
Simply removing it will reopen https://github.com/LLK/scratch-www/issues/693, won't it? How do we ensure it's still required while not invalidating valid numbers?

mewtaylor:
@rschamp ah – it looks like this will work in the develop branch perfectly, but there is one issue with the master branch fix – if you're in the U.S., and you never focus on the phone number field but just move on, it'll do what you're saying. I'll update this master version to reflect that, thanks!

PR #827, opened 2016-08-10
[Develop] Hotfix: Remove `phoneLength` validation

mewtaylor:
See #826

### DIFF
--- a/src/components/forms/phone-input.jsx
+++ b/src/components/forms/phone-input.jsx
@@ -25,8 +25,7 @@ var PhoneInput = React.createClass({
     getDefaultProps: function () {
         return {
             validations: {
-                isPhone: true,
-                phoneLength: true
+                isPhone: true
             },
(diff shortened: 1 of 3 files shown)


rschamp:
Could you explain why this works differently than master? Why won't this just skip validation if the user's phone number is just `+`?

mewtaylor:
I was wrong, it was the same issue as on `master` – fixed now, thanks!

TheGrits:
What if you were to validate all phone numbers according to a global minimum length, and only allow numbers and `+` in the field?

mewtaylor:
@TheGrits you're absolutely right – and I think this fix will do just that. It won't allow a submission without any numbers present, meaning there is a global minimum length (with some validation of phone number structure by country still being done in the `isPhone` check).

TheGrits:
@mewtaylor Ah ok, that makes sense.

PR #829, opened 2016-08-10
[Master] Fix GH-828: check street number range for address

### LINKED ISSUE #828: Valid addresses being rejected as invalid - Teacher Registration
Reported via freshdesk no. 54398 and no. 54703.

> I tried a few times to register/sign up as educator. I got as far as the address page and the error message read "this doesn't look like a real address". It was a real address and I tried variations and to no avail! 

Metadata: 
- User: None 
- Operating System: MacOS Macintosh X 10.8.2 
- Browser: Chrome 49.0.2623.112 
- Flash Version: Flash 22.0 (release 0)

> I am trying to set up a teacher account but when I put in my school's address it will not let me go any further and says that it does not look like a real address but it is the address of my school. 

Metadata: 
- User: (see freshdesk ticket)
- Operating System: Windows NT 10.0 
- Browser: Chrome 52.0.2743.116 
- Flash Version: Flash 22.0 (release 0)

mewtaylor:
This makes use of a custom header used by Smarty Streets (https://smartystreets.com/docs/us-street-api#x-standardize-only) which uses street number ranges, rather than exact addresses, to return results. This can help if, for some reason, we have someone sign up whose post office listing has not been updated, for instance. 

Fixes #828.

### DIFF
--- a/src/lib/api.js
+++ b/src/lib/api.js
@@ -18,9 +18,7 @@ var urlParams = require('./url-params');
 module.exports = function (opts, callback) {
     defaultsDeep(opts, {
         host: process.env.API_HOST,
-        headers: {
-            'X-Requested-With': 'XMLHttpRequest'
-        },
+        headers: {},
         responseType: 'json',
         useCsrf: false
     });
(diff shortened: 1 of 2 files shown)


rschamp:
Since this prevents teachers from registering I think this would be good to hotfix.

PR #834, opened 2016-08-11
[Develop] Fix GH-828: check street number range for address

mewtaylor:
Updates to #830 after testin on staging.


src/lib/api.js
```diff
@@ -52,6 +49,8 @@ module.exports = function (opts, callback) {
                 opts.uri = parts[0] + '?' + qs;
 
             }
+        } else {
+            opts['X-Requested-With'] = 'XMLHttpRequest';
```
rschamp: 1) I think you meant `opts.headers['X-Requested-With']`
2) Might be better to do  

```javascript
if (opts.host === '') {
  defaults(opts.headers, {'X-Requested-With': 'XMLHttpRequest'});
}
```
at the top where we set up defaults. That way we could override this if we wanted to with `{headers: {'X-Requested-With': null}}` if we wanted to.

mewtaylor: whoops thanks!

As for doing that at the top – but then wouldn't we have to override it every time inside `if (opts.hots !== '')`? That's what I was trying to avoid.

PR #836, opened 2016-08-12
Fix GH-760: Add username `not equals` to password validation

### LINKED ISSUE #760: Student one-by-one registration flow lets me use my username as my password
Staging, Chrome, found while testing https://github.com/LLK/scratch-www/pull/744.

I didn't get any validation saying I couldn't use my username as my password (studentclass1):
screen shot 2016-07-26 at 2 16 20 pm

mewtaylor:
for Student Complete Registration. This fixes #760.

### Test Cases ###
* Sign in as a student, and be redirected to the registration page. You should not be able to use your own username as a password
* Regression tests: 
  * Should not be able to use "password" as your password
  * Password should be longer than 6 characters

### DIFF
--- a/src/components/forms/validations.jsx
+++ b/src/components/forms/validations.jsx
@@ -23,6 +23,7 @@ module.exports.validations = {
         return phoneNumberUtil.isValidNumber(parsed);
     }
 };
+module.exports.validations.notEqualsUsername = module.exports.validations.notEquals;
 
 module.exports.validationHOCFactory = function (defaultValidationErrors) {
     return function (Component) {
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -226,6 +226,7 @@ module.exports = {
     ChoosePasswordStep: intl.injectIntl(React.createClass({
         getDefaultProps: function () {
             return {
+                studentUsername: null,
                 showPassword: false,
                 waiting: false
             };
@@ -257,7 +258,7 @@ module.exports = {
                                    validations={{
                                        minLength: 6,
                                        notEquals: 'password',
-                                       notEqualsField: 'user.username'
+                                       notEqualsUsername: this.props.studentUsername
                                    }}
                                    validationErrors={{
                                        minLength: formatMessage({
@@ -266,7 +267,7 @@ module.exports = {
                                        notEquals: formatMessage({
                                            id: 'registration.validationPasswordNotEquals'
                                        }),
-                                       notEqualsField: formatMessage({
+                                       notEqualsUsername: formatMessage({
                                            id: 'registration.validationPasswordNotUsername'
                                        })
                                    }}
@@ -879,7 +880,11 @@ module.exports = {
             };
         },
         onNextStep: function () {
-            this.props.onNextStep();
+            this.props.onNextStep({
+                user: {
+                    username: this.props.studentUsername
+                }
+            });
         },
         render: function () {
             var formatMessage = this.props.intl.formatMessage;
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -226,7 +226,7 @@ module.exports = {
     ChoosePasswordStep: intl.injectIntl(React.createClass({
         getDefaultProps: function () {
             return {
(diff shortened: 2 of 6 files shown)


src/components/forms/validations.jsx
```diff
@@ -23,6 +23,7 @@ module.exports.validations = {
         return phoneNumberUtil.isValidNumber(parsed);
     }
 };
+module.exports.validations.notEqualsUsername = module.exports.validations.notEquals;
```
rschamp: Is this because we can't have multiple of the same validation on one field?

src/components/registration/steps.jsx
```diff
@@ -879,7 +880,11 @@ module.exports = {
             };
         },
         onNextStep: function () {
-            this.props.onNextStep();
+            this.props.onNextStep({
+                user: {
+                    username: this.props.studentUsername
```
rschamp: This seems convoluted - we're passing back the username that got passed in as a prop so that we can just pass it right back in to the next step. It would be more straightforward to pass the username we already have to both. It reminds me of the antipattern of setting state based on props - when you do that you are duplicating the source of truth: https://facebook.github.io/react/tips/props-in-getInitialState-as-anti-pattern.html

src/components/registration/steps.jsx
```diff
@@ -226,6 +226,7 @@ module.exports = {
     ChoosePasswordStep: intl.injectIntl(React.createClass({
         getDefaultProps: function () {
             return {
+                studentUsername: null,
```
rschamp: I don't think there's anything student-specific in here and we may use this in other non-student related contexts later - maybe `username` or `notEquals`?

src/components/forms/validations.jsx
mewtaylor: We can't have different error messages for the same validation on one field, yeah.

src/components/registration/steps.jsx
mewtaylor: Yeah – this is somewhat left over from when I was trying a previous approach of re-using 'notEqualsField`, and having the username be a field that was used in analysis (that was to try and avoid using `notEquals` in two places because of formsy constraints). I think that removing this and just using the prop in the next step is better too.

src/components/registration/steps.jsx
mewtaylor: I think it might good to go with `username` for now – I think it'd be helpful to have a bit more context than `notEquals`, and I'm not sure what else, in other contexts (like reset password or something), we would have that value be different from `username`.

src/components/forms/validations.jsx
rschamp: Oh right the error messages.  There's somewhere else where we hit this issue... it'd be nice if `validations` was more flexible, maybe by optionally taking a list of objects so you could have many of the same type of validations, with different messages for each.

mewtaylor: yeah – this probably goes in the list of things to consider for upstream PRs to formsy we might want to do :)

jwzimmer:
**Testing, Staging**

**Chrome 52, Mac OS X El Capitan**
- I added student accounts using teacher1992 account
- I signed in as d3ourg3804frhj
- I followed the steps and was not able to use my username as password, "password" as password, or less than 6 chars

**Firefox 48, Mac OS X El Capitan**
- I signed in as dfjeiorghejkd
- I followed the steps and was not able to use my username as password, "password" as password, or less than 6 chars

- In the regular New Scratcher join flow, I got a validation for using my username as my password
- In Teacher Account registration, I got a validation for using my username as my password
-

jwzimmer:
- In the regular New Scratcher join flow, I got a validation for using my username as my password
- In Teacher Account registration, I got a validation for using my username as my password
- Used generate student sign up link to get `/classes/101/register/807e430ebef7467d8857a25d5b289e5b`, I got a validation for using my username as my password

**IE 11 on Win 7 VM**
- Went to `/classes/80/register/50c71e56fafa49a783bcf93f4b3fde04`, got a validation for using my username as my password
- Went to `/classes/101/register/807e430ebef7467d8857a25d5b289e5b`, got a validation for using my username as my password

PR #841, opened 2016-08-12
[Master] Hotfix GH-835: Add maxLength validations when needed

### LINKED ISSUE #835: All fields need max length validations that match the server's expectations
Nearly every field has a max length in the DB, so the client side validation should reflect that.

mewtaylor:
This fixes #835 by adding maxlength validations for: first name, last name, organization name, organization role, organization url, gender input, address line 1, address line 2, zip code, city.

This also changes `getCountryOptions` to take in a required argument of `intl`, which is meant to be the currently injected `react-intl` object, so that it can localize the `select country` string on its own. This eliminates hundreds of console warnings, since the previously used `<FormattedMessage />` component would wrap each option in a `span`, which threw a warning for react. 

### Test Cases ###
* These fields should say they're too long after *more than 50* characters are typed:
  * first name, last name, organization name, organization role, city
* These fields should say they're too long after *more than 100* characters are typed: 
  * address line 1, address line 2
* `gender` should say it's too long after 25 characters
* `zip` should say it's too long after 10 characters
* `organization url` should say it's too long after 200 characters

### DIFF
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -35,7 +35,6 @@
         padding: 1rem;
         max-width: 18.75rem;
         min-height: 1rem;
-        max-height: 3rem;
         overflow: visible;
         color: $type-white;
 
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -27,12 +27,14 @@ var Tooltip = require('../../components/tooltip/tooltip.jsx');
 require('./steps.scss');
 
 var DEFAULT_COUNTRY = 'us';
-var getCountryOptions = function (defaultCountry, intl) {
-    if (typeof intl === 'undefined') {
-        intl = defaultCountry;
-        defaultCountry = undefined;
-    }
 
+/**
+ * Return a list of options to give to frc select
+ * @param  {Object} intl           react-intl, used to localize strings
+ * @param  {String} defaultCountry optional string of default country to put at top of list
+ * @return {Object}                ordered set of county options formatted for frc select
+ */
+var getCountryOptions = function (intl, defaultCountry) {
     var options = countryData.countryOptions.concat({
         label: intl.formatMessage({id: 'registration.selectCountry'}),
         disabled: true,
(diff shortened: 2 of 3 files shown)


jwzimmer:
(Still testing! : ) )
Tested on Staging

**Gender, Firefox 48.0**
- 25 char: fffffffffffffffffffffffff
- I put in 25 characters, no validation triggered
- I put in 26 characters by typing an extra char at the end, validation triggered
- I copy-pasted 26 characters as a single string, validation triggered
- I put in 1 character, no validation triggered
- I put in 0 char (with radio button clicked), "field required" validation triggered

**Address Line 1, Address Line 2, Zip Code, Safari Version 9.1.2 (11601.7.7)**
- 100 char: ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
- 10 char: ffffffffff

- 50 char: ffffffffffffffffffffffffffffffffffffffffffffffffff

mewtaylor:
@jwzimmer just to add here what we discussed before:
* I updated the `url` validation field to be 200 in the description, which is what it should be (sorry!)
* for zip code – keeping at 10 seems okay, since the longest non-hyphenated one is 10 digits (Iran), and all hyphenated ones can be written without the hyphen as an option too
* for validation message – agreed it's awkward. I tried to put something in that wouldn't seem like it was calling the person's input "invalid", since that seems like it could be insensitive if there are locales we're forgetting to include. @kaschm any thoughts on this?
* for org > other – should be 50 too. Sorry about that!

jwzimmer:
@kaschm, @mewtaylor for the validations, maybe...?
- "Sorry, this field requires fewer characters" (or "needs"?)
- "Sorry, this field can't contain that many characters"
- "Sorry, this field requires a shorter answer" (or "value")

kaschm:
@jwzimmer how about:

Sorry, you have exceeded the maximum character limit.

?

jwzimmer:
@kaschm sounds good to me, thanks 👍

jwzimmer:
(Not QA-OK yet, pending https://github.com/LLK/scratch-www/pull/846#issuecomment-239933505.)

PR #845, opened 2016-08-15
Add fur, sq and te to language list

mewtaylor:
Part of https://github.com/LLK/scratchr2/pull/3954 in order to add fur, sq and te to the languages list for people to use.


rschamp:
The change to pass the language from intl rather than `window` looks great! Would you mind breaking that out into its own PR? Unless I'm missing something it doesn't look related to the addition/reordering of languages.

mewtaylor:
@rschamp i put it in this one because the default language, if not found, is `0` on the list (which is no longer English). I think it could be done separately, but would require a change here anyways – which is why I found the issue/changed it here in this way in the first place.

rschamp:
Oh ok, then this looks great.

PR #846, opened 2016-08-15
For GH-835: Update max length copy

mewtaylor:
This is an update for #842 to add new copy based on feedback given on PR #841. Thanks @kaschm and @jwzimmer!

### DIFF
--- a/src/l10n.json
+++ b/src/l10n.json
@@ -139,7 +139,7 @@
     "registration.studentUsernameFieldHelpText": "For safety, don't use your real name!",
     "registration.usernameStepTitle": "Request a Teacher Account",
     "registration.usernameStepTitleScratcher": "Create a Scratch Account",
-    "registration.validationMaxLength": "Sorry, that is too many characters for us.",
+    "registration.validationMaxLength": "Sorry, you have exceeded the maximum character limit.",
     "registration.validationPasswordLength": "Passwords must be at least six characters",
     "registration.validationPasswordNotEquals": "Your password may not be \"password\"",
     "registration.validationPasswordNotUsername": "Your password may not be your username",


mewtaylor:
fyi @kaschm @jwzimmer

jwzimmer:
In Chrome, the message is overflowing the validation:
screen shot 2016-08-15 at 5 20 02 pm

I don't think we'd want to push to production with that issue, so I switched the labels from pr - good to merge to pr - needs review to make sure that wouldn't happen, let me know if you would prefer me signal that elsewise, @mewtaylor. : )

mewtaylor:
@kaschm thoughts on taking out the word "maximum"? If we want to keep it in I can change the max height of this though

jwzimmer:
Looks really good

Just a few questions, very minor questions:
1) Do we want the validation message to have a different shaped bubble for some of the fields, e.g. the Type of Org > Other validation is different than the Website validation:
screen shot 2016-08-16 at 9 34 07 am
2) Do we want to be consistent with telling people the max char length in the validation as we decided to do with the Use field, or do we want to treat that as a special case (since it is a more open-ended field)?
screen shot 2016-08-16 at 9 35 55 am

screen shot 2016-08-15 at 5 20 02 pm

mewtaylor:
@jwzimmer the issue with different sized bubbles is related to the offset issue being tracked in https://github.com/LLK/scratch-www/issues/775#issuecomment-236697946. As for the different messages – I think it's okay since that's the only open text area, whereas the others are input fields, but also happy to change it.

jwzimmer:
@mewtaylor Ah, okay re different sized bubbles. : )
For the different messages, I think having them be different is fine, since the fields do have different roles, as you said. So looks good to me. 👍

PR #847, opened 2016-08-16
For GH-835: eliminate max height of validation messsage

mewtaylor:
now that it’s not centered but rather top/down, we don’t need this anymore. This was breaking with the new max character copy.

### DIFF
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -35,7 +35,6 @@
         padding: 1rem;
         max-width: 18.75rem;
         min-height: 1rem;
-        max-height: 3rem;
         overflow: visible;
         color: $type-white;
 


jwzimmer:
Testing, Production

Chrome Version 52.0.2743.82 (64-bit), Mac OS X El Capitan
- Validation messages look good (the difference in the x direction is a known, tracked issue):
screen shot 2016-08-22 at 10 02 07 am
screen shot 2016-08-22 at 9 58 33 am

IE11 on Win 7 VM
- The validation looks good, although I did notice a displaced validation:
screen shot 2016-08-22 at 10 13 10 am

Safari Version 9.1.2 (11601.7.7) on Mac OS X El Capitan
- Validation looks good
screen shot 2016-08-22 at 10 35 33 am

PR #849, opened 2016-08-16
Add missing locale strings for localization

### LINKED ISSUE #840: "Educator Resources" button not translated in Teacher Banner
Chrome 52, Production, Mac OS X El Capitan

To repro, sign into a teacher account, and choose Italian language. Look at the teacher banner on the homepage.

screen shot 2016-08-12 at 2 58 12 pm

@chrisgarrity's comments:
> Educator resources looks like a problem - it is in the translation server, and it is translated into italian.
> So the header on the page is probably translated, but the button in the banner is missing

mewtaylor:
This is part of https://github.com/LLK/scratchr2/pull/3955, adding strings that have been ported over:
* "by" in thumbnails
* "Explore" header on the Explore page

It also does a bit of cleanup on the splash page to format the intl handling as we do on other pages  (i.e. remove `defaultMessage` from being inline, just use the `id`)

[UPDATE] Also fixes #840.

### DIFF
--- a/src/components/thumbnail/thumbnail.jsx
+++ b/src/components/thumbnail/thumbnail.jsx
@@ -1,5 +1,6 @@
-var React = require('react');
 var classNames = require('classnames');
+var FormattedMessage = require('react-intl').FormattedMessage;
+var React = require('react');
 
 require('./thumbnail.scss');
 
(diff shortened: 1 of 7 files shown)


src/components/thumbnail/thumbnail.jsx
```diff
@@ -32,7 +33,10 @@ var Thumbnail = React.createClass({
         if (this.props.creator) {
             extra.push(
                 <div key="creator" className="thumbnail-creator">
-                    by <a href={'/users/' + this.props.creator + '/'}>{this.props.creator}</a>
+                    <FormattedMessage id={'thumbnail.by'} />{' '}
+                    <a href={'/users/' + this.props.creator + '/'}>
+                        {this.props.creator}
+                    </a>
```
rschamp: This is a good example of why we need to upgrade our localization tool — just translating "by" is going to be ambiguous in a lot of languages, and may need the context of the creator username to be accurate.

mewtaylor: It's a temporary fix anyways – the new thumbnail mocks that we'll be using in `Explore` and `Search` can probably start to be used in our front page too.

PR #854, opened 2016-08-17
Redesign Explore and Search pages

technoboy10:
This PR implements a redesign of the explore and search pages. It also adds a `showAvatar` and `cards` property to the Grid and Thumbnail components to make the redesigned elements more usable in other pages. 

Additionally, the Tabs component has been restyled.

**Current Issues**
* The search icon in the search page's masthead does not look right on smaller browser widths
* The design for the category/sort mode area on the explore page on small browser widths could use some TLC

screenshot 2016-08-16 19 52 29
screenshot 2016-08-16 19 35 00


src/views/explore/explore.jsx
```diff
@@ -27,24 +30,29 @@ var Explore = injectIntl(React.createClass({
...
         }
-        var slash = pathname.lastIndexOf('/');
-        var currentCategory = pathname.substring(slash + 1,pathname.length);
-        var typeStart = pathname.indexOf('explore/');
-        var type = pathname.substring(typeStart + 8,slash);
-        if (Object.keys(categoryOptions).indexOf(currentCategory) === -1 || typeOptions.indexOf(type) === -1) {
+        var options = pathname.split('/');
+        var type = options[2];
+        var currentCategory = options[3];
+        var currentMode = options.length > 4 ? options[4] : '';
```
mewtaylor: I think it'd make sense to have the default `'trending'` mode be applied here rather than below in the api request, i.e. here it would say:
`var currentMode = options.length > 4 ? options[4] : 'trending';`

Also – that might make it a bit more clear in the dropdown what the default sort is.

src/views/explore/explore.jsx
```diff
@@ -100,11 +115,17 @@ var Explore = injectIntl(React.createClass({
     },
     getTab: function (type) {
         var classes = classNames({
-            active: (this.props.itemType === type)
+            active: (this.props.itemType === type),
+            inactive: (this.props.itemType !== type)
```
mewtaylor: If it can't be both, you might be able to get away with just having `active` be present or not, and having the `hover` styling for `inactive` just be the default hover styling, perhaps.

src/views/explore/explore.jsx
```diff
@@ -84,14 +93,20 @@ var Explore = injectIntl(React.createClass({
...
+            this.props.itemType + '/' + this.props.category + '/' + value;
+        }
     },
     getBubble: function (type) {
         var classes = classNames({
             active: (this.props.category === type)
         });
         return (
-            <a href={'/explore/' + this.props.itemType + '/' + type + '/'}>
+            <a href={'/explore/' + this.props.itemType + '/' + type + '/' + this.props.mode}>
```
mewtaylor: Ah ok – now I see why the default allows for an empty string. I'm wondering if instead it would work for the three options to then be `['', 'popular', 'recent']` and just have the label for `''` in the dropdown be `'trending'`?

src/views/search/search.scss
```diff
@@ -4,33 +4,278 @@
...
+        }
+
+        .button {
+            display: block;
+            margin: 0 auto;
+        }
+    }
+
+
+    .thumbnail {
```
mewtaylor: There seems to be some duplicate here with explore – should the styling just be moved to `grid` instead, perhaps?

mewtaylor:
@technoboy10 looks awesome, I can't wait for this! I left ya a couple questions/suggestions for the PR when you have a sec. On top of the questions I left inline, the only other one I had was if you had a chance to take a look at `/conference/schedule` view to see if it would be easy to have that use the `tabs.jsx` component too? It'd be great to get that as part of this PR, but if it's too much, we can just open another issue for that.

Other questions not necessarily part of this PR, but that came up when I was looking at it:

@carljbowman – do we want to use this thumbnail formatting on the homepage too (i.e. everywhere on www)?

@rschamp – it looks like the `Thumbnail` component is doing a lot of things. Part of that is having two different stylings for the homepage and for the explore/search pages, but the other part is having it handle studio and project thumbnails in one render method. Should we consider making a base `Thumbnail` presentation component, which is then implemented by `ProjectThumbnail` and `Studio Thumbnail` presentation components?

src/views/explore/explore.jsx
technoboy10: @mewtaylor I had thought about that, but I was concerned about consistency with URLs (if every other sort mode has a specific URL ending, I would expect trending to as well).

carljbowman:
@mewtaylor - The hope is that we will start to design 'thumbnnail's for each object on the site: Project, Studio, User, Comment, Forum Post... Each thumbnail object will have display properties based on metadata and size of the element (relative to parent element & screen size).

For example the `ProjectThumbnail` object will need to display Project Image, Project Title, and Project Creator consistently on across the Homepage, Search, & Explore pages. That said in some places we may want to explore some addtional metadata: i.e. on Search we may want to expose the Remix Count. An addtional variability is responsive behavior, for scoping purpose we may also want to include responsive behavior at the component level.

src/views/explore/explore.jsx
mewtaylor: @technoboy10 that makes sense – @rschamp any thoughts?

thisandagain:
@mewtaylor This is tied to https://github.com/LLK/scratch-api/pull/168 which will need to be rolled back if this isn't going to go out as part of the August deploy.

mewtaylor:
@thisandagain this missed the freeze – my inclination would be to roll back the api change. Happy to fix it up before the deploy though, if @jwzimmer you feel you'd have time to test it.

jwzimmer:
@mewtaylor, I think I'd have time to test this if need be.

src/components/grid/grid.scss
```diff
@@ -1,40 +1,92 @@
...
     $gallery-height: 118px;
 
     .flex-row {
         margin: 0 auto;
         padding: 12px;
-        width: (96px + (4 * $project-width)) / $em;
-        justify-content: flex-start;
-    }
+        width: $cols12;
+        justify-content: space-between;    }
```
mewtaylor: new line before the curly brace

technoboy10: @mewtaylor Oops, fixed!

thisandagain:
@technoboy10 Are you going to open up a new PR?

mewtaylor:
@thisandagain yup! against the release branch

technoboy10:
Sorry, I had to go to class. Will get that done in the next half hour.

PR #858, opened 2016-08-19
Bugfix: align gender/organization type input msg with other messages

mewtaylor:
Because the `input` field is already to the right, they were being transformed too far over. This overrides the transform to be correct, and also scopes some css in more detail.

### DIFF
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -21,43 +21,6 @@
         }
     }
 
-    .validation-message {
-        $arrow-border-width: 1rem;
-        display: block;
-        position: absolute;
-        top: 0;
-        left: 0;
-        transform: translate(16rem, 0);
-        margin-left: $arrow-border-width;
-        border: 1px solid $active-gray;
-        border-radius: 5px;
-        background-color: $ui-orange;
-        padding: 1rem;
-        max-width: 18.75rem;
-        min-height: 1rem;
-        overflow: visible;
-        color: $type-white;
-
-        &:before {
-            display: block;
-            position: absolute;
-            top: 1rem;
-            left: -$arrow-border-width / 2;
-
-            transform: rotate(45deg);
-
-            border-bottom: 1px solid $active-gray;
-            border-left: 1px solid $active-gray;
-            border-radius: 5px;
-
-            background-color: $ui-orange;
-            width: $arrow-border-width;
-            height: $arrow-border-width;
-
-            content: "";
-        }
-    }
-
     .form {
         padding: 3rem 4rem;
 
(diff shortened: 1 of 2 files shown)


src/components/card/card.scss
```diff
@@ -77,6 +40,43 @@
...
+                    border-left: 1px solid $active-gray;
+                    border-radius: 5px;
+
+                    background-color: $ui-orange;
+                    width: $arrow-border-width;
+                    height: $arrow-border-width;
+
+                    content: "";
+                }
+            }
```
rschamp: What's the reason for nesting this under `.form`? You should use as few levels of nesting as possible (but no less) because then your styles apply in more situations without having to redo the CSS.  So here, unless the validation message _shouldn't_ look this way outside of a `.form`, then it shouldn't be specific to `.form`.  Since this is imported by the Card component, then it should be scoped to `.card`, however.

src/components/card/card.scss
```diff
@@ -105,15 +105,19 @@
...
+                .validation-message {
+                    position: relative;
+                    transform: none;
+                    margin: inherit;
+                    width: 100%;
+                    height: inherit;
+
+                    &:before {
+                        display: none;
+                    }
```
rschamp: Same here, this doesn't need the `.form { .row { ... }}` wrapper.

src/components/card/card.scss
mewtaylor: `.form` was already there – I was mostly trying to get this under `.row`, since I don't think we necessarily want it to be styled the same way outside of `.row`. I'm not sure if we'd want to move `.row` outside of `.form` though. Or just move `.validation-message` outside again, working under the pretense that we most likely would never have use for it in another way on `.card`.

src/components/card/card.scss
mewtaylor: This was just to match the wrapping above, so it was scoped the same.

src/components/card/card.scss
rschamp: I think it would make sense to pull `.row` out of `.form`, so long as it doesn't break anything, and I don't think it should.  It would be nice if `.row` wasn't so generically named (`.form-row` would be better) — nesting kind of make sense because of that. But since we know we don't have anything else in `.card` called `.row`, I think it'd be fine/preferable to do that.

I think it'd be better to just leave `.validation-message` where it was unless you have a reason to make it more specific.  If there's another way to use `.validation-message` in `.card`, then that would be the time to make it more specific, but until then it should be more general than specific.

src/components/card/card.scss
rschamp: This actually is another example of why not to make it more specific unless you have to — then every time you want to change that class, you have to remember to repeat the same nesting as you have everywhere else in order for the overrides to work. If you make the original styles as generic as possible then you can override them in other places with just one level of nesting, like `.some-other-class .validation-message`, rather than `.some-other-class .form .row .validation-message`.

jwzimmer:
The edges of the validations are now aligned in Chrome 52, Safari Version 9.1.2 (11601.7.7) on Mac OS X, and IE 11 on Win 7 VM.

Issue #815, opened 2016-08-05
Update Wiki Footer Link to use User's Language


Choco31415:
Currently, when a user clicks on the Scratch Wiki footer link, it takes them to the English wiki. This ignores the user's language. It would be nice if when a user clicks on the Scratch Wiki footer link, they are taken to the wiki of their language, if it exists.

kaschm:
Moved to LLK/scratch-wiki#1

chrisgarrity:
Re-opening this on www because generating the correct link in the footer is something that needs to be handled on the www side of things (not in the wiki). The current list of languages that have their own wiki is:
* German
* Russian
* Dutch (Nederlands)
* Indonesian
* Japanese
* Hungarian 
* French

If a language does not have it's own wiki it should default to the English one: 
https://en.scratch-wiki.info/

This will require a new  localized URL resource (the wiki link and the other languages available), and the build locales script would need to include it on every page since it's in the footer.

Kenny2github:
Though it will happen rarely, what should the procedure be for when a new language is supported? Simply change the localized link for that language away from the enwiki link?

apple502j:
Hmm, they mean this: https://github.com/LLK/scratch-www/blob/develop/bin/get-localized-urls
https://resources.scratch.mit.edu/localized-urls.json contains the links

chrisgarrity:
https://resources.scratch.mit.edu/localized-urls.json is generated based on the actual translation files available on the resources server. I don't think there's any way to automate this new information, so yes, someone would have to manually update the links data when a new wiki is added.

439bananas:
My issue with this is that, there are some languages that the wiki isn't supported in, so those languages would return undefined.scratch-wiki.info

BoomerScratch:
> My issue with this is that, there are some languages that the wiki isn't supported in, so those languages would return undefined.scratch-wiki.info

If the language isn't supported, the link should be test.scratch-wiki.info or scratch-wiki.info.

R4356th:
@benjiwheeler, I could work on this, if you wanted to have this feature.

benjiwheeler:
@R4356th Thanks, that would be great! I don't know how complex the solution will need to be, which is why I haven't labeled it Help wanted.

apple502j:
@R4356th Please do not submit PRs (or at least mark it as draft) - I need to ask Scratch Wiki admins about the change.

R4356th:
[...]

PR #859, opened 2016-08-20
Updated Wiki Footer Link

Choco31415:
This PR addresses #815. Basically, instead of pointing everyone to the English Scratch Wiki, it will read in the Scratcher's language, and point them to a Wiki in their language, if one exists.

As a quick question, the footer component is the correct file to have put this code, yes?

### DIFF
--- a/src/components/footer/www/footer.jsx
+++ b/src/components/footer/www/footer.jsx
@@ -16,19 +16,19 @@ var Footer = React.createClass({
     render: function () {
         // Make it so that the Scratch Wiki url reflects the user's language.
         var wiki_mapping = {
-            "de": "http://scratch-dach.info/wiki/Hauptseite",
-            "hu": "http://hu.scratch-wiki.info/wiki/Kezdőlap",
-            "id": "http://scratch-indo.info/wiki/Halaman_Utama",
-            "jp": "http://jp.scratch-wiki.info/wiki/メインページ",
-            "nl": "http://nl.scratch-wiki.info/wiki/Hoofdpagina",
-            "ru": "http://scratch-ru.info/wiki/Заглавная_страница"
+            'de': 'http://scratch-dach.info/wiki/Hauptseite',
+            'hu': 'http://hu.scratch-wiki.info/wiki/Kezdőlap',
+            'id': 'http://scratch-indo.info/wiki/Halaman_Utama',
+            'jp': 'http://jp.scratch-wiki.info/wiki/メインページ',
+            'nl': 'http://nl.scratch-wiki.info/wiki/Hoofdpagina',
+            'ru': 'http://scratch-ru.info/wiki/Заглавная_страница'
         };
         var wiki_url = null; // The url to the wiki in the user's language.
         if (this.props.intl.locale in wiki_mapping) {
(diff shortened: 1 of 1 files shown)


Choco31415:
Huh, Travis seems to hate double quoted strings. I'll get that fixed.

Choco31415:

chrisgarrity:
I think this may be the general approach we want to take with URLs post-Transifex.
By default Transifex Live pulls out URLs and just leaves a placeholder so that translators don't accidentally modify them. Using the default Transifex behavior, and just replacing the URL ourselves is probably going to be less error-prone than using the 'notranslate-urls' tx-option and then telling the translators to leave it alone until we know there's a translated version.

This is something I want to try with ScratchJr. The one concern I have is whether this gets detected as different strings by Transifex when it has a different URL in the source. I'll be checking that with ScratchJr.

mewtaylor:
@Choco31415 – apologies for the delay! We actually have a method for localizing urls for links that have different language versions.

If you take a look at the [`cards`](https://github.com/LLK/scratch-www/blob/develop/src/views/cards/cards.jsx) view, you'll see links that we've bootstrapped into our `react-intl` system. The links for a page that need to be localized are kept track of in the [`l10n-static.json`](https://github.com/LLK/scratch-www/blob/develop/src/views/cards/l10n-static.json) file in the cards directory, and the localized versions of links are maintained in the [`localized-urls.json`](https://github.com/LLK/scratch-www/blob/develop/bin/lib/localized-urls.json) file.

The `localized-urls.json` file is inspected during `make translations`, where it looks at different languages to see if they have a localized version of a link. It checks for localized links by checking for matching `formatMessage` ids in each language. 

For instance, in `cards.jsx`, to get translated links for the "Starter Cards" pdf, it checks each language (by ISO code) to see if it contains value for the same key id, `"cards.starterLink"`.

The unfortunate issue with this example is that it's in the footer, and not in a specific view. So, we unfortunately can't simply move the links into an `l10n-static.json` file in the `footer` directory. Our system for building translations, in `bin/build-locales`, won't find it there.

Given these issues, I think that, in order to do this correctly, we'd need to create a general `l10n-static.json` file in `src/`, and put the wiki translation in there. This makes the fix for this issue larger than we originally anticipated, because we'd also need to edit the `bin/build-locales` file to correctly handle that new file too. 

I'd be happy to take this on if you'd like to pass it over to me!

Choco31415:
@mewtaylor If you say it would require a major rewrite of some of the code, sure, you I'll pass it over to you. You'll be much more competent in coding up a solution to the wiki footer issue.

@chrisgarrity In ScratchJr, will you be using the same json system that the Scratch website uses for links?

mewtaylor:
@Choco31415 I can take this on, then – thanks again for giving it a go, though! It's really appreciated, and should hopefully be easier to do fixes like this after I do some refactoring, hopefully.

I'll close this PR, but it has been very helpful :)

chrisgarrity:
@Choco31415 We're still in the process of trying things out in ScratchJr, and it's not clear yet whether it will be the same json approach, or something similar but not quite the same.

PR #867, opened 2016-08-22
Adding routes to microworlds

morantsur:
Until now, the new version of the microworlds page was committed, but there was no way to reach the pages. I added a link to them (scratch.mit.edu/microworld/hiphop etc.), and fixes a few issues related to the code being stale. 
Also added a new soccer microworld. 

@ntlrsk @seotts @mewtaylor


morantsur:
I tried fixing the failing tests, but not sure what to do with these errors: 

src/views/microworld/art/art.jsx
  1:5  error  "React" is defined but never used  no-unused-vars

Not sure why, but if I remove this line, the page doesn't compile with this error:
art.jsx:10 Uncaught ReferenceError: React is not defined 

Any ideas?

rschamp:
This is a quirk of JSX — React is required implicitly whenever you have JSX on the page, since when you write `<Component />` this gets transpiled to `React.createElement(Component)`.  Currently we don't make React available as a global, so our solution is to include React whenever we use JSX, and if the `React` var isn't used, we just add [`// eslint-disable-line`](https://github.com/LLK/scratch-www/blob/99c76ee3bfd29e8ffb05cc7fc9a987019a7cef66/src/lib/render.jsx#L4) to the import line

morantsur:
Thanks! That fixed it :)

thisandagain:
@mewtaylor could you work with @morantsur to get this reviewed and landed?

src/views/microworld/soccer/soccer.json
```diff
@@ -0,0 +1,8 @@
+{
+   "title":"Soccer",
+   "description":[
+      "Make your own soccer animation or game!"
+   ],
+   "microworld_project_id":"116297919",
+   "show_forum":false
+}
```
mewtaylor: small thing – could you add a new line at the end of this file?

src/views/microworld/art/art.json
```diff
@@ -6,121 +6,123 @@
...
-            },
-            {
-               "type":"tip",
-               "title":"Change the timing",
-               "thumbnailUrl":"/images/microworlds/hiphop/change-dance-timing.png"
-            }
-         ]
-      }
-   ],
+   "commentedout_content": {
```
mewtaylor: @morantsur where is `commentedout_content` used? I'm having trouble finding it, sorry!

src/routes.json
```diff
@@ -218,5 +218,33 @@
         "pattern": "^/explore/studios/?$",
         "routeAlias": "/explore(?!/ajax)",
         "redirect": "/explore/studios/all"
+    },
+    {
+        "name": "microworld-art",
+        "pattern": "^/microworlds/art",
```
mewtaylor: A @morantsur this is my bad – I forgot we don't need to include the `^` at the beginning anymore, since we add it later in the [build process](https://github.com/LLK/scratch-www/blob/develop/bin/configure-fastly.js#L64-L72). Do you mind removing those before I merge? Sorry!

mewtaylor: Whoops I meant in the `routeAlias` we don't need that, we still need it in the `route`. So, it's not needed in each `routeAlias`, like below.

morantsur:
Done :)

morantsur:
Fixes part of LLK/scratch-www#916

mewtaylor:
@morantsur I assigned you as qa for this – not sure if there's any qa that needs to happen pre-deploy since this is still not linked to in the site, but feel free to add some test cases if it's appropriate!

PR #869, opened 2016-08-22
Bugfix: Add Abkhaz language to custom locales

mewtaylor:
it’s not in `intl-messageformat`. Give it Azerbaijani as a parent language for now, as they have similar enough structures for plualization/alphabet purposes.

Fixes https://github.com/LLK/scratchr2/issues/3973


jwzimmer:
@mewtaylor Abkhaz is now loading, but some of the words/ icons in the editor are overlapping, I think just because Abkhaz is overall longer than English? Should this be it's own issue - is it out of scope for this? (I would think it's out of scope, but I was wondering if the parent language makes any difference, so I thought I'd just check...)
screen shot 2016-08-23 at 12 52 20 pm

mewtaylor:
@jwzimmer yeah +1 that it's out of scope. Thanks!

jwzimmer:
Issue https://github.com/LLK/scratchr2/issues/3976 is out of scope for this case. Since Abkhaz is now loading correctly, this PR is QA-OK.

PR #870, opened 2016-08-23
Fix gh-678: Retain Search Term

### LINKED ISSUE #678: Search - Retain search term in nav bar
After searching, the search term is reflected in the search results header but not in the navigation bar.

TheGrits:
@thisandagain 

Hopefully fixes #678 
I couldn't get the local image running on my Raspberry Pi (just need a bigger SD card, I ran out of room), but according to @St19Galla this doesn't break anything on his image.

The line added works great in the Chrome console, I'm just not sure if it works built in to the page.

### DIFF
--- a/src/views/search/search.jsx
+++ b/src/views/search/search.jsx
@@ -89,7 +89,7 @@ var Search = injectIntl(React.createClass({
     },
     render: function () {
         var formatMessage = this.props.intl.formatMessage;
-        document.getElementsByClassName('input')[0].value = {this.props.searchTerm};
+        document.getElementsByClassName('input')[0].value = this.props.searchTerm;
         
         return (
             <div>
--- a/src/views/search/search.jsx
+++ b/src/views/search/search.jsx
@@ -89,6 +89,7 @@ var Search = injectIntl(React.createClass({
     },
     render: function () {
         var formatMessage = this.props.intl.formatMessage;
+        document.getElementsByClassName('input')[0].value = {this.props.searchTerm};
         
         return (
             <div>


St19Galla:
I haven't been able to test the actual searching portion, as local hosting doesn't allow me to search anything, I just know that the code doesn't break anything on the main page.

As for the unexpected token, I'm not sure where travis is getting upset, you seem to have all your changes closed. I doubt it, but could it be line 121 missing a `,`?

TheGrits:
@St19Galla Travis is pointing to line 92 character 66, the curly bracket. I think I got the problem, thanks for pointing that out!

mewtaylor:
I think it might be better to handle the search term through our redux system, rather than through a call to `document`. Generally, when dealing with values that need to be handled across hierarchies in React, we use redux.

For instance, this is how we handle use of the user session. On the splash page, when a banner is dismissed, we make a redux call to [refresh the session](https://github.com/LLK/scratch-www/blob/develop/src/views/splash/splash.jsx#L172), which then [updates the session](https://github.com/LLK/scratch-www/blob/develop/src/redux/session.js#L63) from the server, which is finally [made available to the navbar](https://github.com/LLK/scratch-www/blob/develop/src/components/navigation/www/navigation.jsx#L347-L356), even though the navbar is a separate component.

@rschamp any thoughts on this?

TheGrits:
@mewtaylor There seems to be something already in place with the new search bar, see #875 (also directly related issue)

mewtaylor:
@TheGrits I think that's slightly different, since it's for a new search bar that exists inside the `search` page. Since this is dealing with the nav bar, which is rendered outside of the scope of the search page, it seems like a separate issue to me.

rschamp:
I agree that this is a job for either redux, in which you would set the current search term and then receive it in the nav bar, or by creating a new prop on `Page` which would eventually pass that value to the search field.

mewtaylor:
@TheGrits how comfortable do you feel giving this a try using redux as mentioned above? Feel free to pass this off to me if you'd like to as well!

TheGrits:
@mewtaylor I'm willing to give it a try, I'm always looking for things to learn about!

mewtaylor:
@TheGrits awesome feel free to drop in a comment if you have any questions!

TheGrits:
@mewtaylor I'm thinking a new PR will be needed, especially considering the merge conflicts. What are your thoughts?

mewtaylor:
@TheGrits Yeah good call – I'll close this one for now, then!

TheGrits:
@mewtaylor So would fetching the term with `window` be okay or not? [That's how it's currently fetched on the search page](https://github.com/LLK/scratch-www/blob/develop/src/views/search/search.jsx#L22)

mewtaylor:
@TheGrits yeah, using `this.props.searchTerm` for the value is still the right place to get the search term from (as you have it already). The change I requested here is more about how that value gets passed to the search input's value field, rather than how we get the search term originally.

Rather than getting the `input` element from `document` in order to pass a value to, the idea would be to create a new redux method in a new js file in the [`redux/`](https://github.com/LLK/scratch-www/tree/develop/src/redux) folder. This method could be similar to something like [`setPermissions`](https://github.com/LLK/scratch-www/blob/develop/src/redux/permissions.js#L50-L56), which would then be called from `this.props.dispatch()` in the search view.

Then, the search term could be [automatically mapped](https://github.com/LLK/scratch-www/blob/develop/src/components/navigation/www/navigation.jsx#L347-L352) as a prop in the nav component, making it available to the nav bar to optionally be displayed as the `value` on the Input component for search.

Does that help at all?

TheGrits:
@mewtaylor That does help a lot, thank you! I'll try to make a PR soon from what I understand, and hopefully we can work forward with that.

PR #871, opened 2016-08-23
Redesign Explore and Search pages

technoboy10:
This PR implements a redesign of the explore and search pages. It also adds a `showAvatar` and `cards` property to the Grid and Thumbnail components to make the redesigned elements more usable in other pages. 

Additionally, the Tabs component has been restyled.

**Current Issues**
* The search icon in the search page's masthead does not look right on smaller browser widths
* The design for the category/sort mode area on the explore page on small browser widths could use some TLC

screenshot 2016-08-16 19 52 29
screenshot 2016-08-16 19 35 00


mewtaylor:
Thanks @technoboy10 !

jwzimmer:
Testing, Staging

Chrome 52
- There's a different amount of white space on the bottom of some project cards on the `explore/projects/all` page. It looks like it may have to do with which characters are in the project title? (I noticed a heart emoticon and "|" in the titles that have the lower usernames.)
screen shot 2016-08-23 at 3 21 28 pm

carljbowman:
@jwzimmer - Thanks for pulling all these together! 

**RE White Space:** This should not be a blocker, but we should def figure out a better solution to in handling special characters. 

**RE Results changing based on Language:** This was by design, search puts a lot of weight on the language of the project. That said we could make the transition when results are loading a little less jarring. 

**RE Blue Background:** This was by design. The site has had that blue background since we started to migrate to WWW. Maybe someday (a long ways out) we can have this be a user defined preference. 

**RE Search Bars:** Again this was by design. We have an issue out for the Navigation search bar to retain the search result, so when that lands we can revisit whether to keep the addtional search bar. 

**RE Studio Thumbnails - IE 11 on Win 7 VM:** I defer to @mewtaylor @technoboy10 on the severity of the issue. Def not awesome, but not sure how difficult this is to fix. 

**RE Not Enough Results:** Seems like a flexbox wrapping issue. I am seeing this on chrome as well. I do not think this is blocking, but certainly awkward. @technoboy10 thoughts?

**RE Responsive Footer:** The choice of which links to show/hide were by design.

jwzimmer:
Based on the above feedback, opening cases for non-blockers that are still issues we want to consider addressing.

mewtaylor:
@jwzimmer @technoboy10 @carljbowman I'll work on the not enough results issue, I have a fix that should be easy to implement.

mewtaylor:
@jwzimmer @carljbowman as for the studio thumbnails – considering the code that touched that part of this system has not changed as a part of this PR, I'm inclined to think it's a staging-only issue at the moment.

jwzimmer:
@mewtaylor Okay I will open a separate issue for studio thumbnails. So that's every issue I brought up addressed:
- You're fixing the not enough results issue
- Studio thumbnails is an unrelated separate case
- Two search bars is a future design consideration, separate case
- Special chars/ placement of username in project cards is a related but non-blocking issue, separate case
- Everything else is by design/ non-issue

/cc @carljbowman @technoboy10

mewtaylor:
@jwzimmer just pushed https://github.com/LLK/scratch-www/pull/877, so the not enough results issue should be fixed soon.

jwzimmer:
@mewtaylor The not-enough-results fix looks good to me in Chrome, Safari on Mac and IE 11 and Firefox on Win 7 VM.

Issue #853, opened 2016-08-16
Revise Community Guidelines to mention personally identifying social media sites


thisandagain:
#### Keep personal info private
For safety reasons, don't give out any information that could be used for private communication - such as real last names, phone numbers, addresses, email address, links to social media sites or chat rooms.

---
Moved from LLK/scratchr2#3923
/cc @ntlrsk @paddle2see

Paddle2See:
Probably should be plural: email addresses

Paddle2See:
The version approved in the last Community meeting:

For safety reasons, don't give out any information that could be used for private communication - such as real last names, phone numbers, addresses, email addresses, links to social media sites or websites with unmoderated chat.

TheGrits:
@mewtaylor I can take this one for you if you'd like.

mewtaylor:
@TheGrits thanks for the help :). Go for it!

PR #873, opened 2016-08-23
Fix gh-853: Update Community Guidelines

TheGrits:
As per @Paddle2See on #853

@mewtaylor


jwzimmer:
### Test case

* [ ] The Community Guidelines should now say:
> **Keep personal info private**
For safety reasons, don't give out any information that could be used for private communication - such as real last names, phone numbers, addresses, email addresses, links to social media sites or websites with unmoderated chat.

jwzimmer:
The text in the PR does not include a period at the end of "Keep personal info private", whereas the version currently on Staging does. However, since the rest of the headings in the Community Guidelines end in periods, this seems like a good deviation from the copy. So, I'm QA-OK'ing this.

screen shot 2016-10-03 at 8 47 59 am

PR #896, opened 2016-08-26
Fix GH-825: Only require zip code for US addresses

### LINKED ISSUE #825: Don't require zip code field for countries without zip codes - Teacher Registration
Reported via freshdesk no. 54399:
> Ghana does not use ZIP codes so I just put in place holders but they are meaningless.

I looked at the address for the US Embassy in Accra, Ghana and it doesn't [list a zip code](http://ghana.usembassy.gov/pao.html).

I'm sure there are other countries that don't have zip codes but for which we're requiring the zip code field...

Ghana:
screen shot 2016-08-09 at 2 40 19 pm

mewtaylor:
Since we only check for the US, only require for that country. Other countries have zip codes, but it seems better to do it this way than codify which countries do/do not have zip codes (in case local practices are not the same as international information). Fixes #825

### DIFF
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -745,7 +745,7 @@ module.exports = {
                                            id: 'registration.validationMaxLength'
                                        })
                                    }}
-                                   required />
+                                   required={(this.state.countryChoice === 'us')} />
                             <GeneralError name="all" />
                             <NextStepButton waiting={this.props.waiting || this.state.waiting}
                                             text={<intl.FormattedMessage id="registration.nextStep" />} />


jwzimmer:
### Testing

Discussed offline with @mewtaylor the implications of verifying the zipcode for American Samoa, we decided it is fine in this case (I brought up whether or not we should treat American Samoa differently than USA, but since we're removing validations = making it easier for every option other than USA to complete the form, we decided it was okay).

Staging, Safari Version 9.1.3 (11601.7.8), Mac OS X El Capitan & IE 11 on Win 7 VM

* I can now skip entering a zip-code for Ghana and other non-USA options
* I can enter a zip-code for non-USA options
* I still have to enter a zip-code for USA

Issue #894, opened 2016-08-26
Mobile Community Guidelines Image Error


St19Galla:
On the Community Guidelines page, the bottom image is not responsive in changing size with the rest of the page, causing a background issue:
<img src="https://cloud.githubusercontent.com/assets/16446551/17989512/d79cce28-6afb-11e6-8895-1d8cf66af848.PNG" width=300px>

Also, ignore the white square over the messages icon - that's a known bug with my phone. Sorry!

TheGrits:
I'm thinking that keeping the image on mobile doesn't make sense. It doesn't carry any important information, so I don't think it is absolutely needed.

coke11:
Staging?

coke11:
@St19Galla Staging?

TheGrits:
@coke11 Nope: 
![screenshot 2016-08-29 at 7 47 20 am](https://cloud.githubusercontent.com/assets/12952370/18050496/ea758fbc-6dbc-11e6-85bb-8e112ae31151.png)

coke11:
@TheGrits 

> Nope

?

PR #902, opened 2016-09-01
Fix gh-894: Community Guidelines Image

TheGrits:
Update to hide page's footer image when the screen is too small.

Before:
screenshot 1

After:
screenshot 2

### DIFF
--- a/src/views/guidelines/guidelines.scss
+++ b/src/views/guidelines/guidelines.scss
@@ -1,4 +1,5 @@
 @import "../../colors";
+@import "../../frameless";
 
 .guidelines {
     .guidelines-footer {
@@ -10,5 +11,11 @@
         dt {
             margin-top: 1.5rem;
         }
+    }
+}
+
+@media only screen and (max-width: $tablet - 1){
+    .guidelines-footer {
+        display: none;
     }
 }


TheGrits:

mewtaylor:
@TheGrits this looks great, and I see your point about hiding the image. Going to assign to @carljbowman for a final say though, in case he had any thoughts on keeping it and making it smaller!

thisandagain:
Since @carljbowman 's OOO, I think this is probably safe to land.

mewtaylor:
@thisandagain @TheGrits sounds good to me thanks!

jwzimmer:
### Testing

Chrome 53 on Mac OS X El Capitan

I noticed that in between when the image disappears and a medium-sized window, the image exists and fits in the view, but is off-centered. I was wondering if we want to fix keeping it centered, remove the image sooner, or leave it as is.

screen shot 2016-09-30 at 1 30 44 pm
screen shot 2016-09-30 at 1 30 01 pm
out

TheGrits:
@jwzimmer The moment the images shows up is set to the tablet screen size value set [here](https://github.com/LLK/scratch-www/blob/develop/src/_frameless.scss#L38) so that there are no hardcoded screen width values.

jwzimmer:
Discussed the image being off-centered from other elements on page with @carljbowman, decided it is fine as is for now.

PR #907, opened 2016-09-01
Fix GH-872: Check for username before sending

### LINKED ISSUE #872: "Oops something went wrong" error with blank username field - Teacher Registration
**Staging**

If you fill out the fields after getting this error, you can continue to the next page, and it doesn't seem to cause an error at the end of registration.

- Just from typing in a few known usernames to test out the validation (I didn't click next, just clicked in between username and password fields, I got this error:
  (404 error)
  (URL in the error was https://[staging]/accounts/check_username//)

screen shot 2016-08-22 at 5 57 02 pm
- Repro'd by trying different names, typing in password field, going back and forth between the two fields (did not click Next):

screen shot 2016-08-22 at 6 08 25 pm
screen shot 2016-08-22 at 6 08 16 pm

In Safari on Mac (see comment above), I'm still seeing this error after clearing cache and history. I see it in **Chrome 52**, too. The repro steps are: go to `educators/register`, click Next, put your cursor in username field, click Next again.
screen shot 2016-08-23 at 1 30 44 pm

mewtaylor:
the `//` for an empty username that gets sent to the backend causes a 500 error. This is an edge case in formsy, as it has not detected a change value yet, so handle it in the username validation method. Fixes #872

### DIFF
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -91,6 +91,13 @@ module.exports = {
         },
         validateUsername: function (username, callback) {
             callback = callback || function () {};
+            if (username.length < 1) {
+                this.refs.form.refs.formsy.updateInputsWithError({
+                    'user.username': formatMessage({id: 'teacherRegistration.validationRequired'})
+                });
+                return callback(false);
+            }
+
             api({
                 host: '',
                 uri: '/accounts/check_username/' + username + '/'
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -91,7 +91,7 @@ module.exports = {
         },
         validateUsername: function (username, callback) {
             callback = callback || function () {};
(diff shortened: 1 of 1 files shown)


src/components/registration/steps.jsx
```diff
@@ -91,6 +91,13 @@ module.exports = {
         },
         validateUsername: function (username, callback) {
             callback = callback || function () {};
+            if (username.length < 1) {
```
rschamp: `if (!username)` might be safer since it would cover null/undefined too (and would avoid a possible "Cannot read property 'length' of undefined" error).

mewtaylor: Ah – I thought that `Input` would be handling that by sending an empty string but sounds good.

rschamp:
We should probably also avoid that 500 in the check username view on scratchr2. But that's not a big deal really.

Issue #787, opened 2016-07-28
Deployment consistently breaks cached pages


thisandagain:
Mac OS, Chrome 51.0.2704.103 (64-bit)

After deployment, navigating to any page within `www` will consistently result in a fatal javascript exception which results in either a blank page being rendered or a blank page with just the primary navigation. Refreshing the page will resolve the issue.

## Repro
- Navigate to any page that has been ported over to `www`
- Deploy `www`
- Navigate to the same page again

## Expected
- Page loads

## Actual
- Fatal javascript exception is thrown which prevents the page for rendering

## Reference
```
common.bundle.js:1 Uncaught TypeError: Cannot read property 'call' of undefined
```

![image](https://cloud.githubusercontent.com/assets/747641/17227099/a99da556-54da-11e6-9671-10f73c338b64.png)

thisandagain:
@rschamp Any thoughts on this? Moving to September milestone.

rschamp:
This wasn't showing up for me in Codetree, I guess because it had multiple assignees? I didn't look at this closely enough to know the root cause yet.

I had suspected that there was a mismatch between a cached view bundle and a fresh common bundle, in which case I would want to use versioned bundle filenames so that it was impossible for a page to load two bundles that didn't match.  The fact that the error is emanating from the common bundle kind of discounts that, but using versioned filenames might be a good idea anyway.

St19Galla:
I can confirm that this still happens on Mac Safari after today's deploy.

rschamp:
Resolved in #910

thisandagain:
Looks like this is happening again after our latest deploy. Re-opening for investigation.

rschamp:
The `Cache-Control: no-cache` header has been set on all the files in S3.  This should cause browsers to actually observe the ETags.  We should keep an eye out, but this should be fixed for the next deploy.

PR #908, opened 2016-09-01
Use hashed bundle filenames to break cache

rschamp:
This is a possible fix for #787. I am still not sure why the error occurs, nor why it's fixed by a refresh. However, I was able to reproduce the issue when deploying locally, and after using this, I wasn't able to anymore.  I did confirm that there was some sort of mismatch going on because the error is raised when a file in the view bundle was attempting to require the render function in the common bundle. This PR ensures that the view bundle and common bundle are always in sync.  That is, a new view bundle won't try to use an old common bundle's functions (or vice versa).

I had to change our template to use ejs rather than mustache because of the way I needed to retrieve the view bundle. But that gives us one less dependency, which I'm ok with.  I also did away with our custom html generator in favor of the html-webpack-plugin which webpack recommends and already provided the paths to the bundle names with the hashes.

It's worth keeping an eye on https://github.com/webpack/webpack/issues/959. The issue we were having looks similar, except that we don't see it locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 ESLINT=./node_modules/.bin/eslint
 NODE=node
 SASSLINT=./node_modules/.bin/sass-lint -v
-S3CMD=s3cmd
+S3CMD=s3cmd sync -P --delete-removed --add-header=Cache-Control:public,max-age=86400
 TAP=./node_modules/.bin/tap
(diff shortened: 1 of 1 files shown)


rschamp:
Just realized this is going to cause another deploy issue if someone loads cached HTML since the old js files will 404 after a deploy.

PR #910, opened 2016-09-08
Add Cache-Control header to all S3 uploads

rschamp:
A possible fix for #787. I haven't been able to reproduce the error locally, and can't seem to cause the error to happen when deploying locally anymore :(

This header should be there in any case, and is recommended when using the other cache headers we have, `ETag` and `Last-Modified`.

I've made the cache length 1 day, just thinking that would be a good balance between giving the benefit of a browser cache and hedging the risk that it lasts too long (given we're having issues with it anyway).  I am open to suggestions for a different length.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 ESLINT=./node_modules/.bin/eslint
 NODE=node
 SASSLINT=./node_modules/.bin/sass-lint -v
-S3CMD=s3cmd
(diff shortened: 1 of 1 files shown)


thisandagain:
Our average session duration is `00:09:31`. I think I'd say even one hour would be ok to persistent the cache between sessions.

thisandagain:
This looks good otherwise though. 👍

PR #911, opened 2016-09-08
Use html-webpack-plugin for html rendering

rschamp:
This came out of another solution for #787, but might be useful anyway.

Removes our custom `mustache-renderer-webpack-plugin` module in favor of the webpack-sanctioned `html-webpack-plugin` to generate HTML. The reasons it might be good to do this:
* This plugin will likely be updated when webpack is updated
* It has many more features than ours. For instance it includes reference to the bundle filename in the case that it is dynamically generated
* It is already published on npm and maintained, so we can reuse it easily when we need to generate HTML in other projects


thisandagain:
It certainly is much more verbose / difficult to read inline, but if we are having major problems with the maintainability of the current mustache approach then this seems fine. @mewtaylor any thoughts?

mewtaylor:
@thisandagain I feel similarly – especially if we don't see that template changing very often, and we think that we can depend on this package more than `mustache` + `our own updates`, then I'm cool with it!

rschamp:
Thanks for the feedback.  I don't think there are any *major* issues with the current plugin, but I do think it will be easier for replicating our deployment strategy elsewhere (e.g. the Scratch Jr. website, next week) if we use a 3rd party plugin.  I think that alone is enough of a reason, so I'll merge this if you guys are cool with it.

I agree that the verbosity is less than ideal. I added a commit to make the paths to the config less verbose by putting our template config options in the same namespace as the `html-webpack-plugin` options (any unused options are still available under `htmlWebpackPlugin.options`).  I also added the other default html-webpack-plugin options to our `template-config.js` so it's explicit that this is the way it works.  It looks like it should be pretty easy to make the `htmlWebpackPlugin` template variable name configurable, so if we really want to we could do that and push it upstream.

PR #912, opened 2016-09-08
Remove interactions with session cookie

rschamp:
The token is not currently used, so remove this entirely.

Instead of retrieving the permission from the session cookie, store permission data from /session/, and cache it in a cookie.  On subsequent page loads, the permission information will be retrieved from the cookie more quickly than the /session/ endpoint returns. When the session changes, the cookie and permissions state is updated to reflect the new state.

Partially resolves LLK/scratchr2#3986

## Test cases
* Logging in as a teacher should display the teacher banner on the splash page
* Logging out from a teacher account should remove the teacher banner
* Logging in as a non-teacher should not display the teacher banner


mewtaylor:
Looks great, thanks!

jwzimmer:
Changing `needs-qa` to `QA-OK` since I tested the same exact test cases for https://github.com/LLK/scratch-www/pull/923#issuecomment-248129698.

PR #913, opened 2016-09-08
Don't set state to undefined API response bodies

rschamp:
This resolves the "Cannot read property 'x' of undefined" errors that occur when the API returns non-JSON responses. For instance if the proxy endpoints return 500 error responses from scratchr2.

We need a way to let the user know when things aren't working in general, but we don't have that.  In the meantime just log the error.


mewtaylor:
LGTM!

rschamp:
@thisandagain FYI — since I believe you were seeing this error last week while testing.

Issue #918, opened 2016-09-13
Update Scratch conference homepage video link


sleggss:
Current homepage reads:
“Watch the keynote live”

[Which will change to:](https://github.com/LLK/scratch-www/issues/864)
“Watch videos of the keynote sessions”

Can we update the link to:
https://youtu.be/alsfSTVn2es?list=PLpfxVARjkP-8chnTrjtDeo88Pcz6-xf_B

/cc @kaschm

coke11:
@TheGrits Ninja! I was going to take this one ;)

PR #919, opened 2016-09-13
Fix gh-918: Update Video Link

TheGrits:
As per @sleggss on #918

(@mewtaylor I see that the issue was assigned to you, but I assumed that it would be OK for me to fix it because of the `help wanted` tag)

### DIFF
--- a/src/views/conference/index/index.jsx
+++ b/src/views/conference/index/index.jsx
@@ -30,7 +30,7 @@ var ConferenceSplash = React.createClass({
                     </p>
                     <p className="sub-button">
                         <b>
-                            <a href="http://bit.ly/scratchmit2016-live" target="_blank">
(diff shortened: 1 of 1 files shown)


mewtaylor:
@rschamp – wondering if we should add in the lint exception for links, like in this example. Thoughts?

TheGrits:
@mewtaylor We had a similar problem here on the guidelines page: [line 33](https://github.com/LLK/scratch-www/blob/develop/src/views/guidelines/guidelines.jsx#L33)

TheGrits:
@mewtaylor I pushed a fix for the lint error, Travis is passing now.

mewtaylor:
Cool thanks @TheGrits !

jwzimmer:
### Testing

Looks good in Chrome Version 53.0.2785.116 (64-bit), Mac OS X El Capitan, and Firefox 49.0.1 on Win 7 VM. 
Link & Text are changed to match as described in https://github.com/LLK/scratch-www/issues/918.

Issue #590, opened 2016-06-21
Initial automated test for scratch-www


jwzimmer:
- [ ] Make a simple initial test now that we know what tools we're moving forward with.
    * I have an almost working mocha script that clicks on the Explore link of the navbar (one of the test cases from the Most Important Workflows decided on in #474); need to finish this up and then check it in.
    * May need to return to evaluation steps from https://github.com/LLK/scratch-www/issues/474 if the tools are not what we want, but we shouldn't really have to... I think we do want to use mocha, selenium, saucelabs, various webdrivers, and perhaps chai/ chai-as-promised.

 - [ ] Test tools should be added to requirements.txt
    * Selenium is in the scratchr2 requirements.txt
    * I have been screen-shotting and labelling everything I am installing for future reference.
- [ ] Test architecture
   * Should enable finding tests easily
   * Should mirror existing repo architecture as much as is reasonable
   - [ ] Naming conventions
   * @mewtaylor and I have worked out the basics of this but I'm sure in practice details will come up that we'll have to decide about.

- [ ] First test should be checked into repo
   - [ ] Document how to run, update test
   - [ ] Feedback from team on extensibility, usability, test's output and utility, etc.

jwzimmer:
This is obviously not happening in the next 3 days, so moving to next milestone.

thisandagain:
Moving to September milestone as per discussion with team last week. /cc @rschamp

jwzimmer:
Discussed with @mewtaylor & @rschamp. Current plan is:
- [ ] Revise Most Important Workflows (see https://github.com/LLK/scratch-www/issues/474) so that we can align them with automated tests that don't leave the homepage, for performance reasons
- [ ] Choose one test to automate
- [ ] Automate it using tap and selenium if possible
  - If this doesn't turn out to be a good strategy, re-visit
- [ ] Work with @mewtaylor & @rschamp to commit the test to scratch-www and set up how it will be run against PR's
  - [ ] Document this process

PR #920, opened 2016-09-14
Fix GH-590: initial automated test

jwzimmer:
This is what I have for the initial automated test per https://github.com/LLK/scratch-www/issues/590#issuecomment-243812155. The initial test is to check that the URL in the Create link in the navbar is correct when the user is signed out (without navigating to the editor, and using selenium and tap, as discussed).
There are some questions I have about it:
- A better solution than brittle XPath (I can continue looking into this; first I wanted to get something working)
- I set it up with a driver for Chrome; should I enumerate through the other available drivers here? How does this work with saucelabs?
- I created an integration folder in the test folder for these kinds of automated tests, but I am open to changing the structure
- Integration with travis? (I think it should be simple?)
- I looked at your existing test files in scratch-www and tried to copy your conventions in terms of naming files, functions, etc., but obviously please let me know of any corrections
- Should I add selenium to [devDependencies](https://github.com/LLK/scratch-www/blob/develop/package.json)? I see that tap is in the list, but you guys use tap for unit tests as well, so I wanted to check what would be appropriate?

/cc @rschamp & @mewtaylor

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
     "sass-lint": "1.5.1",
     "sass-loader": "2.0.1",
     "scratchr2_translations": "git://github.com/LLK/scratchr2_translations.git#master",
+    "selenium-webdriver": "^2.44.0",
     "slick-carousel": "1.5.8",
     "source-map-support": "0.3.2",
     "style-loader": "0.12.3",
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "sass-lint": "1.5.1",
     "sass-loader": "2.0.1",
     "scratchr2_translations": "git://github.com/LLK/scratchr2_translations.git#master",
-    "selenium-webdriver": "^2.44.0",
+    "selenium-webdriver": "2.44.0",
     "slick-carousel": "1.5.8",
     "source-map-support": "0.3.2",
     "style-loader": "0.12.3",
--- a/test/integration/check_navbar_links.js
+++ b/test/integration/check_navbar_links.js
@@ -4,24 +4,26 @@
  * Test cases: https://github.com/LLK/scratch-www/wiki/Most-Important-Workflows#Create_should_take_you_to_the_editor
  */
 
-var tap=require('tap'),  
+var tap=require('tap'),
     seleniumWebdriver = require('selenium-webdriver');
 
 /*
  * Remove question comments after resolving them
  */
-//how to generalize for other browsers... how will this work in saucelabs? Do I need to iterate through the drivers for each browser here?
+//how to generalize for other browsers... how will this work in saucelabs? Do I need to iterate 
+//through the drivers for each browser here?
 //chrome driver
 var driver = new seleniumWebdriver.Builder().withCapabilities(seleniumWebdriver.Capabilities.chrome()).build();
 //open scratch.ly in a new instance of the browser
-driver.get('https://scratch.ly')
+driver.get('https://scratch.ly');
 
 /*
  * Remove question comments after resolving them
  */
 //find the navbar
 //var navbarElement = driver.findElement(seleniumWebdriver.By.id("navigation"))
-//var createLinkSignedOut = navbarElement.findElement(seleniumWebdriver.By.xpath('//li[@class="link create"]/a')) <-- this doesn't work to force findElement to look in the scope of navbarElement
+//var createLinkSignedOut = navbarElement.findElement(seleniumWebdriver.By.xpath('//li[@class="link create"]/a'))
+//^^this doesn't work to force findElement to look in the scope of navbarElement
 
 //find the create link within the navbar
 //the create link depends on whether the user is signed in or not (tips window opens)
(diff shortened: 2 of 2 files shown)


jwzimmer:
I will try to address the issues causing the Travis CI build to fail (I see the list in "Details).

rschamp:
> Should I add selenium to devDependencies? I see that tap is in the list, but you guys use tap for unit tests as well, so I wanted to check what would be appropriate?

Yes, everything that's `require`d should be listed in `devDependencies`.  You can automatically add dev dependencies with `npm install <package name> --save-dev`

test/integration/check_navbar_links.js
```diff
@@ -0,0 +1,44 @@
+/*
+ * Checks that the links in the navbar on the homepage have the right URLs to redirect to
+ *
+ * Test cases: https://github.com/LLK/scratch-www/wiki/Most-Important-Workflows#Create_should_take_you_to_the_editor
+ */
+
+var tap=require('tap'),  
+    seleniumWebdriver = require('selenium-webdriver');
```
rschamp: Style nit: each requirement should be on a new `var` declaration like
```javascript
var tap = require('tap');
var seleniumWebdriver = require('selenium-webdriver');
```

test/integration/check_navbar_links.js
```diff
@@ -0,0 +1,44 @@
...
+
+/*
+ * Remove question comments after resolving them
+ */
+//how to generalize for other browsers... how will this work in saucelabs? Do I need to iterate through the drivers for each browser here?
+//chrome driver
```
rschamp: Using Sauce Connect with webdriver: https://wiki.saucelabs.com/display/DOCS/Updating+Test+Scripts+to+Use+Sauce+Connect
Here is how Travis gets set up with SauceLabs: https://docs.travis-ci.com/user/sauce-connect/

To simplify running these tests locally as well as on Travis, maybe we should just always run them in Sauce?

test/integration/check_navbar_links.js
```diff
@@ -0,0 +1,46 @@
...
+//var createLinkSignedOut = navbarElement.findElement(seleniumWebdriver.By.xpath('//li[@class="link create"]/a'))
+//^^this doesn't work to force findElement to look in the scope of navbarElement
+
+//find the create link within the navbar
+//the create link depends on whether the user is signed in or not (tips window opens)
+
+/*
+ * Remove question comments after resolving them
+ */
+//this xpath is fragile, can i look up by successive attributes instead?
```
rschamp: I think using the fragile XPath is a necessary evil of using Selenium this way. When we change the DOM structure, we will have to update the test, but this is frequently the case with refactors even when not using Selenium.

test/integration/check_navbar_links.js
```diff
@@ -0,0 +1,46 @@
...
+tap.test('checkCreateLinkWhenSignedOut', function (t) {
+    var createLinkSignedOut = driver.findElement(seleniumWebdriver.By.xpath('//div[@id="navigation"]/div[@class="inner"]/ul/li[@class="link create"]/a'));
+    createLinkSignedOut.getAttribute('href').then( function (href) {
+        t.equal('https://scratch.ly/projects/editor/?tip_bar=home', href);
```
rschamp: Is this really the href? We shouldn't be specifying the domain name, it should be something like `/projects/editor/?tip_bar=home`.

test/integration/check_navbar_links.js
```diff
@@ -0,0 +1,46 @@
...
+//open scratch.ly in a new instance of the browser
+driver.get('https://scratch.ly');
+
+/*
+ * Remove question comments after resolving them
+ */
+//find the navbar
+//var navbarElement = driver.findElement(seleniumWebdriver.By.id("navigation"))
+//var createLinkSignedOut = navbarElement.findElement(seleniumWebdriver.By.xpath('//li[@class="link create"]/a'))
+//^^this doesn't work to force findElement to look in the scope of navbarElement
```
rschamp: Does `contains` work? `//li[contains(@class, 'link') and contains(@class, 'create')]/a`?

test/integration/check_navbar_links.js
```diff
@@ -0,0 +1,46 @@
...
+
+var tap=require('tap'),
+    seleniumWebdriver = require('selenium-webdriver');
+
+/*
+ * Remove question comments after resolving them
+ */
+//how to generalize for other browsers... how will this work in saucelabs? Do I need to iterate 
+//through the drivers for each browser here?
+//chrome driver
```
rschamp: Eventually maybe we want to set the browser through an environmental variable and then iterate through running all the tests by switching the variable?  For now I think just one browser is ok.

package.json
```diff
@@ -82,6 +82,7 @@
     "sass-lint": "1.5.1",
     "sass-loader": "2.0.1",
     "scratchr2_translations": "git://github.com/LLK/scratchr2_translations.git#master",
+    "selenium-webdriver": "^2.44.0",
```
rschamp: Please remove the `^` to match the other dependencies. Info on what it means: https://docs.npmjs.com/getting-started/semantic-versioning. And you can set up your npm to never include it with `npm config set save-prefix=''`: https://docs.npmjs.com/misc/config#save-prefix

test/integration/check_navbar_links.js
mewtaylor: In the code, it's set to the relative path, it appears

rschamp: The browser probably supplies the full URL when Selenium looks it up, so the test should check if it ends with the path.

test/integration/check_navbar_links.js
jwzimmer: `contains` works, now using `contains`, thanks.

jwzimmer:
...I think the way I implemented looking only at the href rather than all the URL is bad (too convoluted); I will change it. : )

jwzimmer:
(We will address https://github.com/LLK/scratch-www/pull/920#discussion_r78991538 later (setting up running the tests with travis/ sauce).)

Issue #903, opened 2016-09-01
Update /hoc to include new tutorials


kaschm:
In time for the EU and Africa Code Weeks (early Oct), we should update the current scratch.mit.edu/hoc to the three tutorials we will highlight this year. This page will then be redirected to the activities page in advance of the US Code Week. 

Hide-and-Seek Game --> Make it Fly
Dance, Dance, Dance --> Make Music
Animate Your Name --> [No Change] 

For the two tutorials being updated, the following information needs to be changed:
- Top Section: Thumbnail, Name of tutorial, Link to tutorial, *remove graphical backgrounds*
- 'Activity and Guides' section: new links to cards, facilitator guides, and updated name
- 'Studios' section: new links, update to text

Updated links:

*Make Music* 
- Name of Tutorial: Make Music 
- Link to Tutorial: https://scratch.mit.edu/projects/editor/?tip_bar=music
- Studio link: https://scratch.mit.edu/studios/1999784/
- Thumbnail: grab from studio, above 
- Cards: @ntlrsk please update
- Facilitator Guide: REMOVE THIS LINK AND TEXT

*Make it Fly* 
- Name of Tutorial: Make it Fly 
- Link to Tutorial: https://scratch.mit.edu/projects/editor/?tip_bar=fly
- Studio link: https://scratch.mit.edu/studios/2050750/
- Thumbnail: use thumbnail from above (not animated) 
- Cards: @ntlrsk please update
- Facilitator Guide: REMOVE THIS LINK AND TEXT

thisandagain:
@ntlrsk When do you think we could expect the content (cards & facilitator guides) for this?

kaschm:
updated to remove links to guides, as per @ntlrsk they won't be ready for this push. thanks.

mewtaylor:
@kaschm @ntlrsk this is what it looks like with guides removed and such. The studio image for Make it Fly is a bit low res, so if we have another one available anywhere let me know!
![screencapture-localhost-8333-hoc-1473941482389](https://cloud.githubusercontent.com/assets/1815772/18549402/41191c32-7b1c-11e6-97f3-902d604bb6e8.png)

ntlrsk:
Thanks for working on this! I would like to make the cards available for all three activities - and will give you the cards by early tomorrow morning.

ntlrsk:
I see a couple more things that would be good to change if possible:
[...]

PR #921, opened 2016-09-15
Implement GH-903: Update hoc page with new tutorials

mewtaylor:
Adds:
1. Make it Fly
2. Make Music

Removes:
1. Hide-and-Seek
2. Dance, Dance, Dance

Implements #903.

This also updates the styling scoping for the hoc page to adopt patterns from the Trello CSS Guide. @rschamp I may have misinterpreted the guide in a couple places, so lemme know how it looks to you.


src/views/hoc/hoc.scss
```diff
@@ -8,123 +8,117 @@ $base-bg: $ui-white;
 }
 
 .hoc {
-    .title-banner {
-        &.wbb-bg {
-            background-image: url("/images/hoc2015/hide-bg.jpg");
+    .img,
+    .button {
+        &.card-info {
```
rschamp: From https://gist.github.com/bobbygrace/9e961e8982f42eb91b80#8-miscellany
> Explicitly write out class names in selectors. Don’t concatenate strings or use preprocessor trickery to build a class name. We want to be able to search for class names and that makes it impossible.

Avoid the `&` for this type of thing so our css is as flat as possible.  We might need to adjust the sass linter rules, or disable the sass linter altogether until we've transitioned.

src/views/hoc/hoc.scss
```diff
@@ -8,123 +8,117 @@ $base-bg: $ui-white;
 }
 
 .hoc {
```
rschamp: If we're transitioning to the Trello style, then I think we should avoid the top-level nesting we were using for encapsulation before. Instead of putting everything as a descendant to `.hoc`, the component classes that are specific to the `.hoc` view should be prefixed with `.hoc-`.

src/views/hoc/hoc.scss
```diff
@@ -8,123 +8,117 @@ $base-bg: $ui-white;
...
-            background-image: url("/images/hoc2015/dance-bg.jpg");
+    .img {
+        &.card-info {
+            margin: 10px 10px 5px 10px;
+            border-radius: 5px;
         }
 
-        &.name-bg {
-            background-image: url("/images/hoc2015/name-bg.jpg");
+        &.mod-border {
```
rschamp: This seems like it would be reused elsewhere, so maybe we need an `Image` component to house these styles?

src/views/hoc/hoc.scss
```diff
@@ -8,123 +8,117 @@ $base-bg: $ui-white;
...
+            width: 100%;
         }
 
-        section {
-            border: 0;
-            padding: 10px 0;
-            max-width: $desktop;
+        &.mod-resource,
+        &.mod-studio {
+            margin-right: 15px;
```
rschamp: This seems like it might be handled better by a wrapper component rather than a modifier if it's for formatting things specific to the HOC view.

src/views/hoc/hoc.scss
```diff
@@ -8,123 +8,117 @@ $base-bg: $ui-white;
...
-            }
-
+    .button {
+        &.card-info {
+            margin: 0 10px 10px 10px;
         }
     }
 
-    section {
+    .section {
```
rschamp: `.section` seems pretty broadly named for something defined within `.hoc`. Maybe this should be split out into a separate `Section` component?

src/views/hoc/hoc.scss
```diff
@@ -8,123 +8,117 @@ $base-bg: $ui-white;
...
+        max-width: 600px;
+    }
 
-                img {
-                    border-radius: 5px;
-                    width: 100%;
-                }
-            }
+    .title-banner {
+        &.wbb-bg {
```
rschamp: `.hoc .title-banner.wbb-bg` should become `.hoc-title-banner-wbb-bg` following https://gist.github.com/bobbygrace/9e961e8982f42eb91b80#2-components. This actually seems like a good use of a modifier, like `.hoc-title-banner.mod-wbb`.

src/views/hoc/hoc.scss
rschamp: (See general comment on ignoring annoying things)

src/views/hoc/hoc.jsx
```diff
@@ -30,55 +30,49 @@ var Hoc = React.createClass({
...
                                 <a href="/projects/editor/?tip_bar=name">
-                                    <div className="card-info" onMouseEnter={this.onCardEnter.bind(this, 'name-bg')}>
-                                        <img src="/images/hoc2015/name-tutorial.jpg" alt="" />
-                                        <Button>
-                                            <FormattedMessage
-                                                id='general.tipsAnimateYourNameTitle'
-                                                defaultMessage={'Animate Your Name'} />
+                                    <div className="flex-row-card-info"
+                                         onMouseEnter={this.onCardEnter.bind(this, 'name-bg')}>
+                                        <img className="img card-info"
```
mewtaylor: This actually was an attempt at following that example, which has a default button styling (i.e. here it would be an image styling) with mod, which is `card-info`. I guess it could be called `mod-flex-row-card-info`, though. Seems like if we're going to use `img` as a class, we shouldn't also put it on the mod, though.

src/views/hoc/hoc.scss
mewtaylor: Oh – I thought that part of our customization from Trello would be to keep this nesting for the scoping... I'll take a look at how this might be done without causing conflicts with the way other views/components currently work.

src/views/hoc/hoc.scss
mewtaylor: Ok – yeah I had it like that originally, but changed it due to warnings. They're just warnings though, so I guess it'll be fine for now.

src/views/hoc/hoc.scss
mewtaylor: The image stylings on this page do not currently align with image stylings elsewhere. Consequently, I'd prefer to wait on thinking about an `Image` component until we re-factor elsewhere and start to see the necessity for it.

src/views/hoc/hoc.scss
mewtaylor: 👍

src/views/hoc/hoc.scss
mewtaylor: The way `section` had been styled before this is still pretty specific to HOC, I don't believe many of these stylings for sure apply to other views/components – I think I'd prefer to do what you mentioned above about prefixing `.hoc-` here, and not create a `Section` component until there is a need that spans across views/components.

src/views/hoc/hoc.scss
mewtaylor: I left it as is since it seemed to be doing the modifier. I guess I should've added the `mod-` prefix, though. I can add that in.

jwzimmer:
@mewtaylor I am seeing the new tutorials, Make it Fly and Make Music, but I'm also seeing the two that should have been removed (Dance, Dance, Dance and Hide and Seek).

Staging, Chrome 53, Mac OS X El Capitan
screen shot 2016-10-03 at 1 26 35 pm

mewtaylor:
@jwzimmer fixed in https://github.com/LLK/scratch-www/commit/c44742e66823ddaa7b448d6c24bf3cfd56e0b743, should be up soon.

jwzimmer:
Testing, Staging, Chrome 53 Mac OS X El Capitan

- I'm now seeing studios matching the tutorials:
screen shot 2016-10-03 at 4 20 33 pm
- The Dance Dance Dance tutorial now links to the right place.

Discussing pic with @kaschm and @carljbowman.

mewtaylor:
@jwzimmer image updated in https://github.com/LLK/scratch-www/commit/6bcc7eeae30d35d79bb47a81d18b777265edc9fc

jwzimmer:
Staging, Chrome 53, Mac OS X El Capitan
- The image looks good now
- Links to studios are up (look at the version on production to see it's the right studio) for the new tutorials

Those were the extant issues, so I confirmed those, quickly looked over the rest of the page, and now can QA-OK this.

PR #932, opened 2016-09-30
Fix gh-678: Retain Search Term

TheGrits:
@mewtaylor 
(First PR version is at #870)
This is my first time playing with redux, I'm not sure that I did it correctly.

What I did in the redux was basically combining [this](https://github.com/LLK/scratch-www/blob/develop/src/redux/permissions.js#L50-L56) with something like [this](https://github.com/LLK/scratch-www/blob/develop/src/redux/permissions.js#L17). From what I understand, the navigation component is already retrieving the values [here](https://github.com/LLK/scratch-www/blob/develop/src/components/navigation/www/navigation.jsx#L347-L352).

Please do correct me if I'm wrong, I love learning!


mewtaylor:
@TheGrits thanks for taking a crack at this! I think this needs a bit more work though.

Currently, the `redux/navigation.js` file assumes there is an object called `this.props.searchTerm`, but that object only exists in the `search.jsx` view, so we can't assume it'd be available in the redux file. Additionally, I'm not seeing where `redux/navigation.js` is being required, so I'm not sure that it is connected to the redux system yet...

I think what you'll need to do is:
* Add a reducer, as well as `setSearchTerm` methods to `redux/navigation.js` – see [permissions.js](https://github.com/LLK/scratch-www/blob/develop/src/redux/permissions.js) as an example, but don't worry about `getPermissions` or `storePermissions`, since we really don't need a version of those
* `require` the navigation.js reducer in the main [reducer.js](https://github.com/LLK/scratch-www/blob/develop/src/redux/reducer.js) file. This will be how redux as a system knows about `this.props.searchTerm`, since that's what gets added to the [top level react component](https://github.com/LLK/scratch-www/blob/develop/src/lib/render.jsx#L15-L44).
* `require` the navigation.js reducer file in the search.jsx view (like [this](https://github.com/LLK/scratch-www/blob/develop/src/views/splash/splash.jsx#L9)), and call the `set` method you create once the search term has been retrieved from the URL.

Does that makes sense at all?

TheGrits:
@mewtaylor I figured that I'd still have quite a ways to go. :wink: 
Thanks for helping me along, I hope to update it soon.

TheGrits:
@mewtaylor I've taken another crack at this, but I'm not done yet. I'd greatly appreciate it if you'd help me along again. Thanks!

src/redux/navigation.js
```diff
@@ -0,0 +1,10 @@
+module.exports.navigationReducer = function (state, action) {
+    if(typeof state = 'undefined') {
+        state = '';
+    }
+    return state;
```
mewtaylor: This is covers the base cases that we need of:
* getting an undefined state
* getting a state that is not used by this particular reducer

However, we still need to handle the case in which we get a state that comes from the `setSearchTerm` action below (i.e. some object with a unique `type` and `searchTerm` specified as keys). See [`conference-details`](https://github.com/LLK/scratch-www/blob/develop/src/redux/conference-details.js#L10-L24) for an example – we don't need as many possible states as that has, but something like `SET_DETAILS` but for the search term would be good (and maybe a `KeyMirror` object to set it up as well).

src/views/search/search.jsx
```diff
@@ -12,6 +12,7 @@ var Input = require('../../components/forms/input.jsx');
 var Button = require('../../components/forms/button.jsx');
 var Tabs = require('../../components/tabs/tabs.jsx');
 var Grid = require('../../components/grid/grid.jsx');
+var Navigation = require('../../redux/navigation.js');
```
mewtaylor: We should be referring to this in a way that specifies that it's the redux action module, rather than the `Navigation` Component. See how we import [`details`](https://github.com/LLK/scratch-www/blob/develop/src/views/conference/details/details.jsx#L6) on the conference details page for more details on naming.

Additionally, we still need to call the `setSearchTerm` method in the redux module down below once we know what `this.props.searchTerm` actually is.

TheGrits:
@mewtaylor Sweet, I think it's ready for another review! Something went "click!" in my head as I followed your instructions, thank you so much!

src/redux/navigation.js
```diff
@@ -0,0 +1,24 @@
+var keyMirror = require('keymirror');
+
+var Types = keyMirror({
+    SET_SEARCH_TERM: null
+});
+
+module.exports.navigationReducer = function (state, action) {
+    if(typeof state = 'undefined') {
```
mewtaylor: I think this might be causing the test error – should be a space after `if`, and also it should be `===`

src/redux/navigation.js
```diff
@@ -0,0 +1,24 @@
...
+        state = '';
+    }
+    switch (action.type) {
+        case Types.SET_SEARCH_TERM:
+            return action.setSearchTerm;
```
mewtaylor: this should just be `searchTerm`

src/views/search/search.jsx
```diff
@@ -95,6 +97,9 @@ var Search = injectIntl(React.createClass({
         }
         return allTab;
     },
+    setSearchTerm: function (searchTerm) {
+        this.props.searchTerm = searchTerm;
```
mewtaylor: `props` shouldn't be assigned anything outside of an initialization method (like `getDefaultProps`. And, in order to get this into the `navigation` component, you'd need to instead dispatch `setSearchTerm` from `redux/navigation.js`, otherwise, it won't be available as a value outside of this component. Something like `this.props.dispatch(navigationDetails.setSearchTerm(searchTerm))` – but make sure that you [properly hook up redux](https://github.com/LLK/scratch-www/blob/develop/src/views/conference/details/details.jsx#L86-L94) if you do that too.

src/redux/navigation.js
TheGrits: doh! Thanks for pointing out that gotcha!

src/redux/navigation.js
TheGrits: Oh okay, thanks!

TheGrits:
@mewtaylor Okay, I think I covered everything, this should be ready for another review. Thanks so much for taking all this time to help me along, I'm learning so much about redux!

src/redux/reducer.js
```diff
@@ -4,12 +4,14 @@ var scheduleReducer = require('./conference-schedule.js').scheduleReducer;
 var detailsReducer = require('./conference-details.js').detailsReducer;
 var permissionsReducer = require('./permissions.js').permissionsReducer;
 var sessionReducer = require('./session.js').sessionReducer;
+var navigation = require('./navigation.js').navigation;
```
mewtaylor: This should be `navigationReducer`

src/views/search/search.jsx
```diff
@@ -13,6 +14,8 @@ var Button = require('../../components/forms/button.jsx');
 var Tabs = require('../../components/tabs/tabs.jsx');
 var Grid = require('../../components/grid/grid.jsx');
 
+var navigationDetails = require('../../../redux/navigation.js');
```
mewtaylor: Our convention is to have these called `<type>Actions`, so it'd be best if this was `navigationActions`.

Also, the assignment statement here should be inline with the component assignment statements above, without a space in between.

src/views/search/search.jsx
```diff
@@ -138,4 +144,12 @@ var Search = injectIntl(React.createClass({
     }
 }));
 
-render(<Page><Search /></Page>, document.getElementById('app'));
+var mapStateToProps = function (state) {
+    return {
+        navigationDetails: state.navigationDetails
```
mewtaylor: If the state is called `searchTerm` elsewhere, it should be called `searchTerm` here too

src/views/search/search.jsx
```diff
@@ -138,4 +144,12 @@ var Search = injectIntl(React.createClass({
...
+    };
+};
+
+var ConnectedDetails = connect(mapStateToProps)(Search);
```
mewtaylor: Another naming convention small thing – since this is the search page, it should be called `ConnectedSearch`

src/views/search/search.jsx
```diff
@@ -95,6 +98,9 @@ var Search = injectIntl(React.createClass({
         }
         return allTab;
     },
+    setSearchTerm: function (searchTerm) {
+        this.props.dispatch(navigationDetails.setSearchTerm(searchTerm));
```
mewtaylor: It doesn't look like `setSearchTerm` is being called anywhere, so the dispatch won't happen right now. Since this is a one line method, I think you can get rid of this function `setSearchTerm`, and just call `this.props.dispatch(navigationDetails.setSearchTerm(this.props.searchTerm));` above in `componentDidMount()`

TheGrits:
@mewtaylor 🎉 This should be for final review now! (The error Travis is returning is very strange, I think something went wrong on their end)

TheGrits:
@mewtaylor Aha! I figured out what Travis was giving the error for! 🎉

mewtaylor:
Keeping this at "needs review" so I don't accidentally merge while we're still testing the release.

src/redux/reducer.js
```diff
@@ -4,12 +4,14 @@ var scheduleReducer = require('./conference-schedule.js').scheduleReducer;
 var detailsReducer = require('./conference-details.js').detailsReducer;
 var permissionsReducer = require('./permissions.js').permissionsReducer;
 var sessionReducer = require('./session.js').sessionReducer;
+var navigationReducer = require('./navigation.js').navigation;
```
mewtaylor: this needs to be `navigationReducer`

src/components/navigation/www/navigation.jsx
```diff
@@ -347,7 +349,8 @@ var Navigation = React.createClass({
 var mapStateToProps = function (state) {
     return {
         session: state.session,
-        permissions: state.permissions
+        permissions: state.permissions,
+        searchTerm: state.searchTerm
```
mewtaylor: @TheGrits whoops I missed one thing – because the reducer is called `navigation`, the value will end up being stored in `state.navigation`. So, I think we need to change this to `searchTerm: state.navigation`

src/views/search/search.jsx
```diff
@@ -138,4 +141,12 @@ var Search = injectIntl(React.createClass({
     }
 }));
 
-render(<Page><Search /></Page>, document.getElementById('app'));
+var mapStateToProps = function (state) {
+    return {
+        searchTerm: state.searchTerm
```
mewtaylor: when I tried this locally, it turned out that calling this `searchTerm` conflicts with the `getDefaultProps` method that sets the search term initially. We should probably update the way that works, but for now, could you change this to `navigation: state.navigation`?

TheGrits:
@mewtaylor I have made the changes, the ball should be in your court again. 👍

mewtaylor:
On OSX 10.12:
* in Chrome 54, typing something into the search bar results in a search, and the word I typed stays in the search bar while I'm on the search page
* in FF 49, typing something into the search bar results in a search, and the word I typed stays in the search bar while I'm on the search page too

PR #935, opened 2016-09-30
Fix GH-933: Put label for “Not Required” above input

### LINKED ISSUE #933: Match text format/ location for "Not Required" on different steps of Teacher Registration
Staging, Safari Version 9.1.3 (11601.7.8), Mac OS X El Capitan

From https://github.com/LLK/scratch-www/pull/896:

We added "Not Required" by the Zip Code field since we stopped requiring Zip Codes for non-USA countries. We added it in a different location/ format than we did on other steps (e.g. Website field). We should be consistent.

screen shot 2016-09-30 at 2 05 54 pm
screen shot 2016-09-30 at 2 04 54 pm

/cc @carljbowman

mewtaylor:
Fixes #933 by giving the "Not Required" styling used in organization url (i.e. separated out label elements)

/cc @rschamp

### DIFF
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -741,8 +741,15 @@ module.exports = {
                                         required /> :
                                 []
                             }
-                            <Input label={formatMessage({id: 'teacherRegistration.zipCode'})}
-                                   type="text"
+                            <b className="row-label">
+                                <intl.FormattedMessage id="teacherRegistration.zipCode" />
+                            </b>
+                            {this.state.countryChoice !== 'us' ?
+                                <p className="help-text">
+                                    <intl.FormattedMessage id="teacherRegistration.notRequired" />
+                                </p> : []
+                            }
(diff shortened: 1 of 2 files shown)


src/components/registration/steps.jsx
```diff
@@ -752,7 +759,7 @@ module.exports = {
                                            id: 'registration.validationMaxLength'
                                        })
                                    }}
-                                   required={(this.state.countryChoice === 'us')} />
+                                   required={(this.state.countryChoice === 'us') ? true : 'isFalse'} />
```
rschamp: https://github.com/LLK/scratch-www/blob/develop/src/components/forms/input-hoc.jsx#L14 the "not required" help text is added automatically if the field isn't required. Why is the required state set to `'isFalse'` rather than just `false`?

rschamp: Er, I guess the real question is why is that broken, requiring you to use `isFalse`?

PR #953, opened 2016-10-06
[Master] Hotfix: loosen restrictions on smartystreets check

mewtaylor:
re-fixes #828 by loosening restrictions on the check once again (see https://smartystreets.com/docs/us-street-api#match-strategy)

PR #954, opened 2016-10-06
[Develop] Hotfix: loosen restrictions on smartystreets check

mewtaylor:
See #953


jwzimmer:
@mewtaylor I tried the 3 known addresses that were getting rejected, and they were all accepted. I also tried the media lab and a valid TX address. Some clearly invalid addresses do get past, but I think we are okay with that at this point...? (I think even some simple regex would be more accurate than smarty streets at this setting... for example if we didn't allow non-numeric chars in the US zipcodes... but I don't think we care, esp. since we're not validating non-US addresses.) Random non-US addresses are still accepted (as expected).

This clearly invalid address was accepted:
screen shot 2016-10-06 at 5 36 13 pm

Issue #940, opened 2016-10-03
Missing translation strings


chrisgarrity:
Several languages are 100% translated on pootle (e.g. Catalan and Italian), but there are still strings in English on the site, so they are missing from the po files.

- [ ] **About page**

'Who uses Scratch' and 'Scratch in Schools' from the italian translation:

<img width="444" alt="screen shot 2016-09-20 at 11 01 52 am" src="https://cloud.githubusercontent.com/assets/1320456/18675882/cab23d08-7f21-11e6-8dfd-d04415cf4674.png">
<img width="438" alt="screen shot 2016-09-20 at 11 01 22 am" src="https://cloud.githubusercontent.com/assets/1320456/18675883/cab2e5fa-7f21-11e6-9ab4-87f1b15eae4a.png">

- [ ] **"Projects curated by" on the splash page**

<img width="978" alt="screen shot 2016-09-20 at 11 28 35 am" src="https://cloud.githubusercontent.com/assets/1320456/18676973/74475382-7f25-11e6-9e52-a24eed48c555.png">

- [ ] **The words in the graphics on the statistics page?**

<img width="958" alt="screen shot 2016-09-20 at 11 30 07 am" src="https://cloud.githubusercontent.com/assets/1320456/18677083/b0c98910-7f25-11e6-91c7-954b01d6cc4d.png">
<img width="1009" alt="screen shot 2016-09-20 at 11 29 45 am" src="https://cloud.githubusercontent.com/assets/1320456/18677084/b0ccfe38-7f25-11e6-9355-3fc327d1b703.png">

- [ ] **The Credits page?**
https://scratch.mit.edu/info/credits/

<img width="780" alt="screen shot 2016-09-20 at 11 38 52 am" src="https://cloud.githubusercontent.com/assets/1320456/18677469/e67c5528-7f26-11e6-8fce-e3ef28dc6c4f.png">

/ cc @kaschm

jwzimmer:
From @ntlrsk:
> In preparation for EU Code Week (Oct. 15), noting that a couple of the strings are not included in the translation on the HOC page in Italian, even though Italian showing as 100% translated in all projects. 
> Animate Your Name, ANIMATE YOUR NAME (in two places), and Dance, Dance, Dance are showing up in English on the HOC page. 
> (Those phrases are already available in Pootle for Scratch Website for the cards page.)
> Thanks for investigating.

[...]

Issue #959, opened 2016-10-11
Should credits page be translated?


chrisgarrity:
**The Credits page**
https://scratch.mit.edu/info/credits/

<img width="780" alt="screen shot 2016-09-20 at 11 38 52 am" src="https://cloud.githubusercontent.com/assets/1320456/18677469/e67c5528-7f26-11e6-8fce-e3ef28dc6c4f.png">

Currently it is untranslated, and the strings are not part of the pootle database. Should it be translated?

liam4:
Yes.

TheGrits:
@chrisgarrity This will actually be really tricky to localize. Here's an example:

> The following people have also contributed to the development and support of Scratch over the years: Susan Abend, [a bucket load of names], Vladimir Vuksan, Han Xu, and the many volunteer [Scratch Translators](http://wiki.scratch.mit.edu/wiki/Translators) from around the world.

It doesn't help that the link is in the body of the paragraph, and not at the start or end. What would the preferred way of formatting the localization be? (The whole paragraph, or separate the link from the rest of the paragraph?

chrisgarrity:
@TheGrits It is possible technically to localize text with the link in it, so we could just do the whole paragraph. However, actually coming up with something reasonable in other languages may be a challenge. @thisandagain Would it be ok to split this paragraph into two sentences. The original ending with 'and Han Xu.' Followed by something like:

> Scratch is available in many different languages thanks to the many volunteer [Scratch Translators](http://wiki.scratch.mit.edu/wiki/Translators) from around the world.

thisandagain:
@chrisgarrity That sounds good. Thanks!

PR #960, opened 2016-10-11
Fix gh-940: missing translations

chrisgarrity:
Updated the build-locales script to handle duplicate strings. 

You can spot check the output by looking in the intl directory. 

about.intl.js: the key 'about.whoUsesScratch' should be translated in many languages. 
hoc.intl.js: the key 'hoc.tipsAnimateYourNameTitle' should be in languages other than English

### DIFF
--- a/bin/build-locales
+++ b/bin/build-locales
@@ -86,7 +86,7 @@ for (var v in routes) {
     if (typeof routes[v].redirect !== 'undefined') {
         continue;
     }
-    
+
     views.push(routes[v].name);
     try {
         var subdir = routes[v].view.split('/');
@@ -97,7 +97,8 @@ for (var v in routes) {
             en: ids
         };
         idsWithICU = merge(idsWithICU, localeCompare.idToICUMap(routes[v].name, ids));
-        icuWithIds = merge(icuWithIds, localeCompare.icuToIdMap(routes[v].name, ids));
+        // Note: if lodash.merge gets updated to 4.0 or higher this needs to be mergeWith instead
+        icuWithIds = merge(icuWithIds, localeCompare.icuToIdMap(routes[v].name, ids), localeCompare.customMerge);
     } catch (err) {
         if (err.code !== 'MODULE_NOT_FOUND') {
             throw err;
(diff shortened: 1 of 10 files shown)


mewtaylor:
Related to #961

jwzimmer:
### Testing, Chrome 53, Mac OS X El Capitan

- [ ] **About page** 
  - 'Who uses Scratch' and 'Scratch in Schools' from the italian translation
- [ ] **"Projects curated by" on the splash page**
- [ ] **The words in the graphics on the statistics page?**
  - Not sure if should be translated: https://github.com/LLK/scratchr2/issues/4134
- [ ] **The Credits page?**
  - https://scratch.mit.edu/info/credits/
- [ ] In preparation for EU Code Week (Oct. 15), noting that a couple of the strings are not included in the translation on the HOC page in Italian, or Norwegian:
  - Animate Your Name, ANIMATE YOUR NAME (in two places), and Dance, Dance, Dance are showing up in English on the HOC page
- [ ] Noticed this with Portuguese + Teacher Registration today in Production. The validation string seems to be translated sometimes but not always ("this field is required")
  - See Organization step

chrisgarrity:
The statistics page is actually in scratchr2 not www, so those won't be translated. 
The credits page was completely untranslated rather than a few untranslated strings, so I opened a separate issue #959 to decide whether it should be translated.

jwzimmer:
@chrisgarrity Okay, thanks for explaining 👍

jwzimmer:
@mewtaylor & @chrisgarrity I checked this PR pretty narrowly, just looking to see if it fixed the specific issues raised by @ntlrsk and in #940. Let me know if you think it should have more intense scrutiny. Thanks.

chrisgarrity:
@jwzimmer Did the validation messages get fixed? I wasn't sure those where the same problem. If not we should probably open a new issue for them as it wasn't a symptom of the same bug.

jwzimmer:
@chrisgarrity I opened a new issue for it as https://github.com/LLK/scratch-www/issues/966. : )

PR #961, opened 2016-10-11
[Master] Hotfix - update build-locales script

### LINKED ISSUE #940: Missing translation strings

chrisgarrity:
Fixes the problem where the same literal string is associated with different message keys on multiple pages. Fixes #940

### DIFF
--- a/bin/build-locales
+++ b/bin/build-locales
@@ -86,7 +86,7 @@ for (var v in routes) {
     if (typeof routes[v].redirect !== 'undefined') {
         continue;
     }
-    
+
     views.push(routes[v].name);
     try {
         var subdir = routes[v].view.split('/');
@@ -97,7 +97,8 @@ for (var v in routes) {
             en: ids
         };
         idsWithICU = merge(idsWithICU, localeCompare.idToICUMap(routes[v].name, ids));
-        icuWithIds = merge(icuWithIds, localeCompare.icuToIdMap(routes[v].name, ids));
+        // Note: if lodash.merge gets updated to 4.0 or higher this needs to be mergeWith instead
+        icuWithIds = merge(icuWithIds, localeCompare.icuToIdMap(routes[v].name, ids), localeCompare.customMerge);
     } catch (err) {
         if (err.code !== 'MODULE_NOT_FOUND') {
             throw err;
--- a/bin/lib/locale-compare.js
+++ b/bin/lib/locale-compare.js
@@ -6,7 +6,7 @@ var crypto = require('crypto');
 var fs = require('fs');
 var path = require('path');
 var po2icu = require('po2icu');
-var isArray = require('lodash.isArray');
+var isArray = require('lodash.isarray');
 
 var Helpers = {};
 
--- a/bin/lib/locale-compare.js
+++ b/bin/lib/locale-compare.js
@@ -6,6 +6,7 @@ var crypto = require('crypto');
 var fs = require('fs');
 var path = require('path');
 var po2icu = require('po2icu');
+var isArray = require('lodash.isArray');
 
 var Helpers = {};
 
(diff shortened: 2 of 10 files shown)


bin/build-locales
```diff
@@ -97,7 +97,8 @@ for (var v in routes) {
             en: ids
         };
         idsWithICU = merge(idsWithICU, localeCompare.idToICUMap(routes[v].name, ids));
-        icuWithIds = merge(icuWithIds, localeCompare.icuToIdMap(routes[v].name, ids));
+        // Note: if lodash.merge gets updated to 4.0 or higher this needs to be mergeWith instead
```
mewtaylor: ooh – thanks for this info :)

src/views/splash/splash.jsx
```diff
@@ -215,11 +215,11 @@ var Splash = injectIntl(React.createClass({
                 <Box
                         key="curator_top_projects"
                         title={
-                            'Projects Curated by ' +
+                            formatMessage({id: 'splash.projectsCuratedBy', defaultMessage: 'Projects Curated by '}) +
```
mewtaylor: If this is in `splash/l10n.json`, then we actually don't need to include the `defaultMessage` – we could just add it by id.

package.json
```diff
@@ -59,6 +59,7 @@
     "keymirror": "0.1.1",
     "lodash.clone": "3.0.3",
     "lodash.defaultsdeep": "3.10.0",
+    "lodash.isarray": "^3.0.4",
```
mewtaylor: Could you remove the caret? We want to keep the versioning locked in so that we can manually update packages and make sure they don't break anything – otherwise packages will update every time we deploy.

bin/lib/locale-compare.js
```diff
@@ -6,6 +6,7 @@ var crypto = require('crypto');
 var fs = require('fs');
 var path = require('path');
 var po2icu = require('po2icu');
+var isArray = require('lodash.isArray');
```
mewtaylor: I think tests are failing due to a name mismatch between this and `isarray`

package.json
chrisgarrity: Absolutely, I forgot that --save-dev adds the '^'

src/views/splash/splash.jsx
chrisgarrity: Good to know. I was just using the following 'Learn More' message as a template.

bin/lib/locale-compare.js
chrisgarrity: No, actually I didn't realize there were tests for the script. The new md5 map expects the value to be an array not a string, but I hadn't changed the fixture.

chrisgarrity:
Failing tests should be fixed, and other comments addressed. I'll add another test later, but wanted to make sure it didn't hold anything up.

package.json
mewtaylor: ahhh right yup I forget that just about every time I add a package

src/views/splash/splash.jsx
mewtaylor: ahh ok – that's probably an old string that I have failed to update to the new style we use. Thanks for the heads up that we should update that!

mewtaylor:
@chrisgarrity we should wait to hotfix until traffic goes down, but happy to hotfix it later today.

PR #964, opened 2016-10-12
Fix GH-899: Re-searching should not search for 'undefined'

### LINKED ISSUE #899: Re-searching for same term or no term with Enter causes term to be replaced by "undefined"
On the new Search page, Production, Chrome Version 52.0.2743.82 (64-bit) on Mac OS X El Capitan 10.11.6 

Steps to repro:
- Go to the search page by searching for "anything" in the nav bar
- Results returned will be relevant to "anything"
- Put your cursor in the search bar on the search page and hit the enter key
- Page will refresh with "undefined" as the search term in the query string (URL) and in the search bar on the search page
- Results returned will be relevant to "undefined"

out

/cc @mewtaylor

technoboy10:
This PR resolves #899 by adding a check for an undefined formData.q value (the search page populates the search bar's defaultValue, but not actual value), and searches for `this.props.searchTerm` if the check fails. 

Another option would be to prevent the user from searching the same term repeatedly, but I'm not sure what UX is preferable (having the page react like it normally done when searching, vs saving bandwidth/time by assuming the user doesn't want to search for the same thing again).

### DIFF
--- a/src/views/search/search.jsx
+++ b/src/views/search/search.jsx
@@ -75,7 +75,11 @@ var Search = injectIntl(React.createClass({
         }.bind(this));
     },
     onSearchSubmit: function (formData) {
-        window.location.href = '/search/projects?q=' + formData.q;
+        if (!formData.q){
+            window.location.href = '/search/projects?q=' + this.props.searchTerm;
+        } else {
+            window.location.href = '/search/projects?q=' + formData.q;
+        }
     },
     getTab: function (type) {
         var term = this.props.searchTerm.split(' ').join('+');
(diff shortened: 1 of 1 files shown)


mewtaylor:
Thanks @technoboy10 – this is a great solution. I think that, ideally, the component would recognize `defaultValue` itself and add it to the formdata, but it appears that they're [still working on this issue](https://github.com/christianalfoni/formsy-react/issues/209).

It looks like if we set `value` to `this.props.searchTerm` we'd also have a solution that would eliminate the need for a conditional, but I'm not sure if that would be a better solution or not. What do you think?

technoboy10:
It looks like replacing `defaultValue` with `value` works just as well, and reads a little bit better, I think.

mewtaylor:
On Chrome 53, OSX 10.12:
[ ] – Focus on input field on the search page, don't change the text, and search. It displays the same results as before, with the same search term
[ ] – Focus on the input field on the search page, remove the text, and search with an empty field. It displays the results as if there was no search term

PR #968, opened 2016-10-14
Use no-cache so ETags are used

### LINKED ISSUE #787: Deployment consistently breaks cached pages

rschamp:
Fixes #787 (again).

From https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching

> "no-cache" indicates that the returned response can't be used to satisfy a subsequent request to the same URL without first checking with the server if the response has changed. As a result, if a proper validation token (ETag) is present, no-cache incurs a roundtrip to validate the cached response, but can eliminate the download if the resource has not changed.

This is what we want to happen.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 ESLINT=./node_modules/.bin/eslint
 NODE=node
 SASSLINT=./node_modules/.bin/sass-lint -v
-S3CMD=s3cmd sync -P --delete-removed --add-header=Cache-Control:public,max-age=3600
+S3CMD=s3cmd sync -P --delete-removed --add-header=Cache-Control:no-cache,public,max-age=3600
 TAP=./node_modules/.bin/tap
 WATCH=./node_modules/.bin/watch
 WEBPACK=./node_modules/.bin/webpack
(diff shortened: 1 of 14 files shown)


rschamp:
Just updated to keep the other policies — they still apply.

rschamp:
Also, maybe we should hotfix this?

thisandagain:
This is great. Thanks.

rschamp:
Welp, that force push screwed the commit list up. Reopening as a hotfix.

PR #976, opened 2016-10-19
Implement GH-856: Add new cards and card thumbs to www

### LINKED ISSUE #856: Add 10 new cards to /cards page
New resources! 

We have ten new cards to be added to the /cards page on Scratch. Follow current design pattern (name of theme / img of card / link to PDF). 

Currently blocked on final content from @ntlrsk, including any potential call-out to No Starch Press

mewtaylor:
Fixes #856. This also updates the styling of the `cards.jsx` view to follow the trello style guide, as previously discussed.

@ntlrsk just fyi I moved the cards over to the www repo so that they could be protected by cdn while also be purge-able from the cdn automatically when updates occur.


rschamp:
Looks good to me so long as moving pdfs to www is ok.  Seems to me it should make them easier to manage since even if we do need to link to them in scratchr2, we can just use the `/pdfs/...` urls and will be able to rely on the Fastly purge to update them when necessary.

mewtaylor:
@rschamp absolutely – I talked with @ntlrsk about this already, who was cool with moving them over. I'll remove them from the old repo so we don't have duplication either.

PR #985, opened 2016-10-26
Fix gh-706: "This field is required" when phone number is null

### LINKED ISSUE #706: Phone number field says "enter a valid phone number" rather than "this field is required" - Teacher Registration
Chrome, Production

(The phone number field is now required, so that functionality was fixed in https://github.com/LLK/scratch-www/issues/693; this is just about making the functionality nicer.)

Since you can't totally clear the phone number field (can't delete the +), it always tells you "please enter a valid phone number" rather than "this field is required", even if you have not filled out/ touched the field. That might be confusing/ frustrating for people trying to skip entering a phone number altogether because they think it is optional.

screen shot 2016-07-08 at 10 14 28 am

TheGrits:
Should fix #706 

Separated the null check from the area code check so that when the field is null the 'required' popup will be displayed.

### DIFF
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -455,7 +455,11 @@ module.exports = {
             };
         },
         onValidSubmit: function (formData, reset, invalidate) {
-            if (!formData.phone || formData.phone.national_number === '+') {
+            if (!formData.phone) {
+                return invalidate({
+                    'phone': this.props.intl.formatMessage({id: 'teacherRegistration.validationRequired'})
+                });
+            } else if (formData.phone.national_number === '+') {
                 return invalidate({
                     'phone': this.props.intl.formatMessage({id: 'teacherRegistration.validationPhoneNumber'})
                 });
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -455,13 +455,9 @@ module.exports = {
             };
         },
         onValidSubmit: function (formData, reset, invalidate) {
-            if (!formData.phone) {
+            if (!formData.phone || formData.phone.national_number === '+') {
                 return invalidate({
                     'phone': this.props.intl.formatMessage({id: 'teacherRegistration.validationRequired'})
-                });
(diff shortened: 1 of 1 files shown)


TheGrits:

src/components/registration/steps.jsx
```diff
@@ -455,7 +455,11 @@ module.exports = {
             };
         },
         onValidSubmit: function (formData, reset, invalidate) {
-            if (!formData.phone || formData.phone.national_number === '+') {
+            if (!formData.phone) {
```
mewtaylor: From what @jwzimmer mentioned in the issue, I think that just having the `+` is also a situation in which we'd want to show the `validationRequired` message. I'd recommend keeping the logic statement as it was before, but just changing the message from `validationPhoneNumber` to `validationRequired`, if that makes sense.

TheGrits: @mewtaylor Sounds good! I'll make the change now...

TheGrits:
@mewtaylor I have made the change requested, this one should be ready to go!

mewtaylor:
@TheGrits do you mind adding test cases to this? Feel free to just pull the test cases from the original issue, but it'd be super helpful for when we do qa to have that here too. Thanks!

TheGrits:
@mewtaylor Sure thing!

mewtaylor:
On Chrome 54, OSX 10.12:
- [x] "This field is required" is the message when the phone number field as just a "+" or a "+1"

When I click "Next step" and I haven't checked the checkbox either on that page, then the only error that shows up is the checkbox error – not any phone number error. However, if I check the checkbox, then the "This field is required" field will show up.

If I choose a different country code that's not only 1 digit long, then I will get "Please enter a valid phone number" rather than "this field is required" if the country code is longer than 1 digit.

Considering that country codes are variable in length, and that the validation bubble issue was not in the scope of this, I'm going to mark this as QA-OK, since it handles the case of trying to move forward without attempting to add a phone number at all (since the us country code is the default).

PR #1003, opened 2016-11-07
Add Zulu to Language Menu

EdwardGeneralov:
Scratch share user, See LLK/scratch-flash/issues/1230

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -30,6 +30,7 @@
     "he": "עִבְרִית",
     "hi": "हिन्दी",
     "hr": "Hrvatski",
+    "zu": "isiZulu",
     "is": "Íslenska",
     "it": "Italiano",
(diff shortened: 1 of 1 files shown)


TheGrits:
@chrisgarrity The reason Travis is failing currently is because Zulu is only complete for the editor/blocks, and not the website. https://translate.scratch.mit.edu/zu/ Can languages be added to the editor but not the main site?

EdwardGeneralov:
@TheGrits @chrisgarrity Are Zulu Scratch ?

chrisgarrity:
@EdwardGeneralov @TheGrits The check is failing because we need to add Zulu to the scratchr2_translations package repo. Scratchr2_translations will automatically use the default English for anything that is not yet translated on the translation server, and the spot check of translations should pass.

Issue #980, opened 2016-10-24
Create New Modal Component


mewtaylor:
We currently use `react-modal` to handle some of our modal elements, however it would be nice to have our own modal that can be a bit more customizable to the use cases we need it for. See #979 for a specific, new modal use case. The design for this modal needs to include the following elements:
- An "x" in the top right corner to close the modal (or top left, for rtl)
- Use `onclickoutside` so that clicking the opaque background closes the modal
- Rounded corners
- A blue opacity for the page behind

Needs a mobile design as well

carljbowman:
![r2 0_modal](https://cloud.githubusercontent.com/assets/3409578/19689513/a65ad6fa-9a9b-11e6-96f9-fac56008be24.png)

![r2 0_modal grid](https://cloud.githubusercontent.com/assets/3409578/19689512/a653d076-9a9b-11e6-8c69-0a473ad985d6.png)

kaschm:
Global content for modal:

Tutorial:
Make your own project using this step-by-step guide. 

Activity Cards:
Explore new coding ideas using this set of illustrated cards you can print out.

Educator Guide: 
Use this guide to plan and lead a one-hour Scratch workshop.

ntlrsk:
(Note: we updated the text above.)

carljbowman:
@ntlrsk - What do you think of the banner images?

Animate Your Name
![animate-your-name-banner](https://cloud.githubusercontent.com/assets/3409578/20073117/f51e3dd8-a4f8-11e6-9f50-128d9b90bd93.jpg)

Make it Fly
![make-it-fly-banner](https://cloud.githubusercontent.com/assets/3409578/20073132/fff2e22c-a4f8-11e6-94ff-df9db38d649c.jpg)

Make Music
![make-music-banner](https://cloud.githubusercontent.com/assets/3409578/20073139/092d57f0-a4f9-11e6-9da5-d37ab153060d.jpg)

Race to the Finish
![race-to-the-finish-banner](https://cloud.githubusercontent.com/assets/3409578/20073156/1c27fc48-a4f9-11e6-83e9-957844a7d044.jpg)

Hide and Seek
![hide-and-seek-banner](https://cloud.githubusercontent.com/assets/3409578/20073166/223c7db6-a4f9-11e6-8b49-3545d319ed6f.jpg)

Create a Story
[...]

PR #1008, opened 2016-11-08
Implement GH-980: add new TTT modal component

### LINKED ISSUE #945: Can't sign up from homepage on mobile (iOS) - can't zoom out to click Next button
Using Chrome 53.0.2785.109 on my iPhone 6, Production

I could not create a new account / join Scratch from the www pages (About, homepage) linked to on the mobile homepage. I didn't try all the pages linked to, but I would guess the www pages all have the same issue. I could not zoom out to see the Next button to proceed past the username/ password step in the Join Flow.

I manually went to the URL ending in `/discuss` to get a scratchr2 page and the old navbar in order to create an account. From there, I was able to zoom out enough to get to the Next button and complete the Join Flow.

I would like to see if this happens on other devices/ to other people, just in case it's something unique to my phone. (I wouldn't expect it to be, though.)

To repro:
- On an iPhone, open Chrome
- Go to Scratch homepage
- Click Join Scratch
- In pop-up window (Join Flow), fill out username/ password fields
- Try to click Next button (lower right corner of pop-up)
- Scroll out/ zoom out to try to make Next button visible

img_1250

mewtaylor:
In addition to fixing #980, this also does a couple of things:
1. Update `ReactModal` dependency
2. Re-style the base modal component
3. Create a new iframe modal component, and in so doing fixes #945 

Question left to discuss – right now, the text specific to the `TTTModal` component assumes the strings are made available by whatever `view` is calling it, and include the `FormatMessage` components on it directly. I did this in order to reduce the complexity of the modal's props. However, I would be up for switching it to receiving those strings as props if that seems cleaner to others.

### Test Cases ###
* On a desktop, the modal should show up with blue overlay
* On a tablet/mobile device, the modal should take up the entire screen and be vertically scrollable. The `X` in the top right should follow you as you scroll to exit
* Iframe modals should be horizontally scrollable on smaller screens so that they can still be viewable
* You should still be able to register for Scratch from a desktop
* You should now be able to register for Scratch from tablet/mobile


src/views/splash/splash.jsx
```diff
@@ -344,21 +342,23 @@ var Splash = injectIntl(React.createClass({
...
             <div className="splash">
                 {this.shouldShowEmailConfirmation() ? [
-                    <DropdownBanner key="confirmedEmail"
-                            className="warning"
-                            onRequestDismiss={this.handleDismiss.bind(this, 'confirmed_email')}>
+                    <DropdownBanner
+                        key="confirmedEmail"
+                        className="warning"
+                        onRequestDismiss={this.handleDismiss.bind(this, 'confirmed_email')}
+                    >
```
chrisgarrity: I don't get to comment on unchanged lines, but shouldn't the 'Confirm your email... to enable sharing', just below, be wrapped in a FormatHTMLMessage to be translated?

src/components/modal/ttt/modal.jsx
```diff
@@ -0,0 +1,111 @@
...
+                                    alt="activity-cards-icon"
+                                />
+                                <FormattedMessage id="ttt.activityTitle" />
+                            </div>
+                            <p className="modal-content-ttt-subtitle">
+                                <FormattedMessage id="ttt.activitySubtitle" />
+                            </p>
+                        </div>
+                        <a
+                            href={this.props.tutorialLoc}
```
mewtaylor: Ah thanks :)

src/views/splash/splash.jsx
mewtaylor: that's a good point – I'll add that, thanks!

src/components/modal/iframe/modal.jsx
```diff
@@ -0,0 +1,47 @@
...
+        isOpen: React.PropTypes.bool,
+        onRequestClose: React.PropTypes.func,
+        className: React.PropTypes.string,
+        componentRef: React.PropTypes.string,
+        src: React.PropTypes.string
+    },
+    getDefaultProps: function () {
+        return {
+            className: '',
+            iframeClassName: ''
```
mewtaylor: 👍 yup my bad thanks!

src/components/modal/iframe/modal.jsx
```diff
@@ -0,0 +1,47 @@
...
+        };
+    },
+    render: function () {
+        var iframeClasses = classNames(
+            'modal-content-iframe',
+            this.props.className
+        );
+        return (
+            <Modal
+                key={this.props.componentKey}
```
mewtaylor: I think you're right – I think I was getting caught up in the [`ref` issue](https://github.com/LLK/scratch-www/blob/develop/src/views/splash/splash.jsx#L84) and was doing the same for key, so I'll take that out.

src/components/modal/iframe/modal.jsx
```diff
@@ -0,0 +1,47 @@
...
+                    ref={this.props.componentRef}
+                    src={this.props.src}
+                    className={iframeClasses}
+                />
+            </Modal>
+        );
+    }
+});
+
+module.exports = IframeModal;
```
mewtaylor: It was for me just reducing the amount of line duplication. These were all set up in pretty much the same exact way, and this made it so that all of the `iframe` modals, which there are 4 of, had a couple lines fewer of code.

src/components/modal/base/modal.jsx
```diff
@@ -0,0 +1,58 @@
...
+        setAppElement: ReactModal.setAppElement
+    },
+    propTypes: {
+        className: React.PropTypes.string,
+        overlayClassName: React.PropTypes.string
+    },
+    getDefaultProps: function () {
+        return {
+            className: '',
+            overlayClassName: ''
```
mewtaylor: 👍

mewtaylor:
@rschamp changes posted! I still have `IFrameModal` in there bc of the comment above, let me know what you think though!

src/components/modal/base/modal.jsx
```diff
@@ -0,0 +1,52 @@
...
+            'modal-content',
+            this.props.className
+        );
+        var overlayClasses = classNames(
+            'modal-overlay',
+            this.props.overlayClassName
+        );
+        return (
+            <ReactModal
+                ref="modal"
```
rschamp: As long as we're refactoring this and testing, we should get rid of this string ref — they are deprecated. Instead make it a function like `ref={function (c){ this.modal = c;}.bind(this)}`, and then up above don't refer to `this.refs.modal`, just refer to `this.modal`

src/views/splash/splash.jsx
```diff
@@ -344,25 +343,39 @@ var Splash = injectIntl(React.createClass({
...
+                    title="Make it Fly"
+                    description="Animate the Scratch Cat, the Powerpuff Girls, or even a taco!"
+                    tutorialLoc="/"
+                    activityLoc="/pdfs/cards/Scratch2Cards.pdf"
+                    guideLoc="/pdfs/cards/Scratch2Cards.pdf"
+                    bannerUrl="/images/ttt/make-it-fly-banner.jpg"
+                    thumbUrl=""
+                    modalProps={{
+                        isOpen: true,
+                        ref: 'tttModal'
```
rschamp: If this ref is unused then I think it should be removed.

src/components/registration/registration.jsx
```diff
@@ -36,18 +34,14 @@ var Registration = React.createClass({
...
-                    onRequestClose={this.props.onRequestClose}
-                    className="registration"
-                    style={{content:frameProps}}>
-                <iframe ref="registrationIframe" src="/accounts/standalone-registration/" {...frameProps} />
-            </Modal>
+            <IframeModal
+                isOpen={this.props.isOpen}
+                onRequestClose={this.props.onRequestClose}
+                className="mod-registration"
+                componentRef="registrationIframe"
```
rschamp: I think this will break the end of the registration flow since onMessage refers to this.refs.registrationIframe which won't exist anymore. This is the perfect place for a function ref, like `iframeRef={function(iframe){ this.registrationIframe = iframe; }}`, and then in `onMessage`, `this.registrationIframe.contentWindow.location.reload()`.

src/components/modal/iframe/modal.jsx
```diff
@@ -0,0 +1,39 @@
...
+require('./modal.scss');
+
+Modal.setAppElement(document.getElementById('view'));
+
+var IframeModal = React.createClass({
+    propTypes: {
+        isOpen: React.PropTypes.bool,
+        onRequestClose: React.PropTypes.func,
+        className: React.PropTypes.string,
+        componentRef: React.PropTypes.string,
```
rschamp: String refs are deprecated, and for this usage, we most likely would always want to use a function ref anyway, so I think this should change to `React.PropTypes.func`.

src/views/splash/splash.jsx
```diff
@@ -344,25 +343,39 @@ var Splash = injectIntl(React.createClass({
...
-                           style={{content: emailConfirmationStyle}}>
-                        <iframe ref="emailConfirmationiFrame"
-                                src="/accounts/email_resend_standalone/"
-                                {...omit(emailConfirmationStyle, 'padding')} />
-                    </Modal>
+                    <IframeModal
+                        isOpen={this.state.emailConfirmationModalOpen}
+                        onRequestClose={this.hideEmailConfirmationModal}
+                        className="mod-confirmation"
+                        componentRef="emailConfirmationiFrame"
```
rschamp: Same advice about using a function ref here.

src/views/splash/splash.jsx
mewtaylor: This wasn't supposed to be committed – tower was supposed to exclude this, so I'll remove the whole thing

src/components/modal/base/modal.jsx
```diff
@@ -29,9 +28,15 @@ var Modal = React.createClass({
             'modal-overlay',
             this.props.overlayClassName
         );
+        ReactModal.setAppElement(document.getElementById(this.props.appElementId));
```
rschamp: I think you left this in by mistake?

src/components/modal/iframe/modal.jsx
```diff
@@ -21,11 +20,7 @@ var IframeModal = React.createClass({
             this.props.className
         );
         return (
-            <Modal
-                isOpen={this.props.isOpen}
-                onRequestClose={this.props.onRequestClose}
-                className={this.props.className}
-            >
+            <Modal {...omit(this.props, ['ref', 'src'])}>
```
rschamp: You can exclude `ref` because `ref` is a special attribute, it's not passed in as a prop.

src/components/modal/base/modal.jsx
mewtaylor: yup thanks!

mewtaylor:
@rschamp made those last couple of edits, thanks!

rschamp:
Looks good to me — just noticed @chrisgarrity didn't remove assignment, so leaving it as needs review in case she had more to say.

chrisgarrity:
@mewtaylor @rschamp Nope, just forgot to remove myself. Looks good.

src/components/modal/ttt/modal.jsx
```diff
@@ -0,0 +1,121 @@
...
+
+var TTTModal = React.createClass({
+    propTypes: {
+        title: React.PropTypes.string.isRequired,
+        description: React.PropTypes.string.isRequired,
+        tutorialLoc: React.PropTypes.string.isRequired,
+        activityLoc: React.PropTypes.string.isRequired,
+        guideLoc: React.PropTypes.string.isRequired,
+        thumbUrl: React.PropTypes.string.isRequired,
+        bannerUrl: React.PropTypes.string.isRequired,
```
TheGrits: This is the trailing comma Travis is flagging about, (and pardon my ignorance) I believe that it's a typo?

mewtaylor: 👍 you're totally right, thanks!

TheGrits: @mewtaylor Any time!

jwzimmer:
### Testing - in progress : )

- Looking at https://github.com/LLK/scratch-www/issues/980 to check Modal appearance
  - **Chrome 54, Mac OS 10.11**
  - [x] Modal has an X in upper right corner
    - The X works to close the modal
  - [x] Modal has curved corners
  - [ ] Modal has blue overlay for page behind it, but the shade of blue looks different to me

jwzimmer:
https://github.com/LLK/scratch-www/wiki/Test-Plan-for-Things-to-Try-page

Issue #942, opened 2016-10-04
Clarify Compatibility & Support Info in FAQ


jwzimmer:
Currently we don't make it clear whether we're discussing requirements for the website or the editor (including online editor), so we should clarify that. We don't support the online editor with Linux with any browser except Chrome (latest two versions **for linux** which are generally a bit behind latest two versions for mac).

https://scratch.mit.edu/info/faq/

https://scratch.mit.edu/scratch2download/ (mentions Linux)

(Also, internally, we should clarify the wiki: https://github.com/LLK/scratch-ops/wiki/Compatibility-&-Support)

/cc @thisandagain

jwzimmer:
I updated the wiki to include a note about Scratch website vs. online editor.

My understanding is we support the same things for Scratch website & online editor, except we don't support online editor with Linux.

Current information on website:
> **What are the system requirements for Scratch?**
> To run Scratch 2, you need to be using (1) a Mac, Linux, or Windows computer; (2) a version of Adobe Flash Player released on or after June 15, 2016; (3) a relatively recent web browser: one of the latest two versions of Chrome, Firefox, Safari (Mac or Windows only), Edge (Windows only), or Internet Explorer 10+ (Windows only). If your computer doesn’t meet these requirements, you can try downloading and installing Scratch 1.4, which you can still use to share projects to the Scratch 2 website.

Proposed change:
> > **What are the system requirements for Scratch?**
> To run Scratch 2, you need to be using (1) a Mac or Windows computer; (2) a version of Adobe Flash Player released on or after June 15, 2016; (3) a relatively recent web browser: one of the latest two versions of Chrome, Firefox, Safari (Mac or Windows only), Edge (Windows only), or Internet Explorer 10+ (Windows only). If your computer doesn’t meet these requirements, you can try downloading and installing Scratch 1.4, which you can still use to share projects to the Scratch 2 website. The Scratch website, other than the online editor, can be used on a Linux computer.

Alternately, we could probably just leave out the last sentence to make it less confusing, and just not mention Linux at all. People browsing the website are probably not usually thinking about compatibility like they would w.r.t. a piece of software like the editor, so I'm not sure how relevant or useful that info would be to users.

@kaschm - thoughts on this copy?

TheGrits:
[...]

PR #1012, opened 2016-11-10
Fix gh-942: update compatibility & support FAQ response

jwzimmer:
@mewtaylor Would you mind going over this with me when you have time? I wasn't sure what the comment about updating the `bowser` settings meant exactly, and I am also not sure my repository is actually in the state I want it to be (and I don't want to accidentally break anything!). Thank you. : )

Fixes https://github.com/LLK/scratch-www/issues/942.


mewtaylor:
@jwzimmer this looks good to me! The bowser thing is in scratchr2, so nothing needs to be done here – I can explain what I meant by that another time.

mewtaylor:
@jwzimmer @rschamp if we want this to go into the hotfix, we'll need to re-open this against the right branch. Is that the preference?

rschamp:
I have no preference.  If we do decide to put it into the hotfix, I will suggest just updating the base branch on this PR, you can do that on GitHub now by clicking Edit next to the PR title!

kaschm:
The link to 'Adobe Flash Player' currently goes to the Scratch homepage. I believe it should be something like **https://get.adobe.com/flashplayer/** instead? @mewtaylor

kaschm:
Capturing my issue here: 
https://github.com/LLK/scratch-www/issues/1130

Because it's happening on production, not just staging.

PR #1017, opened 2016-11-14
Fix gh-925: setup initial automated test

### LINKED ISSUE #925: Set up how to run automated integration tests
Continuing initial set-up started in https://github.com/LLK/scratch-www/issues/590. The first test is checked in, but we need to figure out how we'll be running it as part of the development process.
- [ ] Change driver URL to a variable so it isn't hardcoded.
  Re [this line](https://github.com/LLK/scratch-www/blob/develop/test/integration/check_navbar_links.js#L13)
  @rschamp said:
  
  > Maybe it should come from an environment var || "scratch.mit.edu"?
  > and then when we run the tests we'll run it with that var = scratch.ly
- [ ] Set up running with travis/ sauce: https://github.com/LLK/scratch-www/pull/920#discussion_r78991538
- [ ] Maybe set up running with multiple browser drivers: https://github.com/LLK/scratch-www/pull/920#discussion_r78994506

jwzimmer:
Fixes #925.

@mewtaylor Would you mind going over this with me when you have time?

In #925, we said:
- Change driver URL to a variable so it isn't hardcoded --> I added process.env TEST_DOMAIN variable
- Set up running with travis/ sauce --> I am not sure how to do this, but from the conversation in the meeting last Thursday, it sounded like we were going to remove testing dependencies from dev dependencies and set-up running various kinds of tests. I removed selenium as a dependency... not sure how to do the set up for running the tests.
- Maybe set up running with multiple browser drivers --> Haven't addressed yet, seems like we could use sauce to do this and maybe makes sense to do at the same time as the other parts of the set-up? (Ideally, maybe we could pass in drivers as a variable and iterate through them?)


test/integration/check_navbar_links.js
```diff
@@ -7,10 +7,11 @@
 var tap=require('tap');
 var seleniumWebdriver = require('selenium-webdriver');
 
-//chrome driver
-var driver = new seleniumWebdriver.Builder().withCapabilities(seleniumWebdriver.Capabilities.chrome()).build();
+//Passing in the driver as a variable?
+var driver = process.env.DRIVER || seleniumWebdriver.Builder().withCapabilities(seleniumWebdriver.Capabilities.chrome()).build();
```
jwzimmer: I don't think this would work... the proces.env docs said that by default passed-in variables are strings, so I need to look into how to do this properly, but I just wanted to illustrate what I was trying to ask about when I mentioned passing in the drivers as variables...

jwzimmer:
Closing in favor of set-up that @cwillaim has done. : )

Issue #981, opened 2016-10-24
Create HOC Homepage Banner


mewtaylor:
New banner at the top of the homepage that will be present for the two weeks around Hour of Code. This will contain links to the new T3 page, as well as specific resources to get started with (in the form of new tiles, as specified in #979).

Components to use:
- Homepage banner element
- New resource tile (see #979)
- New event row (see #982)

carljbowman:
# Logged-in State
* Includes 'Get Creative with Coding' Banner 
* Includes 'Design a Character' Event Row 

## Desktop - 12col

#### 'Get Creative with Coding' Banner
* 'More things to try' points to the Things to Try page (see #979)
* Resource tiles (everything above dotted line) point to their associated Tutorial
* 'See Cards and Guides' will open a modal with associated resources

#### 'Design a Character' Event Row
* Call to action & copy still being worked on - @kaschm 
* 'See the studio' will point at the studio - @kaschm 
* Video embed URL - @kaschm 

![r2 0_splash-logged-in](https://cloud.githubusercontent.com/assets/3409578/19686543/9285edec-9a8f-11e6-9e16-37689e43a1d5.jpg)

![r2 0_splash-logged-in grid](https://cloud.githubusercontent.com/assets/3409578/19686754/9469edba-9a90-11e6-83b7-d4e1eeaecd0c.jpg)


## Tablet to Desktop - 8col

#### 'Get Creative with Coding' Banner
* Tutorial tiles hidden
* Center align and stack text / 'See things to try' button

#### 'Design a Character' Event Row
* Hidden

![r2 0_splash-logged-in 8col](https://cloud.githubusercontent.com/assets/3409578/19687601/1a16e302-9a94-11e6-9f8b-3d8c5eabb09a.jpg)

![r2 0_splash-logged-in 8col-grid](https://cloud.githubusercontent.com/assets/3409578/19687603/1cb891f0-9a94-11e6-83ae-78099d7f5168.jpg)

## Mobile to Tablet - 6col

#### 'Get Creative with Coding' Banner
* Tutorial tiles hidden
* Center align and stack text / 'See things to try' button

#### 'Design a Character' Event Row
* Hidden
  
![r2 0_splash-logged-in 6col](https://cloud.githubusercontent.com/assets/3409578/19687916/756ddcd2-9a95-11e6-9ca0-1ce7f455cf06.jpg)

[...]

Issue #979, opened 2016-10-24
Create "Things to Try" page


mewtaylor:
New page that will contains resources such as tutorials, educator guides and activity cards for getting started with Scratch.

Components this will need:
- `TitleBanner`
- Resource Tile (new component)
- Resource Modal (new component) (see #980)

carljbowman:
## Things to Try page - Responsive Behavior

### Desktop - 12col
* The vertical size of resource tiles will vary. This variation can be more dramatic based on translation. So rather than setting a fix char length or hiding overflown content, we should use a 'masonry' grid and allow for tiles to paginate asymmetrically

![r2 0_thing-to-try 12col](https://cloud.githubusercontent.com/assets/3409578/19690566/1db1b5a4-9a9f-11e6-9d56-672f9a24c9ab.png)

![r2 0_thing-to-try 12col-grid](https://cloud.githubusercontent.com/assets/3409578/19690547/1142239e-9a9f-11e6-8d58-87d0efd5d9e4.png)

### Tablet to Desktop - 8col
![r2 0_thing-to-try 8col](https://cloud.githubusercontent.com/assets/3409578/19691216/7e37923e-9aa1-11e6-9691-51dbdfa5a7de.png)

![r2 0_thing-to-try 8col-grid](https://cloud.githubusercontent.com/assets/3409578/19691220/80a88028-9aa1-11e6-95cc-59f538e08fe9.png)

### Mobile to Tablet - 6col
_A perfect example of where our current fixed grid system could use some tweaking. Lots of unused space. A flexible grid could help fill the space across more devices, but we would need to design with more variation in mind_

![r2 0_thing-to-try 6col](https://cloud.githubusercontent.com/assets/3409578/19691879/d8f7c7d2-9aa3-11e6-80a0-b6929c230416.png)

![r2 0_thing-to-try 6col-grid](https://cloud.githubusercontent.com/assets/3409578/19691880/db39ffa6-9aa3-11e6-986a-0d4e6d39bda1.png)


### Mobile - 4col
![r2 0_thing-to-try 4col](https://cloud.githubusercontent.com/assets/3409578/19692425/e18cc440-9aa5-11e6-87ac-60a7c6243f30.png)

![r2 0_thing-to-try 4col-grid](https://cloud.githubusercontent.com/assets/3409578/19692431/e9365f6c-9aa5-11e6-9c2e-494464010fc1.png)

## 'See Cards & Guides' Modal - Responsive Behavior

[...]

PR #1020, opened 2016-11-16
Feature GH-981: Add HOC banner component

mewtaylor:
Implements #981.

A couple things to note about this from the specs:
* modals disappear for tablet and below (because the tiles in the banners disappear)
* The teacher banner, for teachers, will appear above the hoc banner

### Test Cases ###
* Banner appears for logged in and logged out users
* Banner appears below teacher banner for teachers
* Tiles in banner show modal when the "see guides" part is clicked
* Modals disappear below our `desktop` breakpoint
* Tiles disappear below our `desktop` breakpoint


src/views/splash/hoc-banner/hoc-banner.jsx
```diff
@@ -0,0 +1,115 @@
...
+                description: formatMessage({id: tiles[2].description}),
+                tutorialLoc: tiles[2].tutorialLoc,
+                activityLoc: formatMessage({id: tiles[2].activityLoc}),
+                guideLoc: formatMessage({id: tiles[2].guideLoc}),
+                thumbUrl: tiles[2].thumbUrl,
+                bannerUrl: tiles[2].bannerUrl
+            }
+        };
+        
+        return [
```
chrisgarrity: I reordered the tiles in based on feedback from @ntlrsk (see https://github.com/LLK/scratch-www/issues/979#issuecomment-258456575) that put Animate Your Name first instead of Make It Fly. We should check whether this change should carry over to the splash page. If so you can just use the first three ttt tiles in order.

src/views/splash/hoc-banner/hoc-banner.jsx
```diff
@@ -0,0 +1,115 @@
...
+    },
+    hideTTTModal: function () {
+        return this.setState({TTTModalOpen: false});
+    },
+    renderTTTTiles: function () {
+        var formatMessage = this.props.intl.formatMessage;
+        
+        var tileObjects = {
+            flyTile: {
+                title: formatMessage({id: tiles[1].title}),
```
rschamp: Not crazy about this kind of data juggling, but it probably won't matter for the short lifetime for this component.  If this were longer term, I'd suggest `thingstotry/ttt` to export an object with named tile data.  Then both views could render an ordered set of those.  Otherwise you have to coordinate reordering in more than one place.

src/views/splash/hoc-banner/hoc-banner.jsx
mewtaylor: ohh I thought that it was in the center here to highlight it more, thanks for checking. @kaschm @carljbowman what should the order be for these? same as on the top row of ttt?

src/views/splash/hoc-banner/hoc-banner.jsx
mewtaylor: Yeah I thought about refactoring this to do that – though the main issue is that it's needed in sequence in ttt and not here. However, if @chrisgarrity 's comment above is correct, then I can just grab this in sequence like we do in ttt (just for the first 3), which might make this a bit less juggling (i.e. essentially saying "we want the 3 most important ones").

mewtaylor:
@rschamp @chrisgarrity – for Nov 21 hotfix, we either need to not merge any December things into `develop` before Monday, or maybe create a separate hotfix branch (which would need to get PRs for the release branch too). Thoughts?

src/views/splash/hoc-banner/hoc-banner.jsx
chrisgarrity: Yes, I think we ended up with this because the ordering on the ttt page was changed after the design was set. Personally I think they should match - I think it's odd that the first three change order when you go from splash to ttt.

rschamp:
@mewtaylor I think a hotfix branch would make the most sense. Can you explain what you mean about getting PRs for the release branch?

mewtaylor:
@rschamp I mean that if there's a PR for a bugfix found during QA that needs to go against the release branch, I'll need to also open a PR against the hotfix branch that we create – unless we just don't merge this until after the release.

chrisgarrity:
@mewtaylor @rschamp A hotfix branch makes sense to me, but I might wait a couple of days any any case to confirm the order of the tiles on the HOC banner.

PR #1025, opened 2016-11-16
Fix gh-1023: Localize Explore Title and Dropdown

### LINKED ISSUE #1023: Explore page header is not translated
The Explore page header and the "Trending, Popular, Recent" drop down are not translated.

Should be able to use the general.explore message id for the header

TheGrits:
Should fix #1023 

Test cases:
-

- `h1` element at the title of the page should now be localized
- dropdown on the right side of the page should now be localized


TheGrits:
@thisandagain @mewtaylor @rschamp (and @chrisgarrity because you opened the issue)

PR #1027, opened 2016-11-16
Fix gh-1026: Trust and Safety Manager URL update

### LINKED ISSUE #1026: Link for Trust & Safety Manager on Jobs page
As discussed with @mewtaylor, on the jobs page on Staging, we updated the jobs available in https://github.com/LLK/scratch-www/pull/989 (which is in the **November 14** milestone) but it looks like the link for the trust & safety manager  position will still not work.

TheGrits:
Should fix #1026 

I have a feeling Travis won't like how long the URL is though...

### DIFF
--- a/src/views/jobs/jobs.jsx
+++ b/src/views/jobs/jobs.jsx
@@ -31,7 +31,7 @@ var Jobs = React.createClass({
                         <h3><FormattedMessage id='jobs.openings' /></h3>
                         <ul>
                             <li>
-                                <a href="http://media.mit.edu/about/opportunities/trust-and-safety-manager-scratch-lifelong-kindergarten">
+                                <a href="http://bit.ly/2f5hABE">
                                     Trust &amp; Safety Manager
                                 </a>
                                 <span>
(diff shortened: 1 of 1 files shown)


jwzimmer:
@mewtaylor if the URL is rejected for making the line too long, could @TheGrits use bitly to shorten the link? The old link for the Community Manager position was a bitly link...

jwzimmer:
@TheGrits I think you can go ahead and try with this shortened URL: http://bit.ly/2f5hABE so it will pass the check, and we'll see if @mewtaylor knows of any reason that isn't a good idea when the PR is reviewed. : )

TheGrits:
@jwzimmer Thanks, I pushed the change!

jwzimmer:
Thanks @TheGrits!

mewtaylor:
bitly is fine – @rschamp did we decide at some point to exempt urls from linting?

rschamp:
I believe they will be when we switch over to eslint-config-scratch.

mewtaylor:
@jwzimmer @thisandagain @rschamp I'd like to cherry pick this into our release tomorrow, since currently people cannot apply for the position through Scratch – thoughts?

thisandagain:
Agreed. Thanks @mewtaylor

PR #1048, opened 2016-11-19
Add Belarusian (Byelorussian) to custom locale

EdwardGeneralov:
Belarusian

### DIFF
--- a/custom-locales.json
+++ b/custom-locales.json
@@ -7,6 +7,10 @@
         "locale": "an",
         "parentLocale": "ca"
     },
+    "be": {
+        "locale": "be",
+        "parentLocale": "ru"
+    },
     "cat": {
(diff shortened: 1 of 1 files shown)


chrisgarrity:
Custom-locale is only needed for languages not defined by [intl-messageformat](https://github.com/yahoo/intl-messageformat/tree/master/dist/locale-data). be is defined, so this change is unnecessary.

EdwardGeneralov:
Invalid

mewtaylor:
@EdwardGeneralov could you explain what you mean by that? Just to reiterate, what @chrisgarrity is saying is that the locale definition for our repo for Belarusian is already covered by a tool we use, `intl-messageformat`, so we don't need to define it again here it appears.

Issue #906, opened 2016-09-01
Redirect /hoc to Activity Page


kaschm:
Before CSEdWeek begins, we should redirect the /hoc page to the new Activities Page https://github.com/LLK/scratch-www/issues/904

ntlrsk:
Will this HOC page still show up anywhere? If so, is it possible to add links to the two educator guides: **Make It Fly** https://scratch.mit.edu/fly/guide and **Make Music** https://scratch.mit.edu/music/guide ?

kaschm:
@ntlrsk we had planned to redirect this page entirely; it won't be available after the redirect. Does this sound ok?

ntlrsk:
@kaschm Yes that is OK with me. Hopefully translators will work on translation for the TTT page - thanks to @chrisgarrity for emailing them.

mewtaylor:
@ntlrsk @kaschm can we get rid of the current hoc page? Or is the plan to try to archive it?

kaschm:
@mewtaylor might be nice to have it in case we want to reuse it for something in the future (for example, Lemann uses a translated version on their site).

mewtaylor:
@kaschm deleting it would mean we’d still have the code for it in github, so if the thought was to keep it to reuse it, we could still find it again even after removing it. Is that okay?

kaschm:
yup, that's fine! thanks. @mewtaylor

ntlrsk:
OK with me too. Thanks for checking.

PR #1050, opened 2016-11-20
Implement GH-906: Redirect hoc page to `/go`

mewtaylor:
Implements #906. For now, it just removes the old HOC page, rather than archiving it. I can change that if needed, but this is nice because it shortens our Fastly url rule by moving something to redirect.

/cc @kaschm


rschamp:
I'd rather it be deleted than archived — archiving means maintaining it.

PR #1057, opened 2016-11-21
[Master] Release HOC #1

mewtaylor:


rschamp:
This looks like the right set of PRs to me

chrisgarrity:
No isiZulu in the menus, but I assume that's because we're including it in the next deploy. Everything else looks like the expected set of changes.

mewtaylor:
yeah sorry @chrisgarrity I miscommunicated on that

PR #1077, opened 2016-11-28
Automated integration test framework for CI

jzoldak:
Addresses #925

@jwzimmer I've been thinking about how we best set up the automated testing for the integration with the backend Scratch APIs, and coded up this approach as a proposal. Would love your feedback.

Important points:
* Dockerized the development environment. This should also make onboarding of new devs easier.
* Using pre-built, supported containers for Selenium standalone and firefox. We could easily extend this to test other browsers also if we want.
* Using the [Nightwatch.js](http://nightwatchjs.org/) Selenium framework.
  * This has a built-in Page Object support, which will make the tests more maintainable.
  * It also exposes helper methods for waiting. We can use these to help with synchronization, which IMO is the trickiest part in any browser-based automation.
* Warning: These tests do take some time to set up and run in CI.
  * They currently increase build time on travis by about 8 minutes (from <3 to <11 minutes)
  * Mitigated somewhat by the fact that they will only run if the other tests pass. E.g., if a PR fails the linters on travis, it fails fast.
  * In the future, we should publish the app container for scratch-www on DockerHub so that it can be pulled as a base. If the node requirements haven't changed, then there will be no need to download and install them, which is the bulk of the time spent.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,11 @@
+.DS_Store
+.dockerignore
+.eslintignore
+.eslintrc
+.sass-lint.yml
+.travis.yml
+build/
+intl/
+node_modules/
+Dockerfile
+README.md
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,5 +7,7 @@
 build/
 intl/
 node_modules/
+reports/
+screenshots/
 Dockerfile
 README.md
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 .DS_Store
 
 # NPM
-/node_modules
+node_modules/
 npm-*
 
 # Build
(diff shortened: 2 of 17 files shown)


jwzimmer:
Wow, thank you for working on this @jzoldak. I will bring it up with @mewtaylor & @cwillaim soon - they have also been working on setting up how to run the tests, so it would be good to look at what they've done and what you've done here and see what direction to go in from there. Thank you!

mewtaylor:
@jzoldak thanks so much again for this PR! It is really cool to check out Nightwatch, and your proposal for docker-ization is really helpful to think about as we start to scale up our end-to-end automation testing.

For now, we're going to close this PR, though it does provide us some grounding and some options moving forward. We're going to eventually land @cwillaim's testing suite as a base line, and take these proposals into account from there. Stay tuned for that PR a little later this month!

PR #1095, opened 2016-12-02
Add a microworlds homepage, and point microworlds pages at the new microworlds projects.

morantsur:
We created a Scratch user - "microworlds", that will control all of the microworlds projects from now on. This changes the projects embedded on the "new" microworlds pages to point at these new projects.

@ntlrsk @seotts


morantsur:
Added another commit, which adds a new microworlds homepage.

src/views/microworldshomepage/microworldshomepage.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+var classNames = require('classnames');
+
+require('../thingstotry/thingstotry.scss');
+
+var MicroworldsHomepage = injectIntl(React.createClass({
+    type: 'MicroworldsHomepage',
+    getInitialState: function () {
+        return {
+            currentTile: Tiles[0]
```
rschamp: It doesn't look like this state is used anywhere, is it?

src/routes.json
```diff
@@ -246,5 +246,12 @@
         "routeAlias": "/microworlds",
         "view": "microworld/soccer/soccer",
         "title": "Soccer"
+    },
+    {
+        "name": "microworlds-homepage",
+        "pattern": "^/microworlds/go/?(\\?.*)?$",
+        "routeAlias": "microworlds/go/?\\??",
```
rschamp: This isn't explained anywhere, and we should fix that. But the `routeAlias` is used for shortening our URL regex for Fastly. You should reuse any `routeAlias` that already exists with the same prefix. Here, it should be `"/microworlds"`.

src/views/microworldshomepage/l10n.json
```diff
@@ -0,0 +1,15 @@
+{
+    "ttt.placeholder": "Placeholder text",
```
rschamp: Is this used anywhere? We shouldn't ask the translators to translate temporary text.

src/views/microworldshomepage/l10n.json
```diff
@@ -0,0 +1,15 @@
...
+    "ttt.tryIt": "Try It",
+    "ttt.open": "Open",
+    "ttt.HipHopTitle": "Hip Hop Dance",
+    "ttt.FashionTitle": "Fashion",
+    "ttt.HouseTitle": "Interactive House",
+    "ttt.MakeMusicTitle": "Make Music",
+    "ttt.SoccerTitle": "Soccer",
+    "ttt.JokeTitle": "Make a Joke",
+    "ttt.ArtTitle": "Make Art"
+}
```
rschamp: If these are specific to microworlds, they should be prefixed with that name instead of reusing another component's prefix so there's not a chance they'll collide with each other, e.g. `"microworlds.HipHopTitle": "Hip Hop Dance"`

src/views/microworldshomepage/microworldshomepage.jsx
```diff
@@ -0,0 +1,72 @@
...
+                    <section className="ttt-section">
+                    </section>
+                    <h1 className="title-banner-h1">
+                        <FormattedMessage id="ttt.title" />
+                    </h1>
+                    <p className="intro title-banner-p">
+                        <FormattedHTMLMessage id="ttt.subTitle" />
+                    </p>
+                </TitleBanner>
+
```
rschamp: Nit: extra line break

src/views/microworldshomepage/microworldshomepage.jsx
```diff
@@ -0,0 +1,72 @@
...
+                <div className="inner">
+                    <section className="ttt-section">
+                        <MasonryGrid >
+                            {this.renderTTTTiles()}
+                        </MasonryGrid>
+                    </section>
+
```
rschamp: Nit: extra line break

src/views/microworldshomepage/microworldshomepage.jsx
```diff
@@ -0,0 +1,72 @@
...
+var MasonryGrid = require('../../components/masonrygrid/masonrygrid.jsx');
+var Page = require('../../components/page/www/page.jsx');
+var TitleBanner = require('../../components/title-banner/title-banner.jsx');
+var TTTTile = require('../../components/ttt-tile/ttt-tile.jsx');
+var Tiles = require('./ttt.json');
+
+var classNames = require('classnames');
+
+require('../thingstotry/thingstotry.scss');
```
rschamp: If you would like to re-use another view's CSS, then that leads me to think we need another type of component to represent those styles, which we would then use in both views.  Ideally each view only will contain the CSS specific to it, and other common CSS should be contained in reusable components.

morantsur:
@rschamp Thank you for your comments!
@mewtaylor helped me address some of these, and I think it makes more sense now. 

I have one more issue - because my tiles don't have a description or modal, I get these warnings in the console:
warning.js:44 Warning: Failed propType: Required prop `description` was not specified in `TTTTile`. Check the render method of `MicroworldsHomepage`
Warning: Failed propType: Required prop `onGuideClick` was not specified in `TTTTile`. Check the render method of `MicroworldsHomepage`.

Any ideas?

src/components/ttt-tile/ttt-tile.jsx
```diff
@@ -44,10 +54,7 @@ var TTTTile = React.createClass({
                     </div>
 
                 </a>
-                <div className="ttt-tile-guides" onClick={this.props.onGuideClick}>
-                    <FormattedMessage id='tile.guides' defaultMessage='See Cards and Guides'/>
-                    <img className="ttt-tile-see-more" src="/svgs/ttt/see-more.svg" />
-                </div>
+                {this.tileGuidesIfExists()}
```
rschamp: I personally find inline rendering easier to understand since all the logic is in one place. Here I would either use the pattern
```jsx
{this.props.onGuideClick && (
    <div ...
)}
```
or add a component to represent the tile guide section and have it only return a component if it receives the `onClick` prop.

If this is based on a pattern that already exists on www then you can ignore that comment :)

src/views/microworldshomepage/microworldshomepage.jsx
```diff
@@ -0,0 +1,59 @@
...
+                    <h1 className="title-banner-h1">
+                        <FormattedMessage id="microworlds.title" />
+                    </h1>
+                    <p className="intro title-banner-p">
+                        <FormattedHTMLMessage id="microworlds.subTitle" />
+                    </p>
+                </TitleBanner>
+                <div className="inner">
+                    <MasonryGrid >
+                        {this.renderTTTTiles()}
```
rschamp: Again, please ignore the following if this is based on a pre-existing pattern in the codebase.

Here I would also expect inline rendering since `renderTTTTiles` receives no arguments so only serves as a level of indirection for a map.

src/views/microworldshomepage/microworldshomepage.jsx
```diff
@@ -0,0 +1,59 @@
...
+                bannerUrl: tile.bannerUrl
+            };
+            translatedTiles.push(
+                <TTTTile
+                    key={key}
+                    {...translatedTile}
+                />
+            );
+        }, this); // don't forget to pass 'this' into map function
+        return translatedTiles;
```
rschamp: If you do merge this into `render`, I would note that the `translatedTiles` and `translatedTile` variables are unnecessary, this could be reduced into a `map` of `Tiles` to `TTTTile` components.

chrisgarrity: @rschamp this is based on what is in the ttt view. However, I realized that the render was originally created to do the reordering of tiles based on the number of columns, then that got moved to the masonry grid component. So this probably should be merged into `render` as should the corresponding code in ttt.

src/views/microworldshomepage/l10n.json
```diff
@@ -0,0 +1,14 @@
+{
+    "tile.tutorial": "Microworld",
+    "ttt.tryIt": "Try It",
+    "ttt.open": "Open",
```
chrisgarrity: These strings are used in the tile (tutorial & tryIt) and modal components(open), but we don't define l10n strings on components so they're in the view. I think ttt.tryIt was supposed to get renamed as tile.tryIt to be consistent with the other strings in the t3tile.

morantsur:
Fixed all comments, after talking to @chrisgarrity.

Thanks!

src/views/microworldshomepage/microworldshomepage.jsx
```diff
@@ -0,0 +1,48 @@
...
+                <div className="inner">
+                    <MasonryGrid >
+                        {Tiles.map(
+                            function (tile, key) {
+                                return <TTTTile
+                                    key={key}
+                                    title={this.props.intl.formatMessage({id: tile.title})}
+                                    tutorialLoc={tile.tutorialLoc}
+                                    thumbUrl={tile.thumbUrl}
+                                    bannerUrl={tile.bannerUrl}
```
chrisgarrity: I just realized that bannerUrl is only used for the modal, so it's another thing that microworlds probably don't need. If you take it out, you'll also need to take the 'required' off the tile prop.

morantsur: You're right - the bannerUrl is not used here. 
I removed it from the microworlds page, but there's no need to remove the isRequired, because it's in the ttt-modal, which I'm not building at all - so it doesn't raise any warnings.

PR #1096, opened 2016-12-05
Fix GH-962: use width for radio buttons, not margin

### LINKED ISSUE #962: Student Registration - No "Male" option, cannot type in "Other" field, can proceed with blank "Other" field in several languages
I noticed this while trying to repro freshdesk no. 58511. If you set the language to Portuguese or Brazilian Portuguese, and follow a student sign-up link, on the demographics step, there is no "male" gender option, and if you choose "other", you cannot type in it, and you can proceed to the next step while it is blank.

I used the link http://scratch.mit.edu/classes/295/register/9490a7eb39b24c99ba6a27d60d5e6e6c.

I registered portu_test_2 (Production, Chrome 53, Mac OS X El Capitan) with no gender (blank "other" option). I wonder if this shows up as 0 in the database @rschamp (like the countries)?

screen shot 2016-10-11 at 3 31 53 pm
screen shot 2016-10-11 at 3 31 17 pm

mewtaylor:
This fixes #962 – the margin was causing flex to overflow the radio buttons when the text was long, causing the male and other options to render on top of each other. This fixes that by giving the radio buttons a fixed with relative to the parent container, and by ensuring that languages with long localizations will still be visible in overflow text.

### Test Cases ###
* In any language (specifically pt_BR, en, and es to test), all three options should show up for this question
* If the input field is selected, it should require text be added in the input before the selection is valid
* If the input field is given a gender, that should be the gender recorded for the user in the db

### DIFF
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -360,24 +360,24 @@ module.exports = {
                                         options={[
                                             {value: 'female', label: formatMessage({id: 'general.female'})},
                                             {value: 'male', label: formatMessage({id: 'general.male'})},
-                                            {value: 'other', label: ''}
+                                            {value: 'other', label: <Input
+                                                className="demographics-step-input-other"
+                                                name="user.genderOther"
+                                                type="text"
+                                                validations={{
+                                                    maxLength: 25
+                                                }}
+                                                validationErrors={{
+                                                    maxLength: formatMessage({
+                                                        id: 'registration.validationMaxLength'
+                                                    })
+                                                }}
+                                                disabled={this.state.otherDisabled}
+                                                required={!this.state.otherDisabled}
+                                                help={null}
+                                            />}
                                         ]}
                                         required />
-                            <div className="gender-input">
-                                <Input name="user.genderOther"
-                                       type="text"
-                                       validations={{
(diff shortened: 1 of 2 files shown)


rschamp:
Not sure how I feel about this solution, although it's a design issue so I would defer to @carljbowman if you already talked to him about it.

The root problem seems to be the rule for the "Other" input box has a negative margin which covers up anything that overflows to a new line.  If we limit Male/Female to 50% of the line, and use `nowrap`, then won't that actually cause an overlap problem, since e.g., if the `Female` option is a long word, then it will overlap the text for `Male`?
image

I think a better way to do it would be to allow the option text to re-flow, but fix the weird negative margin rule for the "Other" input box so that it doesn't depend on a fixed height and cover up the options if they're on more than one line.

mewtaylor:
@rschamp good point – per offline discussion, it now puts the `Input` as a label value, which still maintains the `formData` structure we had before

jwzimmer:
### Testing - in progress

* [ ] In any language (specifically pt_BR, en, and es to test), all three options should show up for this question
* [ ] If the input field is selected, it should require text be added in the input before the selection is valid
* [ ] If the input field is given a gender, that should be the gender recorded for the user in the db

PR #1099, opened 2016-12-06
Fix GH-966: Localize default “required” message

### LINKED ISSUE #966: "This field is not required" translated sometimes but not always - Portuguese - Teacher Registration
I'm not sure if this could be fixed as part of https://github.com/LLK/scratch-www/pull/960, but I don't think it should block that case, since @ntlrsk needs that fix by Friday. 

The string "This field is not required" which pops up for required fields during Teacher Registration has been translated in Portuguese, but the translation doesn't get used in most of the validations. (To repro, go through the steps in Teacher Registration with language set to Portuguese, and trigger the validations by clicking Next without filling in the fields.)

Staging, Firefox 49.0.1, Mac OS X El Capitan, but I also saw it on Production (https://github.com/LLK/scratch-www/issues/940#issuecomment-253255882).

screen shot 2016-10-13 at 3 41 33 pm
screen shot 2016-10-13 at 3 41 03 pm
screen shot 2016-10-13 at 3 40 54 pm

mewtaylor:
Fixes #966. I was trying, if possible, to not introduce localization at the component level in order to keep it in the views – but that seemed a bit too involved given that the fix in this way is only a few lines long.

### DIFF
--- a/src/components/forms/validations.jsx
+++ b/src/components/forms/validations.jsx
@@ -1,4 +1,5 @@
 var defaults = require('lodash.defaultsdeep');
+var intl = require('../../lib/intl.jsx');
 var libphonenumber = require('google-libphonenumber');
 var phoneNumberUtil = libphonenumber.PhoneNumberUtil.getInstance();
 var React = require('react');
@@ -44,5 +45,5 @@ module.exports.validationHOCFactory = function (defaultValidationErrors) {
 };
 
 module.exports.defaultValidationHOC = module.exports.validationHOCFactory({
-    isDefaultRequiredValue: 'This field is required'
+    isDefaultRequiredValue: <intl.FormattedMessage id="teacherRegistration.validationRequired" />
 });
--- a/src/components/forms/validations.jsx
+++ b/src/components/forms/validations.jsx
@@ -45,5 +45,5 @@ module.exports.validationHOCFactory = function (defaultValidationErrors) {
 };
 
 module.exports.defaultValidationHOC = module.exports.validationHOCFactory({
-    isDefaultRequiredValue: <intl.FormattedMessage id="teacherRegistration.validationRequired" />
+    isDefaultRequiredValue: <intl.FormattedMessage id="form.validationRequired" />
 });
--- a/src/l10n.json
+++ b/src/l10n.json
@@ -93,6 +93,8 @@
     "footer.help": "Help Page",
     "footer.scratchFamily": "Scratch Family",
 
+    "form.validationRequired": "This field is required",
+
     "login.forgotPassword": "Forgot Password?",
 
     "navigation.signOut": "Sign out",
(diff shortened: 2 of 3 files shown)


src/components/forms/validations.jsx
```diff
@@ -44,5 +45,5 @@ module.exports.validationHOCFactory = function (defaultValidationErrors) {
 };
 
 module.exports.defaultValidationHOC = module.exports.validationHOCFactory({
-    isDefaultRequiredValue: 'This field is required'
+    isDefaultRequiredValue: <intl.FormattedMessage id="teacherRegistration.validationRequired" />
```
rschamp: I think we should move this string it to general if it's at the component level.

mewtaylor: I'm fine with that – I was thinking we would just do that once it was actually used outside of `teacherregistration`, since it's not used anywhere else yet, but don't feel bound to that.

rschamp: I would say having it in `forms` constitutes it being outside of `teacherregistration`.

mewtaylor:
hmm – git says "everything is up to date", but github is not seeing my recent commit...

src/components/forms/validations.jsx
rschamp: Actually looking again at where this is, it's already used in any component that has a required field, like the login form.
image

jwzimmer:
### Testing

**Chrome 55, Mac OS El Capitan, Staging**
* [ ] The validations should look as expected during student and teacher registration in English
  * Noticed https://github.com/LLK/scratch-www/issues/1123, otherwise they look normal
  * I triggered every "This field is required" validation during teacher registration
* [x] The "This field is required" validations should be consistently translated in Portuguese
  * Noticed https://github.com/LLK/scratch-www/issues/1125
  * However, all the validations are translated in Portuguese now!
* [ ] The "This field is required" validations should be translatable in other languages (meaning, you should not see something like `educatorRegistration.fieldRequired` in any of the languages; it should either be in the language you chose, or, if no translation is available, it should be in English)
  * [ ] Try an RTL language e.g. Hebrew
  * [ ] Try a language with very different characters from English, e.g. Chinese
  * [ ] Try another language, e.g. Brazilian Portuguese

jwzimmer:
@mewtaylor This passed all the test cases, but I'm not sure if #1123 or #1125 were caused by this PR/ should be blockers. Let's discuss that before it's QA-OK'd. Thanks.

jwzimmer:
I'm no longer seeing https://github.com/LLK/scratch-www/issues/1125 post @mewtaylor's fix. I checked with Brazilian Portuguese, Chrome 55, Mac OS El Capitan. 👍

jwzimmer:
I think the other issues are probably not blockers.

PR #1100, opened 2016-12-08
Fix r2-4195: Handle double slash in splash pattern

mewtaylor:
This fixes https://github.com/LLK/scratchr2/issues/4195 by updating the regex to catch two slashes. I decided to intentionally not catch more than 2 – it’ll 404 on 3 or more, which seems appropriate.

We could also redirect on 2 slashes to the homepage, but this seemed just as well.

### DIFF
--- a/src/routes.json
+++ b/src/routes.json
@@ -1,10 +1,16 @@
 [
     {
         "name": "splash",
-        "pattern": "^/{1,2}?$",
+        "pattern": "^/?$",
         "routeAlias": "/?$",
         "view": "splash/splash",
         "title": "Imagine, Program, Share"
+    },
+    {
+        "name": "splash-redirect",
+        "pattern": "^//?$",
+        "routeAlias": "/?$",
+        "redirect": "/"
     },
     {
         "name": "about",
--- a/src/routes.json
+++ b/src/routes.json
@@ -1,7 +1,7 @@
 [
     {
         "name": "splash",
-        "pattern": "^/?$",
+        "pattern": "^/{1,2}?$",
         "routeAlias": "/?$",
         "view": "splash/splash",
         "title": "Imagine, Program, Share"
--- a/src/routes.json
+++ b/src/routes.json
@@ -8,7 +8,7 @@
     },
     {
         "name": "splash-redirect",
-        "pattern": "^//?$",
+        "pattern": "^///?$",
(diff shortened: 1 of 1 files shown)


rschamp:
What do you think about leaving this simple and just removing the endpoint so it 404s instead?

mewtaylor:
I personally like the idea of handling just the double slash – it's a mistake that happens, and that other sites handle for you that i've seen.

rschamp:
If we're going to support it, I think I would rather have `^//+` redirect to `/` than resolving to the homepage, since ideally you only have one URL per page. (At one point Google would dock you for multiple URLs for the same content, but I'm not sure if that is still true.)  Eventually maybe we should add a couple of generic redirect rules
* `^//+(.+)$` to `/\1`
* `^(/.+)//+$` to `\1/`

mewtaylor:
I had it as a redirect before, but was trying to keep the rules simplified – I have no strong preference for keeping it as is if you'd rather see it as its own redirect rule.

rschamp:
In addition to the functional difference, I think I would actually prefer the redirect for simplicity's sake, since then the route pattern is easier to read.

mewtaylor:
Same as in the PR details – I kind of felt that anything more than two slashes should 404.

src/routes.json
```diff
@@ -7,6 +7,12 @@
         "title": "Imagine, Program, Share"
     },
     {
+        "name": "splash-redirect",
+        "pattern": "^//?$",
```
rschamp: Actually this is gonna cause an infinite redirect. It should be `^///?$` in that case.

chrisgarrity:
Testing on El Capitan, Chrome 54, Safari 10.0.1, Firefox 45.5:
* https://scratch.ly// - gets 404, not home page
* https://scratch.ly/// - also gets 404 - as expected

jwzimmer:
@mewtaylor has fixed this, I just checked it & discussed it with him, 3 slashes no longer 404s but anything more than 3 will.

Chrome 55, Mac OS El Capitan, Staging & IE 11 on Win 7 VM, Staging
- 3 slashes redirects to homepage
- 2 slashes redirects to homepage
- 1 slashes redirects to homepage
- 0 slashes redirects to homepage
- 4 slashes 404s
- Random numbers higher than 4, 404

PR #1110, opened 2016-12-13
Provide fallback error messages

### LINKED ISSUE #1107: No error page showing up if no error message is supplied
**TL;DR** Currently, hitting student account creation rate limit - and any other errors in student/ teacher registration where we're not supplying an error message - is silent.

We knew we needed a clearer message for when teachers hit the account creation rate limit (see https://github.com/LLK/scratchr2/issues/4174), but we expected them to run into a page like "Something went wrong" as the heading, and then in bold "There was an error while processing your registration" in the body but no details at the end of registration. However, they actually aren't seeing that page at all - they are just getting "stuck" on the last step. 
@rschamp said:
> we give an "err" but no "msg".

@rschamp & I were discussing the difficulty in distinguishing between the error states users are reporting with teacher/ student accounts. To address that, @rschamp is planning to include the error code in the fallback error message.

rschamp:
Fixes #1107.  Right now I'm adding this into the frozen branch since it could be hotfix-worthy, but if @jwzimmer would like this to respect the freeze I'll update the base accordingly.

In case the response does not supply `msg` or `errors`, provide a default. If `registrationError(s)` is empty, we do not show the error card, which causes "silent" failures.

## Test cases
* The following should still work as expected:
  * Registering as a new teacher
  * Registering as a new student through a student sign-up link
  * Completing registration as a student when logging in for the first time after a teacher creates the student account
* You should see an error page rather than getting stuck on the second step after exceeding the registration rate limit (we can artificially limit this for testing) in the following cases:
  * Registering as a new teacher
  * Registering as a new student through a student sign-up link
* I'm not aware of which type of error could cause the user to become "stuck" during registration completion, but in any case it should not be possible anymore.

### DIFF
--- a/src/views/studentcompleteregistration/studentcompleteregistration.jsx
+++ b/src/views/studentcompleteregistration/studentcompleteregistration.jsx
@@ -86,11 +86,18 @@ var StudentCompleteRegistration = intl.injectIntl(React.createClass({
             method: 'post',
             useCsrf: true,
             formData: submittedData
-        }, function (err, body) {
+        }, function (err, body, res) {
             this.setState({waiting: false});
             if (err) return this.setState({registrationError: err});
             if (body.success) return this.advanceStep(formData);
-            this.setState({registrationErrors: body.errors});
+            this.setState({
+                registrationErrors:
+                    body.errors || {
+                        __all__:
+                            this.props.intl.formatMessage({id: 'registration.generalError'}) +
+                            ' (' + res.statusCode + ')'
+                    }
+            });
         }.bind(this));
     },
     goToClass: function () {
--- a/src/l10n.json
+++ b/src/l10n.json
@@ -110,6 +110,7 @@
     "registration.choosePasswordStepTitle": "Create a password",
     "registration.choosePasswordStepTooltip": "Don't use your name or anything that's easy for someone else to guess.",
     "registration.classroomApiGeneralError": "Sorry, we could not find the registration information for this class",
+    "registration.generalError": "Sorry, an unexpected error occurred.",
     "registration.classroomInviteExistingStudentStepDescription": "you have been invited to join the class:",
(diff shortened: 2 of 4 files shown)


src/views/studentcompleteregistration/studentcompleteregistration.jsx
```diff
@@ -86,11 +86,18 @@ var StudentCompleteRegistration = intl.injectIntl(React.createClass({
...
             useCsrf: true,
             formData: submittedData
-        }, function (err, body) {
+        }, function (err, body, res) {
             this.setState({waiting: false});
             if (err) return this.setState({registrationError: err});
             if (body.success) return this.advanceStep(formData);
-            this.setState({registrationErrors: body.errors});
+            this.setState({
+                registrationErrors:
```
mewtaylor: What's the reason for this one to be `Errors` rather than `Error`?

rschamp: That response returns an object from the Django form errors, so there could be multiple errors.  This view is set up to output the form error object rather than one thing.

jwzimmer:
## Testing - in progress
* The following should still work as expected:
  * [ ] Registering as a new teacher
    * [ ] You should not get "stuck" on any steps during registration
  * [ ] Registering as a new student through a student sign-up link
    * [ ] You should not get "stuck" on any steps during registration
  * [ ] Completing registration as a student when logging in for the first time after a teacher creates the student account
* [ ] You should see an error page rather than getting stuck on the second step after exceeding the registration rate limit (we can artificially limit this for testing) in the following cases:
  * [ ] Registering as a new teacher
  * [ ] Registering as a new student through a student sign-up link

jwzimmer:
#1124 is now fixed, this is QA-OK.

PR #1119, opened 2016-12-16
Remove `routeAlias` from splash redirect

mewtaylor:
It’s not necessary, and I believe is causing confusion for the faslty config.


rschamp:
That makes sense. In fact we might want to remove the alias from all the redirects if they have them to reduce confusion. Since these are only for the "Pass"/"Not Pass" conditions, they should be irrelevant to redirects which don't make it to passing/not passing.

Issue #1130, opened 2016-12-19
Update "Download Flash" link on FAQ 


kaschm:
This is happening on production. 

On the General FAQ (scratch.mit.edu/info/faq), we don't point users to the correct link to download Adobe Flash:

**What are the system requirements for Scratch?**
To run Scratch 2, you need to be using (1) a Mac, Linux, or Windows computer; (2) a version of **Adobe Flash Player** released on or after June 15, 2016; 

The link should go here: https://get.adobe.com/flashplayer/

thisandagain:
Assigning to @mewtaylor for PR review from contributor. :-)

PR #1132, opened 2016-12-20
Fix gh-1130: Update Flash Link

TheGrits:
Updates the URL of the link to download the flash player as per #1130.

Test case:
-
Clicking on the link for the 'Adobe Flash Player' should now lead to `https://get.adobe.com/flashplayer/`


TheGrits:
@mewtaylor @thisandagain @rschamp (and @kaschm as issue author)

src/views/faq/l10n.json
```diff
@@ -13,7 +13,7 @@
...
+    "faq.requirementsBody":"To run Scratch 2, you need to be using (1) a Mac, Linux, or Windows computer; (2) a version of <a href= \"https://get.adobe.com/flashplayer/ \">Adobe Flash Player</a> released on or after June 15, 2016; (3) a relatively recent web browser: one of the latest two versions of <a href=\"http://google.com/chrome/\">Chrome</a> (Mac, Windows, or Linux), <a href=\"http://www.mozilla.org/en-US/firefox/new/\">Firefox</a> (Mac or Windows only), <a href=\"https://support.apple.com/downloads/safari\">Safari</a> (Mac or Windows only), <a href=\"https://www.microsoft.com/EN-US/windows/microsoft-edge\">Edge</a> (Windows only), or <a href=\"https://www.microsoft.com/en-us/download/internet-explorer.aspx\">Internet Explorer 10+</a> (Windows only). If your computer doesn’t meet these requirements, you can try downloading and installing <a href=\"/scratch_1.4\">Scratch 1.4</a>, which you can still use to share projects to the Scratch 2 website. We do not support Chromium.",
```
mewtaylor: @TheGrits it looks like there's some extra spacing in the `href` that we don't want. It appears to be left over from the original, but would you mind removing it?

TheGrits: @mewtaylor I'm noticing that a couple other hrefs in the file have the same issue. Would you like me to remove those spaces too?

mewtaylor: @TheGrits ooh thanks for mentioning that – if you have the time to, adding those in here would be really helpful!

TheGrits:
@mewtaylor I think I got all of them, if you see any more just let me know and I'd be happy to remove them too.

TheGrits:
@mewtaylor Just wondering, does this one still require changes or not?

mewtaylor:
@TheGrits nope, this looks good! my apologies for not getting back to you sooner

TheGrits:
@mewtaylor No worries!

jwzimmer:
### Testing

Chrome 55.0.2883.95, Mac OS El Capitan 10.11.6
* [x] On Staging, the link in the FAQ should redirect to https://get.adobe.com/flashplayer/ instead of redirecting to the FAQ page (which is what happens on Production)
* [x] Other links on the FAQ page should still work as expected
  * I checked random links from the FAQ page; they redirected to reasonable places

PR #1138, opened 2016-12-28
Fix GH-1097: Don’t validate username server side unless it passes client-side checks

### LINKED ISSUE #1097: "Oops something went wrong" with invalid username - Teacher Registration
Production, Chrome 54, Mac OS El Capitan

**TL;DR**
- Error visible to user on the username/ password step of registration
- Error in console
- No known impact on functionality; I completed registration after triggering this error

I'm seeing "Oops something went wrong" & 
"https://scratch.mit.edu/accounts/check_username/[username]/ Failed to load resource: the server responded with a status of 404 (NOT FOUND)" when clicking outside the username field when it has a current error (for invalid chars included in username).

To repro, enter a username like "mr.jwzimmer1990", then click outside the fields. You should see "Oops something went wrong" appear in the UI and a 404 in the console on the username check. 

screen shot 2016-12-05 at 5 25 09 pm

I'm concerned that in some cases this issue could prevent teachers from being able to sign up, if for some reason the error early in the process causes the whole process to fail (I'm trying to repro that).

mewtaylor:
Fixes #1097 by not making the server-side username validation call if the client side validations have failed. It runs the [isValidValue()](https://github.com/christianalfoni/formsy-react/blob/master/API.md#isvalidvalue) check on the input `onBlur` to make sure it is a valid value before trying, preventing a possible 404 from a username with disallowed characters.

This is not the ideal way to do this, since it’s getting the input from an array, and an array buried multiple levels of `refs` deep. However, to make this different would require significant refactor, given our current use of `formsy-react-components`. I'd like to not do that refactor unless we also have a conversation about our registration UX regarding when we want validation to occur generally – i.e., do we want to move to having it occur more `onBlur`, keep as is, only validate on submit, some combination, etc?

### DIFF
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -185,7 +185,6 @@ module.exports = {
                                 <Input className={this.state.validUsername}
                                        type="text"
                                        name="user.username"
-                                       ref="thing"
                                        onBlur={this.onUsernameBlur}
                                        validations={{
                                            matchRegexp: /^[\w-]*$/,
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -136,7 +136,9 @@ module.exports = {
             }.bind(this));
         },
         onUsernameBlur: function (event) {
-            this.validateUsername(event.currentTarget.value);
+            if (this.refs.form.refs.formsy.inputs[0].isValidValue(event.currentTarget.value)) {
+                this.validateUsername(event.currentTarget.value);
+            }
         },
         onValidSubmit: function (formData) {
             this.setState({waiting: true});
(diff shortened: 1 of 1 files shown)


rschamp:
Another way to do this without the deeply nested refs or a significant refactor would be to cache the value of the username field `onChange`, and then check that value `onBlur`.

mewtaylor:
@rschamp actually the deep nesting isn't to get to the value – that's available in the `currentTarget` pretty easily. The nesting is to get access to the formsy method that checks if something is valid, so I don't believe caching in that way would help. We'd need to find a way to make the formsy component referable at the top level, which I think is something that would be easier if we move away from `formsy-react-components`

mewtaylor:
On Chrome 55, OS X 10.12

* I entered an existing username, and saw a request to `check_username` in the network tab. The UI showed an "already exists" error
* I entered a username with a `.` in it, and the UI showed an "invalid character" error. When I clicked outside of the username field, no request was sent to `check_username`, and no "something went wrong" error appeared
* I entered a new username, and clicked outside the username box, and received no error. I was then able to move to the next stage of the registration process with a proper password

PR #1143, opened 2017-01-05
Tx import

chrisgarrity:
Scripts to support transition to Transifex. 

translations pulled from transifex will be put in localizations (.gitignore)
.tx/config: transifex configuration (maintainer account on transifex needed to push/pull), updated by bin/init-l10n-src
bin/import-pootle: convert po data to json files and push to transifex. This is just for the transition. Once we're on transifex this script should be removed.

Commits should probably be squashed on merge.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@ npm-*
 # Locales
 /locales
 /intl
+/translations
+
 
 # Elastic Beanstalk Files
 .elasticbeanstalk/*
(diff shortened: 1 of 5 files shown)


mewtaylor:
@chrisgarrity this is PR one of two, correct? As in, there will be another PR after this one that will remove the initial import script, and then also pair down the `locale-compare` library to only include what's necessary with transifex?

chrisgarrity:
@mewtaylor Correct, this is a transitional PR. We need to keep all the pootle scripts around to do the transfer.

PR #1160, opened 2017-01-31
Fix gh-1159: Add 2017 to Teacher Registration

### LINKED ISSUE #1159: There's no 2017 in the teacher registration dropdown menu for birth year
**TL;DR:** We have 2017 as an option for scratchers signing up from the homepage, but not for teachers. We should be consistent.

If you look at the dropdown from the homepage versus from `/educators/register`, the year 2017 is not included in the latter:
screen shot 2017-01-30 at 4 52 22 pm

Found while testing https://github.com/LLK/scratchr2/pull/4448, but definitely not a blocker.

TheGrits:
Should fix #1159 

# Test cases:
- 2017 should now be an option
- 1916 should no longer be an option

### DIFF
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -324,7 +324,7 @@ module.exports = {
         },
         getYearOptions: function () {
             return Array.apply(null, Array(100)).map(function (v, id) {
-                var year = 2017 - id;
+                var year = new Date().getFullYear() - id;
                 return {value: year, label: year};
             });
         },
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -324,7 +324,7 @@ module.exports = {
         },
         getYearOptions: function () {
             return Array.apply(null, Array(100)).map(function (v, id) {
-                var year = 2016 - id;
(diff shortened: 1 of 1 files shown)


src/components/registration/steps.jsx
```diff
@@ -324,7 +324,7 @@ module.exports = {
         },
         getYearOptions: function () {
             return Array.apply(null, Array(100)).map(function (v, id) {
-                var year = 2016 - id;
+                var year = 2017 - id;
```
rschamp: We shouldn't have hardcoded the year like this — would you please update this to get the current year?

TheGrits: @rschamp 👍 Good thinking! I'm assuming we'd like to be using ReactJS to get the year, or does this typical JS solution work? http://stackoverflow.com/a/6002276/

rschamp: Yep that JS solution works!  React tries not to get involved with "business logic" like this, and  only provides tools for building the UI.

rschamp:
Thanks @TheGrits!

TheGrits:
@rschamp Anytime!

PR #1166, opened 2017-02-01
Fix gh-1157: Search Redirect

### LINKED ISSUE #1157: Redirect `/search/` to `/search/projects/`
Currently, trying to go to `/search/` gives a 403 response, but we should redirect to the project search, similar to what we do for the `/explore` page

TheGrits:
Should fix #1157 

Test cases:
- `/search/` should now redirect to `/search/projects/`

### DIFF
--- a/src/routes.json
+++ b/src/routes.json
@@ -190,7 +190,7 @@
         "name": "search-redirect",
         "pattern": "^/search/?$",
         "routeAlias": "/search(?!/ajax)",
-        "redirect": "/search/projects",
+        "redirect": "/search/projects"
     },
     {
         "name": "terms",
--- a/src/routes.json
+++ b/src/routes.json
@@ -187,6 +187,12 @@
         "title": "Search"
     },
     {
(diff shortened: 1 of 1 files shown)


TheGrits:

src/routes.json
```diff
@@ -187,6 +187,12 @@
         "title": "Search"
     },
     {
+        "name": "search-redirect",
+        "pattern": "^/search/?$",
+        "routeAlias": "/search(?!/ajax)",
```
mewtaylor: we can leave the route alias as `/search` (like in `"search"`) above. The reason that explore is `"/explore(?!/ajax)"` has to do with some other sub-urls that still need to go to scratchr2

TheGrits: @mewtaylor Sounds good, I'll hop to it!

TheGrits:
@mewtaylor Done!

TheGrits:
@mewtaylor lol you ninja-ed me!

PR #1172, opened 2017-02-06
Fix gh-1171: Add Slash to Footer Developers Link

### LINKED ISSUE #1171: Footer > For Developers link should be `/developers/` (currently missing the ultimate slash)
**TL;DR:** In the footer, the general pattern is to end all the links with a slash, but currently the link For Developers is just `/developers`. 

No effect on functionality, but I don't think there's a reason we couldn't be consistent here (and end all the links with `/`).

Found this with a WIP automated test for the footer links on the homepage!

**Repro:**
- Click on the "For Developers" link in the footer of the homepage
- Note the page redirects to `scratch.mit.edu/developers`
- Try other links, e.g. Credits or About Scratch
- Note they redirect to pages with ultimate `/`s

TheGrits:
Should fix #1171 

## Test cases:
- The URL for the 'Developers' link in the footer should now be https://scratch.mit.edu/developers/

### DIFF
--- a/src/components/footer/www/footer.jsx
+++ b/src/components/footer/www/footer.jsx
@@ -76,7 +76,7 @@ var Footer = React.createClass({
                                 </a>
                             </dd>
                             <dd>
-                                <a href="/developers">
+                                <a href="/developers/">
                                     <FormattedMessage id='general.forDevelopers' />
                                 </a>
                             </dd>


mewtaylor:
@thisandagain @rschamp for urls that are on www as opposed to scratchr2, should we stick with no slash? Or do we want to make them all uniform, and therefore all have slashes for now?

thisandagain:
@mewtaylor We should be moving to no slash no I'm fine with the inconsistency between the two codebases as long as it is consistently inconsistent. 😉

mewtaylor:
Ha – well we're not consistent in our inconsistency, so we should remedy that :). However this one is correct, so I think I will close this without merging. Thanks for being so on top of it though, @TheGrits !

jwzimmer:
@mewtaylor I can make a case for making the inconsistency consistent 👍

PR #1181, opened 2017-02-14
Remove Dari (fa_AF)

### LINKED ISSUE #1174: Remove Dari (fa_AF) from the language menu
This may have been a left over from Scratch 1.4. It was never set up in pootle, so there are no active translators. Also, fa_AF is listed as Persian (Afghanistan), not Dari which has its own code prs.

Remove it from the menus. We can add it back if it's requested, and there are people who translate to it.

TheGrits:
Should fix #1174 (if I'm thinking straight)

## Test cases:
Dari (fa_AF) should no longer be a language option in any menus

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -11,7 +11,6 @@
     "cs": "Česky",
     "cy": "Cymraeg",
     "da": "Dansk",
-    "fa-af": "Dari",
     "de": "Deutsch",
     "yum": "Edible Scratch",
     "et": "Eesti",


chrisgarrity:
@TheGrits Thanks for helping. Ordinarily, I'd just go ahead and approve/merge. However, this month we're in the process of switching to a new translation system. There are a few other languages that have some of the same issues as Dari (in the code, but not on the translation server!), and I'm just going to clean them all up as part of the transition to Transifex. So, I'll close this PR to avoid conflicts with the transifex PR that's coming very soon.

TheGrits:
@chrisgarrity No problem, I just saw it in the issues and thought I try to take a swing at it.

Issue #1167, opened 2017-02-02
Automated test for checking the Splash page


jwzimmer:
- Expand https://github.com/LLK/scratch-www/pull/1139 once landed to cover the Splash page
- Details to be added : )

jwzimmer:
**Questions/ to do**

- Running on multiple browsers
- What about varying the size of the browser window? e.g. `browser.manage().window().setSize(1280, 720)`
- Benefit to running on proxy?

PR #1185, opened 2017-02-15
Issue/gh 1167 test splash page

jwzimmer:
Fixes https://github.com/LLK/scratch-www/issues/1167 by adding 3 new files in the test/integration folder for the splash page.

- Checks footer links
- Checks navbar links
- Checks title & first project/ studio in Featured Projects/ Studios homepage rows

Run with `make smoke`:
screen shot 2017-02-15 at 3 22 08 pm


test/integration/test_footer_links.js
```diff
@@ -0,0 +1,334 @@
...
+        //var expectedHref = 'https://www.scratchfoundation.org/';
+        var expectedHref = 'http://www.scratchfoundation.org/';
+        //the href should be at the end of the URL
+        t.equal(url, expectedHref);
+        t.end();
+    });
+});
+
+//quit the instance of the driver
+driver.quit();
```
rschamp: Nit: Files should end with a new line

test/integration/test_navbar_links.js
```diff
@@ -0,0 +1,139 @@
...
+    });
+});
+
+// ==== Search bar ====
+
+tap.test('checkSearchBar', function (t) {
+    var xPathLink = '//input[@id="frc-q-1088"]';
+    // search bar should exist
+    // if it isn't found, it will throw an error, which tap will treat as a failure
+    var checkSearchBar = driver.findElement(seleniumWebdriver.By.xpath(xPathLink));
```
rschamp: If possible, I think it would be better to be explicit about this and assert that there should be one element that matches the path, rather than allowing the raised error to implicitly fail the the test. This way if some other error is causing the failure, you can distinguish that from the element not being on the page.

test/integration/test_project_rows.js
```diff
@@ -0,0 +1,121 @@
...
+//number of tests in the plan
+tap.plan(4);
+
+//load the page with the driver
+driver.get(rootUrl);
+
+//'//div[@id="footer"]/div[@class="inner"]/div[@class="lists"]'
+////li[contains(@class, "link") and contains(@class, "create")]/a
+//not working
+// /div[@class="slick-initialized"]/div[@class="slick-list"]/div[@class="slick-track"]/div[@class="thumbnail"]/a[@class="thumbnail-image"]'
```
rschamp: Nit: remove commented out code (or explain why it's there).

test/integration/test_project_rows.js
```diff
@@ -0,0 +1,121 @@
...
+    var checkScratchDesignStudioRowWhenSignedOut = driver.findElement(seleniumWebdriver.By.xpath(xPathLink));
+    checkScratchDesignStudioRowWhenSignedOut.getText('h4').then( function (text) {
+        //expected value of the title text
+        console.log(text);
+        var expectedTextRegExp = new RegExp('^Scratch Design Studio - ');
+        t.match(text, expectedTextRegExp);
+        t.end();
+    });
+});
+*/
```
rschamp: Nit: remove commented out code or explain why it's there.

test/integration/test_project_rows.js
```diff
@@ -0,0 +1,121 @@
...
+
+//quit the instance of the driver
+driver.quit();
```
rschamp: Nit: newline

package.json
```diff
@@ -25,6 +26,7 @@
     "compression": "1.6.1",
     "express": "4.13.4",
     "express-http-proxy": "0.6.0",
+    "lodash": "4.17.4",
```
rschamp: I think this was added by accident?

test/integration/test_footer_links.js
```diff
@@ -0,0 +1,334 @@
+/*
+ * Checks that the links in the footer on the homepage have the right URLs to redirect to
+ *
+ * Test cases: https://github.com/LLK/scratch-www/wiki/Most-Important-Workflows
+ */
+
+var tap=require('tap');
```
rschamp: Nit: The `=` should have spaces around it.

test/integration/test_navbar_links.js
```diff
@@ -0,0 +1,139 @@
+/*
+ * Checks that the links in the navbar on the homepage have the right URLs to redirect to
+ *
+ * Test cases: https://github.com/LLK/scratch-www/wiki/Most-Important-Workflows
+ */
+
+require('chromedriver');
```
rschamp: Is this line here necessary? `chromedriver` is not in the requirements, so I got errors when I tried to run these tests.

test/integration/test_project_rows.js
```diff
@@ -0,0 +1,121 @@
...
+        //expected value of the title text
+        var expectedText = 'Featured Studios';
+        t.equal(text, expectedText);
+        t.end();
+    });
+});
+
+//checks that the link for a studio makes sense
+tap.test('checkFeaturedStudiosRowLinkWhenSignedOut', function (t) {
+    var xPathLink = '//div[contains(@class, "thumbnail") and contains(@class, "gallery") and contains(@class, "slick-slide") and contains(@class, "slick-active")]/a[@class="thumbnail-image"]';
```
rschamp: I get an error here:
```  checkFeaturedStudiosRowLinkWhenSignedOut
  not ok NoSuchElementError: no such element: Unable to locate element: {"method":"xpath","selector":"//div[contains(@class, "thumbnail") and contains(@class, "gallery") and contains(@class, "slick-slide") and contains(@class, "slick-active")]/a[@class="thumbnail-image"]"}   (Session info: chrome=56.0.2924.87)   (Driver info: chromedriver=2.27.440174 (e97a722caafc2d3a8b807ee115bfb307f7d2cfd9),platform=Mac OS X 10.12.3 x86_64)
```

package.json
jwzimmer: I did not mean to add that. : (

test/integration/test_navbar_links.js
jwzimmer: It's necessary as far as I can tell. We have to pass the chromedriver to selenium webdriver. I thought we weren't going to make testing packages part of package.json > dev dependencies?

test/integration/test_navbar_links.js
jwzimmer: I added an assertion, not sure if it was the best one to choose, but... I used ok() for now. I wanted to get this in a place where it made sense for @paulkaplan & I to talk about it together after our conversation earlier today, so I made a first-pass at addressing the issues in it (there's still 1 failing test, I haven't looked at why it's failing yet).

jwzimmer:
@rschamp re:
> One other thing I wondered, which I may have already asked you about, is if it's possible to capture the result of clicking on the links rather than checking their hrefs? It is possible that we could override the behavior inadvertently, so it would be most correct to check that when you click a link, the browser attempts to navigate to the specific URL you're testing for.

When we first talked about these tests, we had decided not to use the click functionality but to check the hrefs b/c you brought up concerns about performance. I can try click and see how much slower it is, though.

rschamp:
Oh ok, that makes sense. If it is not too much work, I would be curious how much slower it is. If it's not that much slower, I think it would be worth changing it since it would test the behavior better.

jwzimmer:
Post-switching off of selenium promise manager, all tests seem to be working. Note that using "click" is much slower than just checking the href - the file test_footer_links now uses the click function. We tried checking an element on the subsequent page (that the click goes to) versus checking the page's URL without waiting for any part of it to load in case that saved time, but it did not. However, we went with the latter strategy anyway because it removes the tests' reliance on pages not maintained by scratch (e.g. has the ScratchEd site design changed?) and reduces reliance on element locations in a page in general (which is brittle).

screen shot 2017-04-18 at 10 55 37 am

mewtaylor:
This looks good to me – I'm cool with using clicks given the reasons mentioned. However, if that's the direction we're going longer term, I think it'll be important to think about when and how often we check links.

Issue #1153, opened 2017-01-20
Typing "%" into search causes a fatal exception


thisandagain:
Operating System: `Mac OS 10.12.2`
Browser: `Chrome 52.0.2743.116 (64-bit)`

### To Reproduce
- Go to any page on Scratch
- Enter "100% pen" (or anything with a "%") into the search field
- Hit return

### Expected
- See search results

### Actual
- Fatal exception

### Reference
```
search.jsx:138 Uncaught URIError: URI malformed
    at decodeURI (<anonymous>)
    at Constructor.render (search.jsx:138)
```

liam4:
Actually it'd seem that just typing "%" causes the crash. (I'm sure you guys noticed that, though.)

TheGrits:
I guess this is Scratch's version of the blue screen of death!

Can confirm using Chrome 55.0.2883.105 (64-bit) on ChromeOS.

WorldLanguages:
Same on Windows 10 - Chrome 55.0.2883.87

I think Scratch interprets the % as a symbol to go before a symbol code (like %20) and Scratch can't get the symbol "%20[...]" because the correct symbol would be "20[...]

WorldLanguages:
It gets fixed if we use %25 instead of % in the URL.

[](https://scratch.mit.edu/search/projects?q=100%25%20pen)

thisandagain:
In general it appears that we are no longer encoding / decoding uri components properly. Looking quickly at the code, my guess is this was introduced by this PR: https://github.com/LLK/scratch-www/commit/cbd983baf8ea720928c62707bb68676d3b30eef3

@mewtaylor any thoughts?

PR #1193, opened 2017-02-20
Search view fixes

### LINKED ISSUE #1189: Search bar not working - Staging
**TL;DR:** The API request is not getting the search query term.

Found while smoke-testing (https://github.com/LLK/scratchr2/issues/4457).

**Repro:**
- Go to staging
- Search for "duck"
- Search for "duck cat"
- Note the results are the same for either query, and don't seem relevant

rschamp:
* Fixes #1189: Wait for the search term prop before performing the search. Also encode the term sent to the API, so the API request uses a valid URL.
* Also fixes an unlogged error, the same as #1153, but from the search form on the search page (not in the nav bar).
* Finally, clean up keys to resolve warnings from React. There were so many on the search page that my browser was locking up (only affects development).

### DIFF
--- a/src/views/search/search.jsx
+++ b/src/views/search/search.jsx
@@ -54,8 +54,8 @@ var Search = injectIntl(React.createClass({
         term = decodeURI(term.split('+').join(' '));
         this.props.dispatch(navigationActions.setSearchTerm(term));
     },
-    componentDidUpdate: function (nextProps) {
-        if (this.props.searchTerm !== nextProps.searchTerm) this.getSearchMore();
+    componentDidUpdate: function (prevProps) {
+        if (this.props.searchTerm !== prevProps.searchTerm) this.getSearchMore();
     },
     getSearchMore: function () {
         var termText = '';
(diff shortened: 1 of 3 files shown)


thisandagain:
Thanks @rschamp . This LGTM.

src/views/search/search.jsx
```diff
@@ -51,14 +51,16 @@ var Search = injectIntl(React.createClass({
...
             term = term.substring(0, term.indexOf('&'));
         }
-        term = term.split('+').join(' ');
-        this.getSearchMore();
-        this.props.dispatch(navigationActions.setSearchTerm(decodeURI(term)));
+        term = decodeURI(term.split('+').join(' '));
+        this.props.dispatch(navigationActions.setSearchTerm(term));
+    },
+    componentDidUpdate: function (nextProps) {
+        if (this.props.searchTerm !== nextProps.searchTerm) this.getSearchMore();
```
mewtaylor: I could be missing something – but won't `getSearchMore` in this instance be used on the existing `searchTerm`, whereas we want to use it on the new one?

rschamp: Fixed! Just called that argument the wrong thing.

PR #1217, opened 2017-02-23
Fix GH-1196: Don’t localize an empty string

### LINKED ISSUE #1196: Teacher registration Organization step "Other" option overlaps missing message string
I believe there should be no label at all on the last option for Organization type, but instead there's a string being overlapped by a text field. (Chrome 56 macOS 10.12.3)

image

mewtaylor:
Fixes #1196 by removing the other empty string from the localization file, and just adding it to the options list separately.

### DIFF
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -516,11 +516,10 @@ module.exports = {
             'orgChoiceAfterschool',
             'orgChoiceMuseum',
             'orgChoiceLibrary',
-            'orgChoiceCamp',
-            'orgChoiceOther'
+            'orgChoiceCamp'
         ],
         getOrganizationOptions: function () {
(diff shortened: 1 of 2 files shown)


mewtaylor:
@rschamp yeah I think it was basically that the empty string wasn't getting translated (because why), and so when the pull happened, it was getting an empty string in all languages including the source language (English), and therefore defaulting to using the id since the string was empty.

jwzimmer:
Chrome 56, Mac OS El Capitan, Staging

I no longer see the string name, but I cannot type in the Other field when it is checked.

screen shot 2017-02-23 at 10 26 32 am

jwzimmer:
I am not seeing the string name, and I can type in the other field when checked; I went through teacher registration in Spanish & English. (Chrome 56, Mac OS El Capitan, Staging) 👍

PR #1237, opened 2017-03-06
Fix gh-1232: Footer Links space-between

### LINKED ISSUE #1232: Make footer flexbox be `space-between` rather than `space-around`?
Just noticed this while looking at an `InformationPage` locally. Currently, we use `justify-content: space-around` for displaying the footer link lists. Since this centers the columns, we end up with this:
screen shot 2017-02-28 at 8 57 52 am

If we were to move to using `space-between` (filling the width its given, rather than centering within the space its given), we'd get something like this:
screen shot 2017-02-28 at 8 58 11 am

/cc @carljbowman @lifeinchords

TheGrits:
Resolves #1232

### DIFF
--- a/src/components/footer/www/footer.scss
+++ b/src/components/footer/www/footer.scss
@@ -7,7 +7,7 @@
         text-align: center;
         flex-direction: row;
         flex-wrap: nowrap;
-        justify-content: space-around;
+        justify-content: space-between;
 
         dl {
             display: inline-block;


TheGrits:

mewtaylor:
@TheGrits thanks :). I think we're still trying to determine if we want it to be `space-between` or not, so I'm going to hold off on merging until we decide for sure. I'll let you know as soon as we decide!

TheGrits:
@mewtaylor Yep, no problem. I thought that the discussion might have been over since @thisandagain removed the `needs-discussion` tag off of #1232 and added `help-needed`.

PR #1241, opened 2017-03-13
Fix GH-1199: iterate over languages for views without l10n files

### LINKED ISSUE #1199: Terms of Use, Search & Privacy Policy pages language setting immediately resets to English
**TL;DR:** Change the language on the Terms of Use or Privacy Policy page to Farsi. Note that the page refreshes immediately and the language is reset to English.

This isn't Farsi-specific. If you choose Chinese or German the same behaviour exists. (I think probably it will happen for any language besides English.)

Re Terms of Use: On Production, if you choose Farsi, the footer is translated. (We don't translate legal pages, I think, so @chrisgarrity & I weren't expecting the body to be translated, but the footer should still be translatable.)

I saw this on an iPad + Safari, and Mac OS El Capitan + Chrome 56, and Ubuntu 16.04 + Firefox 51.0.1.

mewtaylor:
Fixes #1199 by continuing the iteration over languages even if a view doesn’t have an l10n file. The issue was that only the `en` language object was getting added when the view didn’t have an l10n file, because of the `hasOwnProperty` check. This adjusts it to iterate over languages anyways, and only ouput the `No translations for…` message if it’s for a language of a view that does have an l10n file.

### Test Cases ###
* Change language on the homepage. Go to `/terms_of_use` – the header and footer should be translated, and the langue in the dropdown should match what you changed it to

### DIFF
--- a/bin/build-locales
+++ b/bin/build-locales
@@ -233,7 +233,8 @@ async.forEachLimit(views, 5, function (view, cb) {
     cb();
 }, function (err) {
     if (err) {
-        process.stdout.write('Writing intl files completed with errors\n');
+        process.stdout.write('Writing intl files no haz successes\n');
+        process.exit(1);
     }
     return;
 });
--- a/bin/build-locales
+++ b/bin/build-locales
@@ -189,48 +189,39 @@ async.forEachLimit(views, 5, function (view, cb) {
     var viewLocales = {};
     viewLocales['en'] = merge({}, generalLocales['en'], defaultLocales[view]);
 
-    if ( defaultLocales.hasOwnProperty(view) ) {
-        // merge view specific english strings, first then other languages
-        process.stdout.write('Merging translations for ' + view + '\n');
-        async.forEach(allLangs, function (isoCode, cb) {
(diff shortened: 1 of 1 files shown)


bin/build-locales
```diff
@@ -189,56 +189,47 @@ async.forEachLimit(views, 5, function (view, cb) {
...
+                    return cb();
+                } else {
                     return cb(err);
                 }
-            });
+            }
+            try {
+                viewLocales[isoCode] = merge({}, generalLocales[isoCode], JSON.parse(data));
+            } catch (e) {
+                return cb(e);
```
chrisgarrity: We should probably do something better with errors from JSON.parse. Would it make sense to merge in the English ones before returning the error to the callback?

mewtaylor: 👍 I would be up for causing the whole translations build to exit on an error in this case

chrisgarrity: I agree - let's fail hard on this one. It means that data in the scratch_translations repo is bad and probably we should fix it upstream.

chrisgarrity:
Looking good on staging.

PR #1269, opened 2017-04-12
Fix R2-4470: Only allows birthdays >= 13 yrs old for teacher reg

mewtaylor:
Helps fix https://github.com/LLK/scratchr2/issues/4470 from the UX side of things by preventing moving forward with teacher registration if the birthday given is less than 13.

### Test Cases ###
* Try to register with a birthday less than 13 years before the current date (could be 12 years, 11 months, even). You should get an error bubble
* Try to register with a birthday >= 13 years before current date. You should be able to proceed.


TheGrits:
Potential conflict:
The validation says you need to be 13 years old, but the `l10n.json` file says you need to be over 13 years old?

mewtaylor:
@TheGrits nice catch, thanks for seeing that! I'll update the language now :)

src/views/teacherregistration/l10n.json
```diff
@@ -36,5 +36,6 @@
     "teacherRegistration.howUseScratch": "How do you plan to use Scratch at your organization?",
     "teacherRegistration.emailStepTitle": "Email Address",
     "teacherRegistration.emailStepDescription": "We will send you a confirmation email that will allow you to access your Scratch Teacher Account.",
-    "teacherRegistration.validationEmailMatch": "The emails do not match"
+    "teacherRegistration.validationEmailMatch": "The emails do not match",
+    "teacherRegistration.validationAge": "Sorry, teachers must be at least 13 years old"
```
mewtaylor: @kaschm I added in this language if the user happens to have been born in the 13th year of their life, but has not yet gotten to their birth month. Does this seem alright to you content-wise?

src/components/registration/steps.jsx
```diff
@@ -324,13 +324,26 @@ module.exports = {
         },
         getYearOptions: function () {
             return Array.apply(null, Array(100)).map(function (v, id) {
-                var year = new Date().getFullYear() - id;
+                var year = new Date().getFullYear() - (id + 13);
```
chrisgarrity: Since this is going to eliminate the last 12 years from the options list, I assume getYearOptions is only used for Teachers.

mewtaylor: great catch! I'll update this to not apply to students, which it will right now

jwzimmer:
I think this caused https://github.com/LLK/scratch-www/issues/1278.

jwzimmer:
### Testing
(Firefox 51.0.1, Win 7 VM)
* Most recent birthday I can enter in the dropdown is December, 2004
  * This won't let me click "Next", I get "Sorry, teachers must be at least 13 years old"
* April, 2004 also triggers age validation
* January, 2004 also triggers age validation
  * **But unless I'm counting wrong, you should be 13 years old at that point?**

PR #1271, opened 2017-04-14
Fix gh-875: Double Search Bar

### LINKED ISSUE #875: Do we want to keep double search bars on new Search page?
From discussion with @carljbowman re https://github.com/LLK/scratch-www/pull/871, once we have implemented https://github.com/LLK/scratch-www/issues/678, there will be two search bars with the search term displayed on the new Search page. Do we like that design?

out

TheGrits:
Resolves #875 

## Test cases:
- Search bar in actual search page should be removed
- Search bar in nav-bar should still be there

Cc @carljbowman


TheGrits:

PR #1272, opened 2017-04-17
Alphabetize Routes List

TheGrits:
No issue logged for this, just something I thought I would do because I find it annoying when things aren't sorted.

I didn't change the routes themselves, I just sorted them by copy-pasting.

Feel free to take it or leave it.

## Test cases:
- Just make sure that the pages a routing properly still

### DIFF
--- a/src/routes.json
+++ b/src/routes.json
@@ -1,17 +1,5 @@
 [
     {
-        "name": "splash",
-        "pattern": "^/?$",
-        "routeAlias": "/?$",
-        "view": "splash/splash",
-        "title": "Imagine, Program, Share"
-    },
-    {
-        "name": "splash-redirect",
-        "pattern": "^///?$",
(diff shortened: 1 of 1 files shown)


TheGrits:

mewtaylor:
@TheGrits sorry this has taken a while for me to get to, to the point that there are now new routes that need to be added to this – if you have the time/desire, would you mind doing the following things:

1. Adding the new routes that we have now (like `/tips` for instance)
2. Separating out the `redirect` routes into a separate section, below the alphabetized routes – I was thinking it might be helpful to have those be a separate section

Thoughts?

TheGrits:
@mewtaylor Maybe I should start a new PR to rebase it?

mewtaylor:
@TheGrits yeah that'd be really helpful, thanks so much!

PR #1275, opened 2017-04-19
Add card translations

seotts:
Added translated cards for ca, es, and zh-tw (all cards) and sv (Scratch2 cards)

Addresses scratchr2 issue #4533


bin/lib/localized-urls.json
```diff
@@ -1,23 +1,68 @@
...
+    	"ttt.DressupGuideLoc": "/pdfs/guides/DressupGuide.pdf",
+    	"ttt.StoryActivityLoc": "/pdfs/cards/StoryCards.pdf",
+    	"ttt.StoryGuideLoc": "/pdfs/guides/StoryGuide.pdf"
     },
     "ar": {
         "cards.starterLink": "//cdn.scratch.mit.edu/scratchr2/static/pdfs/help/ar/Scratch2Cards.pdf"
     },
     "ca": {
-        "cards.starterLink": "//cdn.scratch.mit.edu/scratchr2/static/pdfs/help/ca/Scratch2Cards.pdf"
+        "cards.starterLink": "//cdn.scratch.mit.edu/scratchr2/static/pdfs/help/ca/Scratch2Cards.pdf",
```
chrisgarrity: We can't have cards.starterLink multiple times in a single locale, here and line 53. We probably just want to keep the newer one.

bin/lib/localized-urls.json
```diff
@@ -54,5 +121,34 @@
     },
     "sl": {
         "cards.starterLink": "//cdn.scratch.mit.edu/scratchr2/static/pdfs/help/sl/Scratch2Cards.pdf"
+    },
+    "sv": {
+        "cards.starterLink": "/pdfs/cards/sv/Scratch2Cards.pdf"
+    },
+    "zh-tw": {
+        "cards.starterLink": "//cdn.scratch.mit.edu/scratchr2/static/pdfs/help/zh-tw/Scratch2Cards.pdf",
```
chrisgarrity: Another duplicate

seotts:
@chrisgarrity I removed the scratchr2 link for ca. Looking closer, I don't think there's a translation of the Scratch2 starter cards for zh-tw, so I removed both starterLinks for that

bin/lib/localized-urls.json
```diff
@@ -137,7 +135,6 @@
         "cards.petLink": "/pdfs/cards/zh-tw/PetCards.pdf",
         "cards.pongLink": "/pdfs/cards/zh-tw/PongCards.pdf",
         "cards.raceLink": "/pdfs/cards/zh-tw/RaceGameCards.pdf",
-        "cards.starterLink": "/pdfs/cards/zh-tw/Scratch2Cards.pdf",
```
chrisgarrity: Whoops, looks like both copies were removed for zh-tw. We still want this one I think.

chrisgarrity: Never mind, I missed the comment - good catch!

PR #1287, opened 2017-04-25
Use custom VCL for Pass/!Pass conditions

mewtaylor:
We’ve now exceeded our max number of characters for a condition in the Fastly API, and we need to make it larger to accommodate regex conditionals that can match on any of the routes in www currently.

This fixes the issue by moving the conditions – and the states that are affected by it, like setting the backend or cache ttls – to two custom vcl files that are updated via the Fastly API. One is for the `vcl_rev` config, and one is for the `vcl_fetch` config.

I chose to build the strings for these conditions manually rather than adding in a vcl-formatting package due to the decision to go the route of using custom vcl for now. This is because custom vcl can (and should, here) can be snippets of vcl, and most tools i was looking at were designed to build full vcl files rather than partial components.


bin/lib/fastly-extended.js
```diff
@@ -166,5 +166,94 @@ module.exports = function (apiKey, serviceId) {
...
+     */
+    fastly.setResponseTTL = function (condition) {
+        return '' +
+            'if (' + condition + ') {\n' +
+            '    set beresp.ttl = 0s;\n' +
+            '    set beresp.grace = 0s;\n' +
+            '    return(pass);\n' +
+            '}\n';
+    };
+
```
rschamp: Since `setBackend`, `setForwardHeaders`, and `setResponseTTL` only return strings, I don't think they make sense to add to the `fastly` module, since here `set` usually means actually interacting with the API, so I would expect these to actually make requests to the API.

I think these would make more sense in configure-fastly along with the other methods there that build strings.

test/unit/test_fastly_extended_method.js
```diff
@@ -0,0 +1,43 @@
...
+tap.test('testSetTTL', function (t) {
+    var ttl = fastly.setResponseTTL('itsactuallyttyl');
+    t.equal(ttl, '' +
+        'if (itsactuallyttyl) {\n' +
+        '    set beresp.ttl = 0s;\n' +
+        '    set beresp.grace = 0s;\n' +
+        '    return(pass);\n' +
+        '}\n'
+    );
+});
```
rschamp: Thanks for adding these!  Would it be possible to add any that test these methods using the actual routes? That way we could catch PRs that will error on deployment before we actually merge.  Like, one that tests that inputting the result of getAppRouteCondition into these still returns a string (and doesn't throw an exception).

PR #1293, opened 2017-04-27
Fix gh-1292: Dropdown Mobile Un-Friendly

### LINKED ISSUE #1292: Profile Dropdown isn't Mobile Friendly
scratch mit edu- iphone 6 plus

Of course I forgot to test the dropdown on mobile. Sorry!

I'll try to figure out a patch ASAP

Related PR: #1270

TheGrits:
Fixes #1292 

Test cases:
- `min-width: 160px` should not be applied on desktop sized screens
- `min-width: 160px` should be applied to mobile sized screens

### DIFF
--- a/src/components/dropdown/dropdown.scss
+++ b/src/components/dropdown/dropdown.scss
@@ -85,7 +85,7 @@
         }
     }
     
-    @media only screen and (max-width: $mobile - 1) {
+    @media only screen and (max-width: $tablet - 1) {
         min-width: 160px;
     }
 }
--- a/src/components/dropdown/dropdown.scss
+++ b/src/components/dropdown/dropdown.scss
@@ -84,3 +84,7 @@
         }
     }
 }
+
+@media only screen and (max-width: $mobile - 1) {
+    min-width: 160px;
+}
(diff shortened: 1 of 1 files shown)


src/components/dropdown/dropdown.scss
```diff
@@ -84,3 +85,7 @@
         }
     }
 }
+
+@media only screen and (max-width: $mobile - 1) {
```
mewtaylor: I think this needs to be inside a `.dropdown` tag

TheGrits: @mewtaylor Okey dokey! I'll hop to it!

TheGrits:
@mewtaylor That should do the trick!

mewtaylor:
@TheGrits I noticed one more edge case:
screen shot 2017-05-08 at 9 48 59 am
It looks like this is due to the username being hidden in landscape too – so I think that, for the responsive check, we might want to check that it's below `$tablet` rather than below `$mobile`.

TheGrits:
@mewtaylor 👍 Done!

PR #1299, opened 2017-05-02
Fix gh-1294: Unlocalize Browser Links

### LINKED ISSUE #1294: Use non-localized URL for Firefox
Suggestion from translators:

in faq.requirementsBody, replace the Firefox link: `http://www.mozilla.org/en-US/firefox/new/` with a generic one `http://www.mozilla.org/firefox/new/` so the browser will pick the version based on the browser's locale.

Currently translators have to override the en-US link which is potentially error-prone, and creates warnings in transifex that the URL doesn't match.

TheGrits:
Resolves #1294 

## Test cases:
- On FAQ page, the link to the Firefox browser should be `http://www.mozilla.org/firefox/new/`


TheGrits:
@chrisgarrity Done!

PR #1300, opened 2017-05-02
Fix gh-1282 ~ Add "United States" to faq/educators page

### LINKED ISSUE #1282: FAQ: question about privacy is US-centric
The question/answer about privacy `Is Scratch 2.0 (online version) compliant with local and federal data privacy laws?` is specific to the US, but doesn't say that. So it's difficult to localize - should it be translated as is? Should the translator add "United States" to the string? Should translators just ignore it?

Or do we need to update the question so it makes sense for both US and non-US schools. e.g. we care about privacy and comply with US rules, perhaps not other countries if they have stricter standards than US.

St19Galla:
Fixes #1282 

# Test Cases:
Ensure the text says (from @kaschm) : 
> Is Scratch 2.0 (online version) compliant with United States local and federal data privacy laws?

> Scratch cares deeply about the privacy of students and of all individuals who use our platform. We have in place physical and electronic procedures to protect the information we collect on the Scratch website. Although we are not in a position to offer contractual guarantees with each entity that uses our free educational product, we are in compliance with all United States federal laws that are applicable to MIT, a 501(c)(3) organization and the entity that created and maintains Scratch. We encourage you to read the Scratch Privacy Policy for more information.


St19Galla:
@rschamp

PR #1305, opened 2017-05-03
Fix GH-1303: Apply english strings as defaults for views during builds

### LINKED ISSUE #1303: Some translations broken in "meow"
### Repro
- Navigate to splash page
- Change language to "meow"

### Reference
image

@mewtaylor @chrisgarrity @rschamp Any thoughts?

mewtaylor:
This fixes #1303 by applying the english strings for a view as defaults in the event that a view is missing some strings.

If we wanna hotfix this we can too!


jwzimmer:
### Testing

* [x] Partly tested by automated test added in https://github.com/LLK/scratch-www/pull/1314
  * The test checks that no page files in the intl directory are missing any string ids when compared to the list of English string ids for that page
* [ ] English strings should never be used when a localized string is available (except for English); they should be the default only if the string is missing from a given non-English language

jwzimmer:
I added the `blocked` label because we should at least figure out if https://github.com/LLK/scratch-www/issues/1334 is really a blocker.

jwzimmer:
https://github.com/LLK/scratch-www/issues/1334 is not a real bug/ blocker, see https://github.com/LLK/scratch-www/issues/1334#issuecomment-305320895 for explanation 👍

PR #1316, opened 2017-05-18
Fix r2-3978: Use regex in express patterns to limit scope

mewtaylor:
This fixes https://github.com/LLK/scratchr2/issues/3798 by using regex in expressjs patterns, and then also parsing those regex patterns for fastly, such that we can limit the scope of the regex created for a specific view.


rschamp:
QA:

* Visiting explore/projects/all displays the appropriate explore page
* Viewing the explore bar does not result in any errors
* `/explore/ajax/new/<id>/creator` returns an HTML fragment, not a redirect to the explore page
* `/explore/projects/new/<id>/creator` redirects to `/explore/projects/all` (expected)

Issue #1274, opened 2017-04-18
Automated test for checking statistics page


jwzimmer:
**TL;DR:** To replace the manual `/statistics` smoke-test.

Now that https://github.com/LLK/scratch-www/pull/1185 is mostly done, I can work on replacing another manual smoke-test from https://github.com/LLK/scratchr2/wiki/Smoke-Testing-Test-Cases.

This can be done in parallel with https://github.com/LLK/scratch-www/issues/925, which I will work on with @rschamp.

PR #1317, opened 2017-05-18
Fix issue/gh 1274 by adding a test for the statistics page

jwzimmer:
Fixes https://github.com/LLK/scratch-www/issues/1274.

Added a test that checks the 3 of the charts on the Stats page by checking for 3 things:
- the title, which should exist even if the charts don't load (current Staging situation)
- the first label on the X axis, which only exists if the charts loaded
- the first label in the chart legend, which only exists if the charts loaded


test/integration/smoke-testing/test_statistics_page.js
```diff
@@ -0,0 +1,125 @@
...
+
+// to verify the Comment Activity Chart's presence, checks the title,
+// a label on the X axis, and a label in the legend
+tap.test('commentActivityChartShouldExist', function (t) {
+    var chartTitle = 'Monthly Comment Activity';
+    var chartLegendLabel = 'Stacked';
+    var chartXLabel = '01/2008';
+    driver.findElements(seleniumWebdriver.By
+        .css('div.box-head h3'))
+        .then( function (elements) {
```
mewtaylor: Actually sorry just noticed this – I think the style we generally use in www is to newline the closing brace/parentheses, like:
```javascript
.then(function (text) {
    t.equal(text, chartTitle);
})
.then...
```
rather than closing on the same line

jwzimmer:
I don't know why this was working and now isn't.
The problem might be that selenium's wait until elementsLocated function waits until at least 1 element matching is found, rather than all of them, so I'm trying to index into elements in that list that aren't guaranteed to be there?
I attempted to fix this by changing the wait to wait for specific relevant elements rather than the list of all the elements that are chart labels, but the test now just hangs forever, so I went ahead and moved this to the next milestone. : (

jwzimmer:
I used @paulkaplan's [test](https://github.com/paulkaplan/scratch-gui/blob/selenium-tap/test/selenium/test.js) when writing this test, thanks @paulkaplan!

test/integration/smoke-testing/test_statistics_page.js
```diff
@@ -0,0 +1,85 @@
...
+    var productionURL = 'https://scratch.mit.edu/statistics';
+    return driver.get(productionURL);
+});
+
+test('check that Monthly Activity Trends title is present & correct', t => {
+    var chartTitle = 'Monthly Activity Trends';
+    findByCss('div.box-head h3')
+    .then( function (element) {
+        return element.getText('h3');
+    })
```
rschamp: If you like, these can also become arrow functions like `.then(element => element.getText('h3'))`.

paulkaplan:
@jwzimmer Glad that code snippet helped a bit!  maybe we could roll our own tiny library for working with selenium so we could use it over on scratch-3 as well?

Issue #1013, opened 2016-11-10
Remove end-to-end testing package dependencies from repo


mewtaylor:
`selenium-webdriver`, and any related dev dependencies, will tend to involve native dependencies, and take a while to build. Since we will not be doing automated end-to-end tests as part of our continuous integration process and rather spinning them off independently on separate machines, we should remove these dependencies so they do not take up CI time on the install time.

In order to do this, we should also include documentation that specifies that these are required dependencies.

PR #1319, opened 2017-05-22
Issue/gh 1013 remove end to end testing package dependencies from repo

jwzimmer:
Fixes https://github.com/LLK/scratch-www/issues/1013 by removing Selenium from the package.json. I didn't remove TAP since it isn't just used for end-to-end testing.

I added a wiki page listing requirements for automated (end-to-end) testing: https://github.com/LLK/scratch-www/wiki/Automated-Tests.


test/integration/package.json
```diff
@@ -0,0 +1,5 @@
+{
+    "testDependencies": {
+        "selenium-webdriver": "2.44.0"
```
rschamp: Now that this is separate from `package.json`, maybe we should also include `chromedriver`, so that can be tracked (until we go with another way of doing it).

PR #1321, opened 2017-05-23
Fix 1303 Fallout: Apply defaults for untranslated pages too

mewtaylor:
since we’ll need them for nav/footer strings

/cc @chrisgarrity


chrisgarrity:
👍

PR #1329, opened 2017-05-27
Readme update for running standalone client on Windows

LiFaytheGoblin:
From the issues and from my own experience it seems that the standalone client doesn't work well on windows. I added a section "Windows" to the readme, to point windows users to two possible solutions. 
(1) Cygwin has not worked for me personally, but for @cwillisf it did. 
I tried (2) Wubi(uefi) which worked perfectly well. It is a simple .exe wich installs Ubuntu from within Windows and you can decide for Windows or Ubuntu when booting. No partitioning is needed. In Ubuntu you can still access your files from Windows. You have to install things like npm, node and git again for Ubuntu though. I hope my wording is okay, feel free to change it though. I also wasn't sure about how much information I should provide about the two solutions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ make deploy
 ### Current issues with the development
 We're currently in the process of transitioning into this web client from Scratch's existing structure. As we transition, there are going to be some issues along the way that relate to how this client needs to interact with the existing infrastructure to work properly in production.
 
+#### FALLBACK ####
 On top of migrating to using this as our web client, Scratch is also transitioning into using a new API backend, Scratch REST API. As that is also currently in development and incomplete, we are set up to fall back to using existing Scratch endpoints if an API endpoint does not exist – which is where the `FALLBACK` comes in.
 
 Most of the issues we have currently revolve around the use of `FALLBACK`. This variable is used to specify what url to fall back onto should a request fail within the context of this webclient, or when using the `API_HOST`. If not specified in the process, it will not be used, and any request that is not made through the web client or the API will be unreachable.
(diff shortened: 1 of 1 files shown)


mewtaylor:
@cwillisf @LiFaytheGoblin fyi: https://github.com/LLK/scratch-www/issues/1332

README.md
```diff
@@ -102,6 +102,7 @@ make deploy
 ### Current issues with the development
 We're currently in the process of transitioning into this web client from Scratch's existing structure. As we transition, there are going to be some issues along the way that relate to how this client needs to interact with the existing infrastructure to work properly in production.
 
+#### FALLBACK ####
```
liam4: Humble nitpick, other headings in the README don't have the following `####`, i.e.:

```markdown
<!-- Bad -->
#### FALLBACK ####
#### Windows ####

<!-- Good -->
#### FALLBACK
#### Windows
```

mewtaylor: @liam4 nice catch, though I would not say this is a matter of "bad" or "good", just convention. @LiFaytheGoblin would you mind removing the trailing `####`? Probably best to follow the other header conventions for now.

liam4: Yep, I had thought to go with `OLD` and `NEW`, but that didn't seem right, either. `NORMAL` and `YOU` seemed to be putting too much blame or guilt.. writing code comments that you're actually presenting isn't easy :)

St19Galla:
Happy first pull request @LiFaytheGoblin 🎉🎉🎉🎉🎉

README.md
mewtaylor: @liam4 gotcha – I think another way would be to say `Currently in the PR` and `Repo convention`, perhaps :).

mewtaylor:
Haha you beat me to it @St19Galla :).

README.md
LiFaytheGoblin: Thanks for the hint! I can remove these extra ### :)

PR #1358, opened 2017-06-20
[Master] Hotfix: Add redirect for store

mewtaylor:


mewtaylor:
@rschamp nope! You are correct!

mewtaylor:
This appears to be workign on staging, but @jwzimmer I could use a second set of eyes if you have a moment before deploying this!

jwzimmer:
QA:
- I checked with @mewtaylor that the redirect to ending in /password is correct
- Homepage -> store page looks fine on Mac, Chrome 59
- Homepage -> store page looks fine on Chrome, Win 7 VM
- Appending /store to non-homepage URL's (About, Discuss) causes 404 page, seems reasonable
- Appending /store to Explore page (`/explore/projects/all/`) redirects to `/explore/projects/all/`, which seems reasonable

PR #1368, opened 2017-06-26
Update compatibility requirements to clarify support for ChromeOS

thisandagain:
- Clarify support for ChromeOS
- List platforms in order of use from most to least (Windows, ChromeOS, MacOS, Linux)
- Remove confusing line about Chromium
- Remove "Windows" from Safari listing as it is no longer supported


mewtaylor:
@thisandagain do we need to hotfix this, or should it go in with the release?

Also does anyone else read that branch name as sounding like "oreos" in their head, but with "chrome"? Like the most delicious browser cookie snack there ever was?

thisandagain:
@mewtaylor I'd be fine with either but I would like it to get out this week. Would you like me to target the PR against the release branch?

mewtaylor:
@thisandagain if there's a time sensitivity on it I'm fine with having it be a hotfix! Would you mind opening a second PR against the release branch too, so we can just sanity check it on staging/so it goes into staging too?

thisandagain:
Looks good in staging:
https://scratch.ly/info/faq

PR #1380, opened 2017-06-27
FIx layout problems

chrisgarrity:
* removed white-space: no wrap to let buttons flow in Cards section
* Fix download link (download should be part of the link)
* Set the size of the icons to fix alignment


jwzimmer:
### Testing

- [x] Button fix: I can see both buttons aligned vertically on my iPhone 6s, Chrome
- [x] Link fix: I can see that "download" is now part of the link text
- [x] Icon fix: I can see block icon at normal size in IE 11, Win 7 VM

PR #1383, opened 2017-06-28
Fix blocks icon for windows

chrisgarrity:
Note for future reference - set the size for SVGs with height not width if using rem.
IE on Windows 7 doesn’t resize correctly if the width is set with rem. Height works with rem, and exact pixel width works, but that should be avoided.


jwzimmer:
Block icons look fine in IE 11, Win 7 VM

PR #1391, opened 2017-06-29
update /help and /info routes

chrisgarrity:
basing new patterns for /help and /info on /go.

/cc @mewtaylor

### DIFF
--- a/src/routes.json
+++ b/src/routes.json
@@ -153,14 +153,14 @@
     },
     {
         "name": "help",
-        "pattern": "^/help/?$",
-        "routeAlias": "/help",
+        "pattern": "^/help/?(\\?.*)?$",
+        "routeAlias": "^/help/?(\\?.*)?$",
         "redirect": "/tips"
     },
     {
         "name": "info-redirect",
-        "pattern": "^/info/?$",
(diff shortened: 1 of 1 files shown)


src/routes.json
```diff
@@ -153,14 +153,14 @@
     },
     {
         "name": "help",
-        "pattern": "^/help/?$",
-        "routeAlias": "/help",
+        "pattern": "^/help/?(\\?.*)?$",
+        "routeAlias": "^/help/?(\\?.*)?$",
```
mewtaylor: The `^` is not necessary in the `routeAlias`, because it is added by the configuration script

src/routes.json
```diff
@@ -153,14 +153,14 @@
...
         "redirect": "/tips"
     },
     {
         "name": "info-redirect",
-        "pattern": "^/info/?$",
-        "routeAlias": "/info",
+        "pattern": "^/info/?(\\?.*)?$",
+        "routeAlias": "^/info/?(\\?.*)?$",
```
mewtaylor: Same thing here @chrisgarrity – if we can take these out, that'd be great!

PR #1401, opened 2017-07-06
Add error checking and testing for translations

chrisgarrity:
* make build-locales fail if there are errors in the translations
* add test to parse translation files

Also make build-locales quieter if there are no problems.

### DIFF
--- a/bin/build-locales
+++ b/bin/build-locales
@@ -157,7 +157,7 @@ for (var v in routes) {
         localizedAssetUrls[routes[v].name] = {};
 
         var viewUrls = require(l10nStatic);
-        localizedAssetUrls[routes[v].name]['en'] = viewUrls; 
+        localizedAssetUrls[routes[v].name]['en'] = viewUrls;
         var defaultUrls = localizedUrls['en'];
         for (var lang in languages) {
             localizedAssetUrls[routes[v].name][lang] = {};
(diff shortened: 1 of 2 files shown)


chrisgarrity:
I had the same concern, but @thisandagain and @jwzimmer were both fine with adding the test here since we already have the test harness etc.

rschamp:
Would it be possible to move the Jenkins task to a Travis cron job? Then we could avoid checking in bad files and be notified that the build failed the same way as everything else, through Travis.

chrisgarrity:
I think there are two options on www for now:

- Leave as is, test is redundant with build failure. Most developers will be using `npm start` in any case, and you can work around it by running `make clean` followed by `make webpack` to check the build step without translations.
-  Revert build-locales to just warning for translation errors, and keep the test for parsing the translation files (PRs would fail in test). I think we should keep the parse test for the translation files as a back up even if we add tests to scratchr2_translations. I agree that it doesn't feel right to be running tests on an included module, but it is safer.

chrisgarrity:
Final decision - allow build without translations (ignore any failures). PRs etc will fail the test if there are syntax errors in translation files.

bin/build-locales
```diff
@@ -204,8 +203,9 @@ async.forEachLimit(views, 5, function (view, cb) {
         fs.readFile(translationsFile, 'utf8', function (err, data) {
             if (err) {
                 if (err.code === 'ENOENT') {
+                    // ignore missing files for english, Meow, and Edible Scratch
                     if (isoCode !== 'en') {
-                        if (defaultLocales.hasOwnProperty(view)) {
+                        if (isoCode !== 'cat' && isoCode !== 'yum'  && defaultLocales.hasOwnProperty(view)) {
```
mewtaylor: should we combine these with the conditional above?

chrisgarrity: No, we could remove the check/warning. `cat` and `yum` are special cases and we're not going to add the json files, so they will always be missing and we don't need to warn. Everything else should really be there, so we shouldn't see the message except for l10n files that are brand-new and not in Transifex yet. Maybe it would be better to figure out a test that would check for any missing l10n files (ignoring cat and yum).

Kenny2github:
This seems to have been approved - anybody wanna confirm/merge?

chrisgarrity:
Thanks for the nudge @Kenny2github

Kenny2github:
No problem :)

PR #1402, opened 2017-07-07
Fix gh-1397: Update Supporting Organizations

### LINKED ISSUE #1397: Update translations info on /credits page
**TL;DR:** At https://scratch.mit.edu/info/credits, we list Pootle under "Scratch would not be possible without free and open source software, including:" and don't mention Transifex. We should at least remove Pootle, and possibly add Transifex somewhere on the page, if appropriate (Transifex is not free).

/cc @kaschm, @chrisgarrity, @mewtaylor 

screen shot 2017-07-05 at 11 58 36 am

TheGrits:
Resolves #1397 

## Test cases:
- "pootle" should now be removed
- "transifex" should now be added under the organizations that donate their services


TheGrits:

thisandagain:
This also looks great! 🐱 🍰 💥

Issue #575, opened 2016-06-14
[Explore/Search] "Load More" sometimes returns number not divisible by 4


mewtaylor:
If there are more projects, it should return 16 more, which it appears to not do sometimes

mewtaylor:
@thisandagain  this appears to be something with the api – when I have seen it happen, it looks as if the last item returned in the previous "Load More" and the first item returned in the current "Load More" are the same. I think we should close the issue in www.

thisandagain:
Sounds good. Mind opening an issue in API?

PR #1403, opened 2017-07-07
Fix gh-1400: Button Behavior for Search Result State

### LINKED ISSUE #1400: Load More Button Behavior in Search Page
Some search scenarios might be better have the `Load More` button hidden.

**1. No results were found**

screen shot 2017-07-06 at 6 45 33 pm

Would it be better if there's a copy that says no match was found (or something similar) instead of still showing 'load more'?

**2. The search results are less than 16 (limit per search)**

screen shot 2017-07-06 at 7 33 40 pm

**3. Finally, when there are no more Projects/Studios to load upon clicking the button**

(tested using Chrome and Safari on Mac OSX)

carlflor:
Resolves #1400 

- Shows "prompt message" if search term is blank
- Hides Load More Button when:
  - the search results are less than 16
  - there are no more to be loaded
  - Empty Search result
- Added phrase  "Nothing found" for `general.searchEmpty`

*Test Cases*
`scratch` - Returns search result with `Load More` button
`saturated` - Returns more than 16 matches, but hides `Load More` after a few more requests
`flabbergasted` - Returns 16 matches
`pull request` - Returns less than 16 matches (4)
`clean coder` - No search results


thisandagain:
Hi @carlflor thanks for the PR! 

It looks like your PR is not passing some tests however. You can run the tests locally yourself by running `npm test` or look at the output from TravisCI. Please let us know once all tests are passing and then we can take another look!

carlflor:
@thisandagain fixed it! sorry about that! :)

mewtaylor:
@carlflor so sorry for the delay in getting to this, was out of town for a bit and playing catch up after – I'll take a look at this later today!

src/views/search/search.jsx
```diff
@@ -70,9 +70,8 @@ var Search = injectIntl(React.createClass({
         }, function (err, body) {
             var loadedSoFar = this.state.loaded;
             Array.prototype.push.apply(loadedSoFar, body);
-            this.setState({loaded: loadedSoFar});
             var currentOffset = this.state.offset + this.props.loadNumber;
-            this.setState({offset: currentOffset});
+            this.setState({loaded: loadedSoFar, offset: currentOffset});
```
mewtaylor: nice, thanks for this :)

src/views/search/search.jsx
```diff
@@ -96,6 +95,31 @@ var Search = injectIntl(React.createClass({
         }
         return allTab;
     },
+    getProjectBox: function () {
+        var results = <Grid items={this.state.loaded}
```
mewtaylor: We're starting to move to a different style for this kind of multiline – would you mind changing this slightly to something like:
```jsx
var results = <Grid
    items={this.state.loaded}
    itemType={this.props.tab}
    cards={true}
    showAvatar={true}
    showLoves={false}
    showFavorites={false}
    showViews={false}
/>;
```

src/views/search/search.jsx
```diff
@@ -96,6 +95,31 @@ var Search = injectIntl(React.createClass({
...
+                            itemType={this.props.tab}
+                            cards={true}
+                            showAvatar={true}
+                            showLoves={false}
+                            showFavorites={false}
+                            showViews={false} />;
+        var searchAction;
+        if (this.props.searchTerm === '') {
```
mewtaylor: We actually, if you look on production, will show results for a general query when the search term is empty. It appears that there is a separate bug right now for that where you need to click `Load More` to display the first set of results though. I'd recommend removing this check, and I can file a separate bug for the empty search issue.

src/views/search/search.jsx
```diff
@@ -96,6 +95,31 @@ var Search = injectIntl(React.createClass({
...
     },
+    getProjectBox: function () {
+        var results = <Grid items={this.state.loaded}
+                            itemType={this.props.tab}
+                            cards={true}
+                            showAvatar={true}
+                            showLoves={false}
+                            showFavorites={false}
+                            showViews={false} />;
+        var searchAction;
```
mewtaylor: Just to be explicit would you mind setting this to null, i.e. `var searchAction = null;`?

mewtaylor:
One other thing I forgot to mention – we add strings to be translated to a local `l10n.json` file for a page usually, though as you pointed out @carlflor there is no `l10n.json` file for `search`. If you'd like to add that string to the `general` file, you can add it to the `src/l10n.json` top-level translation string file.

mewtaylor:
@carlflor do you have time to take a look at these changes? If not, no worries – I can close this and start a new PR building on your work with these changes.

carlflor:
@mewtaylor Sorry, it took me a long while to accomplish the changes. Let me know if I need to add more. :)

mewtaylor:
@carlflor no worries at all :) – I may not get to this until Monday, but I'll try to get back to you as soon as I can!

mewtaylor:
@carlflor this looks great! I unfortunately just realized that we have one issue though, which is this: https://github.com/LLK/scratch-www/issues/575#issuecomment-226187539

This bug is still around, we are just tracking it in another repository. I wonder if, as a consequence of that bug still being around, we'd need to change this to look only at the length of `body` returned by the request, and see if that length is less than the offset. In the bug mentioned above, the response from the api will still return 16 projects if there are more, but it will overlap with some existing projects and in that overlap will be where the number. 

I am trying to reproduce the bug right now, and having difficulty. If I can't reproduce it in the next week or so, then we could probably merge this in as-is. Thanks for your patience on this!

carlflor:
Sure! No worries. :)

Kenny2github:
@mewtaylor this is waaaay past milestone - any updates?

mewtaylor:
@carlflor sorry for the delay on this – we were having trouble reproducing the bug I had mentioned a few months back, but have recently reproduced it.

Given that, I think the best direction forward with this behavior would be to do this check in the api response, i.e. to see if the length of the body returned by the api request is less than `this.props.loadNumber`. That should help alleviate the issue by isolating the check to the api call, rather than the concatenated list.

I know it's been a while since I've been able to get back to you on this, so if you do not have the capacity to look into this now, no worries! I'd be happy to take it from here if that would be helpful for you, but if you're up to make this adjustment, that'd be awesome (and I'm around to answer any questions you might have too).

carlflor:
Hi @mewtaylor , I'll take a look within the week. Would that be fine?

mewtaylor:
@carlflor absolutely! Thanks!

carlflor:
Hi @mewtaylor , apologies for the delay. Could you check if the changes work for the issue?

src/views/search/search.jsx
```diff
@@ -70,9 +71,9 @@ var Search = injectIntl(React.createClass({
         }, function (err, body) {
             var loadedSoFar = this.state.loaded;
             Array.prototype.push.apply(loadedSoFar, body);
-            this.setState({loaded: loadedSoFar});
             var currentOffset = this.state.offset + this.props.loadNumber;
-            this.setState({offset: currentOffset});
+            var willLoadMore = body.length == this.props.loadNumber;
```
mewtaylor: this should be `===` rather than `==`

carlflor:
@mewtaylor done! :)

src/l10n.json
```diff
@@ -62,6 +62,7 @@
     "general.scratchJr": "ScratchJr",
     "general.scratchStore": "Scratch Store",
     "general.search": "Search",
+    "general.searchEmpty": "Nothing found",
```
mewtaylor: @thisandagain @christanbalch @ntlrsk thoughts on this language for an empty search page?

carlflor:
Cool! Thanks :)

src/l10n.json
mewtaylor: @thisandagain @christanbalch @ntlrsk just wanted to check one more time about empty search language – does this look alright to you?

redmangospros:
is this running, @mewtaylor

src/l10n.json
thisandagain: This looks fine for now. Thanks @mewtaylor

redmangospros:
Great :)

Kenny2github:
@mewtaylor getting close to milestone, any updates?

rschamp:
After merging the ES6 upgrade (#1790), there will be conflicts in the outstanding PRs. Because it will be complicated to merge, we will fix the conflicts ourselves, so don't worry about trying to fix it unless you want to.

carlflor:
Feel free to modify it, I might not be able to look into it any time soon. :)

rschamp:
Thank you @carlflor!

PR #1407, opened 2017-07-10
Fix gh-1399: Misplaced Periods Fix

### LINKED ISSUE #1399: Misplaced periods in text content of Download page?
> # Known Issues
>
> If your offline editor is crashing directly after Scratch is opened, install the Scratch 2 offline editor again (see step 2 above.) This issue is due to a bug introduced in Adobe Air version 14. (released April 2014).

This whole paragraph has a couple of things to note. First:

> This issue is due to a bug introduced in Adobe Air version 14. (released April 2014).

The period before "(released" doesn't look right here. Arguably it could be from the number before it, literally written 14., but we don't call Scratch 2 "Scratch 2.".

I'm thinking it just shouldn't be there at all? That would also make the lowercase R in "released" look correct:

> This issue is due to a bug introduced in Adobe Air version 14 (released April 2014).

Also, another quick thing:

> If your offline editor is crashing directly after Scratch is opened, install the Scratch 2 offline editor again (see step 2 above.) This issue is due to a bug […]

The period is inside of the parenthesis "(see step 2 above.)" here, which I don't think is correct. It looks like it's ending the sentence *inside* of the parenthesis, and not the sentence outside; so it ends up looking like there's a period missing. Instead I think the period should be placed outside of the parenthesis:

> If your offline editor is crashing directly after Scratch is opened, install the Scratch 2 offline editor again (see step 2 above). This issue is due to a bug […]

TheGrits:
Resolves #1399 

## Test cases:
- `This issue is due to a bug introduced in Adobe Air version 14. (released April 2014).` 
Should now read: 
`This issue is due to a bug introduced in Adobe Air version 14 (released April 2014).`

- `If your offline editor is crashing directly after Scratch is opened, install the Scratch 2 offline editor again (see step 2 above.)` 
Should now read: 
`If your offline editor is crashing directly after Scratch is opened, install the Scratch 2 offline editor again (see step 2 above).`

### DIFF
--- a/src/views/download/l10n.json
+++ b/src/views/download/l10n.json
@@ -24,7 +24,7 @@
   "download.otherVersionsOlder": "If you have an older computer, or cannot install the Scratch 2.0 offline editor, you can try installing <a href=\"http://scratch.mit.edu/scratch_1.4/\">Scratch 1.4</a>.",
   "download.otherVersionsAdmin": "If you are a network administrator: a Scratch 2.0 MSI has been created and maintained by a member of the community and hosted for public download <a href=\"http://llk.github.io/scratch-msi/\">here</a>.",
   "download.knownIssuesTitle": "Known issues",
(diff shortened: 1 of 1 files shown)


TheGrits:

jwzimmer:
### Testing

- Looks good in Chrome + Win 7 VM

Issue #1385, opened 2017-06-28
Automated smoke-test for My Stuff & Signing In


jwzimmer:
- Test My Stuff
  - Page structure
  - Project link
  - See Inside button
  - Add to studio button
- Test signing in in scratchr2
- Test signing in in scratch-www

PR #1411, opened 2017-07-12
Fix Gh-1385 - Issue/smoke test signing in

jwzimmer:
It works! Fixes https://github.com/LLK/scratch-www/issues/1385. But.

Things I'm not sure about:
1. passing in the credentials from another file
2. assuming existence of projects, studios, and shared projects
  a. To address this we could make a test user solely for this on Staging & Production, but then where to put the credentials? Last pass and everyone makes their own un-checked-in credentials file?
3. using eval & window
4. we should refactor the helper functions rather than including them in multiple files; I could do that as a separate pr since it involves changing other tests as well as these? or I could do that here?
5. organizing it into two files, (a) my stuff + signing in on scratchr2 and (b) signing in on scratch-www. I did it that way b/c I have to sign in to get to My Stuff anyway, and My Stuff is in scratchr2, so I figured testing the sign in there made sense. But on the other hand, maybe it should be it's own file and get run before the My Stuff test? Would that involve passing the driver from one tap test file to another in order to not re-set the starting point of the browser...?

### DIFF
--- a/test/integration/smoke-testing/test_signing_in_and_my_stuff.js
+++ b/test/integration/smoke-testing/test_signing_in_and_my_stuff.js
@@ -9,15 +9,8 @@
  *  - on a variety of instances (local, staging, prod)?
  */
 
-var path = require('path');
-var fs = require('fs');
-
-var dirPath = path.resolve(__dirname, '../not-tracked/');
-var filePath = dirPath + '/credentials.js';
-var credentials = fs.readFileSync(filePath,'utf8');
-
-var window = {};
-var credentials = eval(credentials);
+var username = process.env.USERNAME;
+var password = process.env.PASSWORD;
 
 var tap = require('tap');
 const test = tap.test;
@@ -45,7 +38,7 @@ const clickButton = (text) => {
     return clickXpath(`//button[contains(text(), '${text}')]`);
 };
 
-var rootUrl = process.env.ROOT_URL || 'https://scratch.ly';
+var rootUrl = process.env.ROOT_URL || 'https://scratch.mit.edu';
 var url = rootUrl + '/users/anyuser';
 
 tap.plan(5);
@@ -65,7 +58,7 @@ test('Sign in to Scratch using scratchr2 navbar', t => {
     .then(() => findByXpath('//input[@name="password"]'))
     .then((element) => element.sendKeys(password))
     .then(() => clickButton('Sign in'))
-    .then(() => findByXpath('//li[contains(@class, "logged-in-user")' 
+    .then(() => findByXpath('//li[contains(@class, "logged-in-user")'
         + 'and contains(@class, "dropdown")]/span'))
     .then((element) => element.getText('span'))
     .then((text) => t.match(text.toLowerCase(), username.substring(0,10).toLowerCase(),
@@ -135,4 +128,3 @@ test('Add To button should bring up a list of studios', t => {
     })
     .then(() => t.end());
 });
-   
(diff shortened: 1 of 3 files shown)


jwzimmer:
(Explanation of labels: I labelled it "pr - needs work" as well as "pr - needs review" because I expect to make changes after reviewing it with someone else. : ) )

test/integration/smoke-testing/test_signing_in_and_my_stuff.js
```diff
@@ -0,0 +1,138 @@
...
+
+var path = require('path');
+var fs = require('fs');
+
+var dirPath = path.resolve(__dirname, '../not-tracked/');
+var filePath = dirPath + '/credentials.js';
+var credentials = fs.readFileSync(filePath,'utf8');
+
+var window = {};
+var credentials = eval(credentials);
```
rschamp: Discussed offline: we should move these to environment variables

test/integration/smoke-testing/test_signing_in_and_my_stuff.js
```diff
@@ -0,0 +1,138 @@
...
+ */
+
+var path = require('path');
+var fs = require('fs');
+
+var dirPath = path.resolve(__dirname, '../not-tracked/');
+var filePath = dirPath + '/credentials.js';
+var credentials = fs.readFileSync(filePath,'utf8');
+
+var window = {};
```
rschamp: Also from offline: This can be removed

jwzimmer:
As discussed with @rschamp, now the username/ password come from the command line rather than reading them in from a separate, uncommitted file. 

I added users that anybody can use for testing on Production/ Staging in our shared passwords list, called "Production Smoke Test user" and "Staging Smoke Test user".

Run the programs like:
```
$ USERNAME=username PASSWORD=password node ./test/integration/smoke-testing/test_signing_in_homepage.js
```
and 
```
$ USERNAME=username PASSWORD=password node ./test/integration/smoke-testing/test_signing_in_and_my_stuff.js
```

test/integration/smoke-testing/test_signing_in_and_my_stuff.js
```diff
@@ -0,0 +1,130 @@
...
+ * https://github.com/LLK/scratchr2/wiki/Smoke-Testing-Test-Cases
+ *
+ * Note that I am using a user with at least 1 project and at least 1 studio
+ *  - I'm not sure how to handle this more elegantly... b/c I don't want to commit a password?
+ *  - I could supply a password for a Staging user I guess... but then people may want to test
+ *  - on a variety of instances (local, staging, prod)?
+ */
+
+var username = process.env.USERNAME;
+var password = process.env.PASSWORD;
```
rschamp: Since these are going to be global environment variables on things like Travis or any other host we use for smoke testing, could you name them something more specific? E.g., `SMOKE_USERNAME/PASSWORD`? Also would you please document them in the README, in the Configuration section? https://github.com/LLK/scratch-www#configuration

test/integration/smoke-testing/test_signing_in_and_my_stuff.js
```diff
@@ -0,0 +1,130 @@
+/*
+ * Tests signing in & My Stuff according to smoke-tests at:
+ *
+ * https://github.com/LLK/scratchr2/wiki/Smoke-Testing-Test-Cases
+ *
+ * Note that I am using a user with at least 1 project and at least 1 studio
+ *  - I'm not sure how to handle this more elegantly... b/c I don't want to commit a password?
+ *  - I could supply a password for a Staging user I guess... but then people may want to test
+ *  - on a variety of instances (local, staging, prod)?
```
rschamp: You can delete some of this since it's resolved I think.

PR #1413, opened 2017-07-13
[Master] Remove SmartyStreets API

rschamp:
We are so lax with this that it's not actually helping us, and we don't have a plan with them anymore.


mewtaylor:
Looks like the one thing left is just handling a lint error since we no longer use log (since it was only used in the smarty streets call) @rschamp

rschamp:
Thanks @mewtaylor, just fixed that and removed the address validation error message from l10n since it's no longer used.

rschamp:
RIP best named API we used.

PR #1416, opened 2017-07-13
Move custom homepage row to new api endpoints

mewtaylor:
This goes along with https://github.com/LLK/scratch-api/pull/292. This can be updated if we choose to migrate the `lovedByFollowing` endpoint off of proxy as well.

### Test Cases ###
* `Projects Shared by Scratchers I'm Following` should load projects when you are signed in on the splash page
* `Projects Loved by Scratchers I'm Following` should load projects when you are signed in on the splash page
* `Projects in Studios I'm Following` should load projects when you are signed in on the splash page
* None of the above rows should load if you are logged out on the splash page

### DIFF
--- a/src/views/splash/splash.jsx
+++ b/src/views/splash/splash.jsx
@@ -40,7 +40,7 @@ var Splash = injectIntl(React.createClass({
                 this.getActivity(this.props.user.username);
                 this.getSharedByFollowing(this.props.user.token);
                 this.getInStudiosFollowing(this.props.user.token);
-                this.getLovedByFollowing(this.props.user.id);
+                this.getLovedByFollowing(this.props.user.token);
                 this.getNews();
             } else {
                 this.setState({sharedByFollowing: []});
@@ -63,7 +63,7 @@ var Splash = injectIntl(React.createClass({
             this.getActivity(this.props.user.username);
             this.getSharedByFollowing(this.props.user.token);
             this.getInStudiosFollowing(this.props.user.token);
-            this.getLovedByFollowing(this.props.user.id);
+            this.getLovedByFollowing(this.props.user.token);
             this.getNews();
         } else {
             this.getProjectCount();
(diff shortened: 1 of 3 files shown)


src/views/splash/splash.jsx
```diff
@@ -106,12 +114,33 @@ var Splash = injectIntl(React.createClass({
             if (!err) return this.setState({featuredGlobal: body});
         }.bind(this));
     },
-    getFeaturedCustom: function () {
+    getSharedByFollowing: function () {
         api({
-            uri: '/proxy/users/' + this.props.session.session.user.id + '/featured'
+            uri: '/projects/following/users',
+            authentication: this.props.session.session.user.token
```
liam4: I'm having trouble figuring out what the value of this is, skimming through the code of this PR and around scratch-www (it looks like this is the first usage of the "authentication" `api` option), and I didn't get any clear results when testing on api-staging with header `X-token` and query `?x-token=` both equal to either the session token or the CSRF token. So I'm curious what the value of this would be?

mewtaylor: You probably didn't get the information you were looking for because the tokens you are referring to are not the ones that will be used for authentication of this endpoint – rather, the token will eventually be retrieved from your session, though that change is not yet on the website.

Also, api-staging is an internal endpoint that we use for testing and is not reflective of the information that is currently on Scratch.

mewtaylor: Oh also I forgot to mention @liam4 – the value is not going to be seen here. This is part of an infrastructure change on the backend to how we handle these rows that we're doing to increase their efficiency, and make sure they're not taking up a lot of resources to be retrieved.

mewtaylor:
Thanks @rschamp ! I split the splash file into container and presentation components, as well as added arguments to those methods as you suggested. What do you think of the split as it is now, no redux?

rschamp:
Thanks for doing that @mewtaylor!

mewtaylor:
@rschamp just FYI I now changed the loved by following row to use the new api endpoint

PR #1421, opened 2017-07-19
Fixes Down Deep Camp Counselors Link

St19Galla:
Just removes the extra `/`


mewtaylor:
On Mac OSX 10.12, Chrome 59:
* Clicking on the Camp counselors studio results in a correct navigation to that studio

PR #1427, opened 2017-07-21
Alphabetize Routes List v2

TheGrits:
Rebase and revamp of #1272 

@mewtaylor

### DIFF
--- a/src/routes.json
+++ b/src/routes.json
@@ -93,12 +93,6 @@
         "title": "DMCA"
     },
     {
-        "name": "donate",
-        "pattern": "^/info/donate/?",
-        "routeAlias": "/info/(cards|communityblocks-interviews|credits|faq|donate)/?$",
-        "redirect": "https://secure.donationpay.org/scratchfoundation/"
-    },
-    {
         "name": "download",
         "pattern": "^/download/?$",
         "routeAlias": "/download",
@@ -258,6 +252,12 @@
         "routeAlias": "/wedo/?$",
         "view": "wedo2/wedo2",
         "title": "LEGO WeDo 2.0"
+    },
+    {
+        "name": "donate-redirect",
+        "pattern": "^/info/donate/?",
+        "routeAlias": "/info/(cards|communityblocks-interviews|credits|faq|donate)/?$",
+        "redirect": "https://secure.donationpay.org/scratchfoundation/"
     },
     {
         "name": "download-redirect",
--- a/src/routes.json
+++ b/src/routes.json
@@ -134,18 +134,6 @@
         "title": "Scratch Community Guidelines"
     },
     {
-        "name": "help",
-        "pattern": "^/help/?(\\?.*)?$",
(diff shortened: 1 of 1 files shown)


src/routes.json
```diff
@@ -247,33 +258,40 @@
         "title": "Scratch Terms of Use"
     },
     {
+        "name": "things-to-try",
+        "pattern": "^/go/?(\\?.*)?$",
+        "routeAlias": "/go/?\\??",
+        "redirect": "/tips"
+    },
```
chrisgarrity: Since we replaced the Things to Try page with Tips, this route is now a redirect. It should probably be with the redirects (and probably renamed things-to-try-redirect). There are a couple of others that have also been converted into redirects `hoc`, `help`. @mewtaylor do you think we should rename them `-redirect` and move them to the redirects section?

mewtaylor: +1 for putting them into the redirect section, and I'm cool with renaming if that's helpful to include `-redirect` but am also fine with keeping them as they are too.

TheGrits:
@chrisgarrity @mewtaylor I've made the changes as discussed in the review!

src/routes.json
```diff
@@ -220,6 +148,41 @@
...
+    {
+        "name": "microworld-soccer",
+        "pattern": "^/microworlds/soccer",
+        "routeAlias": "/microworlds",
+        "view": "microworld/soccer/soccer",
+        "title": "Soccer"
+    },
+    {
+        "name": "microworlds-homepage",
```
mewtaylor: should this go above `microworlds-soccer`? I don't feel strongly about it, just checking!

src/routes.json
```diff
@@ -98,11 +93,17 @@
         "title": "DMCA"
     },
     {
-        "name": "camp",
-        "pattern": "^/camp/?$",
-        "routeAlias": "/camp/?$",
-        "view": "camp/camp",
-        "title": "Down Deep"
+        "name": "donate",
```
mewtaylor: I think that, if `store` is gonna be in the redirect, this probably should be too. Or there's the option of moving `store` into the main views list. I'd be fine with either option, whatever you think is best @TheGrits !

chrisgarrity:
Waiting to merge until after deploy.

src/routes.json
chrisgarrity: Are we alphabetizing by name or by route? 
The name - microworlds-homepage (with 's') would go after microworld-soccer, and that's how I was reading it. So I thought it was in the correct order :)

src/routes.json
TheGrits: I agree, this should be in the redirect section! Thanks!

src/routes.json
TheGrits: @mewtaylor My thought process behind this was exactly what @chrisgarrity said, I thought microworld**s** should be after microworld.

mewtaylor: 👍

TheGrits:
@mewtaylor I've made the change requested!

PR #1430, opened 2017-07-21
[Master] Hotfix: Edits to the scratch camp page

mewtaylor:


mewtaylor:
marking `qa-ok` as @jwzimmer took a look on staging

PR #1482, opened 2017-08-23
Correct “WeDo” case

### LINKED ISSUE #1481: WeDo should be camel case
https://github.com/LLK/scratch-www/blob/6309c306675c69b1b42b67c3b68a39dd720dfa58/src/views/wedo2/l10n.json#L13

The English source strings for the WeDo setup (1.0 and 2.0) should be WeDo instead of wedo.

Kenny2github:
### Resolves
#1481 
### Proposed Changes
Change “Wedo” to “WeDo” in two places
/cc @chrisgarrity


Kenny2github:
@chrisgarrity np! Case is a small change, but it's still a change :)

PR #1514, opened 2017-09-07
Fixes GH-1435/Reduce work for translators on credits page

### LINKED ISSUE #1435: Credits Page: Reduce work for translators
On the credits page, I see huge parts of the text that doesn't need to be translated: Many, many names! However, it still needs to be "translated" via transifex for any changes to be visible! 
So - isn't it possible to store all the names (as far as possible) seperately and to just translate the normal text? 
Updates would be visible more quickly, and it saves translators redundant copy-pasting!

LiFaytheGoblin:
Fixes #1435 

I made the changes @chrisgarrity suggested for the credits page and made as many names as possible non-translate, so translators don't need to retranslate all just because a new name has joined the list. 
I tested it in Firefox and Chromium, with the language English.

For other languages, I go repeating names for some reason, although I did do "make translations". Someone knows why this is?


thisandagain:
Thanks @LiFaytheGoblin ! One issue: it looks like your line length is too long in one place. You can look at the logs from Travis-CI (look for the little red "X" above) to see what the exact error output is. You can also run these tests locally by running `npm test`.

LiFaytheGoblin:
I fixed the problem @thisandagain

PR #1532, opened 2017-09-11
[Master] Hotfix: highlight unread message row in light blue

mewtaylor:
This addresses issues with seeing unread notifications by highlighting them as light blue, along with more opaque icons. 

Here's an example of what it looks like:
screen shot 2017-09-11 at 9 53 17 am

### DIFF
--- a/src/components/social-message/social-message.scss
+++ b/src/components/social-message/social-message.scss
@@ -12,6 +12,10 @@
     opacity: .25;
 }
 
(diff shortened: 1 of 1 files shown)


carljbowman:
Looks good to me! Thanks for making this change!

thisandagain:
Excellent! Thanks @mewtaylor and @carljbowman

PR #1551, opened 2017-09-15
Fix GH-1540: Direct studio activity link to the activity tab

### LINKED ISSUE #1540: "There was new activity in (the studio)" (on Messages page) should link directly to the studio's Activity page
A suggestion from [here](https://scratch.mit.edu/discuss/topic/276191/).

The reasoning is that when there's new activity in a studio, the user is most interested in what that activity is - which can be found out by going to the Activity tab on the studio page. That could be done automatically - just link to the activity tab (which looks to be how it was done before) - but instead, currently, the user has to go to that tab by hand.

mewtaylor:
Fixes #1540 by ensuring that the anchor link for studio activity is directed to the activity tab of the studio. Thanks @towerofnix for bringing this over to github from the suggestions forum!

### DIFF
--- a/src/views/messages/message-rows/studio-activity.jsx
+++ b/src/views/messages/message-rows/studio-activity.jsx
@@ -12,7 +12,7 @@ var StudioActivityMessage = React.createClass({
         datetimeCreated: React.PropTypes.string.isRequired
     },
     render: function () {
-        var studioLink = '/studios/' + this.props.studioId;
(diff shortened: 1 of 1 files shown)


mewtaylor:
On Chrome 60, Mac OSX 10.12:
* Logged in, when to messages. Clicked on studio activity for studio `blahblah` on staging, it took me to the `/activity` endpoint. Same for studio `test3`

Issue #1021, opened 2016-11-16
Setting interface language on About and Explore pages does not persist


chrisgarrity:
This happens on Production & Staging.

Test case:
- on the splash page change your language to Spanish
- Go to Explore page - content should be in Spanish,
- Switch language to something else, e.g., Greek, Explore page is translated to Greek,
- Go to Splash page, content should be in Greek, but has reverted to Spanish.
- Switch language to German, Splash page is in German,
- Go to Explore page, should be in German, it is still in Greek

jwzimmer:
I ran into this when going from the Explore page to Discuss; I changed the language in Explore, but when I went to Discuss, it was in the language I had previously been using.

jwzimmer:
I noticed this on Staging with Farsi, going from the Download Offline Editor page to the Scratch Terms of Use (the language switched back to English on the ToU page).

jwzimmer:
Make sure the repro steps/ test case described in https://github.com/LLK/scratch-www/issues/1199 is fixed when this issue is closed. 👍

WorldLanguages:
@thisandagain I think this was fixed with #1553, should this be closed?

thisandagain:
Resolved in https://github.com/LLK/scratch-www/pull/1553

PR #1553, opened 2017-09-18
GH-1455: apply path explicitly rather than in `defaults`

### LINKED ISSUE #1455: Language change in Explore not applied to all pages
**Steps to reproduce:**

I'll consider you're using English as your Scratch language.

1. Go to the Explore page
2. Scroll into the footer and change the language to any other (French, for example)

**Result:**
All pages except Explore will be in English. Explore will be in French.

mewtaylor:
This should fix #1455 by preventing the path from being changes via opts passed in, even though we don’t pass in a `path` anywhere.

### DIFF
--- a/src/lib/jar.js
+++ b/src/lib/jar.js
@@ -72,9 +72,9 @@ var Jar = {
     },
     set: function (name, value, opts) {
         defaults(opts, {
-            expires: new Date(new Date().setYear(new Date().getFullYear() + 1)),
-            path: '/'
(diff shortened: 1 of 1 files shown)


jwzimmer:
### Testing

- Chrome 61.0.3163.100, Mac OS El Capitan, Staging
- I cannot repro the issue following the steps described in #1021 
- When I set the language to various languages from the Explore page, opening new pages from that page persists the language (from footer links or navbar links)
- Same is true from the About page

Issue #1548, opened 2017-09-14
Release Offline Editor for PT-BR


thisandagain:
### Description
We are testing out a way to package tips content with the offline editor for languages other than English. For this first test, we would like to publish an offline editor build for Brazilian Portuguese before the Scratch Conference in São Paulo on October 5th. Because of this deadline all work must be completed by September 20th in order to hit the freeze / release cycle.

### Scope
- [x] Resolve remaining issues with PT-BR assets ( @leoburd )
    - [x] Resolve improperly scaled images
      - Example: https://cdn.scratch.mit.edu/scratchr2/static/help/pt-br/blocks/move-steps.html
    - [x] Resolve issues with improperly translated strings (many of these are due to LLK/scratchr2#4765)
      - Example: https://cdn.scratch.mit.edu/scratchr2/static/help/pt-br/howto/get-started-6.html
    - [ ] Resolve "blurry" animations **omitting for now**
      - Example: https://cdn.scratch.mit.edu/scratchr2/static/help/pt-br/howto/pgametip10.html
      - Compare to: https://cdn.scratch.mit.edu/scratchr2/static/help/en/howto/pgametip10.html
- [ ] Generate PT-BR offline editor build ( @cwillisf )
- [ ] Get final PT-BR offline editor download links ( @cwillisf )
    - [ ] Windows
    - [ ] MacOS
- [x] Update download links in the `/download` view for the `pt-br` locale ( @chrisgarrity )

towerofnix:
It might be nice if the Download page displayed that the offline editor is "also available in other languages"? (That's certainly not a blocker, though.)

cwillisf:
Updated with examples of each class of issue

cwillisf:
@towerofnix I think that's a good idea once we get past this test phase -- that is, once we have >2 languages ;)

leoburd:
Would it be possible to also generate a version of the offline editor for Linux? Most places in Brazil use either Windows or Linux boxes.

cwillisf:
@leoburd: the normal build process for the offline editor generates 3 builds: one for Windows, one for Mac, and one "cross-platform" build that can be used on older Mac machines, Linux, and other systems.

The cross-platform build is missing some features, though, such as hardware-accelerated graphics effects and hardware extensions. It's also a bit of a pain to run on Linux, since you need to install a very old version of Adobe AIR (version 2.6) -- that's the last version of AIR that Adobe ever shipped for Linux.

[...]

PR #1558, opened 2017-09-19
Link to pt-br downloads if locale is pt-br

chrisgarrity:
Assumptions per discussion with @cwillisf:
- there will be a `pt-br` subdirectory on `/scratchr2/static/sa` for the pt-br specific downloads 
- there will be a `version.xml` file in the pt-br that contains the current version info for the pt-br editor

### Testing ###
see #1548 
- [ ] Download links should be /scratchr2/static/sa/pt-br/* if language selected is pt-br
- [ ] Download links should be /scratchr2/static/sa/* if language selected is any other language

### DIFF
--- a/src/views/download/download.jsx
+++ b/src/views/download/download.jsx
@@ -14,7 +14,7 @@ var SubNavigation = require('../../components/subnavigation/subnavigation.jsx');
 require('./download.scss');
 require('../../components/forms/button.scss');
 
-var Download = React.createClass({
+var Download = injectIntl(React.createClass({
     type: 'Download',
     getInitialState: function () {
         return {
@@ -240,7 +240,6 @@ var Download = React.createClass({
             </div>
         );
     }
-});
+}));
 
-var IntlDownload = injectIntl(Download);
-render(<Page><IntlDownload /></Page>, document.getElementById('app'));
+render(<Page><Download /></Page>, document.getElementById('app'));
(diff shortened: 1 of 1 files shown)


src/views/download/download.jsx
```diff
@@ -233,4 +242,5 @@ var Download = React.createClass({
     }
 });
 
-render(<Page><Download /></Page>, document.getElementById('app'));
+var IntlDownload = injectIntl(Download);
```
mewtaylor: In other views on www, we tend to wrap `React.creatClass({})` with `injectIntl`, like this:

```javascript
var Download = injectIntl(React.createClass({
  // stuff and things
}));
```
I think it might make sense to do that here too to maintain convention, at least for now for consistency.

chrisgarrity: +1

mewtaylor:
On Chrome 60, Mac OSX 10.12:
* Visit `/downloads`, see the download link as `/scratchr2/static/sa/[file]`, which means they are correctly appearing in English
* Visit `/downloads` in Spanish, see the download link as `/scratchr2/static/sa/[file]`, which means they are correctly appearing in English
* Visit `/downloads` in Portuguese, see the download link as `/scratchr2/static/sa/[file]`, which means they are correctly appearing in English
* Visit `/downloads` in PT-BR, see the download link as `/scratchr2/static/sa/pt-br/[file]`, which means they are correctly appearing in pt-br

PR #1559, opened 2017-09-19
Fix GH-1545: add `unread` as a key/value on each message

### LINKED ISSUE #1545: New messages highlighting applies to top N messages regardless of filter setting
Unread messages on the Messages page have highlighting applied to them to distinguish them from older messages. If you have N notifications (according to the counter), then view the Messages page, the top N will be highlighted. 

To repro, switch the filter from All Activity to any other category, and note that the top N will still be highlighted, even though those may not be the same unread/ new messages.

screen shot 2017-09-13 at 2 16 50 pm

mewtaylor:
This fixes #1545 by adding unread as a key/value on each message during the filtering process. This means the unread count is always applied to the list of full messages, and never to a subset (as was happening before to cause #1545).

### Test Cases ###
(See #1545 for repros)
* Visit your messages page with unread messages. Highlighting should only apply to the top `n` rows on the messages page when the `all activity` filter is chosen
  * Choose another filter. If a row is highlighted in `all activity` and is present in that filter too, it should appear highlighted. If a row is not highlighted in `all activity` and is present in that filter too, is should **not** appear highlighted

### DIFF
--- a/src/views/messages/container.jsx
+++ b/src/views/messages/container.jsx
@@ -116,10 +116,20 @@ var Messages = React.createClass({
             )
         );
     },
-    filterMessages: function (messages, typesAllowed) {
+    filterMessages: function (messages, typesAllowed, unreadCount) {
         var filteredMessages = [];
         for (var i in messages) {
-            if (typesAllowed.indexOf(messages[i].type) > -1) {
+            // check to see if the position of the message in the list is earlier
+            // than the unread count. If it is, then the message is totally unread.
+            messages[i].unread = false;
+            if (i < unreadCount) messages[i].unread = true;
+
+            if (typesAllowed.length > 0) {
+                if (typesAllowed.indexOf(messages[i].type) > -1) {
+                    filteredMessages.push(messages[i]);
+                }
+            } else {
+                // if empty, then we're looking at all messages, so just like add the message
                 filteredMessages.push(messages[i]);
             }
         }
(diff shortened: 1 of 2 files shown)


src/views/messages/container.jsx
```diff
@@ -116,10 +116,20 @@ var Messages = React.createClass({
...
+            // than the unread count. If it is, then the message is totally unread.
+            messages[i].unread = false;
+            if (i < unreadCount) messages[i].unread = true;
+
+            if (typesAllowed.length > 0) {
+                if (typesAllowed.indexOf(messages[i].type) > -1) {
+                    filteredMessages.push(messages[i]);
+                }
+            } else {
+                // if empty, then we're looking at all messages, so just like add the message
```
chrisgarrity: I wanted to figure out if there was a way not to iterate over the whole list if typesAllowed is empty. Would it work to copy messages to filteredMessages, and then just set  `filteredMessages[i].unread = true` for the first unreadCount ones. Then the mod-unread class could be set based on whether unread is defined and true: `var className = (typeof message.unread === 'boolean' && message.unread) ? ...`

src/views/messages/presentation.jsx
```diff
@@ -140,14 +140,10 @@ var SocialMessagesList = React.createClass({
             />;
         }
     },
-    renderSocialMessages: function (messages, unreadCount) {
+    renderSocialMessages: function (messages) {
```
chrisgarrity: It's nice that this simplifies the rest of the code and unreadCount doesn't get passed around everywhere 👍

jwzimmer:
### Testing

Chrome 61.0.3163.100, Mac OS El Capitan, Staging
- I checked that a single comment notification shows up highlighted in the relevant categories, and nothing is highlighted otherwise
- I checked that two project activity notifications (love, fave) show up highlighted in the relevant categories, and nothing is highlighted otherwise
- Refreshing the page clears the highlight effect for those messages (they're now read)

PR #1562, opened 2017-09-21
[Master] Release 2.2.25

rschamp:


VutonDesign:
Looks fine to me 👍

Kenny2github:
This seems way past milestone - will this be merged or will it have its milestone pushed back?

rschamp:
@Kenny2github The milestone refers to the date when we stop pushing new features into the release. When we merge and release depends on when everything gets tested. This is an internal date that we use across many projects and repositories, public and private, so even if this public repository looks ready to go, we won't release until everything else is.

PR #1565, opened 2017-09-23
Highlight follow-user user

### LINKED ISSUE #1563: In "(user) is now following you" message rows, (user) should be a highlighted link
Here's what my messages inbox looks like:

"Minty-And-Fresh is now following you"

Note that links to forum topics are highlighted ("Snap! Team development discussion, vol. 2" is highlighted, for example), but links to users just following you aren't ("Minty-And-Fresh" isn't highlighted). So it's not as obvious that I can click on the username to see who followed me, and it's less likely for me to notice the message at all, because it doesn't have a contrasting link in it.

Kenny2github:
### Resolves
Resolves #1563
### Proposed Changes
Undefined for now...
### Reason for Changes
See #1563

### DIFF
--- a/src/views/messages/message-rows/follow-user.jsx
+++ b/src/views/messages/message-rows/follow-user.jsx
@@ -29,7 +29,6 @@ var FollowUserMessage = React.createClass({
                     values={{
                         profileLink: <a
                             href={profileLink}
-                            className="social-messages-profile-link"
                         >
                             {this.props.followerUsername}
                         </a>
--- a/src/views/messages/message-rows/follow-user.jsx
+++ b/src/views/messages/message-rows/follow-user.jsx
@@ -29,7 +29,7 @@ var FollowUserMessage = React.createClass({
                     values={{
                         profileLink: <a
                             href={profileLink}
-                            class="social-messages-profile-link"
+                            className="social-messages-profile-link"
                         >
                             {this.props.followerUsername}
                         </a>
--- a/src/views/messages/message-rows/follow-user.jsx
+++ b/src/views/messages/message-rows/follow-user.jsx
@@ -29,6 +29,7 @@ var FollowUserMessage = React.createClass({
                     values={{
                         profileLink: <a
                             href={profileLink}
+                            class="social-messages-profile-link"
                         >
                             {this.props.followerUsername}
                         </a>
--- a/src/components/social-message/social-message.scss
+++ b/src/components/social-message/social-message.scss
@@ -38,10 +38,10 @@
 }
 
 a.social-messages-profile-link {
-    color: $type-gray;
+    color: $ui-blue;
 
     &:hover {
-        color: darken($type-gray, 10);
+        color: darken($ui-blue, 15);
     }
 }
 


src/views/messages/message-rows/follow-user.jsx
```diff
@@ -29,7 +29,6 @@ var FollowUserMessage = React.createClass({
                     values={{
                         profileLink: <a
                             href={profileLink}
-                            className="social-messages-profile-link"
```
towerofnix: I wonder if removing this class is the best solution? After all, this *is* still a profile link. Maybe having a separate class - something like `.social-messages-focus-link` (on all links that are meant to be "focused", such as who followed you, or a link to a project or studio) - would be better.

Kenny2github: I was wondering the same thing, but (as stated in the OP) I noticed that it was *specifically that class* that made it look like normal text instead of a link - suggesting that it was originally meant to be that way. I could just change the style of that class - but I can't seem to find the root style for links on the messages page. I guess I'll work on that then.

src/views/messages/message-rows/follow-user.jsx
```diff
@@ -29,7 +29,7 @@ var FollowUserMessage = React.createClass({
...
+                            class="social-messages-profile-link"
```
mewtaylor: @Kenny2github just an FYI – `className` is a [convention of ReactJS](https://facebook.github.io/react/docs/dom-elements.html#classname), so switching to use `class` is not ideal if there's a need for this class

Kenny2github: @mewtaylor I was intending to switch it back to `className` but my brain farted

Kenny2github:
@mewtaylor no problem!
I noticed (from your comment on the issue) that it's OS-specific. Unfortunately, I don't have access to any devices that would let me test this on other OSes...

The username could be a separate linked color? I'm not sure that gray is the best choice (my discussion contribution), I'd maybe go with a different shade of blue.

src/views/messages/message-rows/follow-user.jsx
mewtaylor: @Kenny2github lol I know the feeling :)

mewtaylor:
@Kenny2github thanks for that feedback – I can chat with @carljbowman about that and get back to you!

Kenny2github:
This is way past milestone - any consensus? Should I close this for now or keep it open?

mewtaylor:
@Kenny2github thanks for bringing this back up, and sorry for the delay! @carljbowman and I talked about this a bit, and I think we'd rather try and solve this issue by bolding text differently in a way that will work on multiple browsers rather than changing to having this be blue text. 

The reason for this is to make sure that the content of a notification – i.e. the project, studio, etc. – are still the focus of the row as the only thing highlighted in blue. If we also turn the username blue, the worry is that it would take away focus from the what the notification is about, and instead focus too much on who performed it. Does that make sense?

I will open a separate issue about this, but considering the fix for it will need to be more global, it might not be `help-wanted`.

Kenny2github:
@mewtaylor Yeah, I get what you mean. Guess this is resolved, then.

PR #1577, opened 2017-09-26
Change first about line

DeleteThisAcount:
I changed "With Scratch, you can program your own interactive stories, games, and animations — and share your creations with others in the online community." to "With Scratch, you can program your own interactive stories, games, animations and more — and share your creations with others in the online community."

Why?
I think that Scratch can be used for more than the things mentioned in the sentence..

### DIFF
--- a/src/views/about/l10n.json
+++ b/src/views/about/l10n.json
@@ -1,5 +1,5 @@
 {
-    "about.introOne": "With Scratch, you can program your own interactive stories, games, and animations — and share your creations with others in the online community.",
+    "about.introOne": "With Scratch, you can program your own interactive stories, games, animations and more — and share your creations with others in the online community.",
     "about.introTwo": "Scratch helps young people learn to think creatively, reason systematically, and work collaboratively — essential skills for life in the 21st century.",
     "about.introThree": "Scratch is a project of the Lifelong Kindergarten Group at the MIT Media Lab. It is provided free of charge.",
     "about.introParents": "Info for parents",


WorldLanguages:
I don't see this necessary, in my opinion. It never says it **only** can do stories, games and animations.

However I don't see why this should be rejected :)

DeleteThisAcount:
I agree but it can do more than the mentioned stuff.

towerofnix:
PRs/suggestions to change strings like this are totally helpful, but I'd like to note that it *might* be more rare for them to be merged/implemented, since the other Scratch site languages will have to be updated (otherwise, they'll lag behind; inconsistent translations are always a thing to watch out for).

VutonDesign:
@towerofnix Ninja'd
Anyway, This does seem a good idea but would cause a lot of hassle as that line would need to translated into over 70 different languages again. _(I think)_

mewtaylor:
@DeleteThisAcount this definitely brings up a good point about how we think about projects! You're absolutely right that people can create more. However, we most likely cannot merge this, for the reason that @towerofnix mentioned – it would affect translations for many languages.

In general, if you're looking for something to work on, try starting with bugs marked with a `help-wanted` label. Those are things we need someone to take a look at. You can see more about how we prefer people to handle bug-fixing [here](https://github.com/LLK/scratch-www/wiki/Assigning-Yourself-Bugs).

Also, with suggestions like this, it's helpful to open a bug first, so we can talk about that first before exploring fixes.

mewtaylor:
Oh also, as this is your first PR, thanks for getting started with the scratch-www repo!
confetti-horn-emoji

PR #1582, opened 2017-09-27
Added full stops on download page

DeleteThisAcount:
Adds full stops (.)s to the download page.  (Under 1 and 2)

Yes I know that it will need to be in all the languages. Its a really simple change. (I think)


thisandagain:
Thanks for all the suggestions. Again, we would prefer that these be submitted as issues rather than PRs so they can be discussed. If you'd like to submit a PR please do so for items that have the 'help-wanted' label or after discussing with a member of the Scratch Team. Even small changes like this add a lot of work for our translators.

DeleteThisAcount:
Okay sorry.

PR #1587, opened 2017-10-03
Update languages.json

Torotv:
Switch yum To Yumm To Make it should better ?

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -11,7 +11,7 @@
     "cy": "Cymraeg",
     "da": "Dansk",
     "de": "Deutsch",
-    "yum": "Edible Scratch",
+    "yumm": "Edible Scratch",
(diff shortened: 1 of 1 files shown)


thisandagain:
Hi @Torotv thanks for the PR! In the future please file an issue for discussion before submitting a PR. Changes to copy no matter how small need to be considered by the team before we will accept changes. If you are looking for issues to resolve please look for items that have the "help-wanted" label:
https://github.com/LLK/scratch-www/issues?q=is%3Aissue+is%3Aopen+label%3Ahelp-wanted

towerofnix:
Also, "yum" there is the (pretend) language code for Yum/Edible Scratch. It's best to not change language codes, because then translations might be broken.

PR #1590, opened 2017-10-03
[Develop] Hotfix: change message count endpoint

mewtaylor:


towerofnix:
I'm curious what the difference between the two is/was, and why the "proxy" endpoint was being used for a while?

mewtaylor:
@towerofnix one gets the count through the old system, and proxies to the new api. The new one doesn't – there shouldn't be any difference in terms of the value they return though, the difference is just a matter of how we're migrating over from our old system, and making sure we're doing so in a way that can handle load from the site well :).

towerofnix:
Ah yeah, handy!

> There shouldn't be any difference in terms of the value they return though...

I'm noticing that they're returning different values - [`/proxy/users/foo/activity/count`](https://api.scratch.mit.edu/proxy/users/foo/activity/count) is returning 0, while [`/users/foo/messages/count`](https://api.scratch.mit.edu/users/foo/messages/count) is returning 8. Probably not an issue since /proxy/../activity/count isn't being used anymore (as of this PR!), but still a little surprising?

Then again, it looks like just about nothing from the old API is operating at the moment; for example, [comments aren't loading](https://scratch.mit.edu/users/_nix/#comments).

mewtaylor:
@towerofnix I'm seeing comments at the link you provided, are you still not?

Also the difference you see is due to a bug that we're in the middle of resolving at the moment, which is why we haven't merged this change to master yet :).

towerofnix:
@mewtaylor Comments are working now. That was quick!

Alright, I see. Sounds good!

PR #1599, opened 2017-10-08
Center Download Page header

### LINKED ISSUE #1381: Offline editor download page: top paragraph not centered
When the download page is displayed in a wide window, the paragraph at the top of the page ("You can install the Scratch 2.0 editor....") is aligned left and wrapped to an unnecessarily narrow column width.

Note that this problem existed on the old page, but IMO stands out more on the new version of the page.

Screenshot:
image

GreenBayRules:
Fixes #1381 
image

### DIFF
--- a/src/views/download/download.scss
+++ b/src/views/download/download.scss
@@ -40,12 +40,17 @@ $developer-spot: $splash-blue;
             }
 
             .sub-nav {
-                text-align: left;
+                text-align: center;
                 justify-content: flex-start;
+                display: block;
 
                 li {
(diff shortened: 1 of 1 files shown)


GreenBayRules:
Hmmm, why did the travis build fail?

GreenBayRules:
It says this: 
image

GreenBayRules:
That should fix it

thisandagain:
See comments in #1381. I don't think this is quite ready for contribution yet. We need to make a decision regarding when to center and when to keep the left alignment.

PR #1616, opened 2017-10-11
Fix #1612

VutonDesign:
Updates the credits page as requested in #1612.

PR #1617, opened 2017-10-11
Fixes #1612 because of error in #1616 

### LINKED ISSUE #1612: Credits: Add New ST member
@GulpTea needs to be added to the credits (I'm not sure if this is too early to add them to credits, I mean they literally joined 4 days ago).
If this gets help-wanted, I can do this.

GreenBayRules:
I made another pr for the issue because the one from vutondesign did not work.

### DIFF
--- a/src/views/credits/credits.jsx
+++ b/src/views/credits/credits.jsx
@@ -64,7 +64,7 @@ var Credits = React.createClass({
 
                     <li>
                         <img src="//cdn.scratch.mit.edu/get_image/user/246290_170x170.png" alt="Sarah Avatar" />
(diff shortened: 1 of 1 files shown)


GreenBayRules:
okay hopefully travis build passes

VutonDesign:
@GreenBayRules We both passed, But you have 15 commits and I have 7. Should we just use mine?

mewtaylor:
@GreenBayRules it looks like @VutonDesign is still actively working on the original PR for this, so I think it's best to close this one unless @VutonDesign decides to stop working on #1616.

Issue #1569, opened 2017-09-25
Fix failing footer links automated tests


jwzimmer:
While running the automated smoke tests for the September milestone (https://github.com/LLK/scratchr2/issues/4774#issuecomment-331968494), I noticed that the footer link automated tests are now failing*. They should be fixed, since there doesn't seem to be anything wrong with the links themselves - it looks like the tests are the problem.

(*) They did not fail previously. The test for the Tips link still passes, but it looks like it is clicking the Tips in the navbar rather than footer (it should be forced to look in the footer).

jwzimmer:
Fixed in https://github.com/LLK/scratch-www/pull/1624.

PR #1624, opened 2017-10-24
Updated selenium & chromedriver versions, which fixed test failures i…

jwzimmer:
…n test_footer_links. Also fixed the clickFooterLinks function so it only checks links in the footer, to remove false positives (e.g. test finds 'Tips' in navbar and passes).

Fixes https://github.com/LLK/scratch-www/issues/1569.

Changes:
- Update selenium & chromedriver versions in package.json
  - I checked that all smoketests still pass with the updated versions (by running them with `make smoke`)
- Removed initial wait in the clickFooterLinks function; I still don't know if it mattered for the occasional timeout, but the tests passed without it, so I took it out
  - Relatedly, I increased the timeout to 30000 (from 20000) because I saw the `About` link hit the timeout even though that link is clearly working and there doesn't seem to be anything amiss with the test... so I just increased the limit a bit in the hopes that that wouldn't happen again (? bad idea ?)
- I changed the clickFooterLinks function so that the test only checks within the footer. 
  - When I initially noticed the failing tests, the Tips test was still passing, and what was happening was that the test found the Tips link in the navbar and accepted that, because I hadn't passed the found footer element through to the next function. Now that is being passed through.

### DIFF
--- a/test/integration/package.json
+++ b/test/integration/package.json
@@ -1,6 +1,6 @@
 {
     "testDependencies": {
-        "selenium-webdriver": "2.44.0",
-        "chromedriver": "2.27.0"
+        "selenium-webdriver": "2.45.0",
+        "chromedriver": "2.33.0"
     }
 }
--- a/test/integration/package.json
+++ b/test/integration/package.json
@@ -1,5 +1,5 @@
 {
-    "testDependencies": {
+    "devDependencies": {
         "selenium-webdriver": "2.45.0",
         "chromedriver": "2.33.0"
     }
--- a/test/integration/smoke-testing/test_footer_links.js
+++ b/test/integration/smoke-testing/test_footer_links.js
@@ -4,16 +4,15 @@
  * Test cases: https://github.com/LLK/scratch-www/wiki/Most-Important-Workflows
  */
  
-require('chromedriver');
 var tap = require('tap');
-var seleniumWebdriver = require('selenium-webdriver');
-
-// Selenium's promise driver will be deprecated, so we should not rely on it
-seleniumWebdriver.SELENIUM_PROMISE_MANAGER=0;
 
 const {
-    driver
+    driver,
+    webdriver
 } = require('../../helpers/selenium-helpers.js');
+
+// Selenium's promise driver will be deprecated, so we should not rely on it
+webdriver.SELENIUM_PROMISE_MANAGER=0;
 
 var rootUrl = process.env.ROOT_URL || 'https://scratch.ly';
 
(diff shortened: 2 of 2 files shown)


test/integration/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "testDependencies": {
-        "selenium-webdriver": "2.44.0",
-        "chromedriver": "2.27.0"
+        "selenium-webdriver": "2.45.0",
+        "chromedriver": "2.33.0"
```
rschamp: I just noticed these are in `testDependencies` which doesn't seem to be a real thing? Or is it? I would expect `devDependencies`.

jwzimmer: We didn't put them in devDependencies because we didn't want to force everyone to download them. But we did want to document them somewhere, so we put them in testDependencies.

rschamp: I think we're safe from people needing to install them by virtue of this being in a separate `package.json`. I just updated it to `devDependencies` and it didn't install them in the top-level `npm install`.

jwzimmer:
Trying to see why all the integration tests fail for @rschamp with an error like "element not clickable at (x, y)"...

### Running tests on a computer with no tests previously run (Philippines)
* Installed node, which came with npm
  * v8.9.1
* Installed Mac command line tools in order to use git
* Cloned via https link: `git clone https://github.com/LLK/scratch-www.git`
* `npm install tap`
* `npm install selenium-webdriver`
* `npm install chromedriver`
  * At this point, running the tests results in some passing, some failing (previously, the suite failed, which is expected b/c required dependencies had not been installed)
  * It looks like the files that build the driver like [this](https://github.com/LLK/scratch-www/blob/develop/test/integration/smoke-testing/test_statistics_page.js#L8) fail, but the files that build the driver like [this](https://github.com/LLK/scratch-www/blob/develop/test/integration/smoke-testing/test_project_rows.js#L16) pass
  * The failure error is `Error: The ChromeDriver could not be found on the current PATH. Please download the latest version of the ChromeDriver from http://chromedriver.storage.googleapis.com/index.html and ensure it can be found on your PATH.`
    * Loading `test/helpers/selenium-helpers.js` is in the stacktrace
* So I have something set-up incorrectly with where chromedriver is located/ how everything else can find it, but some of the tests run correctly and don't get the error about the element not being visible (which I have seen with the footer links tests, which are passing); so, that doesn't seem like an issue innate to the test suite currently in www > master
* I deleted the scratch-www repo
* I cloned via https the scratch-www repo from my fork of www
* I fetched contents of the branch `issue/add-signing-out`
* `npm install`
* Tests failed for not having selenium & chromedriver
* `npm install selenium-webdriver`
* `npm install chromedriver`
* At this point, I get the same set of pass/ fail as above, and the same error message
  * So it doesn't seem like the issue with the error like "element not clickable at (x, y)" was introduced by https://github.com/LLK/scratch-www/pull/1626 
* I fetched contents of the branch `issue/fix-failing-footer-links`
* At this point I get different tests failing, still for the ChromeDriver not being found on the current PATH, which makes sense b/c I switched which way test_footer_links.js was building the chromedriver in this PR (https://github.com/LLK/scratch-www/pull/1624).
* However, I am still seeing some tests pass, and not seeing the error like "element not clickable at (x, y)".

Issue #1625, opened 2017-10-24
Add signing out to automated tests


jwzimmer:
Re https://github.com/LLK/scratchr2/issues/4785, add functionality to sign out a user to the authentication smoke tests (or possibly make that a separate file for the r2 half, so it doesn't clutter up all the functionality along with My Stuff?).

PR #1626, opened 2017-10-24
Issue/add signing out

jwzimmer:
Fixes https://github.com/LLK/scratch-www/issues/1625.

- Added tests for signing in & out in scratchr2 & scratch-www
  - Renamed the test file `signing_in_homepage` to `signing_in_and_out_homepage`
  - Added a new file for signing in & out from the scratchr2 navbar
    - Note signing in from the scratchr2 navbar also occurs in the My Stuff test file, but I thought it would be better to duplicate that functionality in two files rather than establish some kind of order the smoke tests have to be run in, or to clutter up the My Stuff test file with additional functionality
- Added a findText helper function to selenium-helpers (/cc @paulkaplan, I added you as a reviewer just in case you had any objections to that, since you authored that file)


test/integration/smoke-testing/test_signing_in_and_out_discuss.js
```diff
@@ -0,0 +1,59 @@
...
+    clickText,
+    findByXpath,
+    findText,
+    clickXpath,
+    clickButton,
+    driver
+} = require('../../helpers/selenium-helpers.js');
+
+var username = process.env.SMOKE_USERNAME;
+var password = process.env.SMOKE_PASSWORD;
```
rschamp: Are these environment variables already documented somewhere? If not we should add them.

PR #1629, opened 2017-10-24
Fix GH-1546: Use new filters on api endpoint

### LINKED ISSUE #1546: Display "Load more" button on New Messages page when filter is applied and there are more messages
When a user does not have a filter on the kind of activity being displayed, then Load More will load more messages as long as there are messages to load. If there are no messages to load, the button will not be displayed (which is good).
However, if the user applies a filter, and there are no activities of that type in the amount of total messages that have been loaded on the page, then the page will display no messages, and will not display the Load More button. If there are more messages for that user, the Load More button should still be displayed.

This issue + https://github.com/LLK/scratch-www/issues/1499 together would account for the following situation:
- A user goes to their messages page
- The first N messages are loaded on the page
- The user switches the filter to Project Activity
- There are no project activity notifications in the first N messages
- Currently, the page displays no messages and no Load More button, but it should display the Load More button - then the user would click Load More
- The next N messages would be fetched, but only those matching the filter would be displayed on the page
- If there are no project activity notifications in the 2nd N batch of messages, then this would lead to the user seeing the same blank/ no messages page, and having to click Load More again (which is not obvious); this is where https://github.com/LLK/scratch-www/issues/1499 comes in, since instead the loading state could display (e.g. a spinning load symbol) to indicate that the Load More button click had worked.

I'm not totally sure what UX would be clear, but what we have now could definitely be made clearer. /cc @carljbowman (discussed in person with @mewtaylor)

As reported (with helpful screenshots!) in https://scratch.mit.edu/discuss/topic/276228/.

mewtaylor:
This fixes #1546 by using the new optional `filter` keyword argument that will be added to the messages api endpoint.

### Test Cases ###
* If you change filters, the notifications that show up should only apply to that type of filter
* If you change filters and there are no notifications available for that filter, the load more button shouldn't show up
* If you change filters and there were no notifications of that type in the original list of 40 notifications, you should still see the most recent of that notification type

### DIFF
--- a/src/views/messages/container.jsx
+++ b/src/views/messages/container.jsx
@@ -12,8 +12,7 @@ var Messages = React.createClass({
     type: 'ConnectedMessages',
     getInitialState: function () {
         return {
-            filterValues: [],
-            displayedMessages: []
+            filter: ''
         };
     },
     getDefaultProps: function () {
@@ -26,14 +25,17 @@ var Messages = React.createClass({
         };
     },
     componentDidUpdate: function (prevProps) {
-        if (this.props.user != prevProps.user) {
+        if (this.props.user.username !== prevProps.user.username) {
             if (this.props.user.token) {
                 this.props.dispatch(
                     messageActions.getMessages(
                         this.props.user.username,
                         this.props.user.token,
-                        this.props.messages,
-                        this.props.messageOffset
+                        {
+                            messages: this.props.messages,
+                            offset: this.props.messageOffset,
+                            filter: this.state.filter
+                        }
                     )
                 );
                 this.props.dispatch(
@@ -59,8 +61,11 @@ var Messages = React.createClass({
                 messageActions.getMessages(
                     this.props.user.username,
                     this.props.user.token,
-                    this.props.messages,
-                    this.props.messageOffset
+                    {
+                        messages: this.props.messages,
+                        offset: this.props.messageOffset,
+                        filter: this.state.filter
+                    }
                 )
             );
             this.props.dispatch(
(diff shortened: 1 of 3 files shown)


src/views/messages/container.jsx
```diff
@@ -26,14 +25,17 @@ var Messages = React.createClass({
         };
     },
     componentDidUpdate: function (prevProps) {
-        if (this.props.user != prevProps.user) {
+        if (this.props.user.username !== prevProps.user.username) {
```
chrisgarrity: Just wondering why you needed to add `.username`

mewtaylor: seemed quicker to check username (string) equality than user (object) equality, if that makes sense

jwzimmer:
### Testing

* [ ] If you change filters, the notifications that show up should only apply to that type of filter
* [ ] If you change filters and there are no notifications available for that filter, the load more button shouldn't show up
* [ ] If you change filters and there were no notifications of that type in the original list of 40 notifications, you should still see the most recent of that notification type

jwzimmer:
Marking this as `QA-OK` and `QA-post-deploy` since we should make sure remix notifications do show up on the Messages page post-deploy:
* [ ] Re https://github.com/LLK/scratch-api/pull/355, checking that various social action notifications do show up on the Messages page
  * [ ] Remix does not, but I think that is due to https://github.com/LLK/scratchr2/issues/4856#issuecomment-340919612 - @mewtaylor looked into it and saw no indication it was related to these PRs, but we should verify that the remix notification shows up post-deploy

jwzimmer:
Old remixes that are recently shared do generate a notification on the Messages page still, so I think that's a good indication that the current lack of remix notification is attributable to https://github.com/LLK/scratchr2/issues/4856#issuecomment-340919612.

screen shot 2017-11-01 at 2 42 53 pm

jwzimmer:
I am seeing remix notifications on the Messages page on Production 👍 (as well as on user profiles and the project page).

PR #1639, opened 2017-10-30
Point at GulpTea’s avatar in credits page

mewtaylor:
I noticed while doing qa for #1616 that @TheGrits avatar was pointing at the default avatar – I changed it to point at `GulpTea`'s avatar instead.


TheGrits:
You mean that @GulpTea's avatar was pointing to the default avatar? :wink:

PR #1667, opened 2017-11-14
update read me for automated  tests

jwzimmer:
I just updated the readme in the `/integration` test folder slightly to be more up to date. I added `ROOT_URL` to the table of args. I also explicitly listed the Staging URL - it's listed in the test files, but listing it in the readme might make it slightly more visible, so just wanted to bring that up in case we'd prefer not to list it in the readme...


jwzimmer:
When you have a chance, could you take a look at this @rschamp? No rush.

test/integration/README.md
```diff
@@ -4,23 +4,24 @@
   * See this directory's package.json
 * TAP
   * In the scratch-www repo's package.json
-* Currently, we also require that you download [Chromedriver](https://sites.google.com/a/chromium.org/chromedriver/), but this will be replaced by Saucelabs imminently
+* Currently, we also require that you download [Chromedriver](https://sites.google.com/a/chromium.org/chromedriver/), but this will be replaced by Saucelabs eventually
```
rschamp: Maybe we should just remove references to Sauce altogether, and we can just add it back when we have something? When we do get to it, we may not even end up using that service (although we probably will).

test/integration/README.md
```diff
@@ -4,23 +4,24 @@
...
 | Variable      		| Default               | Description                                 			    |
 | ---------------------	| --------------------- | --------------------------------------------------------- |
+| `ROOT_URL`			| `scratch.ly`			| Location you want to run the tests against
 | `SMOKE_USERNAME`    	| `None` 				| Username for Scratch user you're signing in with to test 	|
 | `SMOKE_PASSWORD`  	| `None`                | Password for Scratch user you're signing in with to test  |
 
 ## Using sauce
-* We're still working on setting this up; more info coming shortly
\ No newline at end of file
+* Not yet set up
```
rschamp: Same here, let's just delete this section.

test/integration/README.md
jwzimmer: Sure, will do.

jwzimmer:
Thanks, @rschamp!

PR #1671, opened 2017-11-15
Upgrade compatibility to node 8

mewtaylor:
This fixes an issue in compilation by upgrading node-sass to v4.6, and also upgrades express in the process.

Callback added to `writeFile` in `get-localized-urls` in order to avoid a deprecation warning during `make translations`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- '4.2'
+- '8'
 sudo: required
 cache:
   directories:
--- a/bin/get-localized-urls
+++ b/bin/get-localized-urls
@@ -36,7 +36,9 @@ https.get('https://resources.scratch.mit.edu/localized-urls.json', (res) => {
     res.on('end', () => {
         try {
             var urlJson = JSON.parse(urlData);
-            fs.writeFile(filename, JSON.stringify(urlJson, null, '  '));
+            fs.writeFile(filename, JSON.stringify(urlJson, null, '  '), (err) => {
+                if (err) process.stdout.write(`Failed writing file: ${err.message}\n`);
+            });
         } catch (e) {
             process.stdout.write(`Failed parsing url data: ${e.message}\n`);
             process.exit(1);
--- a/bin/get-localized-urls
+++ b/bin/get-localized-urls
@@ -36,9 +36,7 @@ https.get('https://resources.scratch.mit.edu/localized-urls.json', (res) => {
     res.on('end', () => {
         try {
             var urlJson = JSON.parse(urlData);
-            fs.writeFile(filename, JSON.stringify(urlJson, null, '  '), (err) => {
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- '4.2'
+- '8'
```
thisandagain: I recommend we also add simply `node` to this to make sure we are always testing against the latest stable release. https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Specifying-Node.js-versions

thisandagain:
```yml
node_js:
    '8'
    'node'
```

bin/get-localized-urls
```diff
@@ -36,7 +36,9 @@ https.get('https://resources.scratch.mit.edu/localized-urls.json', (res) => {
     res.on('end', () => {
         try {
             var urlJson = JSON.parse(urlData);
-            fs.writeFile(filename, JSON.stringify(urlJson, null, '  '));
+            fs.writeFile(filename, JSON.stringify(urlJson, null, '  '), (err) => {
```
thisandagain: Now that you are handling the error properly you can drop the try/catch.

.travis.yml
rschamp: I agree we should do this, but we will need to adjust the config so we only deploy only once. So I think we should do this in a separate PR.

bin/get-localized-urls
mewtaylor: @thisandagain I think the original intention of the `try/catch` was to handle any JSON parsing errors in the line above (and writeFile errors just weren't being seen). I think I should move the `writeFile` line out of the `try/catch`, but still keep it around

bin/get-localized-urls
```diff
@@ -36,7 +36,9 @@ https.get('https://resources.scratch.mit.edu/localized-urls.json', (res) => {
...
+                if (err) process.stdout.write(`Failed writing file: ${err.message}\n`);
```
thisandagain: The `catch` also exits the process with a non-zero exit code. Should this perform equivalently?

bin/get-localized-urls
thisandagain: Then I think for the sake of being terse, I would recommend you keep the existing functionality and simply use `fs.writeFileSync`.

rschamp:
@mewtaylor I don't know what QA we would be able to perform for this. Are there specific test cases you were thinking of?

.travis.yml
rschamp: Captured this in https://github.com/LLK/scratch-www/issues/1675

mewtaylor:
@rschamp goooood point – I think I probably didn't need to put that there, since simply being able to deploy to staging/see a site means it's working. I'll remove the tag!

Issue #1677, opened 2017-11-20
Update localization guide to include info on FormattedHTMLMessage workarounds


mewtaylor:
React-Intl has started to deprecate `FormattedHTMLMessage`, highly recommending that people switch to using `FormattedMessage`, and simply passing values into the component that contain any necessary html in them – see [here](https://github.com/yahoo/react-intl/wiki/API#formathtmlmessage).

We should update the wiki to include this in our style guide, and show workarounds for how contributors can avoid using `FormattedHTMLMessage` when adding localized strings to a page.

chrisgarrity:
@mewtaylor updated wiki section: https://github.com/LLK/scratch-www/wiki/Localization-Guide#removing-formattedhtmlmessage

mewtaylor:
@chrisgarrity looks awesome thanks!

Issue #1674, opened 2017-11-17
Post Learning Resource Designer on /jobs page


ChampikaF:
url: scratch.mit.edu/jobs

request:

Add "Learning Resource Designer, MIT Media Lab, Cambridge, MA" to list of jobs
Link: https://www.media.mit.edu/about/job-opportunities/learning-resources/

GreenBayRules:
I can do this :D

GreenBayRules:
#1676 Done :D

GreenBayRules:
Should be closed since #1676 was merged..

thisandagain:
Resolved in #1676

PR #1673, opened 2017-11-16
Fixes #1633 - Scratch Conference 2018 Website Phase 1

GreenBayRules:
Fixes #1633 
Adds the following:
New Copy (italics = instructions not copy):

Title: Scratch Conference 2018: The Next Generation (replaces Scratch Conferences 2017)
Sub-title: July 26-28, 2018 | Cambridge, MA, USA (replaces 'This year, in celebration...')

[Drop the lighter blue banner 'Learn more about conference dates and locations below.']

Join us for the Scratch@MIT conference, a playful gathering of educators, researchers, developers, and other members of the worldwide Scratch community.

We're planning a very participatory conference, with an entire day of hands-on workshops and lots of opportunities for peer-to-peer discussion and collaboration.

[use date and location styling from current site:]
When: July 26-28, 2018 (with opening reception the evening of July 25)
Where: MIT Media Lab, Cambridge, MA

Interested in offering a session? We invite four types of proposals:

Poster/demonstration (90 minutes). Show off your project in an exhibition setting, alongside other presenters. You will be provided with display space for a poster and table space for a computer or handouts.

Hands-on workshop (90 minutes). Engage participants in hands-on activities, highlighting new ways of creating and collaborating with Scratch.

Interactive panel (60 minutes). Discuss a Scratch-related topic in a panel with three or more people. Your proposal should describe how you will engage the audience during the session.

Ignite talk (5 minutes). Share what you've been doing in a short, lively presentation.

Button: Submit Your Proposal
[link to https://docs.google.com/forms/d/e/1FAIpQLSd7SkuQ-dfW-P3aArSQokK9GkKAUKufTVBHod_ElNIiFE9iBQ/viewform?usp=sf_link]

Deadline for proposals: February 5
Notification of acceptance: March 1

Registration:
Early Bird Registration (March 1-May 1): $200
Standard Registration (after May 1): $300

Questions? Contact the Scratch Team at conference@scratch.mit.edu

to the page.

PR #1679, opened 2017-11-21
[Master] December 1st Hotfix

mewtaylor:
Includes #1673 (thanks @GreenBayRules) and #1674. A couple of small changes from for the conference page PR include:
* changing back the `routeAlias` for the 2017 and 2018 pages – they need to be consistent with the other conference routes to ensure all routes >=2016 use this repo
* changing the localization to use only `FormattedMessage` – this is meant to be part of the standard which we'll eventually have in #1677 

/cc @chrisgarrity @thisandagain @ChampikaF

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -92,8 +92,8 @@ source_file = src/views/microworldshomepage/l10n.json
 source_lang = en
 type = KEYVALUEJSON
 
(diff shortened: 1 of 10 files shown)


thisandagain:
Looks good!

Issue #1238, opened 2017-03-06
Quotes in l10n.json files should all be escaped


chrisgarrity:
faq/l10n.json file has some strings where the quotes are not escaped.
For example,
`"faq.confirmedAccountTitle":"What is a “confirmed” Scratch account?",`
should be
`"faq.confirmedAccountTitle":"What is a \“confirmed\” Scratch account?",`

It doesn't turn out to be an issue until the translation has the quoted word at the beginning of the translation as in Korean:
`"faq.confirmedAccountTitle":""인증된" 스크래치 계정이란 무엇인가요?",`
The two double quotes cause a syntax error when parsing the file.

In addition to fixing all the places in the faq l10n file where quotes aren't escaped it would be good to check whether there are unescaped quotes in any other l10n files.

GreenBayRules:
Is this still an issue? If so I can make a pull request..

thisandagain:
@chrisgarrity ^^

chrisgarrity:
@GreenBayRules yes, this looks like it's still an issue.

GreenBayRules:
Okay, I'll get t o work on it right now :D

GreenBayRules:
PR: #1683 :D

chrisgarrity:
@GreenBayRules description is updated to reflect how we've decided to deal with unicode quotes

GreenBayRules:
Okay thanks, I'll submit a pr in a wihile.

LiFaytheGoblin:
So we want this: `\"` instead of any unicode quotes like `“` or `”`?

chrisgarrity:
@LiFaytheGoblin in our source files `l10n.json` we should only use `\"`. 

Technically, the unicode curly quotes are perfectly fine in json, so if a translator wants to use them, it still works. 

The problem is that it's very easy not to notice that the quotes are the special unicode quotes, so if someone later replaces them with normal double quotes without adding the escape backslash it causes syntax errors in the json file for that language. Any syntax error causes the whole file to get rejected, so all translations for that language are rejected.

PR #1683, opened 2017-11-22
Escape all quotes #1238

GreenBayRules:
Well I just went through all the "l10.json" files that I could find, and escaped the quotes that I saw, hopefully I didn't miss any :D (#1238)


Kenny2github:
It looks like all of the quotes you escaped here were Unicode directional quotes, though, not the string-breaking normal double quotes.

chrisgarrity:
Yes, these are all the unicode quotes. However the original problem occurred when a translator typed in the translation with normal quotes. So one question is whether we should be using the unicode directional quotes if they're likely to get missed in the translations. Escaping all of them seemed like a work-around. @GreenBayRules it's going to take me a little while to check this PR because I want to make sure it's not going to break any of the system that syncs the l10n files with the translation system. I probably won't get to it until after Thanksgiving.

 I guess the alternative would be to write a test that finds any strings that use unicode quotes in json files, and checks that all the translations also have unicode strings. That would be in the https://github.com/LLK/scratchr2_translations repo.

chrisgarrity:
@GreenBayRules unfortunately this isn't going to be as straightforward as I thought - simply escaping the unicode quotes breaks the json. We (Scratch team) need to decide whether we're committed to unicode quotes - in which case we probably need to implement translation checks on the translations repo. Or replace all unicode quotes with escaped normal quotes. In either case, this solution isn't going to work, so I'm going to close the PR. Thanks for helping to investigate!

PR #1684, opened 2017-11-24
Optimize all the things!

as-com:
This PR just catches some low-hanging fruit for better performance in both development and in the browser.

- [x] Don't use `eval` devtool in production (much better parsing performance and smaller bundles)
- [x] Only use UglifyJS in production (faster dev builds)
- [x] Add support for latest Node.js (#1144) for faster build times and less memory usage (oops already covered by #1671)
- [x] Upgrade Webpack to 3.x for even faster builds and optimization opportunities
- [x] Enable caching in babel-loader for faster builds
- [x] Babel plugin for React optimizations
- [x] Figure out what's bloating these bundles
- [x] Minify CSS
- [x] HTML minification
- [ ] Only require large libraries like libphonenumber when needed (remove from common chunk)
- [ ] Lodash
- [ ] UglifyJS parallelization
- [ ] Image optimization

Stretch goals:
- [ ] Hashed filenames for easier cache invalidation
- [ ] Lazy-load these large validation libraries
- [ ] Use `cache-loader` and `thread-loader` for some assets
- [x] Migrate to ES6 modules
  - [x] Webpack tree shaking
  - [x] Webpack scope hoisting
- [ ] Get rid of `React.createClass`
- [ ] React 16
- [ ] TypeScript? (avoid common errors, also produces cleaner/faster code than Babel)

PR #1689, opened 2017-11-27
Don't use eval devtool in production

as-com:
#1684 

This will save *[an alot](http://4.bp.blogspot.com/_D_Z-D2tzi14/S8TRIo4br3I/AAAAAAAACv4/Zh7_GcMlRKo/s1600/ALOT.png)* of bandwidth and parsing time. Also Uglify works much better on code that isn't in strings.

Before:
image
After:
image

This should be tagged "performance," but I don't have the permissions to do so.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -60,7 +60,7 @@ routes.forEach(function (route) {
 // Config
 module.exports = {
     entry: entry,
-    devtool: 'eval',
+    devtool: process.env.NODE_ENV === 'production' ? 'none' : 'eval',
     output: {
         path: path.resolve(__dirname, 'build'),
         filename: 'js/[name].bundle.js'


mewtaylor:
@as-com just an fyi – we likely won't get to review these until after next week. We're a bit busy with CSEdWeek for the next week or two, but we will surely get to these afterwards!

Kenny2github:
@mewtaylor friendly ping now that CSEdWeek is over

mewtaylor:
@as-com I'm noticing when I look at this locally that the non-eval bundles are actually larger than what I currently see in our production bundles. And that's when I run with a production tag to see the non-eval code. So, it's not clear to me that this would help with the bandwidth part of this PR.

amazinigmech2418:
Unfortunately, it says there's invalid JSON in Travis...

as-com:
@mewtaylor So I rebased my change on the current develop branch and ran the build with and without the change. Here are the results:

Left is before, and right is after.

The size of common.bundle.js goes from 1.7 MB down to 642 KB. That's a lot fewer bytes to parse.

Gzipped size also goes down significantly:

PR #1692, opened 2017-11-27
Babel react optimization

as-com:
#1684 

This adds babel-preset-react-optimize, which applies transforms to React code that help with rendering performance.


mewtaylor:
@as-com this is another optimization for which it'd be helpful to include any benchmarks you've been able to see thus far – let me know if you have the time to take a look at those comparisons!

PR #1693, opened 2017-11-27
Simple heuristic for inclusion in common chunk

as-com:
#1684 

Modules that are used in more than 70% of chunks will now be included in the common chunk. This will avoid duplicating modules that aren't explicitly specified in the configuration.

This should be tagged "performance."

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -127,7 +127,14 @@ module.exports = {
             'process.env.API_HOST': '"' + (process.env.API_HOST || 'https://api.scratch.mit.edu') + '"',
             'process.env.SCRATCH_ENV': '"'+ (process.env.SCRATCH_ENV || 'development') + '"'
         }),
-        new webpack.optimize.CommonsChunkPlugin('common', 'js/common.bundle.js'),
+        new webpack.optimize.CommonsChunkPlugin({
+            name: 'common',
+            filename: 'js/common.bundle.js',
+            minChunks: function (module, count) {
+                // Include in common if more than 70% of chunks use it
+                return count / Object.keys(entry).length > 0.7;
+            }
+        }),
         new webpack.optimize.OccurenceOrderPlugin()
     ])
 };


mewtaylor:
@rschamp @chrisgarrity another perspective on this would be helpful – it does increase the size of the common chunk, and decrease the size of pages like splash and about, at least in develop environments. I haven't done an analysis across all pages, but it seems like it could potentially be helpful to be more intentional about the common chunk than only thinking about things that appear in every module.

@as-com – what led to the decision of using `.7` as the threshold? Just wondering if we should consider other thresholds when thinking about this change.

rschamp:
I'm curious if it would work to just get rid of the explicit common entry points, and just set the threshold to `1`?

chrisgarrity:
It would be good to know how big the differences are at a couple of different thresholds. Is there not much difference until you try .7, or is there a sweet spot between .7 and 1?

rschamp:
Another question about this: what effect if any does it have on build time?

I'm not sure I like the non-deterministic (or at least non-explicit) nature of splitting based on Webpack's calculation of module usage. It seems like it could lead to subtle and hard to track down bugs that could change from deploy to deploy.  Additionally because the common lib is currently fairly stable between deploys, we benefit more from browser caching.

With `1`, then you know it will be the truly "common" modules, used by every page, i.e. React, minilog, etc., and this bundle would not change very much.  Basically this would give us an easier way to track these modules without explicitly listing them out ourselves... but actually, this isn't really what we want. We do want something like `.7` — modules used by most of the entries.

To cache more shared code, I think I would be more in favor of including more modules that don't change as much explicitly in `common`, rather than relying on Webpack for this, so that it's more deterministic. What do others think about that?

chrisgarrity:
@rschamp I prefer the deterministic approach. I guess the question is would we review what's in the common chunk often enough - or is there some way to get a notification when something should get included in the common chunk without it being automatic.

as-com:
@thisandagain Sorry, I've been very busy recently. I'll try to get to these tomorrow.

thisandagain:
No problem!

rschamp:
@as-com Thank you for your suggestion and contribution. Closing this for the time being, since we still need to decide how we want to handle this — it has implications on the overall site performance, browser caching, and our internal process for managing modules in the common chunk for this repo.

PR #1698, opened 2017-11-28
[Master] Hotfix: Change avatar to use `uploads` endpoint

mewtaylor:
It also changes some formatting of the jsx, but nothing major. This is mostly for CSEdWeek, to reduce load to our legacy system when we expect increased traffic.

### Test Cases ###
* images on search and explore should still be the same size
* analyze the size of images coming through, if there's any noticeable influence on load time/performance/bandwidth

/cc @thisandagain @colbygk @chrisgarrity

### DIFF
--- a/src/components/grid/grid.jsx
+++ b/src/components/grid/grid.jsx
@@ -64,6 +64,7 @@ var Grid = React.createClass({
                                     href={href}
                                     title={item.title}
                                     src={item.image}
+                                    srcDefault={'https://uploads.scratch.mit.edu/galleries/thumbnails/default.png'}
                                     owner={item.owner}
                                 />
                             );
--- a/src/components/grid/grid.jsx
+++ b/src/components/grid/grid.jsx
@@ -32,33 +32,40 @@ var Grid = React.createClass({
 
                         if (this.props.itemType == 'projects') {
                             return (
-                                <Thumbnail key={key}
-                                           showLoves={this.props.showLoves}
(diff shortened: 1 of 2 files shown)


thisandagain:
This seems reasonable to me. @mewtaylor when testing on `/explore` do you get a sense of how much (kB) this adds to user-facing page load?

thisandagain:
@mewtaylor Happy to test that for myself once it is landed in staging. 😉

mewtaylor:
@thisandagain when looking at the first page of explore, looking at just the img loaded data, it was, according to Chrome devtools,
* Production: 2.3MB data transferred
* Local: 2.4MB data transferred

Through 5 pages of explore data, it was up to:
* Production: 9.7MB data transferred (of images)
* Local: 12.5MB data transferred (of images)

I cycled through a few pages and was finding that, on average, the image data increases by about 25-30% with this change. The base line tends to be 1.5-2.5MB of image data per page, so by rough estimates it will increase to around 1.875MB-3.125MB of image data per page, or an increase of 375KB-625KB.

thisandagain:
@mewtaylor That's not nearly as bad as I worried it might be. Thanks!

rschamp:
I love the onError solution for the fallback images. I get the feeling that we might not revert this after HOC... Would it be a lot of work to capture that behavior in a component to simplify the stuff going on in `Thumbnail`?

```jsx
<FallbackImg src={} fallback={} />
```

PR #1707, opened 2017-11-30
[Develop] Project page caching

colbygk:
Allow caching of specific `^/projects/` pages while maintaining language selection for non-logged in users. Reduce complexity of recv backend selection.

### DIFF
--- a/bin/configure-fastly.js
+++ b/bin/configure-fastly.js
@@ -42,10 +42,6 @@ async.auto({
         // on any of those route conditions.
         var notPassStatement = fastlyConfig.getAppRouteCondition('../build/*', routes, extraAppRoutes, __dirname);
         
-        // For all the routes in routes.json, construct a varnish-style regex that matches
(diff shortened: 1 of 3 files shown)


bin/configure-fastly.js
```diff
@@ -46,18 +47,42 @@ async.auto({
         var passStatement = fastlyConfig.negateConditionStatement(notPassStatement);
         
         // For a non-pass condition, point backend at s3
-        var backendCondition = fastlyConfig.setBackend(
-            'F_s3',
-            S3_BUCKET_NAME,
-            notPassStatement
-        );
-        // For a pass condition, set forwarding headers
-        var forwardCondition = fastlyConfig.setForwardHeaders(passStatement);
```
rschamp: Since we've inlined these two functions, we should remove their definitions to avoid cruft.

colbygk: Good point. Removing those.

PR #1716, opened 2017-12-01
[Develop] temporarily take out curator row and recently shared row

mewtaylor:


GreenBayRules:
Why are you taking the rows out temporarily?

TheGrits:
@GreenBayRules Not sure, but if I had to guess it's to relieve some server strain for Hour of Code week.

mewtaylor:
@GreenBayRules – @TheGrits is correct on this one :). It'll just be for a day or two

PR #1720, opened 2017-12-01
[Develop] Hotfix/cache nav session endpoints

colbygk:
This adds ^/fragment/account-nav.json and ^/session to being cached for logged out users


colbygk:
@rschamp said I could

Issue #1702, opened 2017-11-29
Scratch Conference site - small edits 


me-win:
Hi, we're requesting a few small changes to the Scratch Conference site (scratch.mit.edu/conference):

- Replace the header photo on the page with the one attached (cropped to show the four individuals at the top of the photo). ![_dsc4245](https://user-images.githubusercontent.com/33295461/33396975-27196358-d518-11e7-8236-1b7e73b0d914.jpg)
- Add a colon after Scratch Conference 2018 (It should read "Scratch Conference 2018: The Next Generation")
- Decrease the line size/spacing between “Scratch Conference 2018” and  “The Next Generation.” (Currently, the name does not read as the event name.)
- Please add the following line above the Submit Your Proposal button (left aligned): “Deadline for proposals is February 5, 2018.”

Thanks!

redmangospros:
The Colon : Has Been Deind And Been Asked by @GreenBayRules  And Was Denined By @thisandagain

thisandagain:
Thanks @me-win !

redmangospros:
What DO you Need Help on?

VutonDesign:
@redmangospros Fixing the things stated in the first post.

thisandagain:
Adding `help wanted` label to this for all four items.

redmangospros:
Where Is The Source Code For Confrence?

VutonDesign:
The code for the 2018 conference page is at https://github.com/LLK/scratch-www/tree/develop/src/views/conference/2018/index

redmangospros:
I Can Do The SemliColon :

redmangospros:
Is It Right? #1705

VutonDesign:
It’s right, however your meant to try and make all of the requested changes and not just pick one or two of the easy ones.

redmangospros:
Well, I Did One, So anway.
Where It The COde For the img?

GreenBayRules:
I'll go ahead and do this..

GreenBayRules:
I can't find the location of the header image...

chrisgarrity:
@GreenBayRules the header image is defined in the css as a background image: https://github.com/LLK/scratch-www/blob/bfff01f0246288ccad195a551d39cb4d37ed5222/src/views/conference/2018/index/index.scss#L11
You'll need to add the image to the static images for the [conference/index](https://github.com/LLK/scratch-www/tree/develop/static/images/conference/index)  probably by adding a 2018 directory.

[...]

PR #1724, opened 2017-12-03
Changes requested in #1702

redmangospros:
Adding a semicolon as requested in #1702 to add a semicolon.
This pull requet adds the following changes


GreenBayRules:
Again, as I said before I don't  think you should submit a pull request if you are only finishing part of the issue, you should probably solve the rest of the issue then pull request.

redmangospros:
I will, i will do this.
where is the source code for img.

redmangospros:
#1725

thisandagain:
Tracking full fix in #1727. In the future, please only submit PRs that cover an entire issue unless you ask first. Many thanks!

PR #1730, opened 2017-12-11
[Master] Add back curator and recently shared rows

mewtaylor:
Reverts LLK/scratch-www#1715


thisandagain:
@Kenny2github Yup! CS Education Week ended last night.

Issue #711, opened 2016-07-09
Language selector on Firefox has strange focus ring placement


towerofnix:
Actually it's probably just built builtin focus ring that isn't being hidden.

![Demo](https://cloud.githubusercontent.com/assets/9948030/16709269/f2a0743e-45e1-11e6-8892-a110662b6d7c.png)

I'm running Firefox 50 on OS X El Capitan 10.11.4. There isn't a browser-specific focus ring on Safari or Chrome.

St19Galla:
What is the design preference for this? Hide it as the mouse moves away? :)

coke11:
This happens with almost, every single menu when I run firefox on my mac...

towerofnix:
Fixed by #1759.

PR #1759, opened 2018-01-05
Hide Firefox focus ring on <select> elements (language selector)

towerofnix:
### Resolves:

Fixes the old issue #711.

### Changes:

I don't know why commit 007bb54bf17d62577e85317778db48af4b023656 is in this PR, but it doesn't make any changes.

Quoting the two commit messages:

> Only transition border property since it's the only property which actually animates. (The background image also changes on hover/focus, but that isn't visually effected by transition, so we [don't transition] it.) This is for the next commit. Without this, the focus ring slowly disappears, and the text slowly becomes the color it's meant to be; neither of those are good.

> Hide Firefox focus ring, per [this StackOverflow answer](https://stackoverflow.com/a/18853002/4633828). This isn't a perfect fix - we have to set `color` explicitly, rather than use `inherit`, because we want the text color to be the same when the select element is focused. That means both `color` and `text-shadow`'s color have to be explicitly set to the same thing (otherwise they might not match). Yay, hacks!

### Test Coverage:

Tested locally. The language selector in the site footer appears as expected (no focus ring). Strangely, the "explore" page's selector isn't affected this at all (even though it's also a `<Select>` element).

### DIFF
--- a/src/components/forms/select.scss
+++ b/src/components/forms/select.scss
@@ -12,6 +12,7 @@
         border: 1px solid $active-gray;
         border-radius: 5px;
         background: $ui-light-gray url("../../../static/svgs/forms/carot.svg") no-repeat right center;
+        color: $type-gray;
         padding-right: 4rem;
         width: 100%;
         height: 3rem;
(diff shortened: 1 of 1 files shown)


thisandagain:
Hey @towerofnix. Sorry I didn't see this come in. Looks like you have some minor CI issues to resolve, but in theory this seems fine.

towerofnix:
@thisandagain Cool, fixed that. `npm test` doesn't fail on that locally anymore (I assume travis will work too).

thisandagain:
@mewtaylor Can you R?

Issue #1711, opened 2017-12-01
Scratch 3.0 Preview FAQ Page


carljbowman:
## Overview
We will need a static page that has FAQ content addressing the Scratch 3.0 preview. Most likely we will use our [FAQ page template](https://scratch.mit.edu/info/faq), so there is no real design work. 

## Content
https://docs.google.com/document/d/1DL9rY8tCUCYvKpfmd9xRSpwzFjuvEzW3TuxkpIHZKWU/edit?usp=sharing

## URL
`scratch.mit.edu/preview-faq`

We will use this URL as a pointer in the [Intro modal](https://github.com/LLK/scratch-gui/issues/828) and the [Unsupported Browser modal(s)](https://github.com/LLK/scratch-gui/issues/982).

GreenBayRules:
Could this by any chance become help-wanted after the content has been decided?

thisandagain:
@GreenBayRules Perhaps! We've got a lot of writing to do first, but I'll let you know once we get further on the content. 😄

redmangospros:
Let Me Know too

ChampikaF:
@carljbowman @thisandagain regarding the url - I like the idea of it living at scratch.mit.edu/info/faq-preview but could we also have a redirect from <preview url>/faq ?

thisandagain:
We already have a (general) FAQ so I'd prefer to not use that URL pattern
for this.

ChampikaF:
oops github didn't parse that correctly... what i was proposing was (preview-url)/faq

thisandagain:
Oh! Of course. No problem. :-)

amazinigmech2418:
Can I please be notified when the content is decided? I really want to help.

thisandagain:
@amazinigmech2418 We are going to handle this ourselves. Thanks for the offer.

amazinigmech2418:
You're welcome! I know you said only you and the rest of the ST are doing it, but if you ever need an easier way to update the page, I have a way to make easy API readers for the page...

thisandagain:
@mewtaylor First draft is ready to go!
https://docs.google.com/document/d/1DL9rY8tCUCYvKpfmd9xRSpwzFjuvEzW3TuxkpIHZKWU/edit?usp=sharing 

(updated original issue description)

vutondesign:
@thisandagain Looks Great! However, there seems to be an unfinished section about ScratchX:

[...]

PR #1761, opened 2018-01-08
Implement GH-1711: Add 3.0 preview faq page

mewtaylor:
### Resolves:

Implements #1711

### Changes:

Adds a new page for the 3.0 faq, with a the published faq doc embedded in a `Box` component. Sets a fixed height to `450px` while being responsive in width. Current endpoint: `/preview-faq`

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "babel-loader": "6.2.4",
     "babel-preset-es2015": "6.9.0",
     "babel-preset-react": "6.11.1",
+    "cheerio": "1.0.0-rc.2",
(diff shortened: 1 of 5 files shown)


thisandagain:
Thanks @mewtaylor. Can we drop the "frame" element around the content and try to better integrate it in the page? Ideally we could get closer to the layout from https://scratch.mit.edu/terms_of_use

For the LCL website I remember I used `cheerio` to pull out the HTML from the published page.

thisandagain:
This looks great! 👍

PR #1762, opened 2018-01-10
[Master] Add back in FPC and recently shared rows

mewtaylor:
There appears to have been a merge issue between the hotfixes around CSEdWeek and the implementation of the splash refactor, so this rectifies that error.


mewtaylor:
@rschamp reviewed this offline and said it was good :)

PR #1771, opened 2018-01-19
Upgrade www to es6

mewtaylor:
### Resolves:

This is an initial pass at upgrading our code to use es6 conventions.

### Changes:

Updates all react code to es6, including lint changes. Some of the test code to es6. Currently lint, unit and localization tests are passing – integration tests are broken, though. 

### Test Coverage:

um... yeah... so... everything.

Locally, I tried to get at least every page I could think of to load. I spent more time on:
* Splash
* Teacher registration
* Messages
Since they have more UX than other pages, but I tried everything I could.

PR #1775, opened 2018-01-23
[Develop] Upgrade to ES6

rschamp:
See https://github.com/LLK/scratch-www/pull/1771

Just switched out the source branch so it can more easily be built on staging.


mewtaylor:
@rschamp good call, thanks :)

rschamp:
Things found by the automated smoke tests
[ ] The links in the Featured Studios row go to projects rather than the studios
[ ] Log in menu will not open after clicking
[ ] Join Scratch modal can't be dismissed

chrisgarrity:
Ergonomics:
- [ ] Redux tools not enabled

src/components/accordion/accordion.jsx
```diff
@@ -1,33 +1,35 @@
-var classNames = require('classnames');
-var React = require('react');
+const bindAll = require('lodash.bindall');
```
chrisgarrity: @rschamp should www be using `import` instead of `require` to be consistent with other scratch-* projects?

rschamp: I think we should, but maybe in a separate PR, since we will have to update all the exports too, deal with `default` vs named exports, etc.  Could get hairy.  Also may be a good thing to do at the time that we implement the l10n string extraction system which requires `import`?

PR #1791, opened 2018-01-31
fixes #1764/reduce string duplication in teacher registration

### LINKED ISSUE #1764: Reduce string duplication in teacher registration
The string `Your information will not be displayed publicly, and will be kept confidential and secure` shows up twice in the l10n file, with and without a period at the end.

This causes extra work for translators, and the possibility of inconsistency if it's translated by different people. We should probably replace the two messages `teacherRegistration.orgStepDescription` and `teacherRegistration.addressStepDescription` into one:
```
"teacherRegistration.privacyDescription": "Your information will not be displayed publicly, and will be kept confidential and secure.",
```

LiFaytheGoblin:
### Resolves:
#1764 

### Changes:
merged the strings in the l10n.json as described here: https://github.com/LLK/scratch-www/issues/1764
then I replaced the strings in the teacher registration component with the new string.

### Test Coverage:
npm test was good after I added a new line to credits.jsx (the only reason why I changed that file again, so weird)
I also tested by changing "step" in teacherregistration.jsx to 1, so I didn't need to chose a username and could proceed in the form. I saw that the strings were visible in their new form. (Of course I changed the number back to 0 afterwards!)


LiFaytheGoblin:
Oh no, sorry about the conflict! It looks like I missed an update when I first worked on this and tested it a few days later :( I hope this still works somehow?

chrisgarrity:
@LiFaytheGoblin don't worry about the conflicts. We're in the middle of trying to update www to ES6 version of Javascript. That means that pretty much every outstanding PR will have conflicts. Once we land the ES6 PR, we'll go through and fix up all the PRs.

chrisgarrity:
@LiFaytheGoblin it's failing due to a translation error. I've put in a fix for the translations, so I'll restart the build as soon as they're ready. If all goes well, this should be good to merge.

Issue #1814, opened 2018-02-18
Update wiki footer link in automated smoke tests


jwzimmer:
If the existing automated test suite is going to be continue to be used, the test for the footer links needs to be updated to reflect the recent change to the English Wiki homepage (https://github.com/LLK/scratchr2/issues/4984).

Here's the line that needs to be changed: https://github.com/LLK/scratch-www/blob/develop/test/integration/smoke-testing/test_footer_links.js#L148

The URL `https://wiki.scratch.mit.edu/wiki/Scratch_Wiki_Home` should be changed to `https://en.scratch-wiki.info/wiki/Scratch_Wiki_Home`.

Narulacodes:
i can take this, if this becomes help-wanted
@jwzimmer

ktrvs:
Is the plan still to keep the new wiki links clickable?

Narulacodes:
> Is the plan still to keep the new wiki links clickable?

I think so, they will in some cases.

PR #1815, opened 2018-02-18
Update wiki link in footer

jwzimmer:
(Feel free to close if not needed. : ) )

### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/1814.

### Changes:

The URL `https://wiki.scratch.mit.edu/wiki/Scratch_Wiki_Home` was changed to `https://en.scratch-wiki.info/wiki/Scratch_Wiki_Home` (in https://github.com/LLK/scratch-www/blob/develop/test/integration/smoke-testing/test_footer_links.js#L148).

### Test Coverage:

I ran the file and the other footer link tests still pass. There is one failure, which is this test, because the link in the footer has not yet been updated to the new URL. In the reporting for the test failure, you can see that the old URL was the one found, and the new URL is the one the test is now looking for. (The other tests in this file pass, and I did not make any changes to any other files.)
screen shot 2018-02-18 at 9 29 04 am

### DIFF
--- a/test/integration/smoke-testing/test_footer_links.js
+++ b/test/integration/smoke-testing/test_footer_links.js
@@ -147,7 +147,7 @@ tap.test('clickDiscussionForumsLink', options, function (t) {
 // SCRATCH WIKI
 tap.test('clickScratchWikiLink', options, function (t) {
     var linkText = 'Scratch Wiki';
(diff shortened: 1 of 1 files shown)


jwzimmer:
Whoops, I used an older version of the file, sorry.

PR #1826, opened 2018-03-05
Clarify README and fix link on README to testing guide

### LINKED ISSUE #1823: Read Me is unclear on a number of topics
The Read Me section To Build contains no explanation of what is going on, what it's doing.  It starts by making sure you have npm installed, which is fine, but it doesn't explain the process before having you follow it.

Similarly a number of other verbs have little to no explanation about what they do or why you would do them, including To Test, To Run, To Start, and To Deploy.

BryceLTaylor:
### Resolves:

Resolves #1823 

### Changes:

Update testing guide url
Add explanations to sections of README

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ We’re working to update the [Scratch website](https://scratch.mit.edu) to use 
 We’re currently building Scratch using [React](https://facebook.github.io/react/) and [SCSS](http://sass-lang.com/documentation/file.SASS_REFERENCE.html). Here are some resources to help you get acquainted with how we’re working on the Scratch codebase:
 
 * [Style Guide](https://github.com/LLK/scratch-www/wiki/Style-Guide)
-* [Testing Guide](https://github.com/LLK/scratch-www/wiki/Testing-Guide)
+* [Testing Guide](https://github.com/LLK/scratch-www/wiki/Testing-Guide-for-Bugfixes)
 * [Localization Guide](https://github.com/LLK/scratch-www/wiki/Localization-Guide)
 * [Map of the repository](https://github.com/LLK/scratch-www/wiki/Repo-Map)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,19 +16,29 @@ We’re working to update the [Scratch website](https://scratch.mit.edu) to use
...
 * [Localization Guide](https://github.com/LLK/scratch-www/wiki/Localization-Guide)
 * [Map of the repository](https://github.com/LLK/scratch-www/wiki/Repo-Map)
 
 
 ### Before Getting Started
 * Make sure you have node (v4.2 or higher) and npm [installed](https://docs.npmjs.com/getting-started/installing-node)
+We use npm (Node Package Manager) to maintain and update packages required to build the site.  
+
+### Update Packages
+When making changes it's important to make sure that all of the node packages are up to date. Some changes in the code are needed because of package updates, so make sure you do this. You can update the packages by running this command:
```
rschamp: I think this is a little unclear, how about:

> It's important to make sure that all of the dependencies are up to date because the scratch-www code only works with specific versions of the dependencies. You can update the packages by running this command:

README.md
```diff
@@ -16,19 +16,29 @@ We’re working to update the [Scratch website](https://scratch.mit.edu) to use
...
 
-### To Build
 ```bash
 npm install
+```
+
+### To Build
+To build a version of the server on your local machine you can run this command:
```
rschamp: How about

> To compile the source code into HTML and JavaScript bundles browsers can read, run this command:

README.md
```diff
@@ -16,19 +16,29 @@ We’re working to update the [Scratch website](https://scratch.mit.edu) to use
...
+
+```bash
 npm run build
 ```
+This constructs new files on your machine.  If you want the files to only be in memory for testing purposes, consider skipping to the To Run section below.
```
rschamp: > If you want to run a server that rebuilds the files as you edit them, skip to the To Run section below.

chrisgarrity: One issue is that you need to have built the `intl` directory for the server to run properly. So if you skip the build step, all the text is message ids. Probably the fix is to add a step to the Run section that says `make translations`.

Kenny2github:
Note: It looks like GitHub will block an automatic merge since @rschamp's still-standing review apparently overrides @chrisgarrity's approving review. Might want to dismiss @rschamp's review.

PR #1827, opened 2018-03-08
Move selenium-webdriver and chromedriver packages to main package.json

BryceLTaylor:
Moved all of the packages from a second package.json that was in test/integration/ into the main package.json so that it would update with npm install, ensuring the tests would work. 

They were the only things in test/integration/package.json, which has now been deleted.


thisandagain:
Hmm. These packages were placed outside of the scope of the main `package.json` for a reason: this will greatly increase `npm install` time and may make developing on `www` more difficult for folks on Windows. We should discuss.

rschamp:
Yes I think I misled @BryceLTaylor in making this change, it was part of making getting set up for end to end testing easier. However if we make these dependencies for everyone, it's possible they'll have the same headaches we've had trying to get set up for just the smoke tests.  We've gone a different way now so I think this can be safely closed.

PR #1836, opened 2018-03-21
Add cypress tests

BryceLTaylor:
### Changes:

This sets up Cypress as an end to end test tool to be used on www.  It is self contained in its own folder in the test folder: integration-cypress.  Cypress is intended to replace Selenium and TAP for the www integration tests.

Included is a README file that contains information on how to install and use Cypress.  It also includes two test files that directly replace the smoke tests from the TAP/Selenium tests.  Not all of the old tests are implemented yet.

For now we will not remove the TAP/Selenium tests.

### Test Coverage:

All un-skipped tests that were written pass.


rschamp:
@BryceLTaylor It looks like there are linting errors — we can get you set up with a linter tomorrow.

PR #1839, opened 2018-03-23
Dockerization 

colbygk:
### Resolves:

This allows for using a localized docker development environment that works in concert with scratch-api and scratchr2

### Changes:

Adds Dockerfile, docker-compose.yml and docker_entrypoint.sh

### Test Coverage:

Requires no tests. This is an environment that allows running tests within it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:6
+FROM node:8
 
 RUN mkdir -p /var/app/current
 WORKDIR /var/app/current
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM node:6
+
+RUN mkdir -p /var/app/current
+WORKDIR /var/app/current
+COPY . ./
+RUN rm -rf ./node_modules
+RUN npm install
+
+EXPOSE 8333
+
--- a/README.md
+++ b/README.md
@@ -113,15 +113,18 @@ Setting `FALLBACK=https://scratch.mit.edu` allows the web client to retrieve dat
 
 Additionally, if you set `FALLBACK=https://scratch.mit.edu`, be aware that clicking on links to parts of the website not yet migrated over (currently such as `Explore`, `Discuss`, `Profile`, etc.) will take you to the Scratch website itself.
 
+#### Windows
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:6
```
rschamp: Currently Travis runs under Node 8, so we should match that.

docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+  scratchapi_scratch_network:
+    external: true
+
+services:
+  app:
+    container_name: scratch-www-app
+    hostname: scratch-www-app
+    environment:
+      - API_HOST=http://localhost:8491
+      - FALLBACK=http://scratchr2-app:8080
```
chrisgarrity: Regarding documentation for contributors. Both api and scratchr2 are private, so non-ST folks can't run those containers. Since this docker container assumes that you're running both locally, does this even make sense for a contributor outside the Scratch Team?

colbygk: They don't, but these defaults work for our development environments and I think can be ignored by other folks?

README.md
```diff
@@ -113,5 +113,45 @@ Setting `FALLBACK=https://scratch.mit.edu` allows the web client to retrieve dat
...
+
+A set of [Docker](https://www.docker.com/what-docker) related files are provided to create isolated [container](https://www.docker.com/what-container) environments suitable for development:
+
+* Dockerfile
+* docker-compose.yml
+* docker_entrypoint.sh
+
+Unless you have access to the Scratch REST API repository, this may not be very useful for you.
+
+This currently requires a version of the [Scratch GUI](https://github.com/LLK/scratch-gui) repository to be checked out in parallel to scratch-www, e.g.:
```
chrisgarrity: This is hopefully very temporary. Once the export scratch-gui for the projects page is in the published version of gui, you won't **need** the parallel gui.

FunStudios: Great!

thisandagain:
@colbygk @rschamp What's the status of this? Should this be closed until you have time to fully resolve?

colbygk:
I think it should be okay, but we should hear from @rschamp

rschamp:
@colbygk I updated the docs a bit, LGTM!

Issue #1800, opened 2018-02-02
Update project metadata from 3.0 project view


chrisgarrity:
A user can update the metadata associated with an sb2 project on the 3.0 project view.

PR #1856, opened 2018-04-11
Feature/preview project2

chrisgarrity:
Fixes https://github.com/LLK/scratch-www/issues/1800

Added edit-in-place fields for project title and description. Both validate the maximum length requirements. Edit-in-place elements replace normal title and description if the current user is also the author.

Also added a component to replace @mentions and #hashtags with links when the description is rendered.


src/lib/decorate-text.jsx
```diff
@@ -0,0 +1,30 @@
...
+ * Helper method that replaces @mentions and #hashtags in plain text
+ * 
+ * @param  {string}   text     string to convert
+ * @return {string}            string with links for @mentions and #hashtags
+ */
+module.exports = text => {
+    let replacedText;
+    
+    // Match @-mentions
+    replacedText = reactStringReplace(text, /@(\w+)/g, (match, i) => (
```
rschamp: I don't think `\w` includes hyphens or underscores. The username regex we use to validate usernames during registration is `/^[\w_-]+/`.

src/lib/decorate-text.jsx
```diff
@@ -0,0 +1,30 @@
...
+        <a
+            href={`/users/${match}`}
+            key={match + i}
+        >@{match}</a>
+    ));
+    
+    // Match hashtags
+    replacedText = reactStringReplace(replacedText, /#(\w+)/g, (match, i) => (
```
rschamp: We probably also want to include hyphens and underscores here too? At least hyphens.

src/lib/decorate-text.jsx
```diff
@@ -0,0 +1,30 @@
...
+        <a
+            href={`/search/projects?q=${match}`}
```
rschamp: @thisandagain Mentioned he would like the search to include the `#` so that these could be specific to the hashtag not just the term.

src/redux/preview.js
```diff
@@ -148,3 +301,55 @@ module.exports.getRemixes = id => (dispatch => {
...
+            dispatch(module.exports.setFetchStatus('studios', module.exports.Status.ERROR));
+            dispatch(module.exports.setError(err));
+            return;
+        }
+        if (typeof body === 'undefined') {
+            dispatch(module.exports.setFetchStatus('studios', module.exports.Status.ERROR));
+            dispatch(module.exports.setError('No studios info'));
+            return;
+        }
+        if (body.code === 'NotFound') {
```
rschamp: Probably best caught via `res.statusCode === 404`. `res` is the 3rd argument passed back from api (err, body, res). Ordered that way because most of the time we don't need res, just body.

src/redux/preview.js
```diff
@@ -148,3 +301,55 @@ module.exports.getRemixes = id => (dispatch => {
...
+            dispatch(module.exports.setFetchStatus('project', module.exports.Status.ERROR));
+            dispatch(module.exports.setError(err));
+            return;
+        }
+        if (typeof body === 'undefined') {
+            dispatch(module.exports.setFetchStatus('project', module.exports.Status.ERROR));
+            dispatch(module.exports.setError('No project info'));
+            return;
+        }
+        // TODO: figure out better checking for errors
```
rschamp: I think the `res` argument is the solution here.

src/views/preview/presentation.jsx
```diff
@@ -3,26 +3,42 @@ const injectIntl = require('react-intl').injectIntl;
...
+        user,
+        onFavoriteClicked,
+        onLoveClicked,
+        onUpdate
         // ...otherProps TBD
     } = props;
     const shareDate = (projectInfo.history && projectInfo.history.shared) ? projectInfo.history.shared : '';
-    
+    const lovesClass = loved ? 'project-loves loved' : 'project-loves';
+    const favesClass = faved ? 'project-favorites favorited' : 'project-favorites';
```
rschamp: This is a good use for `classNames`: `classNames('project-loves', {loved: loved})` (and lends itself for using inline for the `className` prop).

src/views/preview/presentation.jsx
```diff
@@ -37,123 +53,208 @@ const PreviewPresentation = props => {
...
+                                                //     id: 'project.titleMaxLength'
+                                                // })
+                                            }}
+                                            validations={{
+                                                // TODO: actual 100
+                                                maxLength: 40
+                                            }}
+                                            value={projectInfo.title}
+                                        /> :
+                                        <div className="project-title">{projectInfo.title}</div>
```
rschamp: Eventually it might be nice to consolidate this kind of switching into a component that accepts an editable prop.

src/views/preview/presentation.jsx
```diff
@@ -37,123 +53,208 @@ const PreviewPresentation = props => {
...
-                                    </div>
-                                </FlexRow>
-                            )}
-                            <div className="project-description">
-                                {projectInfo.description}
+                            </FlexRow>
+                            <div className="project-buttons">
+                                {sessionStatus === sessionActions.Status.FETCHED &&
+                                    Object.keys(user).length > 0 &&
+                                    user.id !== projectInfo.author.id &&
```
rschamp: This logic seems like it might be good to consolidate into a prop.

src/views/preview/preview.jsx
```diff
@@ -10,35 +11,125 @@ const sessionActions = require('../../redux/session.js');
...
+            }));
+        } else {
+            this.setState(state => ({
+                loveCount: state.loveCount + 1
+            }));
+        }
+    }
+    handlePermissions () {
+        // TODO: handle admins and mods 
+        if (this.props.projectInfo.author.username === this.props.user.username) {
```
rschamp: Another one that might be good to consolidate either into a single prop or into a method call like `isAuthor`.

src/components/forms/inplace-input.scss
```diff
@@ -0,0 +1,57 @@
...
+
+    &.pass {
+        border: 1px solid $active-dark-gray;
+    }
+
+    /* IE10/11-specific style resets */
+    &::-ms-reveal, &::-ms-clear {
+        display: none;
+    }
+}
```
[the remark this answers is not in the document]
chrisgarrity: I started doing that, but the www sass lint rules prevent it. If you're fine with removing that from the sass rules I'm happy to start un-nesting.

src/components/forms/inplace-input.jsx
```diff
@@ -0,0 +1,70 @@
...
+    handleBlur (name, value) {
+        if (this.inputRef.props.errorMessages.length === 0) {
+            const jsonData = {};
+            jsonData[name] = value;
+            this.props.handleUpdate(jsonData);
+        }
+    }
+    setRef (input) {
+        this.inputRef = input;
+    }
```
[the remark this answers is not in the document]
chrisgarrity: 👍

src/lib/decorate-text.jsx
chrisgarrity: Actually this raises the question of what we want to allow in hashtags, and internationalization. In `#thebestever!` does the '!' get included in the hashtag. If we allow punctuation and symbols, is it any non-whitespace character, or specific ones. For non-latin characters there's a babel plugin to support Unidcode regex: https://github.com/babel/babel/blob/master/packages/babel-plugin-proposal-unicode-property-regex/README.md

For now let's just support words with underscore and hyphen. I'll open up an issue to discuss what else to include in hashtags. I think that we should at least support non-english characters in the long run - will that work with the search engine?

src/views/preview/preview.jsx
chrisgarrity: Agreed. I've left them as-is for now as I think there will be more statuses once we get into unshared and admin/mods.

src/components/forms/inplace-input.scss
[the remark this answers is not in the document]
rschamp: I'm fine with removing sass linting altogether!

src/redux/preview.js
```diff
@@ -125,6 +160,124 @@ module.exports.getCreditInfo = id => (dispatch => {
...
+            authentication: token,
+            method: 'POST'
+        }, (err, body) => {
+            if (err) {
+                dispatch(module.exports.setError(err));
+                return;
+            }
+            if (typeof body === 'undefined') {
+                dispatch(module.exports.setError('Set loved returned no data'));
+                return;
```
rschamp: Should this be an error? If someone is loving something they already loved, would this be expected?

src/redux/preview.js
```diff
@@ -125,6 +160,124 @@ module.exports.getCreditInfo = id => (dispatch => {
...
+            authentication: token,
+            method: 'DELETE'
+        }, (err, body) => {
+            if (err) {
+                dispatch(module.exports.setError(err));
+                return;
+            }
+            if (typeof body === 'undefined') {
+                dispatch(module.exports.setError('Set loved returned no data'));
+                return;
```
rschamp: Same question here re: errors if no body is returned.

src/redux/preview.js
chrisgarrity: I believe in all cases API returns an object `{userLove: true|false}` as the body.

So, if body is undefined probably something went wrong.

PR #1859, opened 2018-04-20
Gh 1834/add kazahk to language menu

### LINKED ISSUE #1834: Add Kazakh (kk) to the language menu
The folks from Zerde (zerde.gov.kz) have reviewed/translated all the high priority strings, so it's time to add it to the scratchr2 and www menus as 'қазақ' (I've asked the translators to confirm that's the correct name)

LiFaytheGoblin:
### Resolves:
fixes #1834 
Adds Kazakh to the languages.json file.

### Changes:
you can now switch your language to kazakh, hurray!

### QA:
- [ ] Kazakh is available in the footer menu in between Kannada and Kinyarwanda.
- [ ] Navigation and footer switch language to Kazakh when language is selected.

### Test Coverage:
It says it passed the tests with npm test and I also tried in the browser and it did indeed switch the language to kazakh when chosing this in the language menu
grafik


languages.json
```diff
@@ -36,6 +36,7 @@
     "kn": "ಭಾಷೆ-ಹೆಸರು",
     "rw": "Kinyarwanda",
     "ht": "Kreyòl",
+    "kk": "Қазақша",
```
chrisgarrity: One note - this should be between 'kn' and 'rw',

Name-wise Kazakh, would be between Kannada and Kinyawanda.

LiFaytheGoblin: oups, thanks!

languages.json
```diff
@@ -34,8 +34,10 @@
     "is": "Íslenska",
     "it": "Italiano",
     "kn": "ಭಾಷೆ-ಹೆಸರು",
+    "kk": "Қазақша",
     "rw": "Kinyarwanda",
     "ht": "Kreyòl",
+    "kk": "Қазақша",
```
Kenny2github: Looks like you forgot to remove the original key - that's why the build is failing

Issue #1799, opened 2018-02-02
Load sb2 project into 3.0 project view


chrisgarrity:
A user can load an existing sb2 project. They get a preliminary (pre-design) project view including metadata from the sb2 project (title, instructions, etc)

- [x] Title
- [x] Author:
  - [x] Avatar
  - [x] username as link to profile
- [x] share date with 'copyleft' symbol
- [x] remix credit if project is a remix
- [x] description (combination of current description and instructions fields - will need migration)
  - [x] `@mentions` and `#hashtags` should be linked in description
- [x] stats
  - [x] loves - number of loves, logged in user can toggle loved status
  - [x] favorites - number of favorites, logged in user can toggle fav status
  - [x] remixes - numer of remixes
  - [x] views - number of views capped at 100
- [x] list of remixes
- [x] list of studios
- [x] player - gui placeholder

Buttons:
- [ ] ~~share banner with share button if project unshared~~ see #1873
- [x] Remix button if logged in user and not the author
- [ ] ~~See inside (opens GUI)~~ see #1874
- [ ] ~~Add to studio~~ see #1872
- [ ] ~~Report~~ see #1857  
- [ ] ~~Social~~ see #1852 

Excludes comments.

PR #1869, opened 2018-05-01
Initial GUI embed

### LINKED ISSUE #1773: Add 3.0 editor view (new project) to www as preview
As a user going to `/preview/editor` should load the current Scratch 3.0 preview editor with a new empty project (only available on the Media Lab network)

- Define what scratch-gui needs to publish for scratch-www
- Pass intl and state from scratch-www
- No save or load

Depends on https://github.com/LLK/scratch-www/issues/1694

- `/preview/:id/editor` will load the current GUI with the project (`:id`) preloaded

chrisgarrity:
### Resolves:

Fixes #1773, and most of #1799 

### Changes:

- `/preview/editor` will load GUI with an empty project
- `/preview/:id/editor` will load GUI with a project from the projects server (not local data)
- `/preview/:id` loads a Scratch 2 project with just the player from GUI
- can switch in/out of full screen mode for both player and editor
- passes intl object to GUI

### DIFF
--- a/src/views/preview/preview.jsx
+++ b/src/views/preview/preview.jsx
@@ -2,6 +2,7 @@ const bindAll = require('lodash.bindall');
 const React = require('react');
 const PropTypes = require('prop-types');
 const connect = require('react-redux').connect;
+const injectIntl = require('react-intl').injectIntl;
 const Page = require('../../components/page/www/page.jsx');
 const render = require('../../lib/render.jsx');
 
@@ -9,6 +10,8 @@ const PreviewPresentation = require('./presentation.jsx');
 
 const sessionActions = require('../../redux/session.js');
 const previewActions = require('../../redux/preview.js');
+const GUI = require('scratch-gui').default;
+const IntlGUI = injectIntl(GUI);
 
 class Preview extends React.Component {
     constructor (props) {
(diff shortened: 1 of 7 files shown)


src/views/preview/preview.scss
```diff
@@ -1,24 +1,24 @@
...
     height: 100%;
 }
 
-/* override view padding for share banner */
-#view {
-    padding: 0 0 20px 0;
-}
-
-.gui * { box-sizing: border-box; }
+// .gui * { box-sizing: border-box; }
```
rschamp: Cruft

src/views/preview/preview.jsx
```diff
@@ -114,23 +140,35 @@ class Preview extends React.Component {
...
-                projectInfo={this.props.projectInfo}
-                remixes={this.props.remixes}
-                sessionStatus={this.props.sessionStatus}
-                studios={this.props.studios}
-                user={this.props.user}
-                onFavoriteClicked={this.handleFavoriteToggle}
-                onLoveClicked={this.handleLoveToggle}
-                onUpdate={this.handleUpdate}
-            />
+            this.state.inEditor ?
```
rschamp: I think `PreviewPresentation` should handle this logic since it's presentational, and we should just pass all the possible GUI props into that component. It can then handle how the GUI is actually rendered.  This will require moving the `<Page></Page>` piece into there, which is unlike our other views. I think that's ok — the same argument re: presentation applies.

src/views/preview/preview.scss
```diff
@@ -1,24 +1,24 @@
...
-#app,
+/* override view padding for share banner */
+#view {
+    padding: 0 0 20px 0;
+}
+
 .gui {
+    position: absolute;
+    top: 0;
+    z-index: 11;
```
rschamp: What's the need for the z-index here when the GUI is the only thing being rendered?

rschamp:
Marking as needs work pending the fix for Travis.

src/views/preview/preview.scss
chrisgarrity: Probably cruft left over.

PR #1875, opened 2018-05-02
Correct preview route for fastly

chrisgarrity:
### Resolves:

Fastly deploy failure

### Changes:

- use more explicit route
- handle fastly error in the callback correctly
- more succinct parsing of the URL path

Current tests pass


rschamp:
(Pending Travis ✅)

Kenny2github:
@rschamp check passed, time to merge

PR #1888, opened 2018-05-17
Selenium login failure tests

BryceLTaylor:
### Resolves:

All login-failure Selenium tests now pass.  

### Changes:

Make sure that the error messages are visible before getting their text and asserting on that text.
Update the tests to use css selectors to find elements, rather than text of the element.

Relies on https://github.com/LLK/scratch-www/pull/1887

### DIFF
--- a/test/integration/smoke-testing/test-login-failures.js
+++ b/test/integration/smoke-testing/test-login-failures.js
@@ -1,10 +1,8 @@
 const {
-    clickText,
-    findByXpath,
-    clickButton,
     driver,
-    until,
-    By
+    findByCss,
+    clickCss,
+    until
 } = require('../selenium-helpers.js');
 
 var username = process.env.SMOKE_USERNAME;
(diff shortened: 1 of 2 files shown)


test/integration/smoke-testing/test-login-failures.js
```diff
@@ -31,17 +29,17 @@ test('Trying to sign in with no password using scratchr2 navbar', t => {
     var nonsenseusername = Math.random().toString(36)
         .replace(/[^a-z]+/g, '')
         .substr(0, 5);
-    clickText('Sign in')
-        .then(() => findByXpath('//input[@id="login_dropdown_username"]'))
+    clickCss('.dropdown-toggle')
+        .then(() => findByCss('form#login input#login_dropdown_username'))
```
rschamp: If you're eventually targeting an id, you don't have to use anything except for that id.  `#login #login_dropdown_username` should be overly specific since `#login_dropdown_username` alone should identify the thing with that id.  The xpath only used one id, is there a reason you added `form#login`?

test/integration/smoke-testing/test-login-failures.js
```diff
@@ -31,17 +29,17 @@ test('Trying to sign in with no password using scratchr2 navbar', t => {
...
         .then((element) => element.sendKeys(nonsenseusername))
-        .then(() => clickButton('Sign in'))
-        .then(() => driver.wait(until
-            .elementLocated(By.xpath('//form[@id="login"]/button[@type="submit"]'))))
-        .then(() => driver.wait(until
-            .elementLocated(By.xpath('//form[@id="login"]/div[@class="error"]'))))
-        .then(() => findByXpath('//form/div[@class="error"]'))
+        .then(() => clickCss('form#login button'))
+        .then(() => findByCss('form#login .error'))
```
rschamp: Same comment here re: ids, although I think here it's probably ok since all you're doing is converting from xpath to css selectors.  But these can probably be simplified to `#login button` and `#login .error` respectively.

PR #1910, opened 2018-05-31
Run smoke tests with Travis after deployment

BryceLTaylor:
Sets up Travis to run Selenium smoke tests on the branch Travis.  This includes modifying the selenium helpers to be able to run headless and adding a stage to Travis to run the tests after it builds.

The tests will currently only run on the Travis build as they fail intermittently and shouldn't run on everything until they pass more consistently.


rschamp:
For posterity: it's ok to merge this PR with failing commits because they only will fail when pushed to the Travis branch. We will continue resolving the issues around the smoke tests on Travis on the travis branch, and update that to develop and master when we've worked out the kinks.

PR #1919, opened 2018-06-12
Use 16 as offset for teacher registration

thisandagain:
Use 16 as offset for teacher registration (as per GDPR requirements)


BryceLTaylor:
I'm still getting an error that says I must be at least 13 years old to create a teacher account if I put in a birth date after June 2002.
screen shot 2018-06-19 at 4 07 40 pm

BryceLTaylor:
I've opened the issue described above in its own issue: https://github.com/LLK/scratch-www/issues/1929

BryceLTaylor:
There is another issue after you get the "too young" error: https://github.com/LLK/scratch-www/issues/1930

Issue #1916, opened 2018-06-07
Change splash to use 30000000 as a default value of projects


apple502j:
https://github.com/LLK/scratch-www/blob/develop/src/views/splash/splash.jsx#30
We celebrated :tada: 30000000 projects some months ago, but the repo didn't.

### Steps to Reproduce
See the code - but note that it's only used when API died so you can seldom see it.

### Operating System and Browser
(Do you care it?)

Windows 7 Homepremium 64bit firefox 60.0.2

GreenBayRules:
Just wondering what does that project count do, in the site

apple502j:
@GreenBayRules
The screen at the left when logged out I think

apple502j:
merged!

PR #1923, opened 2018-06-13
Change splash to use 30000000

apple502j:
It uses "30000000 projects" by default now!

### Resolves:
#1916 

### Changes:
Changed the default value.

### Test Coverage:
It's difficult to test it, because it has to be without api access.


thisandagain:
Thanks!

PR #1932, opened 2018-06-20
Preview - Implement extension info strip

technoboy10:
### Changes:
This PR implements the extension info strip for the project page by adding an extensions prop to the Presentation component, as well as a new ExtensionChip component for displaying basic extension information. This required adding dependencies for `scratch-storage` and `scratch-parser` to get project information and parse it to get a list of extensions used in the project, as well as changing the Webpack configuration to use babel-loader on `scratch-*` dependencies (like scratch-gui does).

**NB: This PR depends on https://github.com/LLK/scratch-vm/pull/1255 to be merged for projects to take advantage of the extension info strip**
### Known Issues
* There is no sort order for extensions displayed currently, which we may or may not want to add. 
* The "needs connection" bubble for hardware extensions does not link to the extension's settings as had been previously suggested. I wasn't sure if there was a way to do this with the current GUI or not, so that might still need to be implemented.
* sb2 projects and sb3 projects saved before the scratch-vm PR is merged will not display extension chips.


src/views/preview/preview.jsx
```diff
@@ -97,8 +102,35 @@ class Preview extends React.Component {
...
+                !ArrayBuffer.isView(input)) { // taken from scratch-vm
+                    // If the input is an object and not any ArrayBuffer
+                    // or an ArrayBuffer view (this includes all typed arrays and DataViews)
+                    // turn the object into a JSON string, because we suspect
+                    // this is a project.json as an object
+                    // validate expects a string or buffer as input
+                    // TODO not sure if we need to check that it also isn't a data view
+                    input = JSON.stringify(input);
+                }
+                parser(projectAsset.data, false, (err, projectData) => {
```
rschamp: You should handle the error here. Probably the right thing to do is just log it and return, since this is solely so the extension panel can load.  This way we don't have to e.g. validate the project ID, we can just not crash if the projectID is bad.

src/views/preview/preview.jsx
```diff
@@ -41,6 +44,7 @@ class Preview extends React.Component {
         // parts[2]: undefined if no :id, otherwise either 'editor' or 'fullscreen'
         this.state = {
             editable: false,
+            extensions: [],
             favoriteCount: 0,
             loveCount: 0,
             projectId: parts[1] === 'editor' ? 0 : parts[1],
```
rschamp: We already possibly have a projectId, so we should kick off getExtensions with this initial state.

technoboy10: Was this meant to be a comment on line 80? Not sure initializing the extension list to anything but an empty array makes sense.

src/views/preview/preview.jsx
technoboy10: Fixed by 456e99f, but GitHub didn't mark this comment as outdated.

src/views/preview/preview.jsx
technoboy10: Ah gotcha, that makes sense.

technoboy10:
This PR will need to be modified to work with the new location of the extension list as per https://github.com/LLK/scratch-vm/pull/1255#pullrequestreview-130872553

src/views/preview/preview.jsx
```diff
@@ -97,6 +103,38 @@ class Preview extends React.Component {
...
+                    // or an ArrayBuffer view (this includes all typed arrays and DataViews)
+                    // turn the object into a JSON string, because we suspect
+                    // this is a project.json as an object
+                    // validate expects a string or buffer as input
+                    // TODO not sure if we need to check that it also isn't a data view
+                    input = JSON.stringify(input);
+                }
+                parser(projectAsset.data, false, (err, projectData) => {
+                    if (err) {
+                        console.log('Caught project parsing error:', err); // eslint-disable-line no-console
```
rschamp: Instead of `console.log`, use `lib/log`, e.g.: https://github.com/LLK/scratch-www/blob/417791b589fd5dca0616d6e112811c8c279587ca/src/components/errorboundary/errorboundary.jsx#L18

PR #1936, opened 2018-06-25
Setup Selenium tests to be run with Sauce Labs

BryceLTaylor:
### Resolves:

Allows Smoke Tests to be run  with SauceLabs.  

### Changes:

Adds a second function for creating a driver so that it uses Sauce Labs and then checks the environment variable SMOKE_REMOTE to determine which type of driver to make.
Also sets up an package.json to have a smoke-sauce script that sets the timeout to be larger.

### DIFF
--- a/test/integration/selenium-helpers.js
+++ b/test/integration/selenium-helpers.js
@@ -1,6 +1,8 @@
 var webdriver = require('selenium-webdriver');
 
 const headless = process.env.SMOKE_HEADLESS || false;
+const remote = process.env.SMOKE_REMOTE || false;
+const {SAUCE_USERNAME, SAUCE_ACCESS_KEY} = process.env;
 
 const getDriver = function () {
     const chromeCapabilities = webdriver.Capabilities.chrome();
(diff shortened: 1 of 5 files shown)


test/integration/selenium-helpers.js
```diff
@@ -18,7 +20,34 @@ const getDriver = function () {
...
+    return driver;
+};
+// Driver configs can be generated with the Sauce Platform Configurator
+// https://wiki.saucelabs.com/display/DOCS/Platform+Configurator
+const driverConfig = {
+    browserName: 'chrome',
+    platform: 'macOS 10.13',
+    version: '67.0'
+};
+var driver;
```
rschamp: Try not to use `var` — if it's going to change, use `let` rather than `const`.

test/integration/selenium-helpers.js
```diff
@@ -18,7 +20,34 @@ const getDriver = function () {
...
+if (remote === 'true'){
+    driver = getSauceDriver(SAUCE_USERNAME, SAUCE_ACCESS_KEY, driverConfig);
```
rschamp: Please add docs about `SAUCE_USERNAME` and `SAUCE_ACCESS_KEY` to the README.

Issue #1899, opened 2018-05-29
Create warning for users when they open the developer console


thisandagain:
Unfortunately, we've now seen a few instances of users creating malicious scripts and encouraging Scratch users to execute them in the developer console. Because of this we would like to create a warning that appears in the developer console. I've included some examples from other services below (as this has now unfortunately become standard practice):

### Discord
![image](https://user-images.githubusercontent.com/747641/40663575-f8aad162-6326-11e8-97fa-64e036c8fa3c.png)

### Roblox
![image](https://user-images.githubusercontent.com/747641/40663661-2677b524-6327-11e8-83f9-ad1f85751a7d.png)

### Facebook
![image](https://user-images.githubusercontent.com/747641/40663763-6dcdafdc-6327-11e8-83ce-efe321e7b7ec.png)

Assigning to @christanbalch and @dietbacon to draft the copy for the message.

/cc @rschamp @colbygk @chrisgarrity

colbygk:
<img width="1127" alt="screen shot 2018-05-29 at 10 07 10 am" src="https://user-images.githubusercontent.com/192393/40664096-293f2ebc-6328-11e8-900d-78c260983f90.png">

dietbacon:
First draft:

STOP! [in some kind of red image text]

This is part of your browser intended for developers.  If someone told you to copy or paste something here, please be aware that this could allow another person to take over your Scratch account, delete all of your projects, or do many other harmful things.  If you don't understand what exactly you are doing here, you should close this window without doing anything.

JoelGritter:
@thisandagain I'd be interested in developing a solution for this issue if it is open to contributors. The only question I would have would be where to put it. (i.e. is within the render function of the nav-bar a good spot or not: https://github.com/LLK/scratch-www/blob/develop/src/components/navigation/www/navigation.jsx#L185 )

thisandagain:
@JoelGritter That would be great. I think this should be in a new component that's attached to the top level application rather than something like the nav-bar which may not always be used in the future.

JoelGritter:
@seotts @christanbalch @dietbacon 
Is this what you were envisioning or is this _too_ scary?
![image](https://user-images.githubusercontent.com/13542746/40681077-1e6a624e-6356-11e8-8657-c35dbc923b9c.png)

JoelGritter:
@thisandagain Another question: do we want this to be translated or not?

thisandagain:
[...]

PR #1941, opened 2018-06-26
Fix gh-1899: Create Console Warning

JoelGritter:
### Changes:

Created new component for the console warning. The console warning has localized strings in the general l10n.json file, and uses console.log() to write to the console. (console.log() is currently not permitted in the lint rules, so this may need to be changed)

The resulting console warning should look like this:
image

### DIFF
--- a/src/components/consolewarning/consolewarning.jsx
+++ b/src/components/consolewarning/consolewarning.jsx
@@ -1,0 +1,16 @@
+const React = require('react');
+
+class ConsoleWarning extends React.Component {
+    constructor (reactIntl) {
+        super();
+        console.log(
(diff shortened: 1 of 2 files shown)


apple502j:
The code does nothing.

JoelGritter:
@apple502j You're right, it does nothing; but that is purposeful. It doesn't do anything because the component isn't being used elsewhere yet.

JoelGritter:
@rschamp @thisandagain suggested using a component: https://github.com/LLK/scratch-www/issues/1899#issuecomment-392913936

JoelGritter:
Closing for restructuring of solution.

Issue #1950, opened 2018-07-05
Text edit in FAQ


christanbalch:
We need to make a text edit in the [Privacy Policy section of the FAQ](https://scratch.mit.edu/info/faq/#privacy) page.

The portion of the text that needs to change currently reads, "contact email address - If the account holder is younger than 13, we ask for the email address of their parent or guardian....".

We need to change the age to 16 so it reads as, "contact email address - If the account holder is younger than 16, we ask for the email address of their parent or guardian...." instead.

JoelGritter:
I can take care of this one today at some point!

PR #1951, opened 2018-07-05
Fix gh-1950: Update parent email age on FAQ page

JoelGritter:
### Changes:

Changed the age from 13 to 16, as per the issue description.


thisandagain:
Thanks @JoelGritter !

Issue #1974, opened 2018-07-18
Docker does not re-compile on changes


colbygk:
When running within a docker container, webpack-dev-middleware is supposed to watch for changes to the source files, however it does not do so without explicitly specifying `poll` as a `watchOption`.

### Steps to Reproduce

Run with in a docker container, update a `.jsx` file, no re-compile occurs.

### Operating System and Browser

all.

PR #1975, opened 2018-07-18
watchOptions required for use in docker environment

colbygk:
By default, webpack-dev-middleware is supposed to be watching for changes in the local source files, but in practice this seems to not be the case in docker containers, as fsevents are not passed through the NFS mount into a container and it requires explicit polling to enable watching for changes.

This will work in both docker and outside of docker, and tunes the load on a docker container running webpack to not overly burden a local host’s cpu watching for changes. 

### Resolves:

#1974 - Docker does not re-compile on changes

### Changes:

This updates `dev-server/index.js` to specify `watchOptions`. These options include a polling interval of 2500ms and aggregate changes in a 500ms window.  See also: https://webpack.js.org/configuration/watch/

### Test Coverage:

N/A

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,3 +5,4 @@ intl/*
 locales/*
 **/*.min.js
 **/node_modules/*
+scratch-gui/*
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ test:
 	@make tap
 
 lint:
-	$(ESLINT) . --ext .js,.jsx,.json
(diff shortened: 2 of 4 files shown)


colbygk:
Yes, will fix lint and hold off until Friday

colbygk:
Also adjusted the linting rules to ignore node_modules and scratch-gui

benjiwheeler:
This looks good to me, thanks for explaining the .eslintignore scratch-gui entry offline!

PR #1981, opened 2018-07-19
Add landing pages for the micro:bit and EV3 extensions

technoboy10:
This PR adds landing pages for the EV3 and micro:bit extensions, including assets and localization files. The EV3 page will be accessible at /ev3 and the micro:bit page at /microbit.

## Known Issues
* This PR adds a not insignificant amount of technical debt, as much of the code in these two pages is identical. At some point in the future it would be good to refactor these landing pages into something more general.
* The page isn't optimized for browser widths smaller than desktop.


paulkaplan:
@technoboy10 LOOKS WONDERFUL! This all looks good to me. Thanks for following up with filing those issues. The only other issue that we should file is to come back and do an accessibility pass on the buttons, links, images and stuff, there are a number of missing alt/titles. Could you file an issue for that (but no need to update the code, since it is pretty dispersed). 

SHIP IT!

PR #1985, opened 2018-07-21
Add To Studio modal on project page

benjiwheeler:
This introduces the Add To Studio modal.

* src/redux/preview.js has functionality to add and remove the project from studios, and maintains info on the state of oustanding join/leave studio requests
* src/views/preview/preview.jsx uses state props from redux (projectStudios: studios the project is currently a member of; curatedStudios: studios the user is curator or owner of; and studioRequests: the join/leave outstanding request object) and consolidates these into a single array of studios with added boolean properties: includesProject and hasRequestsOutstanding
* this consolidation involves some tricky handling of studios open to all that the project starts out as a member of. We put those studios first, and keep them in place even when the user leaves them, so the list won't be reshuffled unexpectedly and so the user can rejoin them. However, reloading the page after leaving such a studio will make it no longer show up, which makes sense because at that point it's just another one of the thousands of open studios.
* src/components/modal/addtostudio/[container.jsx, presentation.jsx] use the consolidated studios object to render each studio as a button, with visual style indicating whether the project is currently in the studio, or has a join/leave request outstanding.
* Carl designed some nice transitions that will improve on the look of the button states further, but those will come in a future iteration.
* when the user clicks "Okay", if there are requests outstanding, the "Okay" button adds a spinner and reads "Finishing up..." Once there are no requests outstanding, the modal closes.
* note that the user can close the modal three ways: clicking "Close", clicking the "x" button, or clicking on the page outside of the modal. Carl and I felt this would reflect user intention and hopefully not be confusing.
* closing the modal while there are requests outstanding has no effect on the requests.


src/components/modal/addtostudio/modal.scss
```diff
@@ -0,0 +1,201 @@
...
+    height: 1.4rem;
+    width: 1.4rem;
+}
+
+.studio-status-icon--img {
+    height: 1.4rem;
+    width: 1.4rem;
+}
+
+
```
thisandagain: Extra line breaks?

src/components/modal/base/modal.scss
```diff
@@ -4,7 +4,7 @@
 .modal-content {
     position: relative;
     margin: 3.75rem auto;
-    border-radius: 1rem;
+    border-radius: 0.5rem;
```
thisandagain: /cc @carljbowman

static/svgs/modal/add.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
thisandagain: Mind running these through `svgo`?

src/components/modal/base/modal.scss
carljbowman: @benjiwheeler - Thinking we should keep the border-radius 1rem for now. We can circle back on modifying the global level style. It's a little uncertain how this might impact other modals on the site.

It's okay that the design spec is out of sync with WWW.

src/components/modal/addtostudio/modal.scss
benjiwheeler: fixed

src/components/modal/report/modal.scss
```diff
@@ -36,7 +36,7 @@
     width: 80%;
     line-height: 1.5rem;
     font-size: .875rem;
-    
+
```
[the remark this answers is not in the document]
benjiwheeler: It's a necessary line break for lint -- I just removed whitespace on the line.

src/components/modal/base/modal.scss
benjiwheeler: keeping 1rem for now. Chris, let's talk about that option after the commit.

src/components/modal/report/modal.jsx
```diff
@@ -101,21 +102,21 @@ class ReportModal extends React.Component {
...
-                                        <FormattedMessage id="report.CommunityGuidelinesLinkText" />
-                                    </a>
-                                )
-                            }}
-                        />
-                        <Form
-                            className="report"
-                            onSubmit={onReport}
-                        >
+                    <Form
```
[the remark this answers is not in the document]
benjiwheeler: The changes here were from before the project reporting work; they're here because I saw that what I was doing with the base modal and addtostudio scss needed to be reflected here too. I'd prefer to leave them in, if that's ok?

src/redux/preview.js
```diff
@@ -333,26 +381,90 @@ module.exports.getRemixes = id => (dispatch => {
     });
 });
 
-module.exports.getStudios = id => (dispatch => {
-    dispatch(module.exports.setFetchStatus('studios', module.exports.Status.FETCHING));
+module.exports.getProjectStudios = id => (dispatch => {
+    dispatch(module.exports.setFetchStatus('projectStudios', module.exports.Status.FETCHING));
+    api({
+        uri: `/projects/${id}/studios`
```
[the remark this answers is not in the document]
benjiwheeler: In the API, right now it just returns everything. There's currently no pagination. I think the algorithmic complexity and data shouldn't be a problem up to a large number of studios, but maybe we should test that and revisit this question.

src/redux/preview.js
```diff
@@ -333,26 +381,90 @@ module.exports.getRemixes = id => (dispatch => {
...
+            dispatch(module.exports.setError(err));
+            return;
+        }
+        if (typeof body === 'undefined') {
+            dispatch(module.exports.setFetchStatus('projectStudios', module.exports.Status.ERROR));
+            dispatch(module.exports.setError('No projectStudios info'));
+            return;
+        }
+        if (res.statusCode === 404) { // NotFound
+            body = [];
```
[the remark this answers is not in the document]
benjiwheeler: No, it returns an empty list for projects that belong to no studios, including for project ids that don't even exist. Should we still catch this status code? I realize we are not doing a very thorough job of catching failure codes in general in many endpoints.

static/svgs/modal/add.svg
benjiwheeler: done, saved 70%

src/redux/preview.js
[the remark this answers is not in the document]
benjiwheeler: Opened https://github.com/LLK/scratch-api/issues/516

benjiwheeler:
OK, I fixed a lot of lint errors and obviously need to get linting working correctly in my workflow!

@thisandagain and others, please take a look.

src/redux/preview.js
```diff
@@ -333,26 +381,90 @@ module.exports.getRemixes = id => (dispatch => {
...
+        dispatch(module.exports.setProjectStudios(body));
+    });
+});
+
+module.exports.getCuratedStudios = (username, token) => (dispatch => {
+    dispatch(module.exports.setFetchStatus('curatedStudios', module.exports.Status.FETCHING));
     api({
-        uri: `/projects/${id}/studios?limit=5`
+        uri: `/users/${username}/studios/curate`,
+        authentication: token
```
towerofnix: Is this line necessary? It's just for getting the studios a user curates, which does not require an authentication token ([see this, for example](https://api.scratch.mit.edu/users/mres/studios/curate)).

benjiwheeler: Looks like you're right! Fixed.

benjiwheeler:
@towerofnix Saw the issue you linked to. What are your thoughts about using userids vs usernames vs tokens here?

src/redux/preview.js
benjiwheeler: Also, good catch! It's hard to notice that I'm mistakenly providing too much info to the API.

towerofnix:
@benjiwheeler I think it makes sense to store the owner's username in `/studios/<userID>`.`owner`, instead of their user ID, since the ID is just not useful. Or maybe better, have the `owner` property follow the same format as a project `author`, which includes both their username and user ID. That would look like this:

```js
// https://api.scratch.mit.edu/studios/137078
{
    id: 137078,
    title: " SD - strategie - strategies",
    owner: {
        id: 62626,
        username: "s_federici"
    },
    // ↑ Instead of owner: 62626
    ...
}
```

>[..] tokens

I'm not 100% sure what you mean here, but the commit https://github.com/LLK/scratch-www/pull/1985/commits/a60da11f8ae2069dd36e4d42e0001cbd93af78b7 looks like how it's supposed to work to me. (That is, fetching the list of studios a user curates should *not* require an auth token, because that list is public, at least in scratchr2 - `/users/<username>/studios/`.)

src/components/modal/report/modal.jsx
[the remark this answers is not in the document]
chrisgarrity: 👍

src/components/modal/addtostudio/container.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+    componentWillUpdate () {
+        this.checkIfFinishedUpdating();
+    }
+
+    hasOutstandingUpdates () {
+        return (this.props.studios.some(studio => (studio.hasRequestOutstanding === true)));
+    }
+
+    checkIfFinishedUpdating () {
```
rschamp: This name is slightly confusing in the context of `componentWillUpdate`, since it doesn't look like it does anything. I think a better name would be `closeIfFinishedUpdating`.

src/components/modal/addtostudio/modal.scss
```diff
@@ -0,0 +1,193 @@
...
+.mod-addToStudio * {
+    box-sizing: border-box;
+}
+
+.mod-addToStudio {
+    margin: 100px auto;
+    outline: none;
+    padding: 0;
+    width: 36.25rem; /* 580px; */
+    height: 388px; /* 24.25rem; */
```
rschamp: Try to use `rem`s everywhere. In this codebase things are inconsistent, but we should try to be consistent going forward.

src/components/modal/addtostudio/presentation.jsx
```diff
@@ -0,0 +1,119 @@
...
+    intl,
+    isOpen,
+    studios,
+    waitingToClose,
+    onToggleStudio,
+    onRequestClose,
+    onSubmit
+}) => {
+    const contentLabel = intl.formatMessage({id: 'addToStudio.title'});
+    const studioButtons = studios.map(studio => (
```
rschamp: Since we won't be rerendering frequently, this isn't an issue, but in general try to avoid arrow functions within `render` if it's easy. So for this, defining a function above  the `AddToStudioModalPresentation` definition which returns this mapping for a set of `studios`, and then calling that where it's needed would be more performant.

src/components/modal/addtostudio/studio-button.jsx
```diff
@@ -0,0 +1,72 @@
...
+            onClick={onToggleStudio}
+        >
+            <div
+                className={classNames(
+                    'studio-selector-button-text',
+                    {'studio-selector-button-text-selected': includesProject},
+                    {'studio-selector-button-text-unselected': !includesProject}
+                )}
+            >
+                {truncate(title, {length: 20, separator: /[.,:;]*\s+/})}
```
rschamp: Aside: This looks like a good candidate for a helper function, since we probably want most things truncated in the same way.

src/views/preview/preview.jsx
```diff
@@ -196,6 +211,21 @@ class Preview extends React.Component {
             );
         }
     }
+    handleToggleStudio (event) {
+        const studioId = parseInt(event.currentTarget.dataset.id, 10);
+        if (!isNaN(studioId)) { // probably unnecessary test; should always be a number
```
rschamp: I agree that you don't need this test, since you check if it's present with `this.props.studios.find` — non-existent studio ids including `NaN` will be handled by that.

src/redux/preview.js
```diff
@@ -53,9 +57,28 @@ module.exports.previewReducer = (state, action) => {
...
             parent: action.info
         });
-    case 'SET_STUDIOS':
+    case 'SET_PROJECT_STUDIOS':
+        // also initialize currentStudioIds, to keep track of which studios
+        // the project is currently in.
         return Object.assign({}, state, {
-            studios: action.items
+            projectStudios: action.items,
+            currentStudioIds: action.items.reduce((ids, studio) => {
```
rschamp: Curious about the use of an object here rather than an array (or `Set`). Is it always `{<studioId1>: true, <studioId2>: true, ...}`?  The name `currentStudioIds` implies an array, so later on when you use the check `studioId in currentStudioIds`, it looks like a misuse of `in` on an array.

benjiwheeler: Turns out that Sets are not supported as state types in Redux! (Trying to find a reference online now, but failing.) I implemented it with currentStudioIds as a Set, but redux was mysteriously replacing my nice, data-filled set with an empty object.

So yes, currentStudioIds is an object of form `{<studioId1>: true, <studioId2>: true, ...}`. Not an array. What might be a name that would make this clearer?

src/views/preview/preview.jsx
benjiwheeler: <programmer opinion>I think it's bad form to pass on a state that indicates an error on to subsequent code, just because that code happens to handle the error case. What about when others change this in the future?</programmer opinion>

.gitignore
```diff
@@ -4,6 +4,7 @@
 # NPM
 /node_modules
 npm-*
+package-lock.json
```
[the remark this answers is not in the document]
benjiwheeler: Don't understand, need to talk more

Makefile
```diff
@@ -14,10 +14,10 @@ build:
 	@make webpack
 
 clean:
-	rm -rf ./build
-	rm -rf ./intl
 	mkdir -p build
+	rm -rf ./build/*
 	mkdir -p intl
+	rm -rf ./intl/*
```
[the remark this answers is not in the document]
benjiwheeler: The previous way does leave behind the build and intl directories -- the last thing it does is create them. Right?

The reason is that there is a later check for the intl directory, which seems intended to determine whether translations have been built. With an empty -- but existent -- intl directory, this was erroneously failing when running www in docker.

src/_colors.scss
```diff
@@ -35,6 +48,7 @@ $active-gray: hsla(0, 0, 0, .1);
 $active-dark-gray: hsla(0, 0, 0, .2);
 $box-shadow-gray: hsla(0, 0, 0, .25);
 $overlay-gray: hsla(0, 0, 0, .75);
+$transparent: rgba(229, 240, 254, 0);
```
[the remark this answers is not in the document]
benjiwheeler: It's a barely perceptible blue, for the fade-out gradient at the bottom of the Invision mockup. I guess a better name for this might indicate its blueness?

screen shot 2018-07-31 at 11 09 38 am

src/components/modal/addtostudio/container.jsx
benjiwheeler: Agreed.

src/components/modal/addtostudio/modal.scss
benjiwheeler: This is a deliberate decision (per a previous conversation of ours) because of the concern that the bottom of an absolutely positioned element may be impossible to access on small screens with the browser zoomed in.

.gitignore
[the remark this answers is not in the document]
benjiwheeler: Still disagree, but taking it out :)

src/components/modal/addtostudio/studio-button.jsx
benjiwheeler: I'm putting it in lib/truncate.js

Makefile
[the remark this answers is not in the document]
benjiwheeler: I was wrong in my reasoning -- this was because of Docker filesystem mapping. Turned that off for intl.

src/_colors.scss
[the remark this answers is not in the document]
technoboy10: Oh interesting, I think I did the same thing by mixing some other colors together in [#1997](https://github.com/LLK/scratch-www/pull/1997/files#diff-72cf6e9b2977e6c0858134e338fb2026R174). We should maybe try and decide on which way is the right one?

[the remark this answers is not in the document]
carljbowman: @benjiwheeler - Maybe we could grab 30min to pair on cleaning up these colors. 

Looks like there are a few places we have redundancies and other places we could consolidate down into using a core set of the color palette. Theoretically we shouldn't need to add this many new colors to the palette.

src/redux/preview.js
```diff
@@ -62,23 +62,21 @@ module.exports.previewReducer = (state, action) => {
...
-                return ids;
-            }, {})
+            currentStudioIds: action.items.map(item => item.id)
         });
     case 'SET_CURATED_STUDIOS':
         return Object.assign({}, state, {curatedStudios: action.items});
     case 'ADD_PROJECT_TO_STUDIO':
         // add studio id to our studios-that-this-project-belongs-to set.
-        state.currentStudioIds[action.studioId] = true;
+        state.currentStudioIds.push(action.studioId);
```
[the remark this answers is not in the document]
benjiwheeler: Fixed. Using concat.

Issue #1812, opened 2018-02-16
2018 Conference: live stream links July 26


chrisgarrity:
Update site with links to live stream for the Conference Keynote sessions. Will need to hotfix on the morning of the Conference opening.

- [x] live stream link

Narulacodes:
If this becomes help-wanted, in later months or now, i'll be  happy to take this.

Narulacodes:
Ok Thanks. I would really loved to help

PR #1994, opened 2018-07-23
[Master] Conference 2018 Livestream

shickey:
### Resolves:

#1812

### Changes:

Updated landing page header to include link to conference livestream (to be deployed on the 25th in prep for the beginning of the conference on the morning of the 26th)

### DIFF
--- a/src/views/conference/2018/index/index.jsx
+++ b/src/views/conference/2018/index/index.jsx
@@ -18,7 +18,7 @@ const ConferenceSplash = () => (
                 July 26-28, 2018 | Cambridge, MA, USA
             </h3>
             <h3>
-                <a href="http://web.media.mit.edu/events/medialabtalk/">
+                <a href="http://www.media.mit.edu/events/medialabtalk/">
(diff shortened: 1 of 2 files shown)


src/views/conference/2018/index/index.jsx
```diff
@@ -18,10 +18,9 @@ const ConferenceSplash = () => (
                 July 26-28, 2018 | Cambridge, MA, USA
             </h3>
             <h3>
-                (Opening Reception on Evening of July 25)
-            </h3>
-            <h3>
-                Scratch@MIT is sold out!
+                <a href="http://web.media.mit.edu/events/medialabtalk/">
```
chrisgarrity: I don't think this should be `http://web.media.mit.edu/` - I think it should be `https://www.media.mit.edu/`.

`https://www.media.mit.edu/events/` goes to a live page. `https://web.media.mit.edu/events/` gets a 404 not found.

shickey:
@rschamp 👍Done and done!

PR #1997, opened 2018-07-24
Initial implementation for displaying comments on project page

technoboy10:
This PR adds the first iteration of reading comments on the project page.

## Changes
* redux/preview.js was changed to add API calls for fetching a project's top level comments based on a comment offset
* preview/presentation.jsx was restructured to match the spec for the lower half of the project page
* added the preview Comment component to display a single comment
   * This is similar to the Comment component in the projects folder, but different enough in layout that it would have taken more effort to try and use the existing component in this layout
* added the CommentContainer component to handle the logic of displaying a comment and fetching its replies (if they exist)
* modified preview/preview.jsx to pass comment API functions to presentation.jsx

## Known Issues/TODOs
* Comments do not display username links/at-mentions, emojis, or links to Scratch-approved sites. This is due to the current comment API not injecting HTML like scratchr2 does right now
* The load more button still shows up when there are no more comments to load. The API currently does not include comment count, so there is not currently a sensible way to detect when all comments are loaded
* It is currently impossible to directly link to a comment below the first page of comments. scratchr2 does this somewhat inelegantly (it loads more comments until the desired comment is loaded), and the way the new API handles comment replies makes this harder to do here. (thanks @towerofnix for the reminder of how scratchr2 works!)
* Many UI elements (delete, report, create comment, reply) aren't hooked up to any actual API calls or logic
* This PR will need to be squashed and merged because oof that's a lot of commits


towerofnix:
>It is currently impossible to directly link to a comment below the first page of comments. scratchr2 does this with magic some hard to replicate code, so it might be worth revisiting the current behavior of Scratch 2 comment linking

The scratchr2 implementation for this is "keep loading more comments until this one is found", which is not too difficult to replicate, but I do think there should be a better way about doing that... perhaps it's worth comparing to the discussion forum behavior, where you can go to `/post/(post id)`, which will redirect to `/topic/(topic id)/?page=(post page)#(post id)` (or a 404 if that post does not exist). The important difference between comments and forums is that, while both are separated into pages (20 comments/posts or so), forums actually incorporate pages into their UI ("next page"), whereas comments just have a "load more" button.

technoboy10:
@towerofnix Ahhhh that's right, thanks! It's been a minute since I went through that part of the scratchr2 code. :)

Kenny2github:
> This PR will need to be squashed and merged because **oof that's a lot of commits**

src/views/preview/presentation.jsx
```diff
@@ -277,23 +281,51 @@ const PreviewPresentation = ({
...
+                                        <div>comments on/off</div>
+                                    </FlexRow>
+                                    <FlexRow className="create-comment comment">
+                                        <Avatar src="https://cdn2.scratch.mit.edu/get_image/user/3581881_60x60.png" />
+                                        <ComposeComment />
+                                    </FlexRow>
+                                    <FlexRow className="comments-list">
+                                        {comments.map(comment => (
+                                            <CommentContainer
+                                                {...comment}
```
rschamp: We should list these props explicitly because otherwise API changes are implicitly passed as props — it will be easier to handle changes like those if we list what we expect.

src/views/preview/comment/comment-container.jsx
```diff
@@ -0,0 +1,109 @@
...
+        this.setState({
+            expanded: true
+        });
+    }
+
+    render () {
+        const {
+            author,
+            content,
+            datetime_created,
```
rschamp: You could argue we shouldn't be sending this down in the JSON, but would you please convert this to camelCase?

src/views/preview/comment/comment-container.jsx
```diff
@@ -0,0 +1,109 @@
...
+            id,
+            reply_count
```
rschamp: camelCase please

src/views/preview/comment/comment-container.jsx
```diff
@@ -0,0 +1,109 @@
...
+        } = this.props;
+
+        return (
+            <FlexRow className="comment-container">
+                <Comment {...{author, content, datetime_created, id}} />
+                {reply_count > 0 && // eslint-disable-line camelcase
```
rschamp: Then you won't need this exception

src/redux/preview.js
```diff
@@ -59,7 +59,7 @@ module.exports.previewReducer = (state, action) => {
         });
     case 'SET_COMMENTS':
         return Object.assign({}, state, {
-            comments: action.items
+            comments: [...state.comments, ...action.items] // TODO: consider a different way of doing this?
```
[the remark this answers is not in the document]
rschamp: One improvement might be to use `.splice()` along with the `offset` and `limit` arguments that we sent in the request for the comments to put the new items exactly where they should be in the overall list of comments.

src/views/preview/comment/comment-container.jsx
```diff
@@ -0,0 +1,109 @@
...
+            expanded: false,
+            replies: []
+        };
+    }
+
+    componentDidMount () {
+        this.fetchReplies(this.props.id);
+    }
+
+    fetchReplies (id) {
```
rschamp: Eventually we should move this into redux — maybe something like
```js
{
    replies: {
        1: []
        2: [{...reply}, ...]
    }
}
```

src/redux/preview.js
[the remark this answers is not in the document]
rschamp: But I don't think we need to do that now (or maybe ever)

src/views/preview/comment/comment-container.jsx
rschamp: (where `1` and `2` are comment ids)

src/views/preview/preview.jsx
```diff
@@ -73,13 +74,15 @@ class Preview extends React.Component {
...
                 const token = this.props.user.token;
+                this.props.getTopLevelComments(this.state.projectId, this.props.comments.length);
                 this.props.getProjectInfo(this.state.projectId, token);
                 this.props.getRemixes(this.state.projectId, token);
                 this.props.getProjectStudios(this.state.projectId, token);
                 this.props.getCuratedStudios(username);
                 this.props.getFavedStatus(this.state.projectId, username, token);
                 this.props.getLovedStatus(this.state.projectId, username, token);
             } else {
+                this.props.getTopLevelComments(this.state.projectId, this.props.comments.length);
```
rschamp: This is probably something to follow up on, but have you looked to see what happens here when you log in or out while on the page? It looks like the transition would cause comments to load? Seems unexpected.

technoboy10: Hmm, that's a good point, not sure what would happen (although I agree that's probably a follow up). Might be a reason to switch to another data structure in redux, like a Set (or logic in redux that does the same thing).

technoboy10: Oh, I see what you mean (logging in loads the next page of comments, which is... weird). Not sure what would be the right approach for tackling that issue; some sort of pagination state in that component might work?

PR #2000, opened 2018-07-26
Fix width of message select

### LINKED ISSUE #1643: Filter by > Comment Activity too close to filter dropdown arrow - Messages page
This is a fairly minor cosmetic issue, but "Comment Activity" gets cut-off (in Chrome, Mac) or squished (in Firefox, Mac) by the filter dropdown menu arrow because it is longer than the other filter names. 

To repro, go to the Messages page and choose "Comment Activity" as your filter type in Firefox or Chrome. (It may happen in other browsers as well, but those are the two I tried. This is on Staging & Production.)

Chrome
screen shot 2017-10-31 at 6 15 18 pm
Firefox
screen shot 2017-10-31 at 6 11 44 pm

kyleplo:
### Resolves:

Resolves #1643 

### Changes:

When "Comment Activity" is selected, it gets too big for the menu. On mobile and desktop (but not tablet, because there isn't enough room,) this PR changes the width to `110%` to fit the text better.

### Test Coverage:

Desktop Size Chrome:

image
Mobile Size Chrome:

image

### DIFF
--- a/src/views/messages/messages.scss
+++ b/src/views/messages/messages.scss
@@ -24,7 +24,7 @@
     color: $type-gray;
 }
 
-select {
+.messages-title-filter {
     width: 110%;
 }
 
@@ -172,7 +172,7 @@ select {
         width: $cols8;
     }
     
-    select {
+    .messages-title-filter {
         width: 100%;
     }
 }
--- a/src/views/messages/messages.scss
+++ b/src/views/messages/messages.scss
@@ -24,6 +24,10 @@
     color: $type-gray;
 }
 
+select {
+    width: 110%;
+}
+
 .help-block {
     display: none;
 }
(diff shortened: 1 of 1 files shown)


src/views/messages/messages.scss
```diff
@@ -24,6 +24,10 @@
     color: $type-gray;
 }
 
+select {
```
rschamp: We try to avoid using bare element selectors (like `p`, `a` or in this case, `select`), and instead try to target element class names. We could add more `select` elements to the page in the future, and wouldn't want this `width` rule to apply to all of them.

Since the `select` you're targeting already has `100%` as its width, you should be able to adjust its container's width instead. It looks like that container has the class `.messages-title-filter`.

rschamp:
@kyleplo How are you testing the effects of your changes? The recent commit makes the select box fill the entire width of the page.

I just realized that this particular page might be difficult to work on locally since it requires the user to be logged in, so might not be the best fit for an outside contributor like yourself to work on.

PR #2001, opened 2018-07-27
Add sort scheme dropdown to search results page

### LINKED ISSUE #1252: Add sort scheme dropdown to search results page
On the explore page, there is a dropdown for the sorting scheme ("Trending", "Popular", "Recent"). This component should be added to the search results page, defaulting to "Popular".  Changing the choice should change the sort order of the results.

The same design on the Explore page should be used for the dropdown, except that the tags should not appear in the row.

kyleplo:
### Resolves:

Resolves #1252 

### Changes:

> On the explore page, there is a dropdown for the sorting scheme ("Trending", "Popular", "Recent"). This component should be added to the search results page, defaulting to "Popular". Changing the choice should change the sort order of the results.

> The same design on the Explore page should be used for the dropdown, except that the tags should not appear in the row.

This PR adds this dropdown and its functionality, including a new `m` query parameter for filter mode. A new search URL could look like: `/search/projects/?q=cats&m=recent`

### DIFF
--- a/src/redux/navigation.js
+++ b/src/redux/navigation.js
@@ -1,7 +1,6 @@
 const keyMirror = require('keymirror');
 
 const Types = keyMirror({
-    SET_MODE: null,
     SET_SEARCH_TERM: null
 });
 
@@ -10,19 +9,12 @@ module.exports.navigationReducer = (state, action) => {
         state = '';
     }
     switch (action.type) {
-    case Types.SET_MODE:
-        return action.mode;
     case Types.SET_SEARCH_TERM:
         return action.searchTerm;
     default:
         return state;
     }
 };
-
-module.exports.setMode = mode => ({
-    type: Types.SET_MODE,
-    mode: mode
-});
 
 module.exports.setSearchTerm = searchTerm => ({
     type: Types.SET_SEARCH_TERM,
--- a/src/redux/navigation.js
+++ b/src/redux/navigation.js
@@ -1,6 +1,7 @@
 const keyMirror = require('keymirror');
 
 const Types = keyMirror({
+    SET_MODE: null,
     SET_SEARCH_TERM: null
 });
 
@@ -9,12 +10,19 @@ module.exports.navigationReducer = (state, action) => {
         state = '';
     }
     switch (action.type) {
+    case Types.SET_MODE:
+        return action.mode;
     case Types.SET_SEARCH_TERM:
         return action.searchTerm;
     default:
         return state;
     }
 };
+
+module.exports.setMode = mode => ({
+    type: Types.SET_MODE,
+    mode: mode
+});
 
 module.exports.setSearchTerm = searchTerm => ({
     type: Types.SET_SEARCH_TERM,
--- a/src/views/search/search.jsx
+++ b/src/views/search/search.jsx
@@ -18,6 +18,8 @@ const Tabs = require('../../components/tabs/tabs.jsx');
 const Page = require('../../components/page/www/page.jsx');
 const render = require('../../lib/render.jsx');
 
+const ACCEPTABLE_MODES = ['trending', 'popular', ''];
+
 require('./search.scss');
 
 class Search extends React.Component {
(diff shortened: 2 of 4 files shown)


thisandagain:
The search API formally supports the `mode` parameter. For the search results, we would be comfortable supporting both `mode=popular` (default) and `mode=trending`.

kyleplo:
@thisandagain So you don't want the "recent" option?

thisandagain:
@kyleplo Correct. Thanks!

kyleplo:
Is this fixed, or is there anything else I need to do?

rschamp:
@kyleplo Sorry I missed this! I need to review it, you don't need to do anything atm.

src/redux/navigation.js
```diff
@@ -9,13 +10,20 @@ module.exports.navigationReducer = (state, action) => {
         state = '';
     }
     switch (action.type) {
+    case Types.SET_MODE:
+        return action.mode;
```
rschamp: This reducer is poorly named, but is responsible for the search term (as seen below, `SET_SEARCH_TERM` sets the state to the action's `searchTerm`. This means that if you set the state to `mode`, you're setting the search term to `mode`. This prevents the Search page from making the correct API request for search results.

Since the search mode is confined to the search page, I don't think you need to worry about Redux. You can remove the changes you have in this file.

src/views/search/search.jsx
```diff
@@ -23,12 +25,14 @@ class Search extends React.Component {
...
         bindAll(this, [
             'getSearchState',
+            'handleChangeSortMode',
             'handleGetSearchMore',
             'getTab'
         ]);
         this.state = this.getSearchState();
         this.state.loaded = [];
         this.state.loadNumber = 16;
+        this.state.mode = '';
```
rschamp: This is correct - you can just use regular React state (rather than Redux state) to update the mode.

src/views/search/search.jsx
```diff
@@ -47,6 +51,20 @@ class Search extends React.Component {
...
+            mode = query.substring(m + 5, query.length).toLowerCase();
+        }
+        while (mode.indexOf('/') > -1) {
+            mode = mode.substring(0, term.indexOf('/'));
+        }
+        while (term.indexOf('&') > -1) {
+            mode = mode.substring(0, term.indexOf('&'));
+        }
+        mode = decodeURIComponent(mode.split('+').join(' '));
+        this.props.dispatch(navigationActions.setMode(mode));
```
rschamp: Here, instead of dispatching, just use `this.setState({mode: mode})` to set the `mode` state.

src/views/search/search.jsx
```diff
@@ -60,11 +78,20 @@ class Search extends React.Component {
         }
         const start = pathname.lastIndexOf('/');
         const type = pathname.substring(start + 1, pathname.length);
+        const modeOptions = ['trending', 'popular', ''];
```
rschamp: I don't think this belongs in state, it's a constant. Instead of defining it here, just put something like `const ACCEPTABLE_MODES = ['trending', 'popular', '']` at the top of the file, and then refer to `ACCEPTABLE_MODES` instead of `this.state.acceptableModes`.

src/views/search/search.jsx
```diff
@@ -73,7 +100,8 @@ class Search extends React.Component {
         const locale = this.props.intl.locale;
         const loadNumber = this.state.loadNumber;
         const offset = this.state.offset;
-        const queryString = `limit=${loadNumber}&offset=${offset}&language=${locale}&mode=popular${termText}`;
+        const mode = this.state.mode;
```
rschamp: Since you're already referring to `this.state.mode` here, the change from Redux state to local React state should work.

src/views/search/search.jsx
kyleplo: I got a Travis error when I did this https://travis-ci.org/LLK/scratch-www/builds/419141652#L761

rschamp: Oh that's right — none of this needs to happen in `componentDidMount`, it can all happen in the constructor.  And if you do it in the constructor, then you don't need to use `setState`, you can just initialize the state with this value.

BryceLTaylor:
We found an issue when we got into production where search defaulted no category and it not handling that situation very well.
@chrisgarrity fixing wtih https://github.com/LLK/scratch-www/pull/2057

Issue #1992, opened 2018-07-23
Audit "darken" and "lighten" scss 


carljbowman:
### Overview
Sometimes the `darken` and `lighten` scss properties can break with the new colors. This can cause feedback issues like https://github.com/LLK/scratch-www/pull/1990

Likely the fix for `darken`'s will be to bring in all the shades of the 3.0 colors: primary, secondary, and tertiary. 

`ligthen`'s will need some more thought.

Narulacodes:
If we add this, `I'll thake this.`
`Darken` `Lighten`

carljbowman:
@Narulacodes - Thanks for the offer. But I just wrapped up work on this and will be submitting a PR later this morning.

If you are looking for something to help with, look for issues with the `help wanted` label.

carljbowman:
Fixed by https://github.com/LLK/scratch-www/pull/2007#pullrequestreview-142097092

PR #2007, opened 2018-07-31
Remove Darken and Ligthen SCSS

carljbowman:
Updated all the places in the SCSS where we use Darken and Lighten properties.

These properties have been problematic in the past and caused some state issues, e.g. New message(s) hightlight (which was fixed seperately from this issue)

### Resolves:

https://github.com/LLK/scratch-www/issues/1992

### Changes:

Replaced all instances of Lighten and Darken with actually color variables defined by the global color styles sheet.


src/views/tips/tips.scss
```diff
@@ -31,25 +31,23 @@ $base-bg: $ui-white;
...
     margin-right: .75rem;
     background-color: $ui-blue;
     color: $ui-white;
     font-size: 1rem;
-    
+
     &.getting-started-button {
         margin-right: 0;
-        background-color: $darken-button;
+        background-color: $ui-aqua;
```
technoboy10: The button here looks like an actual button now (which it didn't really before), but it seems like it might be a little too high contrast?
screenshot 2018-07-31 15 01 03

src/components/spinner/spinner.scss
```diff
@@ -18,13 +18,13 @@
...
             border-radius: 100%;
-            background-color: darken($ui-white, 8%);
+            background-color: $ui-gray;
             width: 15%;
             height: 15%;
             content: "";
 
             .white & {
-                background-color: $ui-blue-dark    
+                background-color: $ui-blue-dark
```
technoboy10: Missing a semicolon at the end of this line?

src/components/forms/charcount.scss
```diff
@@ -2,7 +2,7 @@
 
 .char-count {
     letter-spacing: 1px;
-    color: lighten($type-gray, 30%);
+    color: $type-gray-75percent;
```
technoboy10: This doesn't look the same, but that may have been intentional.

Before:


After:

carljbowman: Thanks for calling this out, but this update is okay!

PR #2015, opened 2018-08-06
report project POSTs to scratchr2, displays modal reactively

### LINKED ISSUE #1957: Connect report project modal to API
Connect the modal created for https://github.com/LLK/scratch-www/issues/1857 (included in https://github.com/LLK/scratch-www/pull/1900) to the API endpoint for reporting projects.

benjiwheeler:
Resolves #1957.

When tested using Docker, depends on https://github.com/LLK/scratchr2/pull/5133 .

Looks like:

screen shot 2018-08-06 at 12 31 54 pm

screen shot 2018-08-06 at 12 32 14 pm

screen shot 2018-08-06 at 12 32 32 pm

screen shot 2018-08-06 at 2 48 49 pm

screen shot 2018-08-06 at 12 32 41 pm 2

screen shot 2018-08-06 at 12 31 40 pm

Note that this does NOT use a scratch-api endpoint at all; it just uses the existing scratchr2 endpoint.

We ran into difficulties getting scratch-api to read the cookie from scratch-www, in the process of sending a proxy request from scratch-api to scratchr2. We decided to come back to this at a a later time.
 
Why setting headers in www doesn’t work:
* Currently can get csrf token from cookie fine; but can’t get sessionid token that way, because it’s set as http only by scratchr2. So it’s unreadable!

Why sending whole cookie from www to api doesn’t work:
* So, it CAN, it’s just tricky.
* To send cookie, you have xhr options include withCredentials: true. That will send cookie.
* But Chrome won’t show you the response from api, because it wants api to set Access-Control-Allow-Credentials header, and we’re not.
    * But we could! BUT, restifycors will complain and error if you set that header while allowing wildcard incoming domains...

### DIFF
--- a/src/components/forms/validations.jsx
+++ b/src/components/forms/validations.jsx
@@ -28,8 +28,8 @@ module.exports.validationHOCFactory = defaultValidationErrors => (Component => {
         <Component
             validationErrors={defaults(
                 {},
-                defaultValidationErrors,
-                props.validationErrors
+                props.validationErrors,
+                defaultValidationErrors
             )}
             {...omit(props, ['validationErrors'])}
         />
--- a/src/components/modal/report/modal.jsx
+++ b/src/components/modal/report/modal.jsx
@@ -1,10 +1,12 @@
 const bindAll = require('lodash.bindall');
 const PropTypes = require('prop-types');
 const React = require('react');
+const connect = require('react-redux').connect;
 const FormattedMessage = require('react-intl').FormattedMessage;
 const injectIntl = require('react-intl').injectIntl;
 const intlShape = require('react-intl').intlShape;
 const Modal = require('../base/modal.jsx');
+const classNames = require('classnames');
 
 const Form = require('../../forms/form.jsx');
 const Button = require('../../forms/button.jsx');
@@ -12,6 +14,7 @@ const Select = require('../../forms/select.jsx');
 const Spinner = require('../../spinner/spinner.jsx');
 const TextArea = require('../../forms/textarea.jsx');
 const FlexRow = require('../../flex-row/flex-row.jsx');
+const previewActions = require('../../../redux/preview.js');
 
 require('../../forms/button.scss');
 require('./modal.scss');
@@ -68,12 +71,24 @@ class ReportModal extends React.Component {
     constructor (props) {
         super(props);
         bindAll(this, [
-            'handleReportCategorySelect'
+            'handleCategorySelect',
+            'handleValid',
+            'handleInvalid'
         ]);
-        this.state = {reportCategory: this.props.report.category};
+        this.state = {
+            category: '',
+            notes: '',
+            valid: false
+        };
     }
-    handleReportCategorySelect (name, value) {
-        this.setState({reportCategory: value});
+    handleCategorySelect (name, value) {
+        this.setState({category: value});
+    }
+    handleValid () {
+        this.setState({valid: true});
+    }
+    handleInvalid () {
+        this.setState({valid: false});
     }
     lookupPrompt (value) {
         const prompt = REPORT_OPTIONS.find(item => item.value === value).prompt;
@@ -82,17 +97,24 @@ class ReportModal extends React.Component {
     render () {
         const {
             intl,
+            isConfirmed,
+            isError,
+            isOpen,
+            isWaiting,
             onReport, // eslint-disable-line no-unused-vars
-            report,
+            onRequestClose,
             type,
             ...modalProps
         } = this.props;
+        const submitEnabled = this.state.valid && !isWaiting;
+        const submitDisabledParam = submitEnabled ? {} : {disabled: 'disabled'};
         const contentLabel = intl.formatMessage({id: `report.${type}`});
         return (
             <Modal
                 className="mod-report"
                 contentLabel={contentLabel}
-                isOpen={report.open}
+                isOpen={isOpen}
+                onRequestClose={onRequestClose}
                 {...modalProps}
             >
                 <div>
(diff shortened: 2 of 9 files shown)


src/redux/preview.js
```diff
@@ -490,3 +495,28 @@ module.exports.updateProject = (id, jsonData, username, token) => (dispatch => {
         dispatch(module.exports.setProjectInfo(body));
     });
 });
+
+module.exports.reportProject = (id, jsonData, username, token) => (dispatch => {
+    dispatch(module.exports.setFetchStatus('report', module.exports.Status.FETCHING));
+    // scratchr2 will fail if no thumbnail base64 string provided. We don't yet have
+    // a way to get the actual project thumbnail in www/gui, so for now just submit
+    // a minimal base64 png string.
+    defaults(jsonData, {
```
benjiwheeler: I found base64 code for a 1x1 white png, thought this seems safest.

src/redux/preview.js
```diff
@@ -258,6 +260,7 @@ module.exports.getFavedStatus = (id, username, token) => (dispatch => {
 });
 
 module.exports.setFavedStatus = (faved, id, username, token) => (dispatch => {
+    dispatch(module.exports.setFetchStatus('faved', module.exports.Status.FETCHING));
```
benjiwheeler: These seem to have been missing

src/l10n.json
```diff
@@ -172,6 +172,7 @@
     "registration.welcomeStepTitle": "Hurray! Welcome to Scratch!",
 
     "thumbnail.by": "by",
+    "report.error": "There was an error with your report submission. Maybe try again?",
```
benjiwheeler: Who should write/review this text? I just made it up :)

src/components/modal/report/modal.jsx
```diff
@@ -104,68 +120,104 @@ class ReportModal extends React.Component {
 
                     <Form
                         className="report"
-                        onSubmit={onReport}
+                        onInvalid={this.handleInvalid}
```
benjiwheeler: These "onValid"/"onInvalid" properties are a Formsy thing. Our Form component doesn't have specific support for them, but they seem to work fine just passed through this way. Is this kosher?

carljbowman:
Overall everything is looking great. Just a few follow up questions and notes. 

## Failed Submission Feedback
### Current
43729087-275a7736-9975-11e8-8b19-9f72e149a281

### Reasons for Failed Submission
Seems like we could use a few different messages based on why the submission wasn't made:
- No Subject Selected
- Empty Message 
- General for the case the wifi drops or there is no API callback

### Examples
Typically we use the orange callout style for this type of feedback, the same style that is being used in the confirmation (which we should switch, but more on that below). These can be more contextual like a tool tip or appear below the form.

Here is an example in the Teacher Account Registration Flow (on WWW): 
screen shot 2018-08-06 at 2 53 49 pm

Alt view triggered by responsive 
screen shot 2018-08-06 at 2 58 55 pm

### Proposals

reason tip

message tip

general tip

## Confirmation
43729098-31e597ee-9975-11e8-858d-a962c2996eec

We should use a different style to communicate this feedback as to not conflate the tooltip style. Changing the layout and displaying the message should be enough to confirm.

### Proposal
report confirmation

src/redux/preview.js
```diff
@@ -556,3 +561,28 @@ module.exports.updateProject = (id, jsonData, username, token) => (dispatch => {
...
+    });
+    api({
+        host: '',
+        uri: `/site-api/projects/all/${id}/report/`,
+        authentication: token,
+        method: 'POST',
+        json: jsonData,
+        useCsrf: true
+    }, (err, body, res) => {
+        if (err || res.statusCode >= 300) {
```
rschamp: What is the `3XX` case?

src/views/preview/preview.jsx
```diff
@@ -67,6 +69,18 @@ class Preview extends React.Component {
...
         this.addEventListeners();
     }
+    componentWillUpdate (nextProps, nextState) {
+        if (this.props.reportStatus !== nextProps.reportStatus) {
+            this.setState({
+                report: {
+                    ...nextState.report,
+                    confirmed: (nextProps.reportStatus === previewActions.Status.FETCHED),
+                    error: (nextProps.reportStatus === previewActions.Status.ERROR),
+                    waiting: (nextProps.reportStatus === previewActions.Status.FETCHING)
```
rschamp: It's not recommended to set state based on props for the same reason that it's a linter error to `setState` in `componentDidUpdate` — it's possible to cause unnecessary (or infinite) re-renders. Since these are all calculations based on current props, then this is probably best placed in `render` where we pass this state on to the next component.

Since this is coming out of redux, maybe the modal should just get it from there directly?

src/redux/preview.js
```diff
@@ -556,3 +561,28 @@ module.exports.updateProject = (id, jsonData, username, token) => (dispatch => {
...
+    // a way to get the actual project thumbnail in www/gui, so for now just submit
+    // a minimal base64 png string.
+    defaults(jsonData, {
+        thumbnail: 'iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC' +
+            '0lEQVR42mP8/x8AAwMCAO+ip1sAAAAASUVORK5CYII='
+    });
+    api({
+        host: '',
+        uri: `/site-api/projects/all/${id}/report/`,
+        authentication: token,
```
rschamp: Is this needed?

src/redux/preview.js
benjiwheeler: In development, we were seeing that in some cases scratchr2 may return a 3XX, while not adding the report to the db. If we see that, we should consider the report submission to have failed.

benjiwheeler: You could make a case that this should be >= 204 or something else...

benjiwheeler:
@carljbowman Take a look:

screen shot 2018-08-07 at 2 32 30 pm
screen shot 2018-08-07 at 2 32 45 pm
screen shot 2018-08-07 at 2 33 06 pm
screen shot 2018-08-07 at 2 15 07 pm
screen shot 2018-08-07 at 2 46 56 pm

src/redux/preview.js
benjiwheeler: good catch, not needed. removed.

carljbowman:
@benjiwheeler - Looking good! Just one nitpick and question. Thanks for making those updates!

**Nitpick - Border on field with error** 
Could we match the field state state in the Teach Account flow: on validation, the border on the field with an error becomes orange to help clarify which field needs to be modified. 

**Question**
screen shot 2018-08-07 at 2 15 07 pm

Is there just an extra "new line" before the text in the message field? or is the text-alignment off?

benjiwheeler:
@carljbowman Here you go:

screen shot 2018-08-07 at 3 51 59 pm
screen shot 2018-08-07 at 3 30 39 pm

carljbowman:
@benjiwheeler - Woo! That's great. Thanks for making that change.

src/components/forms/validations.jsx
```diff
@@ -28,8 +28,8 @@ module.exports.validationHOCFactory = defaultValidationErrors => (Component => {
         <Component
             validationErrors={defaults(
                 {},
```
benjiwheeler: So there's a bug fix here, I think

the order of these params was wrong, I think... it kept you from being able to override the default message for formsy's isDefaultRequiredValue

@rschamp, LMK if you think it would make sense to make this a separate PR instead. (We'd lose the custom messages here until that was merged in.)

src/views/preview/preview.jsx
benjiwheeler: Done. See new code for modal/report/modal.jsx

src/components/forms/validations.jsx
rschamp: This new order looks correct to me. I think it's fine to include it here.

src/components/modal/report/modal.jsx
```diff
@@ -82,17 +97,25 @@ class ReportModal extends React.Component {
...
             onReport, // eslint-disable-line no-unused-vars
-            report,
+            onRequestClose,
+            reportStatus,
             type,
             ...modalProps
         } = this.props;
+        const isConfirmed = reportStatus === previewActions.Status.FETCHED;
+        const isError = reportStatus === previewActions.Status.ERROR;
+        const isWaiting = reportStatus === previewActions.Status.FETCHING;
```
rschamp: This is looking good! One further improvement would be to move this logic into `mapStateToProps` so the component would receive these as props, `isConfirmed`, `isError` and `isWaiting`.

src/redux/preview.js
rschamp: I think the case I would make is that `res.statusCode !== 200` is the error state. I think this is fine as-is though.

src/components/modal/report/modal.jsx
benjiwheeler: Done.

DeleteThisAcount:
this looks great! is this going to be the 3.0 project player page? is there a repo for this?

antcto:
@benjiwheeler where add the withCredentials for xhr in scratch-www when request with the ur of csrf_token? i can not get the cookie in client browser from api server.

PR #2017, opened 2018-08-07
project page stage margin, width adjustments to match gui css changes

benjiwheeler:
resolves https://github.com/LLK/scratch-gui/issues/2132

must be checked in at the same time as https://github.com/LLK/scratch-gui/pull/2807


src/views/preview/preview.scss
```diff
@@ -1,14 +1,10 @@
...
 
-/* stage size constants 
- * this is a hack right now - stage includes padding of .5rem (8px) for alignment in gui
- * in www the player is placed with margin -.5rem to align the edge.
- * the height is calculated from the actual height on the page (404)
-*/
-$gui-width: 496px;
+/* stage size constants */
+$gui-width: 482px;
+$gui-height: 406px;
```
benjiwheeler: bottom of stage player seemed slightly off at 404px

src/views/preview/preview.scss
```diff
@@ -174,14 +170,13 @@ $stage-height: 404px;
 
     .guiPlayer {
         display: inline-block;
-        margin-left: -.5rem;
```
benjiwheeler: here's where we remove the margin hack! (hard to see because of all the whitespace changes)

src/views/preview/preview.scss
benjiwheeler: I hear you.

"gui-width" seemed to refer to the whole width of the area around the stage, while "stage-width" seemed to refer to the raw width of just the actual stage. The approx. 404px (406 here) included a bunch of other stuff besides the stage, so it seemed to fit more with "gui-" than with "stage-".

How would you feel about "player-width" and "player-height" instead of "gui-"?

src/views/preview/preview.scss
```diff
@@ -29,49 +25,49 @@ $stage-height: 404px;
     .project-title {
         font-size: 1.75rem;
         font-weight: 500;
-        
+
```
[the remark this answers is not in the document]
benjiwheeler: No, my editor just removes extra spaces, and I think that's good form.

src/views/preview/preview.scss
benjiwheeler: I hear you.

gui-width seems to refer to the whole area around the player; stage-width seems to refer just to the stage. 404/406 is the whole height around the player, so it seems to fit more with gui- than with stage- .

I like your "player-" idea, and used that -- what do you think?

src/views/preview/preview.scss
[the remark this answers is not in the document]
benjiwheeler: No, my editor just removes them automatically, and I think removing them is good form -- it prevents unintentional diffs in the future.

benjiwheeler:
Sorry @chrisgarrity, my force-pushing a fixup destroyed your previous comment. 

"gui-width" seemed to refer to the whole width of the area around the stage, while "stage-width" seemed to refer to the raw width of just the actual stage. The approx. 404px (406 here) included a bunch of other stuff besides the stage, so it seemed to fit more with "gui-" than with "stage-".

I like your 'player-" suggestion -- changed to use that. What do you think?

PR #2022, opened 2018-08-12
split account nav into its own component, pass logout callback to GUI

benjiwheeler:
This is a first step in identifying the account nav menu info and styles that will be needed for replicating the menu for GUI, per  https://github.com/LLK/scratch-gui/issues/2816 .

Note that the component can almost handle its own open/closed state, but the navigation component currently overrides this state when there are updates to she session, and that seemed beyond the scope of this component.

### DIFF
--- a/src/components/navigation/www/accountnav.jsx
+++ b/src/components/navigation/www/accountnav.jsx
@@ -21,7 +21,7 @@ const AccountNav = ({
     onClickLogout,
     onClose
 }) => (
-    <div id="account-nav">
+    <div className="account-nav">
         <a
             className={classNames({
                 'user-info': true,
--- a/src/components/navigation/www/accountnav.jsx
+++ b/src/components/navigation/www/accountnav.jsx
@@ -7,7 +7,7 @@ const React = require('react');
 const Avatar = require('../../avatar/avatar.jsx');
 const Dropdown = require('../../dropdown/dropdown.jsx');
 
-require('./navigation.scss');
+require('./accountnav.scss');
 
 const AccountNav = ({
     classroomId,
@@ -21,7 +21,7 @@ const AccountNav = ({
     onClickLogout,
     onClose
 }) => (
-    <React.Fragment>
+    <div id="account-nav">
         <a
             className={classNames({
                 'user-info': true,
@@ -82,7 +82,7 @@ const AccountNav = ({
                 </a>
             </li>
         </Dropdown>
-    </React.Fragment>
+    </div>
 );
 
 AccountNav.propTypes = {
--- a/src/components/navigation/www/accountnav.jsx
+++ b/src/components/navigation/www/accountnav.jsx
@@ -1,0 +1,101 @@
+const classNames = require('classnames');
+const FormattedMessage = require('react-intl').FormattedMessage;
+const injectIntl = require('react-intl').injectIntl;
+const PropTypes = require('prop-types');
+const React = require('react');
+
+const Avatar = require('../../avatar/avatar.jsx');
+const Dropdown = require('../../dropdown/dropdown.jsx');
+
+require('./navigation.scss');
+
+const AccountNav = ({
+    classroomId,
+    isEducator,
+    isOpen,
+    isStudent,
+    profileUrl,
+    thumbnailUrl,
+    username,
+    onClick,
+    onClickLogout,
+    onClose
+}) => (
+    <React.Fragment>
+        <a
+            className={classNames({
+                'user-info': true,
+                'open': isOpen
+            })}
+            href="#"
+            onClick={onClick}
+        >
+            <Avatar
+                alt=""
+                src={thumbnailUrl}
+            />
+            <span className="profile-name">
+                {username}
+            </span>
+        </a>
+        <Dropdown
+            as="ul"
+            className={process.env.SCRATCH_ENV}
+            isOpen={isOpen}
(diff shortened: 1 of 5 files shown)


src/components/navigation/www/accountnav.jsx
```diff
@@ -0,0 +1,101 @@
...
+    isOpen,
+    isStudent,
+    profileUrl,
+    thumbnailUrl,
+    username,
+    onClick,
+    onClickLogout,
+    onClose
+}) => (
+    <React.Fragment>
```
rschamp: I think it would make sense for this to include the wrapping component. What is the reason for leaving it out?

benjiwheeler: Sorry, I don't understand. Which wrapping component? NavigationBox?

rschamp: I mean that this is returning a `Fragment` rather than a self-contained component. Where it's pulled from is a `li` (but sometimes we make that element type configurable with an `as` prop for stuff like this).

benjiwheeler: Hm, I think we need to talk offline, I have several questions

src/components/navigation/www/accountnav.jsx
```diff
@@ -0,0 +1,101 @@
+const classNames = require('classnames');
+const FormattedMessage = require('react-intl').FormattedMessage;
+const injectIntl = require('react-intl').injectIntl;
+const PropTypes = require('prop-types');
+const React = require('react');
+
+const Avatar = require('../../avatar/avatar.jsx');
+const Dropdown = require('../../dropdown/dropdown.jsx');
+
+require('./navigation.scss');
```
rschamp: From offline: this component should import its own styles, rather than the styles of the navigation component

src/components/navigation/www/accountnav.scss
```diff
@@ -0,0 +1,91 @@
+@import "../../../colors";
+@import "../../../frameless";
+
+#account-nav {
```
rschamp: Please scope with a class rather than an id. In this case it's weird because there will practically only be one nav on a page, but since these are components, we shouldn't assume that.  Eventually we should try to get rid of all the id selectors in this repo.

src/components/navigation/www/navigation.jsx
```diff
@@ -437,6 +391,11 @@ const mapStateToProps = state => ({
     searchTerm: state.navigation
 });
 
-const ConnectedNavigation = connect(mapStateToProps)(Navigation);
+const mapDispatchToProps = () => ({});
```
rschamp: Thank you!

src/components/navigation/www/navigation.jsx
```diff
@@ -113,6 +113,8 @@ class Navigation extends React.Component {
     handleCloseLogin () {
         this.setState({loginOpen: false});
     }
+    // NOTE: TODO: continue here. Should move these two functions up to a redux level,
+    // maybe into session...
```
rschamp: 👍

src/views/preview/preview.jsx
```diff
@@ -339,6 +356,7 @@ class Preview extends React.Component {
                     className="gui"
                     projectHost={this.props.projectHost}
                     projectId={this.state.projectId}
+                    onClickLogout={this.handleLogout}
```
rschamp: Does this exist yet?

src/views/preview/preview.jsx
```diff
@@ -315,7 +333,6 @@ class Preview extends React.Component {
                         replies={this.props.replies}
                         reportOpen={this.state.reportOpen}
                         studios={this.props.studios}
-                        user={this.props.user}
```
rschamp: I don't see why this is being removed, is this cruft?

src/components/navigation/www/accountnav.scss
rschamp: Also it looks like we're still styling some things in `navigation.scss` using `.account-nav`, so those would either need to be updated or removed.

benjiwheeler: Re-did with class="account-nav" instead. Removed redundant lines from navigation.scss.

src/components/navigation/www/navigation.jsx
benjiwheeler: 👍🏼

src/views/preview/preview.jsx
benjiwheeler: Yes, it was cruft, unused in preview presentation.

src/views/preview/preview.jsx
benjiwheeler: Yes, it is earlier in the file.

rschamp: Sorry I mean the `onClickLogout` prop on the `GUI` component. Adding this would probably end up passing it down to the final output component and cause a warning.  It's probably fine for this to get merged out of sync though.

rschamp:
@BryceLTaylor This affects the navbar for www pages. For QA we should just make sure the account navigation dropdown still works well on the pages that aren't the preview.

BryceLTaylor:
This seems to have broken logging in.  I believe @benjiwheeler is working on it.

benjiwheeler:
Login issue fixed by https://github.com/LLK/scratch-www/pull/2068 .

PR #2033, opened 2018-08-21
Recongfigure docker external network usage

colbygk:
### Resolves:

Improves development ergonomics when working against docker based local development.

### Changes:

This moves the docker network handling to a separate tool that is part of a new npm module: `scratch-docker`. All other repos that depend on having docker network communication will also use this tool to converge on the named network.

This requires a step to run the network setup ahead of launching docker compose, this is accomplished by using:

```bash
$ make docker-up
and/or
$ make docker-down
```

Looks like I accidentally added this branch to LLK instead of my fork.
Related to: LLK/scratch-api/pull/538

### Test Coverage:

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,24 @@
 ESLINT=./node_modules/.bin/eslint
 NODE= NODE_OPTIONS=--max_old_space_size=8000 node
 SASSLINT=./node_modules/.bin/sass-lint -v
+SCRATCH_DOCKER_CONFIG=./node_modules/.bin/docker_config.sh
 S3CMD=s3cmd sync -P --delete-removed --add-header=Cache-Control:no-cache,public,max-age=3600
 TAP=./node_modules/.bin/tap
 WATCH= NODE_OPTIONS=--max_old_space_size=8000 ./node_modules/.bin/watch
 WEBPACK= NODE_OPTIONS=--max_old_space_size=8000 ./node_modules/.bin/webpack
+
+
+# ------------------------------------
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,11 +1,25 @@
...
 S3CMD=s3cmd sync -P --delete-removed --add-header=Cache-Control:no-cache,public,max-age=3600
 TAP=./node_modules/.bin/tap
 WATCH= NODE_OPTIONS=--max_old_space_size=8000 ./node_modules/.bin/watch
 WEBPACK= NODE_OPTIONS=--max_old_space_size=8000 ./node_modules/.bin/webpack
 
+
+# ------------------------------------
+
+$(SCRATCH_DOCKER_CONFIG):
+	npm install scratch-docker
```
rschamp: Please remove, since it should be a dependency

Makefile
```diff
@@ -1,11 +1,25 @@
...
+
+docker-up: $(SCRATCH_DOCKER_CONFIG)
```
rschamp: Please add to scripts in package.json

Makefile
colbygk: But!  Part of the point of using docker is to have the npm install happen inside the docker container, which may not happen until after you've run docker-compose up (aka make docker-up), which means `scratch-docker` won't be available as a dependency yet.

Makefile
colbygk: Rockin

Issue #2016, opened 2018-08-06
Clip text in Add to Studio buttons


benjiwheeler:
Text is currently limited by number of characters.

Add to or replace this by using CSS properties:

    overflow: hidden;
    text-overflow:ellipsis;

See: https://www.w3schools.com/cssref/tryit.asp?filename=trycss3_text-overflow_hover

LiFaytheGoblin:
also needs white-space: nowrap; for the ... to show up

PR #2037, opened 2018-08-22
2016/add to studio dialog clip studio name

LiFaytheGoblin:
### Resolves:

#2016 

### Changes:

First I added some css to make the studio name in the add-to-studio-dialog on the new project page truncate nicely (thanks Ben for sharing your notes). Then I removed the truncate function from the react code and added a title-attribute to the div of the studio name, so the full name gets displayed on hovering. Then I removed the whole truncate library and also the lodash stuff that was only used by this library.

### Test Coverage:

I ran npm test and it gave me some warnings and errors I was not responsible for. Tested for Chrome and Safari on MacOS.


LiFaytheGoblin:
I'll try to get the sort order right, sorry

Issue #1986, opened 2018-07-21
Show full project/studio title in the hover title text


SheepTester:
Many projects have titles that are truncated, and the only way to see the full title is to load the project page.

It would be nice if the titles showed the full title when you hover over them like YouTube videos.

### Steps to Reproduce

Go to the front page and note that some project titles are cut off. Hovering over them shows nothing.

### Operating System and Browser

ChromeOS 10575.58.0, Chrome 67.0.3396.99

LiFaytheGoblin:
This involves several things:
- [ ] on project pages, hovering over a title should display the full title
- [ ] on studio pages, hovering over a title should display the full title
- [ ] on other pages where there is a truncated project or studio name, eg. the what's happening section, on hovering it should display the full title

LiFaytheGoblin:
Actually, the project name should be cut off at some point.

<img width="1223" alt="screen shot 2018-08-21 at 11 13 04 am" src="https://user-images.githubusercontent.com/5471273/44410841-3f607000-a533-11e8-910d-b162d948766d.png">

LiFaytheGoblin:
With a space:

<img width="985" alt="screen shot 2018-08-21 at 11 15 39 am" src="https://user-images.githubusercontent.com/5471273/44411020-99613580-a533-11e8-85db-05d02dbb1089.png">

LiFaytheGoblin:
Compare with short title

<img width="727" alt="screen shot 2018-08-21 at 11 17 08 am" src="https://user-images.githubusercontent.com/5471273/44411120-cdd4f180-a533-11e8-98d4-f7b27ae56d85.png">

LiFaytheGoblin:
Add to scss:
```
overflow: hidden;
text-overflow: ellipsis;
white-space: nowrap;
```
Add title attribute with full title to visible project title

PR #2043, opened 2018-08-23
Keep title on preview project page from overflowing or otherwise messing up the layout

LiFaytheGoblin:
### Resolves:

Part of #1986 

### Changes:

* added a title attribute to project title so you can see the whole title even when it is not completely displayed on the website
* made it so long titles don't get displayed completely, see comments in scss file

### Test Coverage:
npm test passed
Tested in MacOS Chrome & Safari, looks good for screen with min-width 640px

screen shot 2018-08-23 at 2 32 50 pm

screen shot 2018-08-23 at 2 32 09 pm

### DIFF
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -32,6 +32,14 @@ $stage-width: 480px;
                 transform: translate(22rem, 0);
             }
         }
+
+        &.no-edit {
+            /* titles of projects you don't own should not
+            show the full title if this is too long */
+            overflow: hidden;
+            text-overflow: ellipsis;
+            white-space: nowrap;
+        }
     }
 
     .project-header {
@@ -39,6 +47,8 @@ $stage-width: 480px;
         flex-grow: 1;
         justify-content: flex-start;
         align-items: flex-start;
+        flex-wrap: nowrap;
+        width: 65%; /* Setting a width to the header will keep it from flowing over the usual content of the page */
 
         .inplace-input {
             height: calc(3rem - 4px);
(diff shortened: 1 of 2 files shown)


src/views/preview/preview.scss
```diff
@@ -32,13 +32,23 @@ $stage-width: 480px;
...
+            show the full title if this is too long */
+            overflow: hidden;
+            text-overflow: ellipsis;
+            white-space: nowrap;
+        }
     }
 
     .project-header {
         margin-right: 2rem;
+        width: 65%; /* Setting a width to the header will keep it from flowing over the usual content of the page */
```
rschamp: Where does `65%` come from? I would expect this to be 100% and constrained by the parent component with margins or flex-basis or something like that?

LiFaytheGoblin: I have tried for two hours now to get it to work without setting the width this way but it doesn't. Flex basis seems to require a width also. Changing settings for the button doesn't change the width of that input field either. I think putting the width here is the easiest and for what I have tested safest thing we can do.

LiFaytheGoblin: ACTUALLY this is so interesting that I am now reading more about flex box in general and I might have found a solution. Will investigate further. Hint: min-width 0!

LiFaytheGoblin: OMG RAY IT WORKS!

LiFaytheGoblin:
Found a nice way to do this thanks to https://hackernoon.com/11-things-i-learned-reading-the-flexbox-spec-5f0c799c776b and the hint about min-width

LiFaytheGoblin:
@rschamp Thanks ;D Should I merge?

Issue #2045, opened 2018-08-23
Studio preview icons don't look very nice


LiFaytheGoblin:
So currently the studio preview icons look this way:
<img width="282" alt="screen shot 2018-08-23 at 3 50 52 pm" src="https://user-images.githubusercontent.com/5471273/44548534-55fdf700-a6ec-11e8-967f-d32e5aac2465.png">

But @carljbowman intended them to look differently, which is more like this (what would the studio icon ratio be?):
<img width="175" alt="screen shot 2018-08-23 at 3 50 08 pm" src="https://user-images.githubusercontent.com/5471273/44548495-3d8ddc80-a6ec-11e8-9022-8890c6b1029c.png">

### Steps to Reproduce
Look at the preview page of a project that is in a studio.

### Operating System and Browser
MacOS, Chrome

LiFaytheGoblin:
Note: Ratio should stay the same

LiFaytheGoblin:
Also Remixes will look different
Before:
<img width="291" alt="screen shot 2018-08-24 at 8 43 04 am" src="https://user-images.githubusercontent.com/5471273/44585314-e9313e00-a779-11e8-8edf-646b97e59e73.png">
After:
<img width="274" alt="screen shot 2018-08-24 at 8 44 51 am" src="https://user-images.githubusercontent.com/5471273/44585339-fcdca480-a779-11e8-81c7-e1593c65aaad.png">

LiFaytheGoblin:
For the headlines 'remix', 'studios' etc:

```
    align-self: flex-start;
    margin-left: 14px;
    font-size: 1.2rem;
    font-weight: bold;
```

LiFaytheGoblin:
.thumbnail-column .thumbnail don't need padding

PR #2046, opened 2018-08-24
Fix appearance of studio thumbnails on preview project page

LiFaytheGoblin:
### Resolves:
#2045 

### Changes:
- Changed appearance of thumbnails
- Cutting of title under thumbnail looks better 
- Hovering over title of thumbnail shows full title, this is a change that affects all thumbnails anywhere on the website

### Test Coverage:
npm test passed
Tested on MacOS Chrome and Safari
screen shot 2018-08-24 at 9 27 41 am


src/components/thumbnailcolumn/thumbnailcolumn.scss
```diff
@@ -5,25 +5,26 @@
...
     flex-direction: column;
 
     .thumbnail {
         margin: 7px;
-        border-radius: 4px;
-        box-shadow: 0 0 0 1px $active-gray;
-        background-color: $ui-white;
-        padding-bottom: 4px;
+        // unnecessary:
+        // padding-bottom: 4px;
```
rschamp: `//` is invalid CSS, you should use `/* */` instead. I think SCSS might allow it but we shouldn't depend on that.  Also if it's unnecessary, you can just delete it :)

src/views/preview/remix-list.jsx
```diff
@@ -9,9 +9,9 @@ const RemixList = props => {
     if (remixes.length === 0) return null;
     return (
         <FlexRow className="remix-list">
-            <div className="project-title">
+            <h2>
                 Remixes
-            </div>
+            </h2>
```
rschamp: Please use a specific class name like it was doing before instead of relying on the style of `h2`. Very soon we will need to stop using `h1`, `h2`, etc. everywhere and switch to this method instead, so we may as well not add more debt.

src/views/preview/studio-list.jsx
```diff
@@ -8,10 +8,10 @@ const StudioList = props => {
     const studios = props.studios;
     if (studios.length === 0) return null;
     return (
-        <FlexRow className="remix-list">
-            <div className="project-title">
+        <FlexRow className="studio-list">
+            <h2>
                 Studios
-            </div>
+            </h2>
```
rschamp: Same request here re: `h2`

src/views/preview/preview.scss
```diff
@@ -486,8 +486,16 @@ $stage-width: 480px;
...
+            margin-left: 1rem;
+            font-size: 1.2rem;
+            font-weight: bold;
+            align-self: flex-start;
+        }
+
         .project {
-            margin-bottom: 1.5rem;
+            // not necessary:
+            // margin-bottom: 1.5rem;
```
rschamp: Please remove

src/views/preview/remix-list.jsx
LiFaytheGoblin: Ah, switching to a class name will make this fix way more secure, good idea!

src/components/thumbnailcolumn/thumbnailcolumn.scss
LiFaytheGoblin: I forgot to remove the comments I put there when experimenting ._. will remove them

src/views/preview/remix-list.jsx
LiFaytheGoblin: (h1, h2 and so on tags might be important for screenreaders?)

LiFaytheGoblin: (h1, h2 and so on might be important for screen readers?)

LiFaytheGoblin:
I don't quite get why the build is failing...

Issue #2047, opened 2018-08-24
New project page: Make empty space next to comments smaller


LiFaytheGoblin:
<img width="1015" alt="screen shot 2018-08-24 at 9 38 57 am" src="https://user-images.githubusercontent.com/5471273/44587703-bf7c1500-a781-11e8-9934-5f5a4c9270c3.png">

### Steps to Reproduce
Look at a project page (new preview page)

### Operating System and Browser

MacOS, Chrome, Safari

PR #2048, opened 2018-08-24
2047/project page/white space next to comments

LiFaytheGoblin:
### Resolves:
#2047 

### Changes:
Just changed the width of the comments section a bit to look nicer

### Test Coverage:
Still works as 'good' with other screen sizes as before
npm test passed
MacOS Chrome


src/views/preview/preview.scss
```diff
@@ -124,7 +124,7 @@ $stage-width: 480px;
     }
 
     .comments-container {
-        width: 60%;
+        width: 65%;
```
rschamp: Hmm, I think we need to refactor if we are using magic numbers like these... Ideally the margins, padding, and flex-basis would constrain the proportions.

BryceLTaylor:
Looks good!

Issue #2049, opened 2018-08-24
Logging in on www in development does not appear to refresh session state in react


benjiwheeler:
Logging in on www in development does succeed in creating the session, but the user must reload to get the display components to reflect that.

### Steps to Reproduce

Start in the logged out state on the preview page.
![screen shot 2018-08-24 at 6 11 09 pm](https://user-images.githubusercontent.com/3431616/44610254-620bb680-a7c9-11e8-9775-2d639193719d.png)

Log in. 
![screen shot 2018-08-24 at 6 11 16 pm](https://user-images.githubusercontent.com/3431616/44610264-6a63f180-a7c9-11e8-92ce-1602b70f9d9c.png)

Hm, nothing happens...
![screen shot 2018-08-24 at 6 11 44 pm](https://user-images.githubusercontent.com/3431616/44610283-810a4880-a7c9-11e8-8b13-319de7069370.png)

Weird, if I click on "Sign in" again, it's still spinning...
![screen shot 2018-08-24 at 6 11 27 pm](https://user-images.githubusercontent.com/3431616/44610298-954e4580-a7c9-11e8-92c6-c2e0dbca4f98.png)

But, if I just reload, the login session is fine!
![screen shot 2018-08-24 at 6 11 44 pm](https://user-images.githubusercontent.com/3431616/44610347-c7f83e00-a7c9-11e8-81e0-d997cbf25357.png)

Mac OS 10.13.6, Chrome Version 68.0.3440.106 (Official Build) (64-bit)

benjiwheeler:
resolved by https://github.com/LLK/scratch-www/pull/2068

PR #2068, opened 2018-09-05
moved dispatched navigation function calls to mapDispatchToProps

benjiwheeler:
### Changes:

My commit https://github.com/LLK/scratch-www/commit/3c6530fe859753a3b7b2389659c576e1eaeacfc5 introduced a bug where after the user logs in, the view is not refreshed to reflect the logged in session.

The culprit was the dummy mapDispatchToProps that I introduced (to quiet a js console warning). navigation.jsx was using this.props.dispatch(...) in several places, but the dummy mapDispatchToProps meant that these dispatch calls never went through.

I solved this by making a proper mapDispatchToProps function and putting the dispatched function calls there.

### Test Coverage:

Tested that login and message count fetching work in Mac Chrome.

### DIFF
--- a/src/components/navigation/www/navigation.jsx
+++ b/src/components/navigation/www/navigation.jsx
@@ -54,9 +54,7 @@ class Navigation extends React.Component {
     componentDidMount () {
         if (this.props.session.session.user) {
             const intervalId = setInterval(() => {
-                this.props.dispatch(
-                    messageCountActions.getCount(this.props.session.session.user.username)
-                );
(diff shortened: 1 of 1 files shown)


rschamp:
@BryceLTaylor Please check on the homepage or any other www page:
- signing in works as you expect
- new messages increment the message counter in the menu bar

benjiwheeler:
resolves https://github.com/LLK/scratch-www/issues/2049

PR #2075, opened 2018-09-08
Fix gh-1934: Search bar special chars

### LINKED ISSUE #1934: Search bar only decodes special characters after switching to Studios tab and back
When you enter a search term with special characters, the search bar should show the decoded characters immediately after hitting enter.

Instead after entering a search you must click the Studios tab to see the updated characters.  You can switch back to the projects tab and the characters will be correct.

searchspecialcharacters

### Steps to Reproduce

enter the following things in the search bar and hit enter
  ?q=%D0%9A%D0%BE%D0%BC%D0%B0%D1%80 (should still appear as Комар)
  ?q=cat%3Fcat (should now appear as cat?cat)
  ?q=cat%20cat (should still appear as cat cat)
  ?q=CATCAT (should still appear as cat cat)
Note: the special characters aren't displayed properly
Click the Studios tab
Note: now the characters display correctly.
Click the projects tab
Note: the characters still display correctly

Chrome on Mac

chen-robert:
### Resolves:

Resolves #1934 

The text was encoded once already by the search bar. That means we need to decode it twice in order to decode special characters.

### Test Coverage:

?q=cat%3Fcat (should now appear as cat?cat)
?q=cat%20cat (should still appear as cat cat)

### DIFF
--- a/src/views/search/search.jsx
+++ b/src/views/search/search.jsx
@@ -57,19 +57,28 @@ class Search extends React.Component {
 
     }
     componentDidMount () {
-        const query = window.location.search;
-        const q = query.lastIndexOf('q=');
-        let term = '';
-        if (q !== -1) {
-            term = query.substring(q + 2, query.length).toLowerCase();
-        }
+        const query = decodeURIComponent(window.location.search);
+        let term = query;
+
+        const stripQueryValue = function (queryTerm) {
+            const queryIndex = query.indexOf('q=');
+            if (queryIndex !== -1) {
+                queryTerm = query.substring(queryIndex + 2, query.length).toLowerCase();
+            }
+            return queryTerm;
+        };
+        // Strip off the initial "?q="
+        term = stripQueryValue(term);
+        // Strip off user entered "?q="
+        term = stripQueryValue(term);
+
         while (term.indexOf('/') > -1) {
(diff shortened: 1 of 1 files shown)


chen-robert:
Oh, I think I misinterpreted the bug report.

To clarify, 
`?q=cat%3Fcat` appears as `cat?cat` 
What about `cat%3Fcat` alone? Should it be `cat?cat` or `cat%3Fcat`.
Also, what other flags are there? Should I translate `?q=cat%3Fcat` as well as aliases such as `?query=cat%3Fcat`?

chen-robert:
@benjiwheeler I've implemented the changes. I'm not sure if typing `?q=${searchText}` is user friendly though. Is that intended?

chen-robert:
@rschamp I've fixed the pull request. How should I proceed? It looks like its been sitting here for quite some time.

rschamp:
@chen-robert Sorry about that! @benjiwheeler will re-review soon.

benjiwheeler:
Sorry for the delay, @chen-robert! This looks good to me, and I tested it and it works great.

I think I misinterpreted the bug report too -- and the gif there was a bit confusing. You're right, we shouldn't really be worrying about users actually manually entering "?q=cat%20cat" into the search box!

Issue #2076, opened 2018-09-12
JS srrors on the splash page when running in development locally


benjiwheeler:
Several errors appear in the console on the splash page when running in development locally. These seem to be straightforward bugs where failure cases and prop types aren't being handled right.

![screen shot 2018-09-12 at 6 58 05 pm](https://user-images.githubusercontent.com/3431616/45457666-16f50d00-b6be-11e8-9f08-a6e05a9c42dc.png)

benjiwheeler:
Fixed by https://github.com/LLK/scratch-www/pull/2077

PR #2077, opened 2018-09-12
fixed errors in splash page that appear in develop

benjiwheeler:
See https://github.com/LLK/scratch-www/issues/2076 . 

These seem to be straightforward bugs where failure cases and prop types aren't being handled right.

Here's a fix!


benjiwheeler:
Created https://github.com/LLK/scratch-www/issues/2103

PR #2078, opened 2018-09-12
move login/registration functions and view state to session reducer, pass to gui

benjiwheeler:
### Resolves:

scratch-gui issues https://github.com/LLK/scratch-gui/issues/3015 and https://github.com/LLK/scratch-gui/issues/2816

Should be merged alongside scratch-gui PR https://github.com/LLK/scratch-gui/pull/3070 , though each was designed to be able to be merged without the other and operate without noticeable errors.

### Changes:

Refactors login, registration and account nav menu functionality so that it primarily resides in the session.js reducer. Adds a <Registration /> component and a <LoginDropdown /> component to the GUI editor view, so that the editor can show login and registration views.

Note that the login and registration views do not actually reside in scratch-gui, they only appear to because they show up in the editor view. But they are entirely generated by scratch-www!


src/redux/session.js
```diff
@@ -103,3 +214,52 @@ module.exports.refreshSession = () => (dispatch => {
         return;
     });
 });
+
+module.exports.handleLogIn = (formData, callback) => (dispatch => {
+    dispatch(module.exports.setLoginError(null));
+    formData.useMessages = true; // NOTE: this may or may not be being used anywhere else
```
rschamp: I think this comment should be removed or changed to explain what this is here for — prevent displaying the "Logged in" flash message on the next scratchr2 page that loads.

src/redux/session.js
```diff
@@ -103,3 +214,52 @@ module.exports.refreshSession = () => (dispatch => {
...
+    api({
+        method: 'post',
+        host: '',
+        uri: '/accounts/login/',
+        json: formData,
+        useCsrf: true
+    }, (err, body) => {
+        if (err) dispatch(module.exports.setLoginError(err.message));
```
rschamp: We probably want to callback and return here.

src/redux/session.js
```diff
@@ -103,3 +214,52 @@ module.exports.refreshSession = () => (dispatch => {
...
+        if (body) {
+            body = body[0];
+            if (body.success) {
+                dispatch(module.exports.setLoginOpen(false));
+                body.messages.map(message => { // eslint-disable-line array-callback-return
```
rschamp: Maybe this is copy-pasted, but I think we should heed this linter warning instead of silencing it. We shouldn't use `map` when we mean to do a `for`.

src/redux/session.js
```diff
@@ -103,3 +214,52 @@ module.exports.refreshSession = () => (dispatch => {
...
+                // Update login error message to a friendlier one if it exists
+                dispatch(module.exports.setLoginError(body.msg));
+            }
+        }
+        // JS error already logged by api mixin
+        callback();
+    });
+});
+
+module.exports.handleLogOut = event => (dispatch => {
```
rschamp: While this should remain a thunk, I don't think we should have it handle the event. It should be used in `mapDispatchToProps` and that function should handle `event` and then dispatch this action.

src/views/preview/preview.jsx
```diff
@@ -353,19 +304,40 @@ class Preview extends React.Component {
...
+                        onCompleteRegistration={this.props.handleCompleteRegistration}
+                        onLogIn={this.props.handleLogIn}
+                        onLogOut={this.props.handleLogOut}
+                        onOpenAccountNav={this.props.handleOpenAccountNav}
+                        onOpenRegistration={this.props.handleOpenRegistration}
+                        onToggleLoginOpen={this.props.handleToggleLoginOpen}
+                        onUpdateProjectTitle={this.handleUpdateProjectTitle}
+                    />
+                    <Registration />
+                    <LoginDropdown
```
rschamp: I think this is pretty funny that it works, but also seems brittle since the GUI can shift under www's feet, and it will be hard to ensure the dropdown displays as we expect in all configurations of the GUI. I think it'd be safer to pass this component to the GUI so the GUI can place it where it needs it.

src/views/preview/preview.jsx
```diff
@@ -289,28 +262,6 @@ class Preview extends React.Component {
             loveCount: loves
         });
     }
-    isShared () {
```
rschamp: Thanks for moving these!

src/views/preview/preview.jsx
```diff
@@ -463,26 +451,83 @@ const consolidateStudiosInfo = (curatedStudios, projectStudios, currentStudioIds
...
+    const authorPresent = projectInfoPresent && state.preview.projectInfo.author &&
+        Object.keys(state.preview.projectInfo.author).length > 0;
+
+    return {
+        comments: state.preview.comments,
+        faved: state.preview.faved,
+        fullScreen: state.scratchGui.mode.isFullScreen,
+        // TODO: this value's logic is unfinished. Currently it seems to be
+        // identical to userOwnsProject, but we need to also set this true
+        // for admins and mods.
```
rschamp: We do have this information in session! In `if (permissions.admin === true)`

src/redux/session.js
benjiwheeler: You're right that this was copy-pasted, but it's still good to fix the stuff I touch!

You're also right that map is wrong here. Changed it to forEach and removed the linter override.

src/views/preview/preview.jsx
benjiwheeler: Sometimes the code just...
smell bad

src/views/preview/preview.jsx
benjiwheeler: Changed the logic to use that, take a look please.

src/views/preview/preview.jsx
benjiwheeler: I'll take a crack at this.

src/redux/session.js
benjiwheeler: Hm. Several other of my hacky session.js functions do event.preventDefault() as well. Should I move all these out to components' callbacks?

I see the argument for event handling not belonging in a reducer, but I wonder if it's more brittle...

src/components/navigation/www/navigation.jsx
```diff
@@ -10,13 +10,12 @@ const React = require('react');
 const messageCountActions = require('../../../redux/message-count.js');
 const sessionActions = require('../../../redux/session.js');
 
-const api = require('../../../lib/api');
+// const api = require('../../../lib/api');
```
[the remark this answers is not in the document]
benjiwheeler: My bad, done

src/components/navigation/www/navigation.jsx
```diff
@@ -361,9 +264,12 @@ Navigation.propTypes = {
         educator_invitee: PropTypes.bool,
         student: PropTypes.bool
     }),
-    refreshSession: PropTypes.func,
     searchTerm: PropTypes.string,
     session: PropTypes.shape({
+        accountNavOpen: PropTypes.bool,
+        canceledDeletionOpen: PropTypes.bool,
+        loginOpen: PropTypes.bool,
+        loginError: PropTypes.string,
```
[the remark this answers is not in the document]
benjiwheeler: Let's talk. Maybe there is a better way.

src/components/navigation/www/navigation.jsx
```diff
@@ -305,32 +203,29 @@ class Navigation extends React.Component {
...
+                                    mode="www"
+                                    onClose={this.props.handleCloseLogin}
+                                    onLogIn={this.props.handleLogIn}
+                                />
                             </li>
                         ]) : []}
                 </ul>
                 <Modal
-                    isOpen={this.state.canceledDeletionOpen}
+                    isOpen={this.props.session.canceledDeletionOpen}
```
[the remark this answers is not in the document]
benjiwheeler: Done

src/components/navigation/www/navigation.jsx
```diff
@@ -305,32 +203,29 @@ class Navigation extends React.Component {
...
+                                    onClick={this.props.handleToggleLoginOpen}
                                 >
                                     <FormattedMessage id="general.signIn" />
                                 </a>
-                                <Dropdown
-                                    className="login-dropdown with-arrow"
-                                    isOpen={this.state.loginOpen}
+                                <LoginDropdown
+                                    error={this.props.session.loginError}
+                                    isOpen={this.props.session.loginOpen}
```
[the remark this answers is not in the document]
benjiwheeler: Done

src/components/navigation/www/navigation.jsx
```diff
@@ -269,33 +170,30 @@ class Navigation extends React.Component {
                                 <AccountNav
                                     classroomId={this.props.session.session.user.classroomId}
                                     isEducator={this.props.permissions.educator}
-                                    isOpen={this.state.accountNavOpen}
+                                    isOpen={this.props.session.accountNavOpen}
```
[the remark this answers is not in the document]
benjiwheeler: Done

src/components/navigation/www/navigation.jsx
```diff
@@ -269,33 +170,30 @@ class Navigation extends React.Component {
...
-                                    onClose={this.handleCloseAccountNav}
+                                    onClick={this.props.handleOpenAccountNav}
+                                    onClickLogout={this.props.handleLogOut}
+                                    onClose={this.props.handleCloseAccountNav}
                                 />
                             </li>
                         ] : [
                             <li
-                                className="link right join"
+                                className="link right"
```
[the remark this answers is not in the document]
rschamp: Maybe this was buried — this change will cause smoke tests to begin failing.

rschamp:
> Seems like we should move away from using classes to target specific elements in the smoke test. You should tell me sometime why you're not a fan of id's ... :)

Unless you know for sure that an element will only ever appear once, id's are inappropriate. Therefore id's are an issue for identifying components because there's no guarantee that a component will only be rendered once on a page.  It may be true that in the currently implemented case, there is only a single instance of a certain component, but that is hidden technical debt for when that becomes untrue in the future. For instance, if/when we create a component library, we may want to display multiple menu bar components on one page to demonstrate the different states.

benjiwheeler:
> > Seems like we should move away from using classes to target specific elements in the smoke test. You should tell me sometime why you're not a fan of id's ... :)
> 
> Unless you know for sure that an element will only ever appear once, id's are inappropriate. Therefore id's are an issue for identifying components because there's no guarantee that a component will only be rendered once on a page. It may be true that in the currently implemented case, there is only a single instance of a certain component, but that is hidden technical debt for when that becomes untrue in the future. For instance, if/when we create a component library, we may want to display multiple menu bar components on one page to demonstrate the different states.

Maybe we could use key= params?

rschamp:
Unfortunately `key` is internal to React and isn't rendered. We could consider using `data-*` params instead though.

Issue #2082, opened 2018-09-13
New project page: Layouting of project notes next to player is off


LiFaytheGoblin:
<img width="285" alt="screen shot 2018-09-13 at 1 54 24 pm" src="https://user-images.githubusercontent.com/5471273/45507305-b1595d00-b75f-11e8-84ce-e24efe716b44.png">

The boxes should be at the same y-Position

### Operating System and Browser
MacOS, Chrome

PR #2083, opened 2018-09-13
Fix notfications box, player box, buttons and icons alignments

LiFaytheGoblin:
#2082 

### Resolves:
screen shot 2018-09-13 at 1 54 24 pm

screen shot 2018-09-13 at 1 59 28 pm

### Changes:
screen shot 2018-09-13 at 3 00 57 pm

### Test Coverage:
MacOS, Chrome,
npm test passes

### DIFF
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -192,10 +192,11 @@ $stage-width: 480px;
     .project-notes {
         // not 1.5rem because of stage padding
         margin-left: 1rem;
-        height: $player-height;
+        height: calc(#{$player-height} - 1rem);
         align-items: flex-start;
         flex: 1;
         flex-flow: column;
+        align-self: flex-end;
     }
 
     .share-date {
@@ -209,6 +210,7 @@ $stage-width: 480px;
     .subactions {
         margin-left: 1.5rem;
         justify-content: flex-end;
+        align-items: flex-start;
         flex: 1;
     }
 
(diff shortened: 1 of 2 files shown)


LiFaytheGoblin:
Solution that works with remix credits
screen shot 2018-09-19 at 2 52 53 pm

On IE and FF will fallback to
screen shot 2018-09-19 at 2 52 41 pm

src/views/preview/preview.scss
```diff
@@ -228,6 +232,14 @@ $stage-width: 480px;
         flex-shrink: 1;
     }
 
+    .credit-text {
+        /* Line clamping is not supported in IE, FF yet */
+        display: -webkit-box;
+        overflow: hidden;
+        -webkit-box-orient: vertical;
+        -webkit-line-clamp: 2;
```
rschamp: I don't think we should use any styling that only works in one browser because people will begin reporting it as an issue on the browsers that don't support this.  Additionally since this is not standard CSS, there's no guarantee that it will continue working in the future.

There may be a JS library for this feature, but I don't think this is important enough to bring something in to do it. Maybe we should instead just keep the `max-height` of the box to the height of two lines, add `overflow: hidden` and set the `title` of the box to the entire string?

I also wonder if we even want to clip the contents of this box or do we want to let it get as tall as it needs to to display the entire text, no matter what? Is the clipping in @carljbowman's designs?

LiFaytheGoblin:
@rschamp Sorry I think I can not merge until you give a positive review?

Issue #2067, opened 2018-09-05
Project page comment "fade out" looks incorrect on Safari


BryceLTaylor:
If you load a project page with a lot of comments it should truncate the list so that they aren't all visible.  The last visible entry should have a gradient that fades it out.

On Safari, on either ipad or a mac, it looks awful:

ipad
![ipadgradientcommentsgradientprojectpage](https://user-images.githubusercontent.com/1088291/45105676-c5b7ad00-b102-11e8-82b5-a9b6c10f639f.PNG)

on mac
![screen shot 2018-09-05 at 11 53 03 am](https://user-images.githubusercontent.com/1088291/45105684-cbad8e00-b102-11e8-93bf-f839fd7cbf88.png)

### Steps to Reproduce

Open a project with a lot of comments such as https://scratch.ly/preview/25532916/
Scroll down until you see where the comments get truncated
Note: it looks wrong.

I did not see this on chrome or firefox

### Operating System and Browser

Safari on Mac
Safari on iPad

LiFaytheGoblin:
Reason + Solution https://css-tricks.com/thing-know-gradients-transparent-black/

PR #2099, opened 2018-09-21
Made comments fade out look nice in Safari too

LiFaytheGoblin:
### Resolves:
#2067 

### Changes:
transparent in safari gets a greyish look. fix is to use a white base and set to 0 opacity.
screen shot 2018-09-21 at 10 00 53 am

### Test Coverage:
MacOS, Chrome, Firefox, Safari
npm test passed

### DIFF
--- a/src/views/preview/comment/comment.scss
+++ b/src/views/preview/comment/comment.scss
@@ -1,7 +1,5 @@
 @import "../../../colors";
 
-$cross-browser-transparent: hsla(0, 0, 100%, 0);
-
 .compose-comment {
     width: 100%;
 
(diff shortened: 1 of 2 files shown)


src/views/preview/comment/comment.scss
```diff
@@ -172,7 +174,7 @@
                 position: absolute;
                 bottom: 0;
                 background: linear-gradient(
-                    transparent,
+                    $cross-browser-transparent,
```
rschamp: Is the issue that other browsers use transparent white instead of transparent black? Or do they figure out and use the color being transitioned to/from?

I guess my question is for true cross-browser consistency, should define and use `$ui-light-primary-transparent`?

LiFaytheGoblin: As far as I understood they use white transparent instead of dark transparent. 

Hm, in colors.css you mean? I can do that!

Issue #2086, opened 2018-09-13
Make new project page responsive


LiFaytheGoblin:
There are a few problems with responsiveness of the new project pages like 
- the title field wrapping to the next "line"
- notes and credits not becoming wide enough 
- weird looking love-it and so buttons
- probably more

LiFaytheGoblin:
pay attention to mobile phone vs tablet (see invision)

LiFaytheGoblin:
Problems:
- What to do about nested comments?
- When is the tablet breakpoint?

LiFaytheGoblin:
mod-report: margin 60
modal-content: inherit border radius

LiFaytheGoblin:
Tests:
- [ ] all the screen sizes, what happens in landscape?
- [ ] modals
- [ ] pressing buttons
- [ ] entering text
- [ ] scrolling 
- [ ] warning messages looking good?
- [ ] owning vs not owning project
- [ ] nested comments / comments in general

PR #2105, opened 2018-09-25
If in landscape format on mobile, switch to fullscreen mode

LiFaytheGoblin:
### Resolves:
Part of #2086: The design envisioned the full screen player to appear for mobile devices in landscape mode.

### Changes:
Adds js to react to the orientation of a device to presentation.jsx

### Test Coverage:
Tested as far as possible with MacOS + Chrome dev tools,
npm test passed

### DIFF
--- a/src/views/preview/presentation.jsx
+++ b/src/views/preview/presentation.jsx
@@ -94,7 +94,6 @@ const PreviewPresentation = ({
                                 </a>
                                 <div className="title">
                                     {editable ?
-
                                         <InplaceInput
                                             className="project-title"
                                             handleUpdate={onUpdate}
--- a/src/views/preview/preview.jsx
+++ b/src/views/preview/preview.jsx
@@ -28,8 +28,6 @@ const frameless = require('../../lib/frameless');
 const GUI = require('scratch-gui');
 const IntlGUI = injectIntl(GUI.default);
 
-const isMobileDevice = () => screen.height <= frameless.mobile || screen.width <= frameless.mobile;
-
 class Preview extends React.Component {
     constructor (props) {
         super(props);
@@ -39,7 +37,6 @@ class Preview extends React.Component {
             'handleFavoriteToggle',
             'handleLoadMore',
             'handleLoveToggle',
-            'handleOrientationChange',
             'handlePopState',
             'handleReportClick',
             'handleReportClose',
@@ -51,7 +48,8 @@ class Preview extends React.Component {
             'handleUpdate',
             'initCounts',
             'pushHistory',
-            'renderLogin'
+            'renderLogin',
+            'setScreenFromOrientation'
         ]);
         const pathname = window.location.pathname.toLowerCase();
         const parts = pathname.split('/').filter(Boolean);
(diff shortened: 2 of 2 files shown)


src/views/preview/presentation.jsx
```diff
@@ -94,7 +94,6 @@ const PreviewPresentation = ({
                                 </a>
                                 <div className="title">
                                     {editable ?
-
```
LiFaytheGoblin: Not sure why this line is missing here, must have happend when solving conflicts. Should I add the line back?

src/views/preview/preview.jsx
```diff
@@ -63,6 +68,14 @@ class Preview extends React.Component {
         };
         this.getExtensions(this.state.projectId);
         this.addEventListeners();
+        /* In the beginning, if user is on mobile and landscape, go to fullscreen */
+        if (this.props.playerMode && isMobileDevice()) {
+            if (screen.orientation.type === 'landscape-primary') {
```
LiFaytheGoblin: This way of detecting landscape will need additional testing! Esp. on iPad + Safari

src/views/preview/preview.jsx
```diff
@@ -157,6 +172,19 @@ class Preview extends React.Component {
     handleReportSubmit (formData) {
         this.props.reportProject(this.state.projectId, formData);
     }
+    handleOrientationChange () {
+        /*
+         * If the user is on a mobile device, switching to
+         * landscape format should make the fullscreen mode active
+         */
+        if (this.props.playerMode && isMobileDevice()) {
```
LiFaytheGoblin: I might be able to get rid of the double code

src/views/preview/presentation.jsx
rschamp: This is fine.

src/views/preview/preview.jsx
```diff
@@ -106,9 +111,25 @@ class Preview extends React.Component {
     }
     addEventListeners () {
         window.addEventListener('popstate', this.handlePopState);
+        window.addEventListener('orientationchange', this.setScreenFromOrientation);
     }
     removeEventListeners () {
         window.removeEventListener('popstate', this.handlePopState);
+        window.addEventListener('orientationchange', this.setScreenFromOrientation);
```
rschamp: I think you mean `removeEventListener` here.

src/views/preview/preview.jsx
```diff
@@ -106,9 +111,25 @@ class Preview extends React.Component {
...
+    }
+    setScreenFromOrientation () {
+        /*
+        * If the user is on a mobile device, switching to
+        * landscape format should make the fullscreen mode active
+        */
+        const isMobileDevice = screen.height <= frameless.mobile || screen.width <= frameless.mobile;
+        if (this.props.playerMode && isMobileDevice) {
+            if (screen.orientation.type === 'landscape-primary') {
```
rschamp: `screen.orientation` doesn't look like it has good browser support, have you looked at if this crashes on IE or Edge?  Is there a cross-platform way to detect the same thing?

LiFaytheGoblin: Yeah I could check if screen.height is smaller than screen.width.

src/views/preview/preview.jsx
```diff
@@ -106,9 +111,27 @@ class Preview extends React.Component {
...
+        window.removeEventListener('orientationchange', this.setScreenFromOrientation);
+    }
+    setScreenFromOrientation () {
+        /*
+        * If the user is on a mobile device, switching to
+        * landscape format should make the fullscreen mode active
+        */
+        const isMobileDevice = screen.height <= frameless.mobile || screen.width <= frameless.mobile;
+        if (this.props.playerMode && isMobileDevice) {
+            const isLandscape = (screen.orientation.type && screen.orientation.type === 'landscape-primary') ||
```
rschamp: If `screen` or `screen.orientation` are `undefined` on browsers that don't support these, then this will cause a fatal error, won't it? I think you need to test for them (or just use `screen.height/width` for all).

LiFaytheGoblin: I wanted to be progressive but now I think just going with screen.width is the easiest option.

PR #2107, opened 2018-09-25
Make container smaller for mobile to allow for space left and right

LiFaytheGoblin:
### Resolves:
Part of #2086: According to the design, left and right of page content should be some space on mobile. This is what this pull request does. It will not take full effect unless some of the page content is actually adapted to fit in the page container.

### Test Coverage:
MacOS, Chrome
npm test passed

### DIFF
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -36,8 +36,8 @@ $medium-and-small: "screen and (max-width : #{$tablet}-1)";
         }
 
         @media #{$intermediate} {
-            min-width: 640px;
             width: 90%;
+            min-width: 640px;
         }
     }
 
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -9,7 +9,7 @@ $stage-width: 480px;
 /* screen sizes */
 $small: "screen and (max-width : #{$mobile}-1)";
 $medium: "screen and (min-width : #{$mobile}) and (max-width : #{$tablet}-1)";
-$big: "screen and (min-width : #{$tablet})";
+$intermediate: "screen and (min-width : #{$tablet}) and (max-width : 941px)"; /* 941 because currently breakpoint of .inner in www is 941 */
 $medium-and-small: "screen and (max-width : #{$tablet}-1)";
 
 /* override view padding for share banner */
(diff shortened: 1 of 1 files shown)


LiFaytheGoblin:
Many other pull requests will depend on this.

src/views/preview/preview.scss
```diff
@@ -9,7 +9,7 @@ $stage-width: 480px;
 /* screen sizes */
 $small: "screen and (max-width : #{$mobile}-1)";
 $medium: "screen and (min-width : #{$mobile}) and (max-width : #{$tablet}-1)";
-$big: "screen and (min-width : #{$tablet})";
+$intermediate: "screen and (min-width : #{$tablet}) and (max-width : 941px)"; /* 941 because currently breakpoint of .inner in www is 941 */
```
rschamp: Please name `941` as a descriptive variable, so it's more understandable what `$intermediate` means.

src/views/preview/preview.scss
```diff
@@ -28,6 +28,19 @@ $medium-and-small: "screen and (max-width : #{$tablet}-1)";
 
 .preview {
 
+    .inner {
+        @media #{$medium-and-small} {
+            margin: 0 auto;
+            max-width: 90%;
```
rschamp: I'm curious about `90%` — I think elsewhere we use a static value or `100%`, was that approach not appropriate here?

LiFaytheGoblin: In order to fit the in between sizes I didn't want to use a static value. And I wanted to leave space so I used 90%. However, now that you mention it, I could also try something else... let me check...

LiFaytheGoblin: I guess I could use padding. But I think I would want to get this pull request in after the other changes went in. It is hard for me to test this way if things will work.

LiFaytheGoblin:
I reverted the solution from before because it didn't work out. I made a small fix to the title also because it had wrong spacing. This is what it should look like:

screen shot 2018-10-04 at 10 41 36 am
screen shot 2018-10-04 at 10 42 05 am

The overflowing GUI is a known problem. This is because the width of the GUI is set in GUI, not in www.

PR #2109, opened 2018-09-25
Stats and subactions are responsive and separate components

LiFaytheGoblin:
PR depends on #2108

### Resolves:
Part of #2086: The stats and subactions now position nicely depending on media queries. They are also put in separate components.

### Test Coverage:
MacOS, Chrome
npm test passes

### DIFF
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -255,6 +255,13 @@ $medium-and-small: "screen and (max-width : #{$tablet}-1)";
         line-height: 2rem;
         color: $type-gray;
         font-size: .875rem;
+
+        @media #{$small} {
+            order: 100;
+            margin-top: 1rem;
+            width: 100%;
+        }
+
     }
 
     .subactions {
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -6,6 +6,7 @@ $player-width: 482px;
 $player-height: 406px;
 $stage-width: 480px;
 
+$small: "screen and (max-width : #{$mobile}-1)";
 $medium-and-small: "screen and (max-width : #{$tablet}-1)";
 
 /* override view padding for share banner */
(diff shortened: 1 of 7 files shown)


LiFaytheGoblin:
Argh oh no! I accidentally also took in some commits for the report modal ._. Can these just be left out when merging or so? @rschamp

src/views/preview/preview.scss
```diff
@@ -539,8 +359,3 @@ $medium-and-small: "screen and (max-width : #{$tablet}-1)";
         }
     }
 }
-
-.report-text textarea {
```
LiFaytheGoblin: This will look fine if #2108 goes in first @rschamp

rschamp: Sure — please update the description to say that this PR depends on #2108.

LiFaytheGoblin: done!

src/views/preview/subactions.scss
```diff
@@ -0,0 +1,145 @@
...
+            //         background-color: $active-gray;
+            //     }
+            //
+            //     &:active {
+            //         border: 0 solid transparent;
+            //         box-shadow: inset 0 0 5px $box-shadow-gray;
+            //         background-color: $active-dark-gray;
+            //         padding: calc(.75em + 1px) calc(1.5em + 1px);
+            //     }
+            // }
```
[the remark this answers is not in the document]
LiFaytheGoblin: yep it was like this already

LiFaytheGoblin:
Okay I am kind of worried about merging now @Ray: It said I had merging problems, then locally I merged develop into my branch and resolved the issues which was tricky but it worked. Then I pushed and then it said again that I had merging issues?! just clicked on fix issues or so and copy pasted what I had locally. Is that an okay thing to do?

PR #2117, opened 2018-09-26
Made comments look better on mobile

LiFaytheGoblin:
### Resolves:
Part of #2086 . Looks ok as is but would look better after pull request #2107 .

### Test Coverage:
MacOS, Chrome,
npm test passed


benjiwheeler:
FYI, here is before...
screen shot 2018-10-01 at 5 56 22 pm


...and after (still not with #2107 )

screen shot 2018-10-01 at 5 55 24 pm

PR #2118, opened 2018-09-26
Made project notes look better on mobile

LiFaytheGoblin:
### Resolves:
Part of #2086 . (This already looks ok but kind of depends on Pull request #2107. Edit, actually I changed my mind because I changed 2107 again!)

### Test Coverage:
MacOs, Chrome,
npm test passed

### DIFF
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -200,6 +200,14 @@ $medium-and-small: "screen and (max-width : #{$tablet}-1)";
     .guiPlayer {
         display: inline-block;
         width: $player-width;
+
+        @media #{$small} {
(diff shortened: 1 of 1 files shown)


benjiwheeler:
Looks much better!

Is the `@media #{$medium-and-small}` syntax something we've been using already, or is it something new you're introducing?

LiFaytheGoblin:
@benjiwheeler I introduced it because I often need a query that is for both medium and small. I expect this to change as the focus doesn't lay on only phone and desktop anymore but also considers more of the in between sizes. But for now, this is my solution. Please say if you can think of a better way to do it! There might be!

PR #2136, opened 2018-10-02
New project page: See inside and remix buttons disappear on mobile

LiFaytheGoblin:
On mobile phones (not tablets!) the see inside and remix buttons now disappear.

MacOS, Chrome, npm test passes

### DIFF
--- a/src/views/preview/presentation.jsx
+++ b/src/views/preview/presentation.jsx
@@ -87,7 +87,7 @@ const PreviewPresentation = ({
             { projectInfo && projectInfo.author && projectInfo.author.id && (
                 <Formsy onKeyPress={onKeyPress}>
                     <div className="inner">
-                        <FlexRow className="preview-row force-row">
+                        <FlexRow className="preview-row">
                             <FlexRow className="project-header">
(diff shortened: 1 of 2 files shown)


src/views/preview/presentation.jsx
```diff
@@ -124,20 +124,22 @@ const PreviewPresentation = ({
                                     }
                                 </div>
                             </FlexRow>
-                            <div className="project-buttons">
-                                {/* TODO: Hide Remix button for now until implemented */}
-                                {(!userOwnsProject && false) &&
-                                    <Button className="button remix-button">
-                                        Remix
+                            <MediaQuery minWidth={frameless.mobile}>
```
chrisgarrity: It seems inconsistent that in `lib/frameless` we define desktop, tablet and mobile. Then in preview css we define screen sizes: small, medium, big (along with medium-and-small). It's a bit of technical debt that I think we should address at some point.
/cc @rschamp 

Also - what does mobile really mean? lots of tablets are also 'mobile' in the sense of having a cellular connection. And there are tablets that are big enough to be 'desktop'.

Issue #2147, opened 2018-10-04
Unshared project banner should be responsive


LiFaytheGoblin:
As can be seen in #2115, the unshared project banner should be responsive. Currently it looks like this:
<img width="317" alt="screen shot 2018-10-04 at 1 21 57 pm" src="https://user-images.githubusercontent.com/5471273/46491350-7e434000-c7d8-11e8-96ed-6faf18538d7f.png">

Mainly, the flex row that the banner is should stay a row on mobile and not become a column. Align-items should be center. There should be an equal amount of space above and below the text in the banner.

(MacOS, Chrome)

PR #2148, opened 2018-10-04
Share banner now looks good on mobile too

LiFaytheGoblin:
### Resolves:
#2147 

### Changes:
Overwritten some flex row styles and changed button styles to be flex row
screen shot 2018-10-04 at 1 52 09 pm

### Test Coverage:
MacOS, Chrome, npm test passed

### DIFF
--- a/src/views/preview/share-banner.scss
+++ b/src/views/preview/share-banner.scss
@@ -7,14 +7,27 @@ $navigation-height: 50px;
     width: 100%;
     overflow: hidden;
     color: $ui-orange;
+
+    /* overwrite styles for flex row and preview row */
+    .flex-row {
+        flex-direction: row;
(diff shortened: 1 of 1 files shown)


src/views/preview/share-banner.scss
```diff
@@ -7,14 +7,27 @@ $navigation-height: 50px;
...
+
+    /* overwrite styles for flex row and preview row */
+    .flex-row {
+        flex-direction: row;
+    }
+
+    .preview-row {
+        margin-top: 0 !important;
+        padding: .5rem;
+        align-items: center !important;
```
rschamp: We shouldn't need `!important` except as a last resort. I will play around with some ways to avoid this.

PR #2150, opened 2018-10-04
Add comments to projects and replies to comments

paulkaplan:
This is a bit of a work-in-progress, but it includes all the major parts of posting comments and replying to comments on the preview page, including
- commenting/replying only visible to logged in users
- comment posting with submitting/rejected states
- seeing newly created comments/replies get added to the thread in the right place

A few things are more WIPy:
- only a few of the more common error responses are ported over, did not look up the messages for all of the possible responses
- needs localization (/cc @chrisgarrity), it already contained a fair number of hard-coded strings, so i'm going to defer this until i talk to chris about how to do it.

I've still got some various cleanup things, but I wanted @rschamp to get a chance to take a look early

Here is a brief demo (although things may have changed a bit UI wise, like the replies are no longer contained in the parent bubble). 

new-comment-demo

For @BryceLTaylor, I put together this list of tests to go through:

As a logged out user
- I should not see the "leave a comment" text field for project comments
- I should not see “reply” link in comments

As a logged in user
- I should see new comment form above the comment list
- - Clicking cancel clears the textarea, form remains
- - Clicking post submits, makes comment appear at the top of the comment list, makes the textarea clear
- - - See “comment posting error states” for failure modes
- I should see the reply link on each comment
- Clicking reply makes a reply text area appear with post/cancel buttons
- - Cancel makes the textarea go away
- - Post submits the reply, makes the text area turn into a comment
- - See “comment posting error states” for failure modes

Comment posting error states
- There are a number of submission errors that can return after posting. These are shown with a popover on the textarea without clearing the content. (tooltip UI is not in for this yet and several error messages are missing, but a few common error messages are handled, like posting too quickly, or using inappropriate words)
- character limiting does not yet prevent typing, just counts down

---

Update: I changed the comment reducer actions to include `topLevelCommentId` to simplify some of the reducer code, and also fixes https://github.com/LLK/scratch-www/issues/2151

### DIFF
--- a/src/redux/preview.js
+++ b/src/redux/preview.js
@@ -87,6 +87,19 @@ module.exports.previewReducer = (state, action) => {
             comments: [...state.comments, ...action.items] // TODO: consider a different way of doing this?
         });
     case 'SET_COMMENT_DELETED':
+        if (action.topLevelCommentId) {
+            return Object.assign({}, state, {
+                replies: Object.assign({}, state.replies, {
+                    [action.topLevelCommentId]: state.replies[action.topLevelCommentId].map(comment => {
+                        if (comment.id === action.commentId) {
+                            return Object.assign({}, comment, {deleted: true});
+                        }
+                        return comment;
+                    })
+                })
+            });
+        }
+
         return Object.assign({}, state, {
             comments: state.comments.map(comment => {
                 if (comment.id === action.commentId) {
@@ -96,30 +109,13 @@ module.exports.previewReducer = (state, action) => {
             })
         });
     case 'ADD_NEW_COMMENT':
-        if (action.comment.parent_id) {
-            let topLevelParent = action.comment.parent_id;
-
-            // If this is a nested reply, we need to look up which top level comment
-            // to put this new reply under.
-            if (!state.replies[topLevelParent]) {
-                Object.keys(state.replies).forEach(topLevelCommentId => {
-                    state.replies[topLevelCommentId].forEach(reply => {
-                        if (reply.id === action.comment.parent_id) {
-                            topLevelParent = topLevelCommentId;
-                        }
-                    });
-                });
-            }
-
-            if (state.replies[topLevelParent]) {
-                const replies = JSON.parse(JSON.stringify(state.replies));
-                // Replies to comments go at the end  of the thread
-                replies[topLevelParent] = replies[topLevelParent].concat(action.comment);
-                return Object.assign({}, state, {replies: replies});
-            }
-
-            log.error('Could not find top level parent to put reply in');
-            return state;
+        if (action.topLevelCommentId) {
+            return Object.assign({}, state, {
+                replies: Object.assign({}, state.replies, {
+                    // Replies to comments go at the end  of the thread
+                    [action.topLevelCommentId]: state.replies[action.topLevelCommentId].concat(action.comment)
+                })
+            });
         }
 
         // Reply to the top level project, put the reply at the beginning
@@ -232,14 +228,16 @@ module.exports.setStudioFetchStatus = (studioId, status) => ({
     status: status
 });
 
-module.exports.setCommentDeleted = commentId => ({
+module.exports.setCommentDeleted = (commentId, topLevelCommentId) => ({
     type: 'SET_COMMENT_DELETED',
-    commentId: commentId
+    commentId: commentId,
+    topLevelCommentId: topLevelCommentId
 });
 
-module.exports.addNewComment = comment => ({
+module.exports.addNewComment = (comment, topLevelCommentId) => ({
     type: 'ADD_NEW_COMMENT',
-    comment: comment
+    comment: comment,
+    topLevelCommentId: topLevelCommentId
 });
 
 module.exports.getProjectInfo = (id, token) => (dispatch => {
--- a/src/redux/preview.js
+++ b/src/redux/preview.js
@@ -95,6 +95,38 @@ module.exports.previewReducer = (state, action) => {
                 return comment;
             })
         });
+    case 'ADD_NEW_COMMENT':
+        if (action.comment.parent_id) {
+            let topLevelParent = action.comment.parent_id;
+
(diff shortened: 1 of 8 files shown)


src/redux/preview.js
```diff
@@ -95,6 +95,38 @@ module.exports.previewReducer = (state, action) => {
                 return comment;
             })
         });
+    case 'ADD_NEW_COMMENT':
+        if (action.comment.parent_id) {
+            let topLevelParent = action.comment.parent_id;
+
+            // If this is a nested reply, we need to look up which top level comment
+            // to put this new reply under.
```
benjiwheeler: Why not look up the parent's parent't parent's... parent instead? If we haven't introduced errors, that should work, right? And be faster (though I'm not sure how much it really matters since we only do all this once per added comment)

src/views/preview/comment/comment.jsx
```diff
@@ -1,89 +1,136 @@
...
-                    ) : null}
-                    <span className="comment-report">
-                        Report {/* TODO internationalize */}
-                    </span>
-                </div>
-            </FlexRow>
+class Comment extends React.Component {
+    constructor (props) {
+        super(props);
+
```
benjiwheeler: most of our components seem to not have whitespace here, FWIW

src/views/preview/comment/compose-comment.jsx
```diff
@@ -1,26 +1,185 @@
...
+            this.setState({status: ComposeStatus.EDITING, error: null, message: ''});
+
+            // Add the username, which isn't included right now from scratch-api
+            if (body.author) body.author.username = this.props.user.username;
+
+            this.props.onAddComment(body);
+        });
+    }
+    handleCancel () {
+        this.setState({message: '', error: null, status: ComposeStatus.EDITING});
```
benjiwheeler: Could these be on separate lines?

src/views/preview/comment/compose-comment.jsx
```diff
@@ -1,26 +1,185 @@
...
+                body = {rejected: 'serverError'};
+            }
+
+            if (body.rejected && this.state.status === ComposeStatus.SUBMITTING) {
+                this.setState({status: ComposeStatus.REJECTED, error: body.rejected});
+                return;
+            }
+
+            // Clear the text field on successful submission
+            this.setState({status: ComposeStatus.EDITING, error: null, message: ''});
```
benjiwheeler: Could these be on separate lines?

src/views/preview/comment/compose-comment.jsx
```diff
@@ -1,26 +1,185 @@
...
+                parent_id: this.props.parentId || '',
+                comentee_id: this.props.comenteeId || ''
+            }
+        }, (err, body, res) => {
+            if (err || res.statusCode !== 200) {
+                body = {rejected: 'serverError'};
+            }
+
+            if (body.rejected && this.state.status === ComposeStatus.SUBMITTING) {
+                this.setState({status: ComposeStatus.REJECTED, error: body.rejected});
```
benjiwheeler: Could these be on separate lines?

paulkaplan:
@benjiwheeler thanks for the review. I changed the reducer code to be much simpler by including the `topLevelCommentId` in the action so we know right away whether the action is applying to a top level comment or a nested comment. This also fixes the problem where you couldn't delete nested comments.

src/views/preview/presentation.jsx
```diff
@@ -313,10 +315,21 @@ const PreviewPresentation = ({
                                     <h4>Comments</h4>
                                     {/* TODO: Add toggle comments component and logic*/}
                                 </FlexRow>
+
+                                <FlexRow className="comments-root-reply">
+                                    {isLoggedIn &&
```
rschamp: This will also need to toggle based on whether comments are turned on or off for the project. But there is no way for you to know that info atm (logged here https://github.com/LLK/scratch-api/issues/584)

src/views/preview/preview.jsx
```diff
@@ -165,8 +166,11 @@ class Preview extends React.Component {
                 });
             });
     }
-    handleDeleteComment (id) {
-        this.props.handleDeleteComment(this.state.projectId, id, this.props.user.token);
+    handleAddComment (comment, topLevelCommentId) {
+        this.props.handleAddComment(comment, topLevelCommentId);
+    }
+    handleDeleteComment (id, topLevelCommentId) {
+        this.props.handleDeleteComment(this.state.projectId, id, topLevelCommentId, this.props.user.token);
```
rschamp: Seeing these handlers points out that we have an odd convention here of passing `handleX` as a prop. Usually we use `onX` for props, `handleX` for the internal handler.

src/views/preview/preview.jsx
```diff
@@ -525,8 +531,11 @@ const mapStateToProps = state => {
 };
 
 const mapDispatchToProps = dispatch => ({
-    handleDeleteComment: (projectId, commentId, token) => {
-        dispatch(previewActions.deleteComment(projectId, commentId, token));
+    handleAddComment: (comment, topLevelCommentId) => {
+        dispatch(previewActions.addNewComment(comment, topLevelCommentId));
```
rschamp: It's a little weird right now with adding comments done partly in the component and partly in redux, while both the request and the state manipulation is done in redux for deleting comments. Reading the file right here, it's confusing that one needs a token but the other doesn't.  Are you planning to move the comment add request into Redux?

src/views/preview/comment/compose-comment.jsx
```diff
@@ -31,16 +31,21 @@ const CommentErrorMessages = {
     isFlood: "Woah, seems like you're commenting really quickly. Please wait longer between posts.",
```
chrisgarrity: I know this is WIP, so just a note.... These strings will all need to be localized. They should be defined in `src/l10n.json` as they'll be used in more than one view (eventually). I would give them keys like `comment.isBad`, to indicate they're comment strings.

Issue #2133, opened 2018-10-02
New project page: cleanup


LiFaytheGoblin:
- [x] `$medium-and-small: "screen and (max-width : #{$tablet}-1)";` should be in frameless, also  https://github.com/LLK/scratch-www/pull/2136#pullrequestreview-161190276
- [x] Subactions.jsx: Ray sais about shareDate variable: "This would probably be more appropriate as a shareDate prop passed into the component. But I think it's probably preferable that we do a cleanup pass in a separate PR since it's clearer how this was refactored right now."

#2109

PR #2153, opened 2018-10-05
Move media queries used in responsive project page to frameless

LiFaytheGoblin:
### Resolves:
Part of #2133 

### Changes:
The media query variables in various scss files used in the responsive project page have been moved to frameless to keep them consistent accross all files, easier to update, less double code.

Media queries like this are used all over the website tho. Would it make sense to replace them all by the variables from frameless? If so we should make an issue for that.

### Test Coverage:
MacOS, Chrome, npm test passed

### DIFF
--- a/src/_frameless.scss
+++ b/src/_frameless.scss
@@ -38,6 +38,29 @@ $desktop: 942px;
 $tablet: 640px;
 $mobile: 480px;
 
+/* Media Queries */
+
+/* Width */
+/*
+* ... small | medium     | intermediate | big ...
+* ... medium-and-smaller |
+*          ... intermediate-and-smaller |
+*/
+
+$small: "only screen and (max-width : #{$mobile}-1)";
+$medium: "only screen and (min-width : #{$mobile}) and (max-width : #{$tablet}-1)";
+$intermediate: "only screen and (min-width : #{$tablet}) and (max-width : #{$desktop}-1)";
+$big: "only screen and (min-width : #{$desktop})";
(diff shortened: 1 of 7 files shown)


src/components/modal/base/modal.scss
```diff
@@ -32,7 +26,7 @@ $medium-height: "only screen and (min-height : #{$mobile} + 1) and (max-height :
         outline: none;
     }
 
-    @media #{$small}, #{$medium}, #{$big} {
+    @media #{$intermediate-and-smaller} {
```
chrisgarrity: So much clearer!

src/_frameless.scss
```diff
@@ -38,6 +38,29 @@ $desktop: 942px;
...
+* ... medium-and-smaller |
+*          ... intermediate-and-smaller |
+*/
+
+$small: "only screen and (max-width : #{$mobile}-1)";
+$medium: "only screen and (min-width : #{$mobile}) and (max-width : #{$tablet}-1)";
+$intermediate: "only screen and (min-width : #{$tablet}) and (max-width : #{$desktop}-1)";
+$big: "only screen and (min-width : #{$desktop})";
+
+$medium-and-smaller: "only screen and (max-width : #{$tablet}-1)";
```
chrisgarrity: I like this naming - it's clear that it's everything below a certain size.

rschamp:
@chrisgarrity Would you please take on getting this merged?

Issue #2160, opened 2018-10-08
Cannot search with words containing percent


apple502j:
We cannot search with `100%pen`.

### Steps to Reproduce
1. Search `100%pen` projects
2. Fail and error page

### What's wrong?
https://github.com/chen-robert/scratch-www/blob/develop/src/views/search/search.jsx#L81

%pe is invalid so it raises URIError.

### Operating System and Browser
On Windows 10 Firefox 62

apple502j:
Duplicate of #1153
However, it was closed.

chen-robert:
I'll take a look at this.

PR #2161, opened 2018-10-09
Fix gh-2160: Search with %

chen-robert:
### Changes:

Because query terms only sometimes need to be decoded, we attempt URI decoding and silence the error if it fails. This lets us still search with queries like `cat%25cat` when we want to, but now falls back to no-decoding when an invalid string is presented (e.g. `100%pen`).

### DIFF
--- a/src/views/search/search.jsx
+++ b/src/views/search/search.jsx
@@ -78,7 +78,7 @@ class Search extends React.Component {
         while (term.indexOf('&') > -1) {
             term = term.substring(0, term.indexOf('&'));
         }
-        term = decodeURIComponent(term.split('+').join(' '));
+        term = decodeURIComponent(term);
         this.props.dispatch(navigationActions.setSearchTerm(term));
     }
     componentDidUpdate (prevProps) {
--- a/src/views/search/search.jsx
+++ b/src/views/search/search.jsx
@@ -80,7 +80,7 @@ class Search extends React.Component {
         }
         try {
             term = decodeURIComponent(term);
-        } catch (e){
+        } catch (e) {
             // Error means that term was not URI encoded and decoding failed.
             // We can silence this error because not all query strings are intended to be decoded.
         }
--- a/src/views/search/search.jsx
+++ b/src/views/search/search.jsx
@@ -78,7 +78,12 @@ class Search extends React.Component {
         while (term.indexOf('&') > -1) {
             term = term.substring(0, term.indexOf('&'));
         }
-        term = decodeURIComponent(term);
+        try {
+            term = decodeURIComponent(term);
+        } catch (e){
+            // Error means that term was not URI encoded and decoding failed.
+            // We can silence this error because not all query strings are intended to be decoded.
+        }
         this.props.dispatch(navigationActions.setSearchTerm(term));
     }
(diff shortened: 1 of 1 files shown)


src/views/search/search.jsx
```diff
@@ -78,7 +78,12 @@ class Search extends React.Component {
         while (term.indexOf('&') > -1) {
             term = term.substring(0, term.indexOf('&'));
         }
-        term = decodeURIComponent(term.split('+').join(' '));
+        try {
+            term = decodeURIComponent(term);
+        } catch (e){
```
benjiwheeler: i know this is a tiny quibble, but could you put a space before the `{` ?

src/views/search/search.jsx
```diff
@@ -78,7 +78,12 @@ class Search extends React.Component {
         while (term.indexOf('&') > -1) {
             term = term.substring(0, term.indexOf('&'));
         }
-        term = decodeURIComponent(term.split('+').join(' '));
+        try {
```
benjiwheeler: Why was it safe to take out the `.split('+').join(' ')` ? Because that's part of decoding uri's anyway? Do we know why it was there? Would be good to test a few strings and confirm they decode as we expect

benjiwheeler:
This looks good, thanks for identifying this case that we weren't handling!

src/views/search/search.jsx
chen-robert: I believe it is because we have already decoded the URI. Regardless, including that piece of code introduces a bug when we try searching with `+`. For example, `cat+dog` gets translated to `cat dog`.

src/views/search/search.jsx
chen-robert: I've fixed it. This didn't show any errors in my ESLint though?

chen-robert:
@benjiwheeler I've fixed the issues. Could you please review the pr again?

benjiwheeler:
Will do!

FYI, one piece of our workflow (I just joined the team recently, this is new to me and was not obvious) is that we use github's "assignee" and "labels" info to tell each other "I updated, can you take a look" or "this needs revision". So after you updated, you could assign me, take off the "needs work" label, and put on the "needs review" label.

chen-robert:
I don't think I have permission to edit the labels or assignees, sorry about that. I'm not part of the Scratch team - I just work on this for fun.

benjiwheeler:
So I tested this out. It's a big improvement for some searches, like your example of "100%pen" (or even "100% pen" -- both crash the page right now, but not with your improvement).

I'm not sure we should stop parsing the `+` symbol, but I agree with you that parsing it into a space introduces a bug, and from my testing, there's no search that would be prevented if we take that parsing out -- you can always use `%20` for a space in the querystring. I did a quick search for existing search links in the wild, by googling `site:https://scratch.mit.edu/search/projects`, and didn't find any that use a `+`. I agree that we should take it out. 

Before approving, however, I just want to run this by @thisandagain .

thisandagain:
This seems like an ok change to me. Thanks @chen-robert

benjiwheeler:
Awesome! I'm merging it! Thanks, @chen-robert !

PR #2164, opened 2018-10-09
Comment report action and modal flow for deleting

paulkaplan:
This PR implements the following:
- Fix the "delete" action for nested comments (first commit)
- Include a confirmation modal for deleting comments, that allows you to report instead
- Add action for reporting comments, including a confirmation modal
- Fix proptype warnings from previous PR

image

image
image

Some of the finer details about the report/delete flow are not yet implemented, e.g. whether the comments are removed instantly, having different text in the report modal depending on project ownership... I will come back to these in a later design pass.

Also /cc @chrisgarrity, introduced some new strings for these modals, tried to follow the other modal string conventions

### DIFF
--- a/src/views/preview/comment/comment.jsx
+++ b/src/views/preview/comment/comment.jsx
@@ -7,6 +7,8 @@ const FlexRow = require('../../../components/flex-row/flex-row.jsx');
 const Avatar = require('../../../components/avatar/avatar.jsx');
 const FormattedRelative = require('react-intl').FormattedRelative;
 const ComposeComment = require('./compose-comment.jsx');
+const DeleteCommentModal = require('../../../components/modal/comments/delete-comment.jsx');
+const ReportCommentModal = require('../../../components/modal/comments/report-comment.jsx');
 
 require('./comment.scss');
 
@@ -15,10 +17,18 @@ class Comment extends React.Component {
         super(props);
         bindAll(this, [
             'handleDelete',
+            'handleCancelDelete',
+            'handleConfirmDelete',
+            'handleReport',
+            'handleConfirmReport',
+            'handleCancelReport',
             'handlePostReply',
             'handleToggleReplying'
         ]);
         this.state = {
+            deleting: false,
+            reporting: false,
+            reportConfirmed: false,
             replying: false
         };
     }
@@ -33,7 +43,37 @@ class Comment extends React.Component {
     }
 
     handleDelete () {
+        this.setState({deleting: true});
+    }
+
+    handleConfirmDelete () {
+        this.setState({deleting: false});
         this.props.onDelete(this.props.id);
+    }
+
+    handleCancelDelete () {
+        this.setState({deleting: false});
+    }
+
+    handleReport () {
+        this.setState({reporting: true});
+    }
+
+    handleConfirmReport () {
+        this.setState({
+            reporting: false,
+            reportConfirmed: true,
+            deleting: false // To close delete modal if reported from delete modal
+        });
+
+        this.props.onReport(this.props.id);
+    }
+
+    handleCancelReport () {
+        this.setState({
+            reporting: false,
+            reportConfirmed: false
+        });
     }
 
     render () {
@@ -45,7 +85,8 @@ class Comment extends React.Component {
             content,
             datetimeCreated,
             id,
-            projectId
+            projectId,
+            reported
         } = this.props;
 
         return (
@@ -71,7 +112,10 @@ class Comment extends React.Component {
                                     Delete {/* TODO internationalize */}
                                 </span>
                             ) : null}
-                            <span className="comment-report">
+                            <span
+                                className="comment-report"
+                                onClick={this.handleReport}
+                            >
                                 Report {/* TODO internationalize */}
                             </span>
                         </div>
@@ -79,7 +123,8 @@ class Comment extends React.Component {
                     <div
                         className={classNames({
                             'comment-bubble': true,
-                            'comment-bubble-deleted': deleted
+                            'comment-bubble-deleted': deleted,
+                            'comment-bubble-reported': reported
                         })}
                     >
                         {/* TODO: at the moment, comment content does not properly display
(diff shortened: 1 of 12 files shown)


benjiwheeler:
Re: the screenshots you included, is the third one supposed to be the same as the first one? not sure what it is supposed to depict

src/views/preview/comment/comment.jsx
```diff
@@ -113,6 +158,24 @@ class Comment extends React.Component {
...
+                {this.state.deleting ? (
+                    <DeleteCommentModal
+                        isOpen
+                        key="delete-comment-modal"
+                        onDelete={this.handleConfirmDelete}
+                        onReport={this.handleConfirmReport}
+                        onRequestClose={this.handleCancelDelete}
+                    />
+                ) : null}
+                {this.state.reporting || this.state.reportConfirmed ? (
```
benjiwheeler: for clarity, could you put parentheses around `this.state.reporting || this.state.reportConfirmed` ? Not sure about other folks, but that helps me feel confident of the way the order of operations is being handled

src/views/preview/comment/top-level-comment.jsx
```diff
@@ -33,6 +34,12 @@ class TopLevelComment extends React.Component {
         this.props.onDelete(commentId, this.props.id);
     }
 
+    handleReportReply (commentId) {
+        // Only apply topLevelCommentId for reporting replies
```
benjiwheeler: sorry, i don't follow this comment. what does "apply" mean in this context?

src/views/preview/comment/comment.jsx
```diff
@@ -7,6 +7,8 @@ const FlexRow = require('../../../components/flex-row/flex-row.jsx');
 const Avatar = require('../../../components/avatar/avatar.jsx');
 const FormattedRelative = require('react-intl').FormattedRelative;
 const ComposeComment = require('./compose-comment.jsx');
+const DeleteCommentModal = require('../../../components/modal/comments/delete-comment.jsx');
```
benjiwheeler: Why does comment.jsx live in the views/preview dir instead of a components/comment dir? Seems like that would make more sense

src/views/preview/comment/comment.jsx
```diff
@@ -33,9 +43,39 @@ class Comment extends React.Component {
     }
 
     handleDelete () {
+        this.setState({deleting: true});
+    }
+
+    handleConfirmDelete () {
+        this.setState({deleting: false});
```
benjiwheeler: is there a way we could wait for onDelete() to finish before setting state.deleting to false?

src/views/preview/comment/comment.scss
```diff
@@ -150,7 +150,7 @@
                 content: "";
             }
 
-            &.comment-bubble-deleted {
+            &.comment-bubble-deleted, .comment-bubble-reported {
```
benjiwheeler: this is a bit confusing to me. Am I right that this means:

an element with both .comment-bubble AND .comment-bubble-deleted
OR
.comment-bubble-reported (that appears under an element with) .comment-bubble
?

src/views/preview/comment/compose-comment.jsx
```diff
@@ -183,7 +183,7 @@ ComposeComment.propTypes = {
     onAddComment: PropTypes.func,
     onCancel: PropTypes.func,
     parentId: PropTypes.number,
-    projectId: PropTypes.number,
+    projectId: PropTypes.string,
```
benjiwheeler: should maybe be `PropTypes.oneOfType([PropTypes.string, PropTypes.number])`

src/views/preview/comment/comment.jsx
paulkaplan: Yeah, that was a future improvement from that first comment PR. That which lives in components/comment is actually used for a different thing. I was going to do some of those cleanup things in a separate PR.

src/views/preview/comment/top-level-comment.jsx
paulkaplan: I can change the comment. It means it is adding topLevelCommentId to the call to `onReport` as an additional param. I guess I meant it in a `Function.apply` sense.

src/views/preview/comment/comment.jsx
paulkaplan: That is a pending TODO comment in the preview reducer to add pending states to the API calls

src/views/preview/comment/comment.scss
paulkaplan: Eek yeah that is wrong, I'll fix it

src/views/preview/comment/compose-comment.jsx
paulkaplan: I think it is always a string, judging from the other places that have a propType for it (presentation.jsx)

src/views/preview/comment/comment.jsx
paulkaplan: Fixed in 237c6ec

src/views/preview/comment/comment.scss
paulkaplan: Fixed in 997d0b5

src/views/preview/l10n.json
```diff
@@ -6,5 +6,13 @@
...
-    "preview.videoMotionChip": "Video Motion"
+    "preview.videoMotionChip": "Video Motion",
+
+    "comments.report": "Report",
+    "comments.delete": "Delete",
+    "comments.reportModal.title": "Report Comment",
+    "comments.reportModal.reported": "The comment has been reported, and the Scratch Team has been notified.",
+    "comments.reportModal.prompt": "Are you sure you want to report this comment?",
+    "comments.deleteModal.title": "Delete Comment",
+    "comments.deleteModal.body": "Delete this comment? If the comment is mean or disrespectful, please click Report instead to let the Scratch Team know about it."
```
chrisgarrity: As mentioned in slack, these should be in `src/l10n.json` as comments will eventually be used on multiple views.

src/views/preview/comment/comment.jsx
```diff
@@ -71,15 +112,19 @@ class Comment extends React.Component {
                                     Delete {/* TODO internationalize */}
```
chrisgarrity: It's probably a good time to localize 'Delete' and 'Report'.

src/views/preview/comment/comment.jsx
```diff
@@ -71,15 +112,19 @@ class Comment extends React.Component {
...
                     </FlexRow>
                     <div
                         className={classNames({
                             'comment-bubble': true,
-                            'comment-bubble-deleted': deleted
+                            'comment-bubble-deleted': deleted,
+                            'comment-bubble-reported': reported
                         })}
                     >
                         {/* TODO: at the moment, comment content does not properly display
```
chrisgarrity: Should also localize 'reply' (At line 145) - I can't leave a comment on it as it's not in the diffs.

paulkaplan: Fixed in c19adee

src/views/preview/comment/comment.jsx
paulkaplan: Fixed in c19adee

src/views/preview/l10n.json
paulkaplan: Fixed in c19adee

src/views/preview/comment/top-level-comment.jsx
rschamp: This could be clearer if `commentId` was renamed to `replyId` here.

rschamp:
Assigning to @benjiwheeler since he had a previous "needs work".

PR #2165, opened 2018-10-09
fix modal prop

benjiwheeler:
### Resolves:

An issue with react-modal (https://github.com/reactjs/react-modal/issues/697) was causing errors when we passed the prop `bodyOpenClassName` a value of `null`. It works when the value is `undefined` or the prop is omitted.

### Changes:

Now, when we are not setting `bodyOpenClassName` to a custom value, we simply omit the prop entirely.

### DIFF
--- a/src/components/modal/base/modal.jsx
+++ b/src/components/modal/base/modal.jsx
@@ -23,15 +23,14 @@ class Modal extends React.Component {
         return this.modal.portal.requestClose();
     }
     render () {
-        // only set bodyOpenClassName prop if this.props.useStandardSizes is true.
-        // (careful -- setting bodyOpenClassName to null here causes an error, because
-        // ReactModal does not correctly handle the case of a null bodyOpenClassName)
-        const bodyOpenClassNameProp = this.props.useStandardSizes ?
-            {bodyOpenClassName: classNames('overflow-hidden')} : {};
         return (
             <ReactModal
                 appElement={document.getElementById('app')}
-                {...bodyOpenClassNameProp}
+                // bodyOpenClassName can be blank string, but must not be null here; a null value causes
+                // an error, because ReactModal does not correctly handle the case of a null bodyOpenClassName
+                bodyOpenClassName={classNames({
+                    'overflow-hidden': this.props.useStandardSizes
+                })}
                 className={{
                     base: classNames('modal-content', this.props.className, {
                         'modal-sizes': this.props.useStandardSizes
--- a/src/components/modal/base/modal.jsx
+++ b/src/components/modal/base/modal.jsx
@@ -23,13 +23,16 @@ class Modal extends React.Component {
         return this.modal.portal.requestClose();
     }
     render () {
+        // only set bodyOpenClassName prop if this.props.useStandardSizes is true.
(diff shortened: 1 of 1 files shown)


src/components/modal/base/modal.jsx
```diff
@@ -26,7 +26,10 @@ class Modal extends React.Component {
         return (
             <ReactModal
                 appElement={document.getElementById('app')}
-                bodyOpenClassName={this.props.useStandardSizes ? classNames('overflow-hidden') : null}
+                // only set bodyOpenClassName prop if this.props.useStandardSizes is true.
+                // (careful -- setting bodyOpenClassName to null here causes an error, because ReactModal
+                // does not correctly handle the case of a null bodyOpenClassName)
+                {...(this.props.useStandardSizes ? {bodyOpenClassName: classNames('overflow-hidden')} : {})}
```
fsih: Could you define this object as a variable above the return statement? There doesn't seem to be a great solution but I think that'll be a little more readable.

benjiwheeler:
Is this the right syntax?

src/components/modal/base/modal.jsx
```diff
@@ -23,10 +23,15 @@ class Modal extends React.Component {
...
+        // only set bodyOpenClassName prop if this.props.useStandardSizes is true.
+        // (careful -- setting bodyOpenClassName to null here causes an error, because
+        // ReactModal does not correctly handle the case of a null bodyOpenClassName)
+        const bodyOpenClassNameProp = this.props.useStandardSizes ?
+            {bodyOpenClassName: classNames('overflow-hidden')} : {};
         return (
             <ReactModal
                 appElement={document.getElementById('app')}
-                bodyOpenClassName={this.props.useStandardSizes ? classNames('overflow-hidden') : null}
+                {...bodyOpenClassNameProp}
```
rschamp: Optional classes based on the truthiness of props is a good job for `classNames`:
```jsx
                bodyOpenClassName={classNames({
                    'overflow-hidden': this.props.useStandardSizes
               })}
```

benjiwheeler: What will the value of the prop be if it omits overflow-hidden? An empty string? That might work. I only tested null (throws an error when switching between project page and editor) vs. undefined (no error).

rschamp: I would have to check, but I'm pretty sure it's an empty string since class names are supposed to be strings.

benjiwheeler:
Tried it the way @rschamp suggested, and it worked, sort of. Now, instead of using ReactModal's default bodyOpenClassName (`"ReactModal__Body--open"`), it uses a blank string. I can't see a difference in functionality, though, so I think we should do it this way.

rschamp:
I think that might be a consequence of the way the Modal component is set up, so probably isn't an issue? Especially since the current behavior should be that it's using `null` instead of `"ReactModal__Body--open"`?

benjiwheeler:
@rschamp Yes, I think it's not an issue. Looks like `ReactModal__Body--open` is just used so that you can set it in your css if you like, it's not given any styling by default.

PR #2179, opened 2018-10-12
Fix Select Menu in Messages Section

### LINKED ISSUE #1643: Filter by > Comment Activity too close to filter dropdown arrow - Messages page

sehgalvibhor:
### Resolves:

Resolves #1643

### Changes:

Fixes the text slicing in the select menu of the Messages section. Potential reason can be due to different indentation spacing/styles across different browsers. This PR changes the text-indent property to left-padding potentially fixing it for all the browsers (Chrome, IE and Mozilla tested)

### Test Coverage:

Chrome (Desktop):
image

Chrome (Mobile):
image

Firefox(Desktop):
image

Firefox(Mobile):
image

IE (Desktop):
image

IE (Responsive):

image

### DIFF
--- a/src/components/forms/select.scss
+++ b/src/components/forms/select.scss
@@ -16,7 +16,6 @@
         width: 100%;
         height: 3rem;
         margin:0px;
-        list-style-position:outside;
         padding-left: 1rem;
         color: $type-gray;
         font-size: .875rem;
--- a/src/components/forms/select.scss
+++ b/src/components/forms/select.scss
@@ -15,7 +15,9 @@
         padding-right: 4rem;
         width: 100%;
         height: 3rem;
-        text-indent: 1rem;
+        margin:0px;
+        list-style-position:outside;
+        padding-left: 1rem;
         color: $type-gray;
         font-size: .875rem;
         appearance: none;
--- a/src/components/forms/select.scss
+++ b/src/components/forms/select.scss
@@ -13,9 +13,9 @@
         border-radius: 5px;
         background: $ui-light-gray url("../../../static/svgs/forms/carot.svg") no-repeat right center;
         padding-right: 4rem;
+        padding-left: 1rem;
         width: 100%;
         height: 3rem;
-        padding-left: 1rem;
         color: $type-gray;
         font-size: .875rem;
         appearance: none;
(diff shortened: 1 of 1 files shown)


src/components/forms/select.scss
```diff
@@ -15,7 +15,8 @@
         padding-right: 4rem;
         width: 100%;
         height: 3rem;
-        text-indent: 1rem;
+        margin:0px;
```
benjiwheeler: I understand why we want to remove text-indent, but please test on a few browsers and include screenshots, since I'm not sure why this was added in the first place!

But I'm not sure why we need, or want, `margin: 0`. That seems to slightly change the height of the whole Messages header bar. If I leave it out, the menu seems to look good...

Also:
* Put a space between the `:` and the `0`}
* change `0px` to just `0` (that's what we use in the other CSS, and IMHO it's a better practice)

benjiwheeler:
AND, thanks for the contribution, this is going to make the menu look better!

sehgalvibhor:
> Could you take a look at whether or not we really need `margin: 0`?
> 
> AND, thanks for the contribution, this is going to make the menu look better!

Thanks for the review.
I checked for the margin property, it's not required. Made the necessary changes.

benjiwheeler:
Great, thanks for the fix!

benjiwheeler:
Great, thanks!

Last thing -- unfortunately our automated tests are failing because they're suuuppppeeerr strict about little details like the order that we put css settings...

So it wants you to put the `padding-left` line right after the `padding-down` line, above `width`. (See below.)

Would you mind making that one change, then I'll merge this!

screen shot 2018-10-14 at 12 36 47 am

sehgalvibhor:
Yeah, I was curious about why the CI was failing but could not really infer it. I'll quickly make the changes and update this PR.

benjiwheeler:
Thanks for this!

Issue #1318, opened 2017-05-21
User menu too thin


towerofnix:
At a full-width (greatest number of columns) page, the item "Account Settings" is split across multiple lines:

!["Profile \n My Stuff \n Account \n settings \n Sign out"](https://cloud.githubusercontent.com/assets/9948030/26285040/f7e7c5fc-3e1d-11e7-8868-118dbc84c77d.png)

Relatedly, at a min-width (least number of columns) page, the dropdown is too small to clearly present *any* of its items:

!["Profile \n My \n Stuff \n Accoun- \n setting- \n Sign \n out"](https://cloud.githubusercontent.com/assets/9948030/26285051/2684e174-3e1e-11e7-81e0-f4ea307f10f6.png)

Browser: Chrome Version 58.0.3029.110, macOS 10.12.4.

TheGrits:
Fixed in #1293

benjiwheeler:
This was not fixed in #1293 -- that change only affected screen sizes with tablet-or-less widths.

It is fixed by PR https://github.com/LLK/scratch-www/pull/2181

thisandagain:
Resolved in LLK/scratch-www#2181

towerofnix:
No big deal but that PR isn't merged yet. This should be (re-)opened and the PR should be changed to include "Resolves #1318".

Issue #2180, opened 2018-10-12
Account menu disappears on mobile when logged in


benjiwheeler:
Width 860px:
![screen shot 2018-10-12 at 9 33 20 am](https://user-images.githubusercontent.com/3431616/46872407-fd9cc900-ce01-11e8-8f00-4c927be409fe.png)

Width 520px:
![screen shot 2018-10-12 at 9 33 28 am](https://user-images.githubusercontent.com/3431616/46872401-fa094200-ce01-11e8-87af-3737457f4fa9.png)

tested on Mac Chrome

PR #2181, opened 2018-10-12
made account menu have min-width, changed max-width to rem from px

benjiwheeler:
Resolves https://github.com/LLK/scratch-www/issues/1318

Maintains status of closed issue https://github.com/LLK/scratch-www/issues/1292

Note: hard to test on mobile widths because of https://github.com/LLK/scratch-www/issues/2180

Before: 
screen shot 2018-10-12 at 9 23 00 am

After:
screen shot 2018-10-12 at 9 22 52 am

Before: 
screen shot 2018-10-12 at 9 24 40 am

After:
(Note: this is slightly wider, and uglier, than a username this width would previously have)
screen shot 2018-10-12 at 9 24 31 am

Before and after, unchanged:
screen shot 2018-10-12 at 9 24 19 am

### DIFF
--- a/src/components/dropdown/dropdown.scss
+++ b/src/components/dropdown/dropdown.scss
@@ -9,7 +9,8 @@
     border-radius: 0 0 5px 5px;
     background-color: $ui-blue;
     padding: 10px;
-    max-width: 260px;
+    min-width: 9rem;
+    max-width: 16.25rem;
     overflow: visible;
     color: $type-white;
     font-size: .8125rem;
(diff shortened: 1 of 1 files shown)


thisandagain:
@chrisgarrity Can you please take a look at this? Not sure how this got ignored for so long. Sorry about that. 😥

benjiwheeler:
Good point, Chris. There's no perfect way around this, but the greater width makes more languages stay single-line, and the failover formatting looks OK to me (maybe shorter line height would be ideal, but OK for now)

screen shot 2018-10-15 at 3 59 38 pm

PR #2182, opened 2018-10-14
Fix gh-2156: Use variables for media queries for consistency

### LINKED ISSUE #2156: Use variables for media queries for consistency
After #2153: Make sure the new media query variables from src/_frameless.scss are used everywhere in www instead of the repeatedly written media queries in the various scss files to ensure consistency.

Example:
Instead of "@media only screen and (min-width : #{$mobile}) and (max-width : #{$tablet}-1)"
use "@media #{$medium}"

sjgllghr:
### Resolves:
Resolves #2156 

### Changes:

Replaces media queries in .scss files to use variables defined in src/_framless.scss. (e.g. replace instances of `@media only screen and (max-width: $mobile - 1)` with `@media #{$small}`)

### Test Coverage:

Tested by running in multiple browsers (Firefox, Chrome, Safari) at various window sizes and comparing to the behavior and appearance of the develop branch. Also ran `make test`

### DIFF
--- a/src/views/conference/2016/expect/expect.scss
+++ b/src/views/conference/2016/expect/expect.scss
@@ -25,7 +25,7 @@
             margin-top: 1.2rem;
         }
 
-        @media only screen and (max-width: $tablet - 1) {
+        @media #{$medium-and-smaller} {
             img {
                 width: 50%;
             }
@@ -37,7 +37,7 @@
             }
         }
 
-        @media only screen and (max-width: $desktop - 1) {
+        @media #{$intermediate-and-smaller} {
             .uneven {
                 flex-direction: column;
                 align-items: center;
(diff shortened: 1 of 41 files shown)


src/views/conference/2016/expect/expect.scss
```diff
@@ -172,7 +172,7 @@
             }
```
benjiwheeler: Could you also have the line above:

`         @media only screen and (min-width: $mobile) and (max-width: $desktop - 1) {`

use the shorter query variables? Or maybe introduce a `$medium-and-intermediate` variable using this query?

sjgllghr:
Sure, just to make sure I’m seeing the same thing you are, when I search for “only screen and” I only see three results outside of the definitions, two like the one you pointed out  earlier and this one:
https://github.com/LLK/scratch-www/blob/24d07f6139c18f2dfd832f2af9cbd3c83985f05c/src/views/splash/beta/top-banner.scss#L72

I’m not quite sure how to handle this one since it has a hardcoded value in it. 

Are there more usages I am missing?

benjiwheeler:
@sjgllghr No, those are the only ones I see as well. I think you can leave that weird 1200px one, at some point we'll try to figure out why it was written that way!

Issue #1897, opened 2018-05-28
Move /info/quotes/ to scratch-www


apple502j:
It is still in scratch-www. Should it be migrated?
### Steps to Reproduce
Open https://scratch.mit.edu/info/quotes/ and it has old header

### Operating System and Browser
Windows 8.1
Chrome 66.0.3359.181

apple502j:
I can help if it's marked as help wanted

PR #2186, opened 2018-10-15
Migrate Parents Page

### LINKED ISSUE #1248: Migrate Parents Page

apple502j:
### Resolves:
Resolves #1248

### Changes:
Migrates parents page.

Differences

* Used YouTube embed instead of TED.
* We don't translate CleanSpeak as it's a name of software.
* Replaced old "help page" links with tips
* Removed quotes ( see #1897 )


thisandagain:
Thanks @apple502j! This is great. We would love to get this landed before 3.0 goes live so I greatly appreciate the effort. @chrisgarrity could you take a look at the code here?

.tx/config
```diff
@@ -156,3 +156,7 @@ source_file = src/views/microbit/l10n.json
 source_lang = en
 type = KEYVALUEJSON
 
+[scratch-website.parents-l10njson]
```
chrisgarrity: Note: I've added https://github.com/LLK/scratchr2_translations/issues/24 to remember to add the corresponding config in scratchr2_translations.

src/views/parents/parents.jsx
```diff
@@ -0,0 +1,259 @@
...
+                                }}
+                            />
+                        </p>
+                    </div>
+                </FlexRow>
+            </section>
+        </div>
+        <TitleBanner className="faq-banner">
+            <div className="inner">
+                <section id="faq">
```
chrisgarrity: The `FAQ` link in the masthead jumps the page to just below the FAQ header on the page. It may need an additional div with the id `faq` like teacher accounts on the educators page: https://github.com/LLK/scratch-www/blob/6d8235c7e8085e04abc960ea63d9b648d837789b/src/views/teachers/landing/landing.jsx#L131

The developers page also uses `TitleBanner` for the faq, so the css for that page might help as it does jump to the page with the 'FAQ' header showing.

apple502j:
Why don't you either merge or refuse this? No reply for a month... 😢

thisandagain:
Hi @apple502j. Sorry we didn't notice that you had made updates! I'm flagging for re-review by @chrisgarrity now.

Issue #2168, opened 2018-10-10
don't pass userOwnsProject to gui, pass flags about specific functionality


benjiwheeler:
Currently, www's preview.jsx passes a userOwnsProject prop to scratch-gui's containers/gui.jsx component, which passes it on to menu-bar.jsx . But it's unclear what userOwnsProject means in practice.

In https://github.com/LLK/scratch-www/pull/2163 , @rschamp commented:

>  I'm not sure if the API to GUI should be userOwnsProject because it's not clear what should happen if that's true. Instead it would be nice if the things we told GUI were what GUI could do if the user owns the project. E.g., canSave canCreate canRemix canSaveAsCopy etc.

In https://github.com/LLK/scratch-gui/pull/3303 , re: `MenuBar` getting passed the userOwnsProject prop, @rschamp commented:

> What does userOwnsProject have to do with a MenuBar component? I know I could go investigate, but it would be nice if I didn't have to, and it would tell me what it does through the name itself. It looks like the main thing it allows is enabling the Save Now button, so I think it would be better named canSave or enableSaveNow to match enableCommunity? can- seems like a more React-y prefix, so maybe enableCommunity should become canSeeCommunity instead.

PR #2187, opened 2018-10-15
set explicit capabilities like “can____” for GUI and project page

benjiwheeler:
### Resolves:

Partially resolves https://github.com/LLK/scratch-www/issues/2168 ; 

### Changes:

sets several variables:
```
canCreateNew
canRemix
canReport
canSave
canSaveAsCopy
canShare
```

Question: are there any booleans we should add? Do all of these make sense?

### Test Coverage:

No functions added; no tests added

Tried using Mac Chrome

### DIFF
--- a/src/views/preview/preview.jsx
+++ b/src/views/preview/preview.jsx
@@ -538,7 +538,7 @@ const mapStateToProps = state => {
         canReport: isLoggedIn && !userOwnsProject,
         canSave: userOwnsProject,
         canSaveAsCopy: false,
-        canShare: true,
+        canShare: userOwnsProject && state.permissions.social,
         comments: state.preview.comments,
         faved: state.preview.faved,
         fullScreen: state.scratchGui.mode.isFullScreen,
--- a/src/views/preview/preview.jsx
+++ b/src/views/preview/preview.jsx
@@ -337,6 +337,8 @@ class Preview extends React.Component {
                         addToStudioOpen={this.state.addToStudioOpen}
                         assetHost={this.props.assetHost}
                         backpackOptions={this.props.backpackOptions}
+                        canAddToStudio={this.props.canAddToStudio}
+                        canReport={this.props.canReport}
                         comments={this.props.comments}
                         editable={this.props.isEditable}
                         extensions={this.state.extensions}
@@ -382,6 +384,12 @@ class Preview extends React.Component {
                         assetHost={this.props.assetHost}
                         backpackOptions={this.props.backpackOptions}
                         basePath="/"
+                        canCreateNew={this.props.canCreateNew}
+                        canRemix={this.props.canRemix}
+                        canReport={this.props.canReport}
+                        canSave={this.props.canSave}
+                        canSaveAsCopy={this.props.canSaveAsCopy}
+                        canShare={this.props.canShare}
                         className="gui"
                         projectHost={this.props.projectHost}
                         projectId={this.state.projectId}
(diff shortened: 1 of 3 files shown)


src/views/preview/preview.jsx
```diff
@@ -513,8 +528,17 @@ const mapStateToProps = state => {
         userPresent;
     const authorPresent = projectInfoPresent && state.preview.projectInfo.author &&
         Object.keys(state.preview.projectInfo.author).length > 0;
+    const userOwnsProject = isLoggedIn && authorPresent &&
+        state.session.session.user.id === state.preview.projectInfo.author.id;
 
     return {
+        canAddToStudio: isLoggedIn && userOwnsProject,
+        canCreateNew: false,
```
rschamp: Isn't this always true?

src/views/preview/preview.jsx
```diff
@@ -513,8 +528,17 @@ const mapStateToProps = state => {
...
+        canRemix: false,
```
rschamp: I'd think this would be `!userOwnsProject`.

src/views/preview/preview.jsx
```diff
@@ -513,8 +528,17 @@ const mapStateToProps = state => {
...
+        canReport: isLoggedIn && !userOwnsProject,
+        canSave: userOwnsProject,
+        canSaveAsCopy: false,
```
rschamp: This would also be `userOwnsProject`, no?

src/views/preview/preview.jsx
```diff
@@ -513,8 +528,17 @@ const mapStateToProps = state => {
...
+        canShare: true,
```
rschamp: Only if `userOwnsProject` (and if the user `isSocial`).

src/views/preview/preview.jsx
benjiwheeler: canCreateNew? It will be true, but currently we have File->new disabled, until that PR goes up.

src/views/preview/preview.jsx
benjiwheeler: well yes, eventually! but not for this PR. right?

src/views/preview/preview.jsx
benjiwheeler: see above

src/views/preview/preview.jsx
benjiwheeler: agreed, but for later

benjiwheeler:
Did you want to discuss anything besides these current/future values?

src/views/preview/preview.jsx
rschamp: Um yes, we should keep it disabled until the GUI tells us what ID to switch to.

src/views/preview/preview.jsx
rschamp: Yes, right!

src/views/preview/preview.jsx
rschamp: This one I think we can do now though. We have all of this information available. It would just make the new project page act more similar to the existing one.

rschamp: But I think this is fine to leave for now.

rschamp:
No that was all. You're right we should enable these as they're ready to test.

src/views/preview/preview.jsx
benjiwheeler: I see your point. Changed to `canShare: userOwnsProject && state.permissions.social`

benjiwheeler: Changed after discussion

src/views/preview/preview.jsx
```diff
@@ -382,6 +384,12 @@ class Preview extends React.Component {
                         assetHost={this.props.assetHost}
                         backpackOptions={this.props.backpackOptions}
                         basePath="/"
+                        canCreateNew={this.props.canCreateNew}
+                        canRemix={this.props.canRemix}
+                        canReport={this.props.canReport}
```
rschamp: Whoops — just noticed this now after reviewing the GUI PR. We don't need to tell GUI about this.

benjiwheeler: Right.

PR #2190, opened 2018-10-15
Use the proper API route for reporting projects

paulkaplan:
/cc @benjiwheeler @colbygk this PR moves the last request from preview that was hitting `/site-api` directly to hitting the correct proxy route. I also added the token so that the API accepts the request, and I think because of fixes colby made to `/proxy` it now goes through correctly. This worked on my local setup, but once it is deployed we should just double check that it works on staging (i don't see why it wouldn't, but these proxy routes are finnicky. 

/cc @rschamp This closes the card we discussed today about "API used for all proxy routes"

### DIFF
--- a/src/redux/preview.js
+++ b/src/redux/preview.js
@@ -660,7 +660,7 @@ module.exports.reportComment = (projectId, commentId, topLevelCommentId, token) 
     });
 });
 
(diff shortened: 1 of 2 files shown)


benjiwheeler:
Great! Yay! LGTM. So it seems like API doesn't need to change at all?

PR #2192, opened 2018-10-16
Add missing l10n files to tx config

chrisgarrity:
Add views that weren't in the tx config. Without this, new strings in those files will never get translated.

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,6 +1,6 @@
 [main]
 host = https://www.transifex.com
-lang_map = zh_CN:zh-cn, zh_TW:zh-tw, pt_BR:pt-br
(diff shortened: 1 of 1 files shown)


benjiwheeler:
LGTM

.tx/config
```diff
@@ -1,6 +1,6 @@
 [main]
 host = https://www.transifex.com
-lang_map = zh_CN:zh-cn, zh_TW:zh-tw, pt_BR:pt-br
+lang_map = zh_CN:zh-cn, zh_TW:zh-tw, pt_BR:pt-br, es_419:es-419, aa_DJ:aa-dj
```
benjiwheeler: My one thought: any chance you could add a comment to this line to help clarify what it means? What are the variations that only differ by capitalization?

PR #2194, opened 2018-10-16
Add admin capabilities to preview project comments

paulkaplan:
This PR implements the admin view for comments and the "restore" action. Easiest to review each commit separately, since they each deal with a pretty separate part of the code. The first commit moves away from the `deleted` and `reported` boolean flags I had been using to a more general `visibility` string flag which can represent all the different visibility states. The second commit introduces a switch to use the admin routes for listing comments, which is on by default (i.e. will "show hidden comments" always for admins, for now). The third commit adds the restore functionality. The fourth commit fixes an issue where replies would remain visible even when the parent is deleted (and vice versa for restoring).

/cc @BryceLTaylor testing this PR i did the following manual tests:
- deleting/reporting as a non-admin user is the same, and they cannot restore comments, nor can comments previously hidden be seen
- for admin users, they will by default see all hidden comments along with the respective status. there is a restore button which will restore the comment to be visible
- As discussed with mods and @kathymakes, replies don't get a "restore" option if the parent is not visible. This is because the parent comment visibility status overrides the child when it is toggled. Children do get the "restore" option when their parent is visible.


src/views/preview/comment/comment.jsx
```diff
@@ -105,27 +112,44 @@ class Comment extends React.Component {
...
-                            ) : null}
-                            <span
-                                className="comment-report"
-                                onClick={this.handleReport}
-                            >
-                                <FormattedMessage id="comments.report" />
-                            </span>
+                            {visible ? (
+                                <React.Fragment>
+                                    {canDelete ? (
```
rschamp:
```suggestion
                                    {canDelete && (
```

src/views/preview/comment/comment.jsx
```diff
@@ -105,27 +112,44 @@ class Comment extends React.Component {
...
+                                        <span
+                                            className="comment-delete"
+                                            onClick={this.handleDelete}
+                                        >
+                                            <FormattedMessage id="comments.delete" />
+                                        </span>
+                                    ) : null}
```
rschamp:
```suggestion
                                    )}
```

src/views/preview/comment/comment.jsx
```diff
@@ -105,27 +112,44 @@ class Comment extends React.Component {
...
+                                    >
+                                        <FormattedMessage id="comments.report" />
+                                    </span>
+                                </React.Fragment>
+                            ) : (
+                                <React.Fragment>
+                                    <span className="comment-visibility">
+                                        <FormattedMessage id={`comments.status.${visibility}`} />
+                                    </span>
+                                    {canRestore && (
```
rschamp: Would be good to be consistent in how to conditionally render. I like this way better personally.

PR #2196, opened 2018-10-16
Make share banner work on preview

paulkaplan:
Sorry to steal this one from you @benjiwheeler!

Three things:
- Use `projectInfo.is_published` to tell if a project is published. Using the share time does not work if the project was subsequently unshared...
- Wire up the share button to update the published status through the API
- Move strings over to preview l10n file

One thing that is funky is that the API changes the `isPublished` attribute in the db to `is_published` when responding. But you still need to use `isPublished` to update through the API... /cc @colbygk not so happy about this...


paulkaplan:
@rschamp updated with your comment, as well as a few style fixes

PR #2197, opened 2018-10-17
Pass to and receive from GUI info about project creation lifecycle; handle url changes

benjiwheeler:
Wait to merge until https://github.com/LLK/scratch-gui/issues/3391 is merged

* Sets and provides props to gui:
  * canSave: now also true when user is logged in and project page is showing default project
  * enableCommunity: instead of always enabling, only enable when the project actually has a project/community page
* Can reset project page state to reflect change in which project is showing]
  * updates url when gui calls back with new project id

### DIFF
--- a/src/views/preview/preview.jsx
+++ b/src/views/preview/preview.jsx
@@ -403,6 +403,7 @@ class Preview extends React.Component {
                         canRemix={this.props.canRemix}
                         canSave={this.props.canSave}
                         canSaveAsCopy={this.props.canSaveAsCopy}
+                        canSaveNew={this.props.canSaveNew}
                         canShare={this.props.canShare}
                         className="gui"
                         projectHost={this.props.projectHost}
@@ -430,11 +431,12 @@ Preview.propTypes = {
         visible: PropTypes.bool
     }),
     canAddToStudio: PropTypes.bool,
-    canCreateNew: PropTypes.bool,
+    canCreateNew: PropTypes.bool, // NOTE: rename these?
     canRemix: PropTypes.bool,
     canReport: PropTypes.bool,
     canSave: PropTypes.bool,
     canSaveAsCopy: PropTypes.bool,
+    canSaveNew: PropTypes.bool,
     canShare: PropTypes.bool,
     comments: PropTypes.arrayOf(PropTypes.object),
     faved: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


src/views/preview/preview.jsx
```diff
@@ -312,6 +324,29 @@ class Preview extends React.Component {
...
+            if (projectId === 0) {
+                newUrl = `/${parts[0]}/editor`;
+                this.props.resetProject();
+            } else {
+                newUrl = `/${parts[0]}/${projectId}/editor`;
+                this.fetchCommunityData();
+            }
+            history.pushState(
+                `project ${projectId}`,
+                `project ${projectId}`,
```
rschamp: Might be more useful to us in the future if this state was something we could use easily, like `{projectId: projectId}`.

src/views/preview/preview.jsx
```diff
@@ -422,13 +459,15 @@ Preview.propTypes = {
         visible: PropTypes.bool
     }),
     canAddToStudio: PropTypes.bool,
-    canCreateNew: PropTypes.bool,
+    canCreateNew: PropTypes.bool, // NOTE: rename these?
```
rschamp: This comment looks like cruft to me... rename what in what way?

src/views/preview/preview.jsx
```diff
@@ -312,6 +324,29 @@ class Preview extends React.Component {
...
+            const parts = window.location.pathname.toLowerCase()
+                .split('/')
+                .filter(Boolean);
+            let newUrl;
+            if (projectId === 0) {
+                newUrl = `/${parts[0]}/editor`;
+                this.props.resetProject();
+            } else {
+                newUrl = `/${parts[0]}/${projectId}/editor`;
+                this.fetchCommunityData();
```
rschamp: In both of these cases, a change to the URL should trigger a change to the `projectId`, which should cause the "init" process to happen in `componentDidUpdate` and trigger these that way.

src/views/preview/preview.jsx
benjiwheeler: Done

src/views/preview/preview.jsx
benjiwheeler: removed

src/views/preview/preview.jsx
benjiwheeler: Done, take a look

src/views/preview/preview.jsx
```diff
@@ -66,7 +68,7 @@ class Preview extends React.Component {
             extensions: [],
             favoriteCount: 0,
             loveCount: 0,
-            projectId: parts[1] === 'editor' ? 0 : parts[1],
+            projectId: parts[1] === 'editor' ? '0' : parts[1],
             addToStudioOpen: false,
             reportOpen: false
         };
```
benjiwheeler: FYI, the call to getExtensions below should maybe be removed... I tested without it, the page loads the same and shows extensions, they just show up a moment later (because they aren't being loaded until after projectInfo is fetched)

Issue: https://github.com/LLK/scratch-www/issues/2216

PR #2198, opened 2018-10-17
Comment minor fixes

paulkaplan:
A few minor fixups after seeing the comments on staging

1. Add a missing comment status message
2. Show comment delete for admins as well as project owners
3. Only show report for logged in users

### DIFF
--- a/src/l10n.json
+++ b/src/l10n.json
@@ -238,6 +238,7 @@
     "comments.status.delbyclass": "Deleted by class",
     "comments.status.hiddenduetourl": "Hidden due to URL",
     "comments.status.markedbyfilter": "Marked by filter",
+    "comments.status.censbyunconstructive": "Censored unconstructive",
     "comments.status.suspended": "Suspended",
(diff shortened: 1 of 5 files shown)


src/l10n.json
```diff
@@ -238,6 +238,7 @@
     "comments.status.delbyclass": "Deleted by class",
     "comments.status.hiddenduetourl": "Hidden due to URL",
     "comments.status.markedbyfilter": "Marked by filter",
+    "comments.status.censbyunconstructive": "Censored unconstructive",
```
benjiwheeler: could this (and others above, ideally) name use camelCase instead of all-lowercase? camelCase seems to be the norm above...

benjiwheeler:
Nice, this looks very clean! One tiny suggestion.

src/l10n.json
paulkaplan: Yeah they are that way because they being concatted directly from the server response

PR #2206, opened 2018-10-18
Add * to comments from staff on preview

paulkaplan:
This PR will show the staff * next to staff usernames on comments, once that information starts coming from the API on comments.

### DIFF
--- a/src/views/preview/comment/comment.jsx
+++ b/src/views/preview/comment/comment.jsx
@@ -111,7 +111,9 @@ class Comment extends React.Component {
                         <a
                             className="username"
                             href={`/users/${author.username}`}
-                        >{author.username}</a>
+                        >
+                            {author.username}{author.is_staff ? '*' : ''}
+                        </a>
                         <div className="action-list">
                             {visible ? (
                                 <React.Fragment>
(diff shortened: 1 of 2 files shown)


src/views/preview/comment/comment.jsx
```diff
@@ -214,6 +216,7 @@ Comment.propTypes = {
     author: PropTypes.shape({
         id: PropTypes.number,
         image: PropTypes.string,
+        is_staff: PropTypes.bool,
```
rschamp: Even though this is what we use, we probably would want to clean up this name in the API so it was just `isScratchTeam` or `scratchteam`?

paulkaplan: hmm yeah, although as with `isPublished` vs. `is_published`, I think it gets even more confusing when you start renaming properties...

rschamp: Kinda... this is the job of our models in the API though. There are plenty of other places where we don't blindly pass though database fields.

paulkaplan: Ok, well let me know what it should be and I'll change it here and in my PR against API https://github.com/LLK/scratch-api/pull/614

paulkaplan:
@rschamp updated

PR #2213, opened 2018-10-19
Integrate www-gui localization

### LINKED ISSUE #2172: Integrate gui and www selected language
www and gui maintain the current locale as state, but they're not synchronized.

Changing the language in gui should update the locale for www, and changing the locale in www should set the locale for gui.

chrisgarrity:
### Resolves:

Fixes #2172 (along with https://github.com/LLK/scratch-gui/pull/3431)

### Changes:
* Initialize `locales` in redux using methods exported by GUI.
* pass GUI method to set the `scratchlanguage` cookie

NOTE: The  does not reload the page, so switching back to the project page will not reflect language changes made in gui until the next page reload. Reloading the page will lose project state so we don’t want to do that.

### Test Coverage:

- [ ] Change language in www, load preview, then see inside: the gui should load in the currently selected language (the language needs to be in both gui and www)
- [ ] View a project in the editor: you can switch languages.
- [ ] View community after changing languages in gui: the preview will still be in the previous language. On reload it will show the new language.

### DIFF
--- a/src/views/preview/preview.jsx
+++ b/src/views/preview/preview.jsx
@@ -12,6 +12,7 @@ const render = require('../../lib/render.jsx');
 const storage = require('../../lib/storage.js').default;
 const log = require('../../lib/log');
 const EXTENSION_INFO = require('../../lib/extensions.js').default;
+const jar = require('../../lib/jar.js');
 
 const PreviewPresentation = require('./presentation.jsx');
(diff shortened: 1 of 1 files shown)


chrisgarrity:
I've marked this blocked so that the gui changes can land first.

BryceLTaylor:
This was tested in a recent bug hunt.  Looks good!

Issue #2093, opened 2018-09-19
Redesign Credits Page - Phase 1


kathymakes:
## Overview

We want to make some changes to how collaborators are represented, make sure the list of folks working on Scratch is up-to-date, and clean up the overall design of the page. 

Related links
>**[Invision prototype
](https://invis.io/PVO4F19TJ3G)[Github #2061
](https://github.com/LLK/scratch-www/issues/2061)**

## Content
Here's a google doc that has all copy + relevant links for the page.

> **[Credits Page Content](https://docs.google.com/document/d/1qjnMru-wJCcFina7zdhNA3HqPBtqsZN6HLVrFYosMKQ/edit#)**

Note that the text + images in the drawings might not be up to date.

## Designs

More style variations right now on [Invision](https://invis.io/PVO4F19TJ3G).

Although not totally finalized yet, here's the design I feel works best. White background gives brightness and contrast to Scratch team pictures, also to text-heavy parts of page. 

### Sizing
(More details in drawing)

- avatar size: 80px
- avatar font size: 12px
- avatar line height: 16px
- header font size: 32px
- header font weight: bold (700)
- body font size: 16px
- body line height: 24px
- text max width: 520px

![image](https://user-images.githubusercontent.com/8203939/45772894-e578c600-bc16-11e8-9325-7aa81cd955f5.png)

### Responsive

Proposal for a multi-column grid scaling down.
- Margins on body content shrink
- One column of images wrap with each breakpoint.
- On the smallest size (320), all font sizes shrink down
- On the smallest size (320), avatar blocks get smaller as well.

<img width="1072" alt="credits responsive" src="https://user-images.githubusercontent.com/8203939/45772185-050eef00-bc15-11e8-8c20-10bcf302688a.png">

kathymakes:
More detailed spec of sponsor logos:

![image](https://user-images.githubusercontent.com/8203939/46957887-89169400-d066-11e8-87e5-2120dc54ed05.png)

PR #2220, opened 2018-10-21
Redo the credits page.

picklesrus:
See https://github.com/LLK/scratch-www/issues/2093 for details.

### Resolves:
#2093 
_
### Changes:

Update the credits page

screen shot 2018-10-21 at 2 59 22 pm
screen shot 2018-10-21 at 2 59 10 pm
screen shot 2018-10-21 at 2 58 52 pm
screen shot 2018-10-21 at 2 58 34 pm


src/views/credits/credits.jsx
```diff
@@ -2,458 +2,158 @@ const React = require('react');
 const render = require('../../lib/render.jsx');
 const FormattedHTMLMessage = require('react-intl').FormattedHTMLMessage;
```
chrisgarrity: `FormattedHTMLMessage` has been deprecated. Whenever we update strings that use it, we replace them with normal `FormattedMessage` strings with placeholders. The wiki has [instructions for doing the conversion.](https://github.com/LLK/scratch-www/wiki/Localization-Guide#removing-formattedhtmlmessage)

src/views/credits/l10n.json
```diff
@@ -1,13 +1,15 @@
...
   "credits.researchersTitle": "Scratch Researchers",
-  "credits.researchersBody": "<a href=\"https://scratch.mit.edu/info/research/\">Research on Scratch</a> is being conducted by members of the MIT Scratch Team and researchers at other universities, including Yasmin Kafai (who collaborated on the <a href=\"http://www.nsf.gov/awardsearch/showAward?AWD_ID=0325828\">initial NSF Scratch grant</a>) at the University of Pennsylvania Graduate School of Education, Karen Brennan (who leads the <a href=\"http://scratched.gse.harvard.edu/\">ScratchEd project</a>) at the Harvard Graduate School of Education, Benjamin Mako Hill at the University of Washington, Andrés Monroy Hernández at Microsoft Research, Mimi Ito and Crystle Martin at the University of California, Irvine, Quinn Burke at College of Charleston, Deborah Fields at Utah State University, and Kylie Peppler at Indiana University.",
+  "credits.researchersBody" : "<a href=\"https://scratch.mit.edu/info/research/\">Research on Scratch</a> is being conducted by members of the MIT Scratch Team and researchers at other universities, including:",
```
chrisgarrity: This will need a link placeholder, and the link text will need to be defined.

src/views/credits/l10n.json
```diff
@@ -1,13 +1,15 @@
...
+  "credits.researchersContributors" : "Yasmin Kafai (who collaborated on the <a href=\"http://www.nsf.gov/awardsearch/showAward?AWD_ID=0325828\">initial NSF Scratch grant</a>) at the University of Pennsylvania Graduate School of Education, Karen Brennan (who leads the <a href=\"http://scratched.gse.harvard.edu/\">ScratchEd project</a>) at the Harvard Graduate School of Education, Benjamin Mako Hill at the University of Washington, Andrés Monroy Hernández at Microsoft Research, Mimi Ito and Crystle Martin at the University of California, Irvine, Quinn Burke at College of Charleston, Deborah Fields at Utah State University, and Kylie Peppler at Indiana University.",
```
chrisgarrity: Yuk, this one's a mess for translation. I guess we don't update researchers that often, but each time we do it requires translators to retranslate the whole string. And it has the NSF link that should be a place holder. @kathymakes could we take a look at redoing the researchers so that it's not such a burden for translators?

src/views/credits/l10n.json
```diff
@@ -1,13 +1,15 @@
...
   "credits.acknowledgementsTitle": "Acknowledgements",
   "credits.acknowledgementsContributors": "The following people have also contributed to the development and support of Scratch over the years:",
   "credits.acknowledgementsTranslators": "With the help of <a href=\"http://wiki.scratch.mit.edu/wiki/Translators\">Scratch Translators</a> around the world, Scratch is available in many languages.",
```
chrisgarrity: Another HTML string that should have a placeholder, there are probably others in the unchanged parts of l10n.

paulkaplan:
@picklesrus looks like @chrisgarrity noticed some blocking issues with the way this is localized. Could you work with her to fix it up?

picklesrus:
I've converted everything to FormattedMessage!
I didn't change the researchers message though. Still need to chat with Kathy about that.  Since this is my first www PR, I imagine there will be code issues, so I think it should be okay to review in the meantime.

src/views/credits/credits.scss
```diff
@@ -1,39 +1,205 @@
...
+@media only screen and (min-width: 1041px) and (max-width: 1280px) {
+    .credits {
+        .logoGrid {
+            max-width: 864px;
+         }
+    }
+}
+@media only screen and (min-width: 769px ) and (max-width: 1040px) {
+    .credits {
+        text-align: left;
```
paulkaplan: I do not think the text-align lefts are necassary

paulkaplan: I do not think the text-align lefts are doing anything here.

src/views/credits/credits.scss
```diff
@@ -1,39 +1,205 @@
...
+        .acknowledgeContent,
+        .midHeader {
+            margin: 0 44px;
+        }
+    }
+}
+@media only screen and (max-width: 364px) {
+    .credits {
+        text-align: left;
+        font-size: .6rem;
```
paulkaplan: I do not think the font-size here does anything, because the font sizes are set on `p` to 1rem. Which element is this supposed to control the font size on?

src/views/credits/credits.scss
```diff
@@ -1,39 +1,205 @@
...
+        span {
+            display: inline-block;
+            width: 144px;
+            font-size: .875rem;
+            word-wrap: break-word;
+
+        }
+    }
+
+    .acknowledgeContent {
```
paulkaplan: Nit: The convention for css class names is `-` separated.

src/views/credits/credits.scss
```diff
@@ -1,39 +1,205 @@
...
+            }
+
+            span {
+                width: 96px;
+            }
+        }
+
+        .logoGrid {
+            display: flex;
+            margin: 0 auto;
```
paulkaplan: Several of these are redundant with the logoGrid props at the top.

src/views/credits/credits.scss
```diff
@@ -1,39 +1,205 @@
...
+        align-items: center;
+    }
+
+    .title-banner-h1 {
+        font-size: 2rem;
+    }
+
+}
+
+@media only screen and (min-width: 1041px) and (max-width: 1280px) {
```
paulkaplan: One thing we can pair on together quickly is that it is easiest to build up your styles from smallest size to largest, and just use `min-width: nextSize` as you go down the stylesheet and get bigger and bigger. that way you do not have to set max-widths as well, and you can take advantage of the cascading-ness of css.

src/views/credits/credits.scss
```diff
@@ -1,39 +1,205 @@
...
+            border-radius: 8px;
             background-color: $ui-white;
-            padding: 20px;
-            width: 85px;
-            height: 85px;
+            width: 80px;
+            height: 80px;
+        }
+
+        span {
```
paulkaplan: Styling spans is not a great idea here for two reasons:
1. You are using a span in a specific way (looks like for each team members name) and so it would make more sense to use a class name, making the CSS easier to understand. It is better to use class names than to imply meaning from the nesting of selectors (i.e. this implicitly means/assumes that an li > span is a contributor name)
2. React automatically create spans in certain situations (when you do, for example `<div>text</div>`, the "text" will get wrapped in a span), and so styling all spans may have unintended consequences because the selector is too broad (which also ties in with 1)

src/views/credits/credits.scss
```diff
@@ -1,39 +1,205 @@
...
-        margin: 10px 0;
-        width: 188px;
-        text-align: center; 
+        margin: 0;
+        padding-bottom: 32px;
+        text-align: center;
+        line-height: 1.25rem;
 
         img {
-            margin: 10px 23px;
```
paulkaplan: Similar to the comment below about spans, you probably want to use a class name here. You don't want to rely on li > img "meaning" avatar, it makes it hard to change and makes it easy for others to make mistakes.

paulkaplan: In fact, this comment basically applies to any place where you are nesting an element selector underneath another element selector. Again, very happy to pair on this to fix!

picklesrus:
Thanks for pairing!  Will wait til Friday to merge :)

PR #2223, opened 2018-10-23
Responsive about page!

### LINKED ISSUE #891: Mobile About Page Video Bug
The video on the About page does not seem to be responsive on my iPhone 5s Safari creating a bad background issue:

Something to consider is whether it would be better to have the video shrink size to fit or move to beneath/above the intro text when the screen size is small.

Also, ignore the white box as messages, that's a known bug on my phone :) Sorry!

sehgalvibhor:
### Resolves:

Closes #891

### Changes:

About page does not seem to be responsive, keeps flowing out of the div. Changing the layout to grid ensures that the page is responsive, video auto flows to the next row in case of reduction in screen size.

### Test Coverage:

Tested on Chrome:

Full Page:
image

Responsive Page:
image

Iphone 5:
image

Tested on Firefox:

Full page:
image

Responsive:
image

### DIFF
--- a/src/views/about/about.scss
+++ b/src/views/about/about.scss
@@ -2,14 +2,14 @@
 
 .about {
     .masthead {
-        display: flex;
-        flex-wrap: no-wrap;
+        display: grid;
+        grid-template-columns: repeat(auto-fit, minmax(461px, 1fr));;
         justify-content: space-between;
         align-items: stretch;
 
         div {
             display: inline-block;
-            width: calc(50% - 10px);
+            width: calc(100% - 10px);
 
             ul {
                 margin: 0;
--- a/src/views/about/about.scss
+++ b/src/views/about/about.scss
@@ -30,10 +30,10 @@
             }
 
             iframe {
+                display: block;
                 border: 1px solid $ui-gray;
                 width: 100%;
                 height: 290px;
-                display: block;
             }
         }
     }
--- a/src/views/about/about.scss
+++ b/src/views/about/about.scss
@@ -2,14 +2,15 @@
 
 .about {
     .masthead {
-        display: grid;
-        grid-template-columns: repeat(auto-fit, minmax(461px, 1fr));;
+        display: flex;
+        flex-wrap: no-wrap;
         justify-content: space-between;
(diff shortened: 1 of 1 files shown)


benjiwheeler:
This is helpful! And/but, I believe (@LiFaytheGoblin, can you confirm?) we have chosen not to use CSS grid for now, because compatibility is not quite widely enough established.

https://caniuse.com/#feat=css-grid shows very good adoption so far, but I think we are looking for higher adoption before we move to using grid.

benjiwheeler:
Thanks, that does look like an improvement on the current version! Maybe you can update the code here to use that approach, and I can take a look.

Also -- just a suggestion -- you might want to look at some of the recent changes to preview.scss by @LiFaytheGoblin , and see if there are patterns she used that you might want to use here.

sehgalvibhor:
Sure, I'll check the changes and see if I can replicate them in the about page too.
Thanks!

LiFaytheGoblin:
The flex component MediaQuery could be used to display the video on different places depending on the screen size @sehgalvibhor

benjiwheeler:
@sehgalvibhor If you're not able to take a look at replicating the media query pattern right now, maybe we should just merge in your current changes, since they're an improvement? Then we can always come back and do a more thorough change later.

sehgalvibhor:
@benjiwheeler : Yeah I think I might just need a bit more time to play around with it and test it properly. We can merge the changes however I think Travis is failing. Will have to re-order the CSS properties right?

joker314:
It looks like it!

```
  33:17  error  Expected `display`, found `border`  property-sort-order
  34:17  error  Expected `border`, found `width`    property-sort-order
  35:17  error  Expected `width`, found `height`    property-sort-order
  36:17  error  Expected `height`, found `display`  property-sort-order
```
 So `display` then `border` then `width` then `height`.

sehgalvibhor:
@joker314 : Thanks for reviewing. Fixed it!
@benjiwheeler : We can merge for now.

benjiwheeler:
Thanks, @sehgalvibhor (and @joker314)!

PR #2226, opened 2018-10-23
Extract strings for localizing the preview page

paulkaplan:
@chrisgarrity This PR pulls out the rest of the strings that were written directly into the preview page.


src/views/preview/extension-chip.jsx
```diff
@@ -19,7 +19,7 @@ const ExtensionChip = props => (
             }
             {props.hasStatus && (
                 <div className="extension-status">
-                    Needs Connection
+                    <FormattedMessage id="preview.needsConnection" />
```
chrisgarrity: OK, this is technical debt, but maybe we should deal with it before we get any deeper. Eventually this will be the project view, the string ids probably should be 'project.something'. If we change them later, they will all be untranslated at that point. The system should fill in the translations from previous translations, but they will be 'unreviewed'. If we rename them all now it will be less churn for translators later.

paulkaplan:
@chrisgarrity updated the string ids

PR #2230, opened 2018-10-24
Showing specific comment by url hash on preview

paulkaplan:
This PR enables loading a single comment into view by the existing URL hash scheme (#comment-<id>). The functionality is different from 2.0 project page because it does not load up all comments page-by-page until it finds the right one. Instead, it loads the requested comment specifically. If the comment is a top-level comment, it is shown and replies are loaded. If the requested comment is a reply, the associated top-level comment is loaded and shown, then the replies are loaded. In either case, the requested comment is highlighted and then scrolled into view.

/cc @chrisgarrity because you were mentioning URL schemes, this is one that has to be maintained because other parts of the app are generating links (comment reports, notifications).

We discussed adding a button to "go back" to showing all the comments, but I'm going to defer that until further investigation into how this feature is used.

### DIFF
--- a/src/redux/preview.js
+++ b/src/redux/preview.js
@@ -399,6 +399,35 @@ module.exports.getTopLevelComments = (id, offset, isAdmin, token) => (dispatch =
     });
 });
 
+module.exports.getCommentById = (projectId, commentId, isAdmin, token) => (dispatch => {
+    dispatch(module.exports.setFetchStatus('comments', module.exports.Status.FETCHING));
+    api({
+        uri: `${isAdmin ? '/admin' : ''}/projects/comments/${commentId}`,
+        authentication: isAdmin ? token : null
+    }, (err, body) => {
+        if (err) {
+            dispatch(module.exports.setFetchStatus('comments', module.exports.Status.ERROR));
+            dispatch(module.exports.setError(err));
+            return;
+        }
+        if (!body) {
+            dispatch(module.exports.setFetchStatus('comments', module.exports.Status.ERROR));
+            dispatch(module.exports.setError('No comment info'));
+            return;
+        }
+
+        if (body.parent_id) {
+            // If the comment is a reply, load the parent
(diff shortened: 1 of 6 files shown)


src/redux/preview.js
```diff
@@ -399,6 +399,35 @@ module.exports.getTopLevelComments = (id, offset, isAdmin, token) => (dispatch =
...
+        }
+        if (!body) {
+            dispatch(module.exports.setFetchStatus('comments', module.exports.Status.ERROR));
+            dispatch(module.exports.setError('No comment info'));
+            return;
+        }
+
+        if (body.parent_id) {
+            // If the comment is a reply, load the parent
+            return dispatch(module.exports.getCommentById(projectId, body.parent_id, isAdmin, token));
```
rschamp: Will this, in turn load the sibling replies after the parent loads?

src/views/preview/preview.jsx
```diff
@@ -65,11 +65,18 @@ class Preview extends React.Component {
...
+        const commentHashPrefix = '#comment-';
+        const singleCommentId = window.location.hash.indexOf(commentHashPrefix) !== -1 &&
+            parseInt(window.location.hash.replace(commentHashPrefix, ''), 10);
+
         this.state = {
             extensions: [],
             favoriteCount: 0,
             loveCount: 0,
             projectId: parts[1] === 'editor' ? '0' : parts[1],
+            singleCommentId: singleCommentId,
```
rschamp: Does this belong in state? Will it change? We could probably do this where we actually render the view and pass it in as a prop.

src/redux/preview.js
paulkaplan: Yup! It is a recursive call that bottoms up when you load the parent_id, because it definitely does not have a parent_id. So it ends up at the calls below which load the replies

src/views/preview/preview.jsx
paulkaplan: I was planning to leave it as state because we may add a way to clear out this view and instead just view all comments.

PR #2231, opened 2018-10-24
Decorate comment text to add @username links and scratch-domain links

paulkaplan:
Implements the feature for converting scratch domain links and @usernames to /users/username links in the comments.

I did this by enhancing the `decorateText` helper to optionally convert scratch domain link text to react links. There was an issue in the documentation for decorateText, which was actually returning an array, not a string. For comments, each element of the resulting array is either rendered directly if it is a component (i.e. link) or put inside `<EmojiText ...>` since it may already contain an imageified emoji (handled by the server). 

This code also makes more explicit what kind of conversions we expect in each part of the app, since you have to pass an object of flags indicating what you want.

### DIFF
--- a/src/views/preview/presentation.jsx
+++ b/src/views/preview/presentation.jsx
@@ -238,7 +238,11 @@ const PreviewPresentation = ({
                                             value={projectInfo.instructions}
                                         /> :
                                         <div className="project-description">
-                                            {decorateText(projectInfo.instructions)}
+                                            {decorateText(projectInfo.instructions, {
+                                                usernames: true,
+                                                hashtags: true,
+                                                scratchLinks: false
+                                            })}
                                         </div>
                                     }
                                 </FlexRow>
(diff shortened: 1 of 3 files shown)


src/views/preview/presentation.jsx
```diff
@@ -238,7 +238,11 @@ const PreviewPresentation = ({
                                             value={projectInfo.instructions}
                                         /> :
                                         <div className="project-description">
-                                            {decorateText(projectInfo.instructions)}
+                                            {decorateText(projectInfo.instructions, {
+                                                usernames: true,
+                                                hashtags: true,
+                                                scratchLinks: false
```
chrisgarrity: I know that we don't link scratch links in project descriptions on 2.0, but was that a design decision, or just because we didn't go back and add it when linking was added to comments? /cc @carljbowman

PR #2284, opened 2018-11-06
Avoid displaying an error when /news returns a 500

rschamp:
This situation probably occurs most frequently when running dev servers while offline, but could also happen if the API is having issues. 500 responses from the API should not take down the homepage so drastically, and are also sometimes unavoidable while working offline.

### DIFF
--- a/src/views/splash/splash.jsx
+++ b/src/views/splash/splash.jsx
@@ -71,7 +71,10 @@ class Splash extends React.Component {
     getNews () {
         api({
             uri: '/news?limit=3'
-        }, (err, body) => {
+        }, (err, body, resp) => {
+            if (resp.statusCode !== 200) {
+                return log.error(`Unexpected status code ${resp.statusCode} received from news request`);
+            }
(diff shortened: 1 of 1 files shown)


benjiwheeler:
It's not clear to me what this improves. When I run the current code vs the changed code, each time setting the network to offline right before this request goes out, I see no visible difference in the page. There is some difference in console errors; is that the idea?

rschamp:
Without a connection to the outside network, scratch-api should return a 500 response for /news. Setting to offline mode in your browser won't prevent the API from doing the fetch over the network. The failure mode is the homepage displaying a blue "Scratch has crashed" screen instead of the homepage.

benjiwheeler:
Got it! Will take another look

benjiwheeler:
LGTM. There are probably even more places where we aren't checking the response status or content before returning it...

rschamp:
Most definitely. I just wanted this to get into develop since I needed it in order to do anything on an airplane.

PR #2286, opened 2018-11-06
Update chromedriver so tests pass more reliably

BryceLTaylor:
### Resolves:

test_footer_links tests were failing due to a problem in the version of chromedriver. 

### Changes:

Updates the package.json in test/integration to include the latest and working version of chromedriver.

### Test Coverage:

`SMOKE_USERNAME=[username] SMOKE_PASSWORD_[password] npm run smoke` passing in a valid staging username and password.  The tests should now pass.

### DIFF
--- a/test/integration/package.json
+++ b/test/integration/package.json
@@ -1,6 +1,6 @@
 {
     "dependencies": {
         "selenium-webdriver": "3.6.0",
-        "chromedriver": "2.37.0"
+        "chromedriver": "2.43.1"
     }
 }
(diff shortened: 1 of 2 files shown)


BryceLTaylor:
I realized it would be a good idea to update the version of chrome in the same PR.  Saucelabs does not support Mojave yet, so that update will have to come later.

paulkaplan:
Ah good call. This actually explains why it was passing on sauce (which does let you set the chrome version) but failing locally.

PR #2289, opened 2018-11-07
Pass authorId and authorUsername to gui

benjiwheeler:
Should be merged before: https://github.com/LLK/scratch-gui/pull/3681

### Resolves:

Resolves, partially: https://github.com/LLK/scratch-gui/issues/3597

### Changes:

Defines props on preview.jsx called `authorId` and `authorUsername`; passes them to gui, so they can be used to show creator info.


src/views/preview/preview.jsx
```diff
@@ -452,6 +452,8 @@ class Preview extends React.Component {
                     <IntlGUI
                         hideIntro
                         assetHost={this.props.assetHost}
+                        authorId={this.props.authorId}
```
rschamp: I believe thumbnailUrl is one of the attributes of `state.preview.projectInfo.author`? If not, it should be! Like I said in https://github.com/LLK/scratch-gui/pull/3681, I think we should pass that in as `authorThumbnailUrl` rather than using this `id` to construct it in the GUI.

benjiwheeler: Sorry, do you mean 1. "should" as in "ought to change so that..." or 2. "should" as in "already probably does"? 

1. Not sure I agree -- `user` models returned by the API can include the fields `joined` date, `bio`, `country`, etc. (See `routes/users/model.js` in API). Do we really want all that to be fetched by getProjectInfo(), even if it means an additional db query on every project fetch?

2. Unless I totally misunderstand the code, `author` as returned by getProjectInfo() only includes a small fraction of the full `user` model.

towerofnix: >Also, whether or not we make author have the other info that API has access to, API does not currently know anything about user profile thumbnails, as far as I can see.

Not sure if you're saying that the API doesn't know about user thumbnails in general, or just the `/projects` endpoints, but - it does return thumbnails in `/users/<username>`, i.e. in `profile.images['90x90']` (and `['50x50']`, etc).

benjiwheeler: I see what you mean! OK, it seems like API could return author thumbnails...

src/lib/user-thumbnail.js
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Generate a thumbnail url for a particular userid, with width and height.
+ * @param {string} userId userId for the user whose thumbnail we want
+ * @param {number} width desired thumbnail width; defaults to 32
+ * @param {number} height desired thumbnail height; defaults to width.
+ * @returns {string} thumbnail url string
+ */
+const thumbnailUrl = (userId, width, height) => (
+    `/get_image/user/${userId}_${width ? width : 32}x${height ? height : (width ? width : 32)}.png`
```
rschamp: This is _probably_ a temporary measure, but we should set it up as if it's permanent in case it doesn't end up temporary. If this goes live, we need these images to be cached, so this should use a hostname. In production we're using `cdn2.scratch.mit.edu` for `get_image` URLs. On Fastly we refer to this as "static", so maybe `process.env.STATIC_HOST`? Follow the pattern of `API_HOST` etc., to make sure it's set correctly in Travis as well. In development, we can set it to `""` and you'll get whatever you were seeing here without any host.

src/views/preview/preview.jsx
rschamp: No this PR shouldn't wait for that, we should log it as a follow up.

src/lib/user-thumbnail.js
benjiwheeler: OK, take a look

benjiwheeler:
@rschamp How does this look? I can do a follow-up to change hardcoded production cdn urls to use STATIC_HOST. And we could take a look together at how to switch to using cdn.scratch.ly.

src/views/preview/preview.jsx
```diff
@@ -485,6 +490,9 @@ class Preview extends React.Component {
 
 Preview.propTypes = {
     assetHost: PropTypes.string.isRequired,
+    authorId: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
```
rschamp: I see why now, but this still feels more confusing than correct, even though it's technically correct... but fixing it in `mapStateToProps` might be equally confusing (I think you could add `|| null` to the end of all the assignments leading up to `authorId` and `authorUsername`).

Would you please add a short comment to clarify here?

```suggestion
    // If there's no author, this sometimes ends up being `false`
    authorId: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
```

benjiwheeler: Done

PR #2290, opened 2018-11-07
Add "project not available" view for when project body cannot be loaded

paulkaplan:
This PR implements a special view to be rendered when the server "cannot find" the project that is being requested. This happens when the project id does not exist, or is not visible to the given user.

image


src/views/preview/project-not-available.jsx
```diff
@@ -0,0 +1,24 @@
+const React = require('react');
+const FormattedMessage = require('react-intl').FormattedMessage;
+const FlexRow = require('../../components/flex-row/flex-row.jsx');
+
+require('./project-not-available.scss');
+
+const ProjectNotAvailable = () => (
```
towerofnix: The language here is generic - the same text and image used on every scratchr2 404 page. It also looks like none of the code here is specific to the project page. Should this be separated into a generic 404 view, which can be required and used as needed?

paulkaplan: @towerofnix good point! Moved to a component

src/redux/preview.js
```diff
@@ -38,7 +38,8 @@ module.exports.getInitialState = () => ({
     projectStudios: [],
     curatedStudios: [],
     currentStudioIds: [],
-    moreCommentsToLoad: false
+    moreCommentsToLoad: false,
+    projectNotAvailable: false
```
benjiwheeler: FWIW, I prefer using positive names rather than negative names where possible -- e.g., `projectAvailable` here. They make me less confused when mentally parsing lines like `projectNotAvailable: !!action.info` below.

("FWIW" == "I want to say that but I don't expect you to change it")

src/redux/preview.js
```diff
@@ -51,7 +52,8 @@ module.exports.previewReducer = (state, action) => {
         return module.exports.getInitialState();
     case 'SET_PROJECT_INFO':
         return Object.assign({}, state, {
-            projectInfo: action.info
+            projectInfo: action.info ? action.info : {},
+            projectNotAvailable: !!action.info
```
benjiwheeler: Maybe I'm not following, but doesn't this do the opposite of what we want? If `action.info` is a bunch of project data, won't `!!action.info` be `true`? As in, we'd think the project is not available?

src/redux/preview.js
```diff
@@ -318,6 +320,7 @@ module.exports.getProjectInfo = (id, token) => (dispatch => {
         if (typeof body === 'undefined') {
             dispatch(module.exports.setFetchStatus('project', module.exports.Status.ERROR));
             dispatch(module.exports.setError('No project info'));
+            dispatch(module.exports.setProjectInfo(null));
```
benjiwheeler: Now that you're handling falsy params to `setProjectInfo`, seems like this could be (a tiny bit) cleaner like:

```
        if (body) {
            dispatch(module.exports.setFetchStatus('project', module.exports.Status.FETCHED));
        } else {
            dispatch(module.exports.setFetchStatus('project', module.exports.Status.ERROR));
            dispatch(module.exports.setError('No project info'));
        } 

        dispatch(module.exports.setProjectInfo(body));
```

paulkaplan:
@benjiwheeler thanks for catching that. I was think I would prefer to keep the error logic separate in an early return, just to keep the code as clear as possible. I also added a check for 404 response types, which can occur without an undefined body, but are the same type of error.

src/redux/preview.js
paulkaplan: Usually I would agree. The reason I used the negative here is that it feels like it more accurately captures the state. I wanted to avoid "404" or technical language like that which would imply something more specific than what is actually happening. Does that make sense?

benjiwheeler: Sure, I can see that perspective.

src/redux/preview.js
benjiwheeler: Just to be clear, is the `body` truthy or falsy when `statusCode === 404`? I'm still wondering if you can simplify the code per my suggestion above!

paulkaplan: Body seems to be truthy in that case

Issue #2281, opened 2018-11-05
Project Page: First time users have "Getting Started" tutorial open


chrisgarrity:
- [x] logged out users that open a new project should have getting started open by default. 
- [x] make loading tutorial by name work in www/gui

PR #2302, opened 2018-11-12
Issue/2281 gettingstarted

chrisgarrity:
### Resolves:
#2281 

### Changes:
- allow a query string on the route
- `create` on the navbar opens the getting started tutorial if the user is not logged in
- replaced any other use of `tip_bar=getStarted` with `tutorial=getStarted` (with the exception of Tips page)
- did not replace other uses of `tip_bar=home` as those should probably go to the new tips/ideas page.

### Test Coverage:

Manual testing. Note that the links still go to `/projects/editor` not `/preview/editor` so they don't have to be edited again before launch. You have to manually replace `projects` with `preview` to see the changes.


src/routes.json
```diff
@@ -199,7 +199,7 @@
     },
     {
         "name": "preview",
-        "pattern": "^/preview(/editor|(/\\d+(/editor|/fullscreen)?)?)?/?$",
+        "pattern": "^/preview(/editor|(/\\d+(/editor|/fullscreen)?)?)?/?(\\?.*)?$",
```
benjiwheeler: are we positive that querystring is part of the route, in both node AND fastly?

benjiwheeler: Chris pointed me to https://docs.fastly.com/guides/vcl-tutorials/vcl-regular-expression-cheat-sheet which is helpful!

PR #2327, opened 2018-11-19
Hotfix/hoc banner

chrisgarrity:


src/views/splash/presentation.jsx
```diff
@@ -487,14 +491,42 @@ class SplashPresentation extends React.Component { // eslint-disable-line react/
...
+                                            />
+                                        ] :
+                                        []
+                                }
+                                
                             </MediaQuery>
                         ]) : []
                     }
+                    {featured.shift()}
+                    {featured.shift()}
```
rschamp: In general we should avoid this, it's mutating a variable while it's rendering. But in this case it doesn't look like it matters. It's kind of confusing though, I'm not sure what it's doing.

chrisgarrity: @rschamp the design called for a 'full bleed' banner after the first two rows. Given how the rows are rendered, this renders the first two rows, closes the wrapper that limits the width, then renders the rest of the featured rows in a second wrapper after the full width banner.

I agree it's kind of ugly. If there would be a better way to break out of the width restriction, that would be great. Alternatively we may want to try to avoid full width middle banners. Middle banners that are the width of the wrapper are easy to insert in between featured rows.

PR #2336, opened 2018-11-19
Allow user to exit "single comment mode" and see all comments.

paulkaplan:
This visually appears like a "load more comments" button for now, but has the impact of unsetting the #comments-<id> hash in the url and resetting the comment state, showing all the comments (in paginated view)

To test locally: 
- load a project with a comment id in the url, locally with the seed data you could use `/preview/12#comments-10`
- click the "see all comments button"
- see that the comments are now loaded chronologically, and you get the top level comment form back

*Note: I changed comments-16 to comments-10 above; my local data didn't have 16* - @benjiwheeler


benjiwheeler:
LGTM. I mentioned the non-working back button in https://github.com/LLK/scratch-www/issues/2337

PR #2340, opened 2018-11-20
Show details about why a project is not public

paulkaplan:
Show more details about why a project is not public, like censored and trashed.

Follow up the project info request with a request to the visibility endpoint to find out if the project is trashed or censored. The project just not being published is handled by the existing code.

This PR generalizes the ShareBanner to a more generic "Banner" that is then filled with the relevant content.

/cc @colbygk this relies on an API endpoint:`/users/:username/projects/:id/visibility` that is expected to return `{censored: bool, deleted: bool, reshareable: bool, censorMessage: string}`

One thing we do need the endpoint to check is if the html is escaped and making sure it is unescaped correctly. right now it expects the html to be unescaped, like some of the other endpoints that return html (like formatted comments).


paulkaplan:
Ok cool, i'm going to mark this as "blocked" pending the endpoint, but i guess it could be merged without it, nothing bad would happen...

paulkaplan:
I'm going to go ahead and merge this even though it wont work right until the endpoint is in because it is blocking other work

Issue #2200, opened 2018-10-17
'Copy link' should work, and help user see that it works


benjiwheeler:
@kchadha writes:

> The ‘copy link’ button has no visual feedback that anything happened when you clicked on it... 

Does this need design?

kchadha:
I think it also has no functionality attached to it...? There was nothing copied to my clipboard when I clicked on it.

benjiwheeler:
Resolved by https://github.com/LLK/scratch-www/pull/2343

kathymakes:
We can do a tooltip here like we do with comment "copy link." 
<img width="240" alt="screen shot 2018-12-03 at 5 35 22 pm" src="https://user-images.githubusercontent.com/8203939/49405918-00d97480-f722-11e8-87d3-50dcad6ba71f.png">

fill: white
font color: UI blue
border: 1px UI blue at 50% opacity

rschamp:
This is non-trivial until we have a reusable tooltip component in www :/

PR #2343, opened 2018-11-21
Implement the copy link functionality

paulkaplan:
Implement the "copy project link" button functionality on the new project page. It was visible but not doing anything.

This does not implement any further design, like showing a confirmation popover, but just makes sure the button works. I used a very slim package to implement copy that does not include any fallbacks for IE, since we do not need to support that here.


benjiwheeler:
Resolves https://github.com/LLK/scratch-www/issues/2200

PR #2344, opened 2018-11-21
Show comment actions on hover and fix report translation

paulkaplan:
Resolves two minor issues with comments

1. The actions (delete/report/restore) should only show on hover, that just had gone unimplemented. /cc @kathymakes the hover target is the entire comment
2. There was an incorrect translation string for reporting comments, causing this 
image, fixed it

### DIFF
--- a/src/views/preview/comment/comment.jsx
+++ b/src/views/preview/comment/comment.jsx
@@ -158,7 +158,7 @@ class Comment extends React.Component {
                                             className="comment-report"
                                             onClick={this.handleReport}
                                         >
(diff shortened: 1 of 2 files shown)


benjiwheeler:
This looks good, and you can totally merge it as is. 
But I did notice that the mouseover area is the entire comment and its reply box below, not just the comment itself. In both these screenshots, the mouse is over the new comment below, but the top comment's Report appears:

image

screen shot 2018-11-22 at 12 01 49 am

paulkaplan:
@benjiwheeler that feels right to me

Issue #2240, opened 2018-10-25
display notification after sharing, including special text for first time sharing


benjiwheeler:
On current production site, first time sharing shows notification text: "Congratulations on sharing your project! Other people can now try it out, give comments, and remix it." Subsequent project shares then show "Your project is now shared."

We should replicate a similar experience in 3.0.

This presents a complication, though: currently, shared status is updated for the project via a simple project metadata PUT to an API endpoint. API doesn't do any complex querying to see if the share is the user's first share ever.

There are a few ways we could address this:
1. change the project info updating endpoint to do complex querying if the project is set to be shared
2. create a different, sharing-only endpoint that does the logic from # 1
3. Skip the complex logic entirely and instead use another property of the user as a basis for which message we show, such as whether or not they are a scratcher

I vote for # 3!

carljbowman:
@benjiwheeler - Thanks for logging this issue!

There is a lot more that we can do to create a better share experience! We have some addtional ideas for both the first share experience but also more generally improve the share flow to prompt adding a title and project information if that information is missing. 

This work will also intersect with a larger initiative of improving the overall early experiences (onboarding) with Scratch. There are a few things in 2.0, like New User (which needs a new name) becoming a Scratch, but it's rather obscured. Totally agree, we can make these first time experiences more celebratory and exciting! 

Likely before we get to this particular work in this issue, we have a lot to do on the very first touch point with first time users, the sign up flow.

DeleteThisAcount:
just gonna say this here


the new scratchers share message is so encouraging for some reason, i would love if it showed more often lol

kathymakes:
Thanks @carljbowman! Yeah, that sums up where we are with design for now - hoping for a holistically thoughtful experience for new scratchers! I would be OK with showing a "Congratulations" message either the first project or the first few projects until they become a scratcher. Can draw up

kathymakes:
For now (release), we can keep the same simple visual styling. @benjiwheeler 
![project page success alert](https://user-images.githubusercontent.com/8203939/47669267-e8a19300-db80-11e8-9c3c-12f756b3db6d.png)

PR #2346, opened 2018-11-25
Display green "Shared" banner after user shares; bring CSS into line with spec

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/2240

### Changes:

Displays green "Shared" banner on project page after user shares, which persists until user browses away or makes a change.

### Test Coverage:

none


benjiwheeler:
Thanks for the suggestions! Take a look.

benjiwheeler:
Here's what this looks like now:
screen shot 2018-11-28 at 6 30 27 pm

(Note that new scratchers will see a longer message than this; this is what Scratchers will see:)
screen shot 2018-11-28 at 6 30 38 pm

paulkaplan:
@benjiwheeler there are some lint errors

benjiwheeler:
Arrgh, ok, take a look now!

paulkaplan:
@benjiwheeler why were the CSS changes needed? Some of them smell a bit, like having fixed heights on buttons (what if the text wraps?). I thought the banner looked ok with a single line of text with no button, like the "Note: your project is in the trash" one?

src/views/preview/presentation.jsx
```diff
@@ -128,12 +130,28 @@ const PreviewPresentation = ({
...
         }
-    } else if (canShare && !isShared) {
-        banner = (<Banner
-            actionMessage={<FormattedMessage id="project.share.shareButton" />}
-            message={<FormattedMessage id="project.share.notShared" />}
-            onAction={onShare}
-        />);
+    } else if (canShare) {
+        if (isShared) {
+            if (justShared) {
```
paulkaplan: Can't you just add justShared to the isShared if statement (i.e. `if (isShared && justShared) { ...`) to reduce the nesting and allow you to put the helpful comment that is currently at the bottom of this block alongside that condition?

benjiwheeler: No, because the `else` for the unshared state (below) depends only on `isShared` and not on `justShared`. I did go back and forth on whether the way I have it now is better or worse than doing 

```
if (canShare && isShared && justShared) {
    if (isNewScratcher) {
        // ...
    } else {
        // ...
    }
} else if (canShare && !isShared) {
   // ...
}
```

...which is slightly more redundant, but maybe easier to read?

src/views/preview/project-view.jsx
```diff
@@ -665,6 +673,7 @@ const mapStateToProps = state => {
             state.permissions.admin === true),
         isLoggedIn: isLoggedIn,
         isAdmin: isAdmin,
+        isNewScratcher: isLoggedIn && state.permissions.new_scratcher,
```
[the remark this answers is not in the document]
benjiwheeler: Good point. Yeah, as helpful as these individual flags (and the `can`- flags) are, they can be a little misleading -- like how it's hard to know exactly what `canCreateNew` means, since in a sense, any user can create new at any time.

It could be better to give these more elaborate names, such as `isLoggedInNewScratcher`, but even for me (who prefers verbosity more than most!) that seems like overkill.

So I appreciate your noting this.

benjiwheeler:
The CSS changes are to bring the banner in line with the invision specs that I know of.

I changed that `height` to a `min-height` and added a few more mins and maxes.

PR #2367, opened 2018-11-29
Additional project data for Moderators/Admins on the project page

chrisgarrity:
### Changes:

Adds some project data that is useful for the moderation team. Only shows if the current user is an admin.

### Test Coverage:

Current tests run.

Manual Testing:
- [ ] When logged in as admin you can see Moderation Info section:
screen shot 2018-11-29 at 2 09 18 pm
- [ ] When not an admin, or not logged in, 'Moderation Info' is not visible

_Note: the number of scripts counted is the number of stacks, not just stacks with hat blocks. Mod team confirmed that this is fine._


src/views/preview/project-view.jsx
```diff
@@ -205,14 +209,29 @@ class Preview extends React.Component {
                                 }
                             });
                         }
+                        const sprites = projectData[0].targets.length - 1; // don't count stage
+                        const scripts = projectData[0].targets
+                            .map(target =>
+                                Object.values(target.blocks)
+                                    .filter(block => block.topLevel).length
```
chrisgarrity: I checked with the mod team and top-level blocks is fine with them even though it's a change from 2.0

PR #2375, opened 2018-12-03
[Master] Add project data to adventure game tutorial

rschamp:
/cc @chrisgarrity @christanbalch


rschamp:
👍 from @colbygk offline

PR #2393, opened 2018-12-04
project page will no longer vanish when project metadata update fails on the backend

benjiwheeler:
One of the weird behaviors identified in the issue https://github.com/LLK/scratch-api/issues/638 is that currently, if you edit project metadata on a censored project, the project disappears:

dec-04-2018 15-29-21

That's because we're only checking for status code == 500, and so we are acting in the code like status code === 404 is valid, and setting the project metadata to the response body (which is empty)

### DIFF
--- a/src/redux/preview.js
+++ b/src/redux/preview.js
@@ -754,9 +754,9 @@ module.exports.updateProject = (id, jsonData, username, token) => (dispatch => {
             dispatch(module.exports.setError('No project info'));
             return;
         }
-        if (res.statusCode === 500) { // InternalServer
+        if (res.statusCode >= 400) { // API responding with error
             dispatch(module.exports.setFetchStatus('project', module.exports.Status.ERROR));
-            dispatch(module.exports.setError('API Internal Server Error'));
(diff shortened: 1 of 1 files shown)


src/redux/preview.js
```diff
@@ -754,7 +754,7 @@ module.exports.updateProject = (id, jsonData, username, token) => (dispatch => {
             dispatch(module.exports.setError('No project info'));
             return;
         }
-        if (res.statusCode === 500) { // InternalServer
+        if (res.statusCode >= 400) { // InternalServer
```
rschamp: We should keep everything else up to date with this check. 500 was literally an Internal Server Error, so let's update this comment and the handler below to keep up to date with the new error codes being checked.  "API Error response" as the error that gets set, maybe?

rschamp:
@benjiwheeler Would you please try rewriting the title of this so it's easier to understand what it does?

src/redux/preview.js
benjiwheeler: You're right, I was too hasty!

benjiwheeler:
Take a look now

PR #2404, opened 2018-12-05
Admin panel for the project page

rschamp:
~~Just in case we don't do anything else,~~ _As was the plan_, this "proxies" a standalone admin panel onto the new project page in an iframe.

~~Requires https://github.com/LLK/scratchr2/pull/5219 to work.~~ (merged!)
Requires https://github.com/LLK/scratchr2/pull/5220 to work.

This re-arranges some styles in the existing admin panel on the homepage so it can be re-used on the project page.

Then, listen for events (added in the PR above) that tell us when a modal is opened or closed in the iframe'd page and expand the panel and iframe accordingly so we can see/hide the rest of the page to interact with the modal.

### DIFF
--- a/src/components/adminpanel/adminpanel.jsx
+++ b/src/components/adminpanel/adminpanel.jsx
@@ -1,9 +1,8 @@
 const bindAll = require('lodash.bindall');
+const classNames = require('classnames');
 const connect = require('react-redux').connect;
 const PropTypes = require('prop-types');
 const React = require('react');
-
-const Button = require('../forms/button.jsx');
 
 require('./adminpanel.scss');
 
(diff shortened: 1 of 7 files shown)


src/components/adminpanel/adminpanel.jsx
```diff
@@ -23,71 +22,45 @@ class AdminPanel extends React.Component {
...
+                            {this.props.children}
+                        </div>
+                    </React.Fragment>
+                ) : (
                     <span
                         className="toggle"
                         onClick={this.handleToggleVisibility}
                     >
-                        x
+                        &gt;
```
benjiwheeler: this makes for a pretty small click target. might want to use an enclosing circle or something, as a follow-on at some point.

src/components/adminpanel/adminpanel.jsx
```diff
@@ -23,71 +22,45 @@ class AdminPanel extends React.Component {
...
-                >
+        return (
+            <div
+                className={classNames(
+                    'admin-panel', this.props.className, {
+                        hidden: !this.state.showPanel
+                    }
+                )}
+            >
+                {this.state.showPanel ? (
```
benjiwheeler: I couldn't see the panel content when open, because scratchr2 gave a 404. I assume this is waiting for a change on scratchr2's side? 

Url: http://localhost:8080/scratch2/2000000157/adminpanel/

image

src/components/adminpanel/adminpanel.jsx
rschamp: Sure! This is just what already existed for the admin panel we have on the homepage.

PR #2406, opened 2018-12-05
show embed view

### LINKED ISSUE #2388: need /embed view that can be used to embed projects in an iframe
Many 2.0 projects are shared in iframes currently, all around the web in play-only mode.

We should make an endpoint that provides such a view.

Features:
* it should be as small and cacheable as possible
* redirects the current embed URL: `/projects/embed/:id`
* handles the URL `/projects/:id/embed`
* will use the 'greenflag overlay' getting implemented for project page
* Logo on the bottom, not too prominent, watermarked styling, links back to the splash page.
* remove or disable the fullscreen button
* green flag/stop buttons position should match gui

#### Enhancements that are not being considered for the initial version
* should the title be present?
* should there be a watermark/overlay that communicates something about what Scratch is?
* should the user's name and avatar be present?
* should there be some link back to the project's Scratch page?

chrisgarrity:
### Resolves:
Fixes #2388 

### Changes:
* detect `embed` in the URL
* initialize GUI in embed mode if detected
* match `embed` in the project page route
* add fastly config rules to redirect `/projects/embed/:id` to `/projects/:id/embed`

### Test Coverage:
Manual Testing:
- [ ] Embed view has the project loaded fullscreen, with the Scratch logo instead of fullscreen toggle

Post Deploy testing:
- [ ] `/project/embed...` urls get redirected to the new style

### DIFF
--- a/bin/configure-fastly.js
+++ b/bin/configure-fastly.js
@@ -217,6 +217,52 @@ async.auto({
             if (err) return cb(err);
             cb(null, redirectResults);
         });
+    }],
+    embedRedirectHeaders: ['version', function (cb, results) {
+        async.auto({
+            requestCondition: function (cb2) {
+                var condition = {
+                    name: 'routes/projects/embed (request)',
+                    statement: 'req.url ~ "^/projects/embed/(\\d+)"',
+                    type: 'REQUEST',
+                    priority: 10
+                };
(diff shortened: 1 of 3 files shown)


benjiwheeler:
Also, let's watch this closely and test it after deploy to confirm that the redirecting works

Issue #2328, opened 2018-11-19
Need loading state and result state for Remix


benjiwheeler:
As a follow-on to implementing the Remix button on the project page: https://github.com/LLK/scratch-www/issues/2283

... we should design:
1. what we want the user to see as the Remix is happening (e.g., button says "Remixing...")
2. what we want the user to see once the Remix is finished (e.g., banner) 

Related to https://github.com/LLK/scratch-www/issues/2339

kathymakes:
___
> **1. what we want the user to see as the Remix is happening (e.g., button says "Remixing...")**
___
We can include something in the Remix button that is dimmed out **like this**:
As a state change, this feels most concrete + obvious?
<img width="645" alt="screen shot 2018-11-19 at 2 54 47 pm" src="https://user-images.githubusercontent.com/8203939/48731568-58002500-ec0b-11e8-8531-05b20deb071a.png">

I put the button fill at 60% opacity.
<img width="1279" alt="screen shot 2018-11-19 at 3 01 20 pm" src="https://user-images.githubusercontent.com/8203939/48731887-19b73580-ec0c-11e8-9e48-ff0fe6058ae7.png">

This is my preference, though we could also think about animations (this is a little hacky but...):

**animates like this:**
![remixing](https://user-images.githubusercontent.com/8203939/48731523-3dc64700-ec0b-11e8-947f-c1ea79306c2f.gif)
(hacky animation - you get the idea here)


___
> **2. what we want the user to see once the Remix is finished (e.g., banner)**
___

What might be a good solution is to just show a green banner right after they've remixed...

![my project - remixed - green](https://user-images.githubusercontent.com/8203939/48739276-7709b180-ec21-11e8-8803-c6f6003636c7.png)

...and then on subsequent visits show the regular unshared banner.

![my project - share banner](https://user-images.githubusercontent.com/8203939/48739270-740ec100-ec21-11e8-9d04-6d23c47a2497.png)

@benjiwheeler We can continue to discuss and chat with mods as well!

kathymakes:
Currently (and I was unable to get an image of this) there is a pop up window after you remix for the first time which encourages you to add your own ideas. I wonder if we want to do something similar?

This is only from the GUI in 2.0.

There is also this interstitial state and alt text...
<img width="592" alt="screen shot 2018-11-19 at 5 40 21 pm" src="https://user-images.githubusercontent.com/8203939/48739507-365e6800-ec22-11e8-88aa-74c79519bbe5.png">

PR #2408, opened 2018-12-05
Remix button progress spinner & banner message showing that user has remixed

benjiwheeler:
### Resolves:

With https://github.com/LLK/scratch-gui/pull/3965 , resolves https://github.com/LLK/scratch-www/issues/2328

### Changes:

Adds new banner to project page that congratulates user on remixing project

show a green banner right after they've remixed:

image

...and then on subsequent visits show the regular unshared banner:

image

Also, show remixing progress state in the Remix button:

dec-05-2018 17-41-10

And show a tooltip:

image


benjiwheeler:
Made follow-on issue for counterclockwise spin: https://github.com/LLK/scratch-www/issues/2411

src/views/preview/presentation.jsx
```diff
@@ -136,6 +139,18 @@ const PreviewPresentation = ({
                 message={embedCensorMessage(visibilityInfo.message)}
             />);
         }
+    } else if (justRemixed && isNewScratcher) {
+        banner = (
```
chrisgarrity: Should the just remixed banner show for all users, or just new Scratchers? it looks like it'll only be newScratchers

src/views/preview/project-view.jsx
```diff
@@ -103,11 +106,20 @@ class Preview extends React.Component {
...
+            if (this.state.justShared) {
+                this.setState({ // eslint-disable-line react/no-did-update-set-state
+                    justShared: false
+                });
+            }
         }
         if (this.state.projectId === '0' && this.state.projectId !== prevState.projectId) {
             this.props.resetProject();
-            this.setState({justShared: false}); // eslint-disable-line react/no-did-update-set-state
+            if (this.state.justRemixed || this.state.justShared) {
```
chrisgarrity: is this `if` needed, or should it always reset these states (like the project)

benjiwheeler: Unfortunately, `setState` will cause a re-render even if the values you set it to are identical to the current values. So this just avoids an unnecessary re-render, but will always re-render if `justRemixed` or `justShared` were true.

see https://itnext.io/react-setstate-usage-and-gotchas-ac10b4e03d60
image

src/views/preview/presentation.jsx
benjiwheeler: You're right, checked with designers, fixed

Issue #1858, opened 2018-04-18
Handle view count for new project page


chrisgarrity:
Use green flag prop from LLK/scratch-gui#3968 to call end point in LLK/scratch-api#476

towerofnix:
> **If the new count is not over 100,** call the API endpoint to increment view count

Just to clarify, what does this imply? It sounds like it means that, under the new project page, projects would be limited to having 100 views, but I don't understand why that'd be intended.

thisandagain:
Assigning to @rschamp for triage / assignment to Ben (?)

joker314:
As a slightly secondary question, will there be checks made server-side to enforce the 100-view limit, or just client side?

colbygk:
Hi @joker314 and @towerofnix A guiding principle of Scratch is to encourage constructive learning and positive community interactions. Gaming views is a part of the system we have to spend considerable resources on avoiding (in both database transactions and informational storage), and yet, even if we had a perfectly non-game-able system, it makes for a competitive metric that we'd like to avoid leading to negative experiences within the wider community.

benjiwheeler:
Closed by the combination of four PRs, referenced above

PR #2417, opened 2018-12-06
pass onGreenFlag to GUI; when called, call API to register project view

benjiwheeler:
### Resolves:

Part of resolving https://github.com/LLK/scratch-www/issues/1858

relies on https://github.com/LLK/scratch-gui/pull/3981 and https://github.com/LLK/scratch-api/pull/650

### Changes:

Passes a function callback to GUI, for it to call when green flag clicked. When handling this callback, post to the API endpoint responsible for registering project views.

### DIFF
--- a/src/redux/preview.js
+++ b/src/redux/preview.js
@@ -884,3 +884,17 @@ module.exports.updateProjectThumbnail = (id, blob) => (dispatch => {
         dispatch(module.exports.setFetchStatus('project-thumbnail', module.exports.Status.FETCHED));
     });
 });
+
+module.exports.logProjectView = (id, username) => (dispatch => {
+    dispatch(module.exports.setFetchStatus('project-log-view', module.exports.Status.FETCHING));
+    api({
+        uri: `/users/${username}/projects/${id}/views`,
+        method: 'POST'
+    }, (err, body, res) => {
+        if (err || res.statusCode !== 200) {
+            dispatch(module.exports.setFetchStatus('project-log-view', module.exports.Status.ERROR));
+            return;
+        }
+        dispatch(module.exports.setFetchStatus('project-log-view', module.exports.Status.FETCHED));
+    });
+});
--- a/src/redux/preview.js
+++ b/src/redux/preview.js
@@ -885,11 +885,14 @@ module.exports.updateProjectThumbnail = (id, blob) => (dispatch => {
     });
 });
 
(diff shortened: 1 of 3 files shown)


src/redux/preview.js
```diff
@@ -884,3 +884,17 @@ module.exports.updateProjectThumbnail = (id, blob) => (dispatch => {
         dispatch(module.exports.setFetchStatus('project-thumbnail', module.exports.Status.FETCHED));
     });
 });
+
+module.exports.logProjectView = (id, username) => (dispatch => {
+    dispatch(module.exports.setFetchStatus('project-log-view', module.exports.Status.FETCHING));
+    api({
+        uri: `/users/${username}/projects/${id}/views`,
+        method: 'POST'
```
towerofnix: I noticed this doesn't take an authentication token.. hypothetically, could I `post` the endpoint above with many different usernames but the same project ID, to artificially increase the project's number of views? It seems like this should require an authentication token. (Actually, it sounds like it really ought to be at the endpoint `/projects/${id}/views`, since you should never be able to post that anybody *else* has viewed a project (only yourself), and the server can detect who is sending the request by viewing the user associated with the authentication token.)

benjiwheeler: The naming is a bit confusing -- this is (as I understand it) only used for recording that this particular user viewed this particular project.

We still might want to require an auth token, though -- I think you might be right that that would be better.

Re: the endpoint location, we're moving towards endpoint paths that include the current user's username, even though that's redundant (if you're sending an auth token), because that can help us avoid unnecessary database lookups.

rschamp: Good catch @towerofnix. Without authentication, the view will not be recorded appropriately. The username in this URL should be the username of the owner of the project. The URL scheme `/users/{owner}/project/{project id}` is a convention that makes it easier to do certain types of validation on the API.

@benjiwheeler The username looks like it's coming from the currently logged in user, but it should be the owner of the project. And I think we need to add `{withCredentials: true}` for this to function properly. I don't think the endpoint will have given any errors about these things though since it's designed to not return errors.

benjiwheeler:
This brings up a larger point, which is how do I test that this is working? I saw the request go through to scratchr2, but (as you point out) it doesn't give info on the result... which is in redis only, right? I guess I could put in a log message with the redis key, and attach to the redis container and inspect it?

benjiwheeler:
@rschamp So, should we change the API endpoint to require auth?

PR #2424, opened 2018-12-07
Allow admins to toggle comments

paulkaplan:
Fixes an issue where admins could not disable comments.

### DIFF
--- a/src/views/preview/presentation.jsx
+++ b/src/views/preview/presentation.jsx
@@ -57,6 +57,7 @@ const PreviewPresentation = ({
     canRestoreComments,
     canSave,
     canShare,
+    canToggleComments,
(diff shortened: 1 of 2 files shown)


src/views/preview/presentation.jsx
```diff
@@ -489,7 +490,7 @@ const PreviewPresentation = ({
                                 <div className="comments-container">
                                     <FlexRow className="comments-header">
                                         <h4><FormattedMessage id="project.comments.header" /></h4>
-                                        {userOwnsProject ? (
+                                        {canToggleComments ? (
```
rschamp: There's a lint error above because this was the only place in the file `userOwnsProject` was used.

PR #2427, opened 2018-12-07
Add moderator info for sb2 projects

chrisgarrity:
### Resolves:

smoketest issue

### Changes:
Handle moderator information for sb2 projects

### Test Coverage:
Manual testing:
- sb3 projects show correct number of sprites and # of top blocks
- sb2 projects show correct number of sprites and hat blocks
- sb1 projects show zero for both.

### DIFF
--- a/src/views/preview/project-view.jsx
+++ b/src/views/preview/project-view.jsx
@@ -97,8 +97,8 @@ class Preview extends React.Component {
             justShared: false,
             loveCount: 0,
             modInfo: {
-                scripts: 0,
-                sprites: 0
+                scriptCount: 0,
+                spriteCount: 0
             },
             projectId: parts[1] === 'editor' ? '0' : parts[1],
             reportOpen: false,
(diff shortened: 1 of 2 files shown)


src/views/preview/project-view.jsx
```diff
@@ -243,13 +243,23 @@ class Preview extends React.Component {
...
                             });
                         }
-                        const sprites = projectData[0].targets.length - 1; // don't count stage
-                        const scripts = projectData[0].targets
-                            .map(target =>
-                                Object.values(target.blocks)
-                                    .filter(block => block.topLevel).length
-                            )
-                            .reduce((accumulator, currentVal) => accumulator + currentVal, 0);
+                        let sprites = 0;
```
benjiwheeler: can we rename these to something like `numSprites` and `numScripts`, even just in this block? It would help make their purpose clearer, even if we leave their names under `modInfo` the same.

src/views/preview/project-view.jsx
```diff
@@ -243,13 +243,23 @@ class Preview extends React.Component {
...
+                        let scripts = 0;
+                        if (projectData[0].projectVersion) {
+                            if (projectData[0].projectVersion === 3) {
+                                sprites = projectData[0].targets.length - 1; // don't count stage
+                                scripts = projectData[0].targets
```
benjiwheeler: For future reference, I tested this logic with:
```
[
{blocks: [{topLevel: true},{topLevel: false}]},
{blocks: [{topLevel: false},{topLevel: false}]},
{blocks: [{topLevel: true},{topLevel: true}]},
{blocks: [{topLevel: true},{topLevel: true}]},

].map(target =>
                                        Object.values(target.blocks)
                                            .filter(block => block.topLevel).length
                                    )
                                    .reduce((accumulator, currentVal) => accumulator + currentVal, 0);
```

it worked!

src/views/preview/project-view.jsx
chrisgarrity: renamed everything to `...Count`

Issue #2072, opened 2018-09-07
Preview page icons not updating correctly


benjiwheeler:
Several reports in the 9/7/18 smoke test:

"The heart icon doesn’t change color when I heart a project. I can heart a project many times (the counter goes up at least). FF on a mac"

"Clicking on the “love it” button causes the counter to increment, but does not highlight the “heart” icon and will increment forever"

"Favoriting and reloading does not save the UI state that you have favorited a project. The star does not stay yellow to indicate that you have favorited the project. Instead it allows you to re-favorite the project rather than toggle off your favorite."

benjiwheeler:
Seems like there has been some progress:
* I can no longer heart a project more than once / counter only goes up by 1, then acts as a toggle
* heart icon does highlight

However, several of these problems remain:
* state (whether i have hearted and/or starred project) is not maintained across page loads (is that an issue with what's available to the backend in the db?)
* the heart counter number does not stay the same across page loads
* the star number does persist, but allows a second favoriting to appear to occur after page reload, allowing you to appear to increment by 2 (though the second increment does not stick)

benjiwheeler:
11/30/18 Bug hunt turned up further repro's of these problems:

@BryceLTaylor: "I can star a project and it still has 0 stars"
@paulkaplan: "Still can get more than 1 heart by clicking quickly"

benjiwheeler:
There’s 2 ways I’ve found to solve this, and they both have (different) pros and cons from a UI perspective.

#1: the “client solution”: all clicks have an immediate, correct-looking response on the client side. You can click away to your heart’s content, toggling the state in a way that seems correct.
drawback: since we’re sending these clicks to the server, if you click multiple times fast, sometimes the server responds with an understanding that’s out of step with the client, and suddenly overrides local state.

![clientway](https://user-images.githubusercontent.com/3431616/49686821-56b66f80-fac8-11e8-893f-5edf51432d6e.gif)



#2: the “server solution”: on the client side, we only respond to user’s clicks when the client and server already agree about the state. So there are no sudden surprises when server responds. However, this also means that until server responds, any clicks after the first one seem to have no effect.

![serverway](https://user-images.githubusercontent.com/3431616/49686822-5ddd7d80-fac8-11e8-9e25-9eb0622f196d.gif)

Note that in both solutions, these side effects are only noticeable if you click fast and have a slow connection. On a fast connection, you really have to work to see them. I think both types of solution are used by other companies, in different circumstances.

PR #2428, opened 2018-12-08
fix love and favorite misbehavior

benjiwheeler:
fix love and favorite misbehavior by recording initial state, and referencing initial count

### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/2072

### Changes:

Adds `initialLoved` and `initialFaved` key-values to the preview reducer state, so that we can compare client state of those toggles to the original values, and set the resulting client-side love and favorite counts accordingly.

### Test Coverage:

None. Man, test debt


paulkaplan:
@benjiwheeler i'm confused about what this does now (apart from the user-select part)? What does tracking clientFaved give us?

benjiwheeler:
I clarified in the definition. This PR:
* allows icons to update immediately, matching count numbers
* disables selection of love/favorite numbers: prevents awkward highlighting

PR #2430, opened 2018-12-10
Use falsey check for shareDate instead of checking for null.

paulkaplan:
It is set to empty string elsewhere when the value does not exist, not null.


paulkaplan:
@benjiwheeler that is a good point. In this case I don't think that is possible, because the whole page only shows when projectInfo has loaded, and the share date comes with that request all at once.

Issue #2330, opened 2018-11-19
Update hardware extension landing page for launch


ericrosenbaum:
Currently each of the hardware extension landing pages (microbit, EV3, WeDo) has a "getting started" section that includes a step that says "Use the Scratch 3.0 editor," with a link to the beta:

![screen shot 2018-11-19 at 2 00 10 pm](https://user-images.githubusercontent.com/567844/48728651-84b03e80-ec03-11e8-9478-7642144738b6.png)

For launch, we should change this link, and consider: 
 - What it should link to (probably a link that opens the editor directly)
- How to update the wording (perhaps "Use the Scratch editor")

PR #2443, opened 2018-12-11
Update extension landing pages

ericrosenbaum:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/2330

### Changes:

- Getting Started instructions link no longer points to beta
- Getting Started instructions link to editor opens the tutorial for EV3 and WeDo

- Update starter projects for EV3 and WeDo, with new links, titles, descriptions, and thumbnails.
- Starter project links for EV3, WeDo and microbit no longer point to beta
- Starter project links for EV3 and Wedo also open the tutorial

Note that the starter project links will not work on staging, because they use production project ids.


src/views/ev3/l10n.json
```diff
@@ -18,12 +18,12 @@
...
-    "ev3.distanceInstrumentTitle": "Distance Instrument",
-    "ev3.distanceInstrumentDescription": "Move your body in front of the sensor to make music.",
-    "ev3.spaceTacosTitle": "Space Tacos",
-    "ev3.spaceTacosDescription": "Build your own controller to catch tacos in space.",
+    "ev3.starter1Title": "Play Basketball",
+    "ev3.starter1Description": "Move in front of the distance sensor to bounce the ball.",
+    "ev3.starter2Title": "Make Music",
+    "ev3.starter2Description": "Press the buttons to play saxophone and drums.",
+    "ev3.starter3Title": "Space Tacos",
+    "ev3.starter3Description": "Build your own controller to catch tacos in space.",
```
rschamp: Wondering why these keys are changing — I think it's better for the ID to be specific to the content so that we don't re-use the id unnecessarily. E.g. if we change what starter1 is just by updating the translation string, it's possible for the English string to say something else, but old translations will still say the previous content until the new translation is done and reviewed. In that case it would be better for us to display English but with the correct content.

src/views/ev3/l10n.json
```diff
@@ -47,7 +47,7 @@
     "ev3.imgAltWaitForWindows": "Windows will notify you when the EV3 is ready.",
     "ev3.imgAltEnterPasscodeMac": "Enter the passcode into the connection request window opening on your Mac.",
     "ev3.imgAltPlugInMotor": "To find port A: hold the EV3 with the screen and buttons facing you, with the screen above the buttons. Port A is on top, and it is the left-most one",
-    "ev3.imgAltWaveHello": "A Scratch project with a waving fairy.",
-    "ev3.imgAltDistanceInstrument": "A Scratch project with a guitar.",
-    "ev3.imgAltSpaceTacos": "A Scratch project with Scratch Cat and a taco in space."
+    "ev3.imgAltStarter1": "A Scratch project with a basketball.",
+    "ev3.imgAltStarter2": "A Scratch project with musical instruments.",
+    "ev3.imgAltStarter3": "A Scratch project with Scratch Cat and a taco in space."
```
rschamp: Same here re: the ids.

src/views/ev3/l10n.json
ericrosenbaum: good point- updating the ids!

Issue #2444, opened 2018-12-11
Remixes list links to /preview


paulkaplan:
Going to any project on the new project page that has remixes, see that the thumbnail column of remixes all link to `/preview/:id`. seems to be related to this code: https://github.com/LLK/scratch-www/blob/develop/src/components/thumbnailcolumn/thumbnailcolumn.jsx#L13-L15

PR #2445, opened 2018-12-11
Fix hard-coded /preview/id url for remix list.

paulkaplan:
### Resolves:

Fixes https://github.com/LLK/scratch-www/issues/2444

### Changes:

Makes the itemType `projects` so that the URLs are correct

### DIFF
--- a/src/components/thumbnailcolumn/thumbnailcolumn.jsx
+++ b/src/components/thumbnailcolumn/thumbnailcolumn.jsx
@@ -12,7 +12,7 @@ const ThumbnailColumn = props => (
     <FlexRow className={classNames('thumbnail-column', props.className)}>
         {props.items.map((item, key) => {
             const href = `/${props.itemType}/${item.id}/`;
-            if (props.itemType === 'preview') {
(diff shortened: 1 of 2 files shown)


src/components/thumbnailcolumn/thumbnailcolumn.jsx
```diff
@@ -12,7 +12,7 @@ const ThumbnailColumn = props => (
     <FlexRow className={classNames('thumbnail-column', props.className)}>
         {props.items.map((item, key) => {
             const href = `/${props.itemType}/${item.id}/`;
-            if (props.itemType === 'preview') {
+            if (props.itemType === 'projects') {
```
chrisgarrity: very small nit, but this might be better as 'project' rather than 'projects' (the studio-list version is singular 'studio'

paulkaplan: Oh it needs to be projects for the URLs to work, since it goes to `/itemType/id`, we also use itemType: 'projects' in other places.

Issue #2354, opened 2018-11-28
Update project page metadata with information loaded via the API


rschamp:
The Scratch 2 project page has this metadata, which we should think about keeping:
```
<title>{ projectTitle } on Scratch</title>
<meta name="description" content="{ projectTitle } on Scratch by { username }" />
<meta property="og:title" content="Scratch - { projectTitle }"/>
<meta property="og:description" content="{ projectInstructions|truncatewords:50 }" />
```

paulkaplan:
Not going to close this because the new meta tags are in but will not be shown until we use a meta-tagless template, which i think is going to be done in the sentry change @rschamp ?

rschamp:
Correct

PR #2448, opened 2018-12-12
Use custom meta tag info for project page.

paulkaplan:
Note this _will not_ work until the page uses a template that does not include these meta tags. That is because react-helmet does not override any existing info in the head.

Thus, this was tested by removing that info from the template and seeing it work, then putting the template back.

This will be part of https://github.com/LLK/scratch-www/issues/2354, but again, it depends on being rendered into a template that does not already use those metatags /cc @rschamp


rschamp:
Oh except for the conflict.

Issue #2299, opened 2018-11-09
UX: Downloads Page


kathymakes:
### Overview
We are redesigning the [Downloads page](https://scratch.mit.edu/download). At a high level, we want to follow the patterns we use for the Extensions pages (example: [Microbit](https://scratch.mit.edu/microbit)) and cut down the content.

Here are more notes and info: [Google Doc](https://docs.google.com/document/d/1otHd3j_Dpa9_6Y_nVCxP5b1DN1yCZWTSntAdI5I_quQ/edit)

### Design
Still a work in progress. Logging to track this work.

![1280px](https://user-images.githubusercontent.com/8203939/48578031-a4cfbd00-e8e6-11e8-8ab3-5b82fa6ba93c.png)

### Still need

- [ ] Finalized copy for the page
- [ ] Screenshots, if we decide to include

### Responsive

Most of our pages that follow the extensions page formula aren't totally optimized for stretching smaller. But here is one potential framework to follow.

This is probably a Phase 2, or lower priority item (especially because this page isn't very useful to folks on phones or tablets.)

Still, would be nice to bake responsive into our styles in the process of building it out, and show it a little TLC!

<img width="1355" alt="responsive" src="https://user-images.githubusercontent.com/8203939/48578087-b913ba00-e8e6-11e8-84bd-f4ad0dc3a2b1.png">

kyleplo:
Maybe mobile screen sizes shouldn't show the download buttons?

benjiwheeler:
We could show something else -- like "Please view this page on a Windows or
Mac computer to download."

What logic are we correctly using to determine which OS to highlight? What
does it currently show for, say, Chromebooks?

kathymakes:
Yeah - good point @kyleplo ! It would be much better to tell them up front than to redirect to the app store (and then not be able to do anything.) 

I think for Chromebook, maybe they could both be dimmed out?

MasterOfTheTiger:
Wait. I do not see a GNU/Linux option. Electron supports it right?

thisandagain:
@masterofthetiger Linux will not be available at launch, but we are working with some partners of Linux (Raspberry Pi in particular) support for later in 2019.

kathymakes:
### A few changes...

The current illustration of GUI might communicate, "Here is the scratch 3.0 editor," but if we wanted to emphasize that this is a downloadable editor, we can alternatively try this illustration:

[...]

PR #2449, opened 2018-12-12
New Download page

picklesrus:
### Resolves:

Much of the work toward  #2299 but not quite all.
TODOs:
- Replace image next to title
- Replace screenshots
- Text and images for install section isn't final yet.

### Changes:

Moves the current download page into downloads/scratch2.
Replaces the existing one with the version for scratch 3, assuming we don't have apps ready in the app store at the very beginning. See #2299.


src/views/download/download.scss
```diff
@@ -6,123 +6,171 @@
...
+    .title {
+        margin-bottom: 1rem;
+        font-size: 2rem;
+    }
+
+    .centered {
+        display: flex;
+    }
+
+    .self-center {
```
benjiwheeler: instead of naming this class after what it does, I suggest naming it after the type of component/tag that uses it. In this case, you might call it 'icon'?

src/views/download/download.scss
```diff
@@ -6,123 +6,171 @@
...
-        &.masthead {
-            margin-bottom: 0;
-            background-color: $ui-blue-dark;
-            padding-bottom: 0;
+    .title {
+        margin-bottom: 1rem;
+        font-size: 2rem;
+    }
+
+    .centered {
```
benjiwheeler: instead of naming this class after what it does, since it's only used in a few places, I suggest naming it after the type of component/tag that uses it. In this case, you might call it 'legacy-link'?

src/views/download/download.scss
picklesrus: Done! Actually, there already was an icon for exactly this... so I just moved it there and removed this class altogether :)

src/views/download/download.scss
picklesrus: yes! Done!

Issue #2421, opened 2018-12-07
Educator's Page Update


carljbowman:
### Overview
Modify the content of the page by using preexisting components. A larger rethink and redesign of the page will come in the future after the launch of 3.0.

### Mock
[On InVision](https://projects.invisionapp.com/d/main#/console/16172669/335334873/preview)

![for educators](https://user-images.githubusercontent.com/3409578/49736282-b000d900-fc57-11e8-85bf-48a71bdb141b.png)

### Assets & Content
Icons: [educator-page-assets-v2.zip](https://github.com/LLK/scratch-www/files/2669361/educator-page-assets-v2.zip)

Text: [Google Doc](https://docs.google.com/document/d/1-zjRLysSDJ6LEKawIlom5twuV09YUFOjiftxX0gaDuA/edit)

/cc @ChampikaF @ntlrsk @thisandagain

apple502j:
May I work on it?

ntlrsk:
Looking great! Two minor edits if possible when implementing the text: 1) In the "Connecting with Educators" section, for the first sentence make the link on "**Teaching with Scratch**" (instead of "Join the Teaching")  2) Under **News and Updates**, for the second sentence, can drop "for educators"  so it will just read: "Sign up to receive **updates and tips** from the Scratch Team."

kyleplo:
Also, when do you plan on updating the "my classes" and related pages to scratch-www

picklesrus:
@apple502j I think it is probably best if I keep this one. If this wasn't on the critical path for 3.0 and we weren't so close to the deadline this would have been a great one for you to work on though!  Thank you for the offer!

picklesrus:
Also, see comments about the teacher page on https://github.com/LLK/scratch-www/pull/2447

picklesrus:
Things left todo before calling this fixed
- [ ] Finalize link to cards
- [ ] Finalize link to guides

PR #2451, opened 2018-12-12
minor educator pages fixes

picklesrus:
### Resolves:

The spacing parts of #2421.


src/views/teachers/landing/landing.scss
```diff
@@ -310,6 +311,7 @@ $story-width: $cols3;
     .masthead-info {
         p {
             width: $cols4;
+            text-align: center;
```
benjiwheeler: Is this left over from something else?

picklesrus: Yep. Removed!

Issue #2362, opened 2018-11-28
Update Tips to Ideas page


carljbowman:
### Overview
When we launch Scratch 3.0 we will need to update the Tips page. All the current content for 2.0 will not longer be relevant. At a high-level this update will include replacing 2.0 content with 3.0 content, some UX changes, and reframing the page as "Ideas."

### ToDo
- [x] Get remaining 3 tutorials and assets up to GUI (KO & Eric)
  - [x] KO to handoff JSON and assets to Eric for the following tutorials (KO)
    - [x] Animate a Character
    - [x] Make a Catch Game
    - [x] Video Sensing
  - [x] Eric to create a PR and merge (Eric)
- [x] Create new URL “/ideas” (Eric)
- [x] Create redirect for “/tips” and “/go” to “/ideas” (Eric)
- [x] Change text/URL in navbar and footer from “Tips” to “Ideas” (Eric)
  - [x]  On www
  - [x]  On scratchr2
- [x] Update “Tips page” language on Educators page (Eric)
- [x] Implement new layout “Ideas” page (Eric)
- [x] Finish masthead illustration and share with Eric (Carl)
- [x] Finish starter projects illustration and share with Eric (Carl)
- [x] Create and share Tutorial modal images with Eric (KO)
- [x] Share Tutorial tile images with Eric (KO)
- [x] Upload Activity Card PDFs and Educator Guides PDFs to AWS and share with Eric (Natalie)

#### Page
![v3 ideas page](https://user-images.githubusercontent.com/3409578/49652119-21921a80-f9ff-11e8-930d-f1c69282e8cd.png)

#### Activity Modal
![v3_ideas page modal](https://user-images.githubusercontent.com/3409578/49311076-17779580-f4ae-11e8-9565-d8d3e16daa76.png)

### Content

**Copy** — [Google Doc](https://docs.google.com/document/d/1k_Xkbi0Blt-8oN17QzvB0xlTe7VfjWL_6a_V9pY02JU/edit)

**Illustrations & Icons** — [ideas-page-assets-v3.zip](https://github.com/LLK/scratch-www/files/2663772/ideas-page-assets-v3.zip)

**Tutorial Tile Images** — @kosiecki17 to provide

**Tutorial Modal Images** — @kosiecki17 to provide

ntlrsk:
@ericrosenbaum @carljbowman  Note: wording change: Activity Cards --> Coding Cards

carljbowman:
[...]

PR #2455, opened 2018-12-12
Ideas page

ericrosenbaum:
### Resolves:

Progress toward https://github.com/LLK/scratch-www/issues/2362

### Changes:

- Update the tips page and rename it to "ideas".


src/views/ideas/l10n.json
```diff
@@ -0,0 +1,69 @@
...
+    "ttt.HideAndSeekDescription": "Make a hide-and-seek game with characters that appear and disappear.",
+    "ttt.FashionTitle": "Fashion Game",
+    "ttt.FashionDescription": "Make a game where you dress a character with different clothes and styles.",
+    "ttt.PongTitle": "Pong Game",
+    "ttt.PongDescription": "Make a bouncing ball game with sounds, points, and other effects.",
+    "ttt.DanceTitle": "Let's Dance",
+    "ttt.DanceDescription": "Design an animated dance scene with music and dance moves.",
+    "ttt.CatchTitle": "Catch Game",
+    "ttt.CatchDescription": "Make a game where you catch things falling from the sky.",
+    "ttt.VirtualPetTitle": "Virtual Pet",
```
chrisgarrity: These all stick around as things translators have to translate. It's not a big deal for the folks that have already done the tips page, but for new languages (e.g., igbo), it will be extra work they don't really need to do.  We should remove any that we know we're not going to use. It would be ok to keep the ones that will have a new tutorial eventually.

src/views/ideas/l10n-static.json
```diff
@@ -0,0 +1,15 @@
+{
+    "cards.ScratchCardsAllLink": "https://resources.scratch.mit.edu/www/cards/en/scratch-cards-all.pdf",
+    "cards.nameCardsLink": "https://resources.scratch.mit.edu/www/cards/en/name-cards.pdf",
+    "cards.animateCardsLink": "https://resources.scratch.mit.edu/www/cards/en/animate-cards.pdf",
+    "cards.musicCardsLink": "https://resources.scratch.mit.edu/www/cards/en/music-cards.pdf",
+    "cards.storyCardsLink": "https://resources.scratch.mit.edu/www/cards/en/story-cards.pdf",
+    "cards.chaseCardsLink": "https://resources.scratch.mit.edu/www/cards/en/chase-cards.pdf",
+    "cards.videoCardsLink": "https://resources.scratch.mit.edu/www/cards/en/video-cards.pdf",
+    "guides.nameGuideLink": "https://resources.scratch.mit.edu/www/guides/en/NameGuide.pdf",
```
chrisgarrity: So, unfortunately the cards and the guides did not use the same naming conventions on the resources server, so the script that generates the names (based on filename) uses uppercase for guides: `guides.NameGuideLink`. We should probably fix this. I'm going to have to revise the script that generates the data for the cards in any case. I'll look into whether I can standardize the Guides link.

chrisgarrity: On second thoughts, let's make them work with the script as is. So the ones that exist already need to be capitalized:
- `guides.NameGuideLink`
- `guides.MusicGuideLink`
- `guides.StoryGuideLink`

The script that generates the list simply takes the filename without extension, and appends `Link`, so we need to make sure we coordinate with Natalie on naming the ones that are new.

src/views/ideas/ideas.jsx
```diff
@@ -0,0 +1,252 @@
...
+            currentTile: tile,
+            TTTModalOpen: true
+        });
+    }
+    handleHideTTTModal () {
+        this.setState({
+            TTTModalOpen: false
+        });
+    }
+    renderTTTTiles () {
```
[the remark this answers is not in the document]
ericrosenbaum: all set

src/views/ideas/l10n.json
ericrosenbaum: good point, removed most of the unused ones

src/components/modal/ttt/modal.scss
```diff
@@ -1,97 +1,62 @@
...
 }
 
-.button.white.mod-ttt-item {
-    border: 1px solid $ui-blue;
-    box-shadow: none;
+.ttt-description {
+    margin-top: 0;
 }
 
-@media #{$small} {
```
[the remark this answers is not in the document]
ericrosenbaum: I'm going to file a follow-up issue for fixing responsive stuff on this page

PR #2469, opened 2018-12-13
Hide copyright

### LINKED ISSUE #2446: Should we show "(copyleft) Unshared" in copyright date field for unshared projects?
Seems to me that this looks pretty weird:

screen shot 2018-12-11 at 4 18 19 pm

apple502j:
### Resolves:

Resolves #2446 

### Changes:
Check `isShared` to see if it should appear.


benjiwheeler:
I think we should remove "Unshared" as well -- is there any deliberate design that calls for this? It was part of what I pointed out in the original issue

PR #2480, opened 2018-12-13
Make list of languages on www consistent with gui

### LINKED ISSUE #2395: Make list of languages consistent across platforms
Scratchr2 and www have a different list of supported languages from Scratch 3.0. Right now the editor (gui) will ignore any locale that it doesn't have translated and just use English. When should languages that are not supported in 3.0 get removed from the menus on www and scratchr2?

Some of the languages were started many years ago and were never maintained, such as Asturian where even the editor is less than half translated.

I think the languages fall into three buckets:
* languages that have < 50% of the 2.0 editor and a minimal amount of the website - these were probably in 1.4 and have not been maintained since then
* languages that have > 50% of the 2.0 editor and a minimal amount of the website - probably people who mainly use Scratch 2.0 offline
* languages that meet our minimums for 2.0, but that are < 90% reviewed on 3.0

I would recommend that we drop languages in the first bucket, keep languages in the third assuming that people are still translating, and we should discuss the second.

chrisgarrity:
### Resolves:

Fixes #2395 

### Changes:
start to depend on `scratch-l10n`

So far only the supported locales is used.

There is one small difference between scratch-l10n and the `languages.json` that was removed.
languages was:
`{locale: ‘languagename’, …}` key value pairs.

l10n exports:
`{locale: {name: ‘languagename’},…}

That required an additional change to the language chooser.

### Test Coverage:

manual testing..
@rschamp I assigned this to you because it included fastly config changes.


rschamp:
For Fastly, this looks good so long as the `languages` imported from `l10n` have keys that are their language codes, like the json that was being imported before.

PR #2482, opened 2018-12-14
Update tests

ericrosenbaum:
Update the integration tests to reflect the change from the Tips page to the Ideas page.

### DIFF
--- a/test/integration/smoke-testing/test_footer_links.js
+++ b/test/integration/smoke-testing/test_footer_links.js
@@ -162,10 +162,10 @@ tap.test('clickStatisticsLink', options, t => {
 
 // ==== SUPPORT column ====
 
-// TIPS PAGE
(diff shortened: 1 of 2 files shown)


paulkaplan:
@ericrosenbaum thanks for doing this, I was about to file this. One thing I noticed is that /tips does not redirect to /ideas, which feels pretty important. Do you know if that is filed?

ericrosenbaum:
@paulkaplan @chrisgarrity and I updated the routes (e.g. https://github.com/LLK/scratch-www/blob/develop/src/routes.json#L393 is I think the specific thing you are asking about). That should redirect /tips to /ideas, right? 

I'm not sure I understand the whole system, but maybe that change is not on staging (even though it's merged), because the smoke tests failed, so it didn't deploy- they failed because the test I'm fixing here failed.

PR #2497, opened 2018-12-17
Add flag for dynamic meta tags in the template

paulkaplan:
Resolves https://github.com/LLK/scratch-www/issues/2354 by allowing routes to declare `dynamicMetaTags: true`, in which case we do not render the default meta information, allowing the dynamic ones to take over.

@rschamp I know we mentioned using a separate template, but I didn't want to duplicate some of the things that are shared, like favicon, google site verification, etc, so I did it as a flag.


paulkaplan:
@rschamp thumbs up

PR #2498, opened 2018-12-17
Fix some spacing issues on the download page. Adds a more padding und…

picklesrus:
…er heading titles and makes the text in the older versions section have normal text styling.

See screenshots below for comparison of old and new. The one on top is the new one, the second is the old one.

screen shot 2018-12-17 at 3 06 40 pm
screen shot 2018-12-17 at 3 06 12 pm
screen shot 2018-12-17 at 3 05 54 pm
screen shot 2018-12-17 at 3 05 46 pm

### DIFF
--- a/src/views/download/download.jsx
+++ b/src/views/download/download.jsx
@@ -191,9 +191,9 @@ class Download extends React.Component {
                         <h2 className="title">
                             <FormattedMessage id="download.olderVersionsTitle" />
                         </h2>
-                        <h3 className="faq-question">
+                        <p>
                             <FormattedMessage id="download.olderVersions" />
-                        </h3>
+                        </p>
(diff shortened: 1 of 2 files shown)


picklesrus:
The screenshot you have looks right to me. The  "Looking for ..." text is normal instead of a header/bold/etc and there's more space between the "Older Versions" line and the looking for text. Sorry, maybe I put my screenshot in the wrong order?

benjiwheeler:
OOHhhh, I was thinking of them as "before" and then "after"

FWIW the spacing looks off on the new version... seems like much more margin between "Older Versions" and "Looking for...?" than between "Install Scratch Desktop" and "Download Scratch Desktop"

Issue #2484, opened 2018-12-14
A remixed project without changes has the wrong thumbnail


BryceLTaylor:
If you remix a project and then immediately, without making changes, leave the page, the project will have just the cat on a white background for the thumbnail.

### Steps to Reproduce

Sign in
Navigate to a project you don't own
See inside
Click "Remix" at the top of the page
Note: the project gets a new number and the title says "remix" in it
Click the "my stuff" folder icon in the top bar
Note: on the my stuff page the thumbnail for the project is just the cat on a white backdrop.
If you ever save the project it gets the correct thumbnail.

### Operating System and Browser

Mac Chrome

benjiwheeler:
@thisandagain can you help me understand why this issue is critical?

thisandagain:
@benjiwheeler As we've seen in the past when thumbnails are not properly working in the "My Stuff" page, users can become highly disturbed and worry that their work did not properly save which results in a flood of bug reports and help desk tickets.

benjiwheeler:
Confirmed that this is now working in scratch.ly

PR #2500, opened 2018-12-17
pass onUpdateProjectThumbnail to gui when on project page

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/2484 
(together with https://github.com/LLK/scratch-gui/pull/4111 , but the two do not depend on each other)

### Changes:

Passes onUpdateProjectThumbnail function prop to `PreviewPresentation` and from there to `GUI`, so that the project page player has access to it (not just in editor view).

### Test Coverage:

none


benjiwheeler:
Confirmed that this is working in scratch.ly

PR #2502, opened 2018-12-17
Fix the ideas page's getting started SVG

ericrosenbaum:
The illustration for the "getting started" section of the ideas page was slow to fully load on some browsers (including mac Safari and iOS safari). 

This version of the SVG is smaller and does not contain an embedded bitmap image.


ericrosenbaum:
@benjiwheeler good call on svgo- I want to just get this change in though since it does fix the bug. I'll file another issue for doing an SVG optimization pass throughout www.

Issue #2496, opened 2018-12-17
Concerns about aggregating loves/favorites on the project page


BryceLTaylor:
@thisandagain writes:
We decided to only show 100+ on views, but I’m worried that aggregating the loves and favorites into “human” readable strings (e.g. 1K) may cause some serious issues with the community at launch.
![image11](https://user-images.githubusercontent.com/1088291/50097841-bc9da800-01e8-11e9-9d67-cd943933e4d3.png)

paulkaplan:
We decided to remove the approximation in  love/favorite numbering

PR #2505, opened 2018-12-18
Remove approx liking/favoriting format

paulkaplan:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/2496

### DIFF
--- a/src/views/preview/stats.jsx
+++ b/src/views/preview/stats.jsx
@@ -2,7 +2,6 @@ const PropTypes = require('prop-types');
 const React = require('react');
 const FlexRow = require('../../components/flex-row/flex-row.jsx');
 const classNames = require('classnames');
-const approx = require('approximate-number');
 
 const CappedNumber = require('../../components/cappednumber/cappednumber.jsx');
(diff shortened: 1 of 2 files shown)


src/views/preview/stats.jsx
```diff
@@ -16,20 +15,20 @@ const Stats = props => (
             key="loves"
             onClick={props.onLoveClicked}
         >
-            {approx(Math.max(0, props.loveCount), {decimal: false})}
+            {Math.max(0, props.loveCount)}
```
chrisgarrity: We should probably localize the numbers with `FormattedNumber` from `react-intl`: https://github.com/yahoo/react-intl/wiki/Components#number-formatting-components
```
<FormattedNumber value={Math.max(0, props.loveCount)} />
```

paulkaplan: Do we actually want commas and stuff? I thought we said we just wanted to go back to just numbers

chrisgarrity: Some non-latin languages actually use different numbers, but to avoid risk let's just file a followup issue to really localize the numbers.

PR #2515, opened 2018-12-19
Set isRemixing to true and then start remixing process

paulkaplan:
This makes it feel snappier to start the remixing project because it does not wait for the GUI to notify


paulkaplan:
@benjiwheeler I only noticed a difference with projects that took a long time to serialize

Issue #2513, opened 2018-12-19
Some screen widths cause stage to mess up 


benjiwheeler:
When the project page is viewed on mobile screen widths, the regular-sized stage seems to push our layout into a weird state. 

### Steps to Reproduce

1. View project page on a phone (I used a Pixel 2), or use Chrome's device simulation mode and choose a phone layout
2. See weirdness!

![image](https://user-images.githubusercontent.com/3431616/50246364-2366c100-03a3-11e9-950f-d53ebc8b95ca.png)

### Operating System and Browser

Mojave and current Chrome, on a mac;
And, current Android and Chrome on a Pixel 2

benjiwheeler:
Should consider @paulkaplan's change at https://github.com/LLK/scratch-gui/pull/3945 in addressing this

benjiwheeler:
I would advocate for giving this a priority 3.

PR #2518, opened 2018-12-20
Fix page width by hiding overflow, letting stage width be flexible

benjiwheeler:
### Resolves:

Resolves, at least partially: https://github.com/LLK/scratch-www/issues/2513
Resolves https://github.com/LLK/scratch-www/issues/2853
(https://github.com/LLK/scratch-gui/pull/4827 will also help resolve, but this PR does not require it)

### Changes:

* Flexibly allows the project page player stage width to be controlled by gui, rather than enforced to a particular width
* changes overflow handling so that the page width stays matching the window width (rather than becoming out of step, as seen in the screenshot from #2513)

### Before

image

image

### After

image

image

### Test Coverage:

None

### DIFF
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -479,6 +479,7 @@ $stage-width: 480px;
         line-height: 1.5rem;
         flex: 1;
         overflow-wrap: break-word;
+        word-break: break-word;
     }
 
     .project-description:last-of-type {
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -365,7 +365,7 @@ $stage-width: 480px;
     .guiPlayer {
         display: inline-block;
         position: relative;
-        width: $player-width;
+        max-width: $player-width;
         z-index: 1;
 
         $alert-bg: rgba(255, 255, 255, .85);
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -38,9 +38,11 @@ $stage-width: 480px;
 
     .inner {
         margin: 0 auto;
+        overflow: auto;
 
         @media #{$medium-and-smaller} {
-            max-width: 90%;
+            // subtract page padding
+            max-width: calc(100% - 1rem);
         }
 
         @media #{$intermediate} {
@@ -146,7 +148,8 @@ $stage-width: 480px;
         flex-grow: 1;
 
         @media #{$medium-and-smaller} {
-            min-width: calc(100% - 2rem);
+            // subtract margin and border
+            min-width: calc(100% - 2rem - 4px);
         }
     }
 
@@ -421,7 +424,7 @@ $stage-width: 480px;
         border-radius: 8px;
         background-color: $ui-blue-10percent;
         padding: .75rem;
-        width: calc(100% - 1.5rem);
+        width: calc(100% - 1.5rem - 2px);
         flex-wrap: nowrap;
         align-items: center;
         justify-content: flex-start;


rschamp:
This happened when I loaded a project that crashed. I don't think we want this to be possible:
image

benjiwheeler:
OK, I used `max-width` instead of `width` (and instead of eliminating the line completely).

rschamp:
image
This is what happens when I put a long unbroken string of letters into the description. I'm not sure if this would happen if a project was loaded (something is up with my local environment), but still doesn't look great in this case.

benjiwheeler:
> This is what happens when I put a long unbroken string of letters into the description

Yikes!

Hmm, it seems to be a problem that already exists. Here's the develop branch running locally:

In Chrome:
image

In Safari:
image

I've separately done some work to improve the breaking of long strings, in investigating https://github.com/LLK/scratch-www/issues/2853 (which I stopped working on because it wasn't prioritized) . If we want to address that here, there's a reasonable step forward: we could add one line of css to `.project-description`: 
```
        word-break: break-word;
``

This seems to improve things:

Chrome:
![image](https://user-images.githubusercontent.com/3431616/78092186-1786ee80-739d-11ea-8d01-1d45fdf326a9.png)

Safari:
![image](https://user-images.githubusercontent.com/3431616/78092201-27063780-739d-11ea-8830-9e7c18e8b837.png)


FWIW, just to compare production to this PR (regardless of whether we add `word-break: break-word;`)

here is a long description string with the current, existing production code:
![image](https://user-images.githubusercontent.com/3431616/78057689-221d9580-7355-11ea-8e6d-9b5bdcff963e.png)

...and with this PR's change:
![image](https://user-images.githubusercontent.com/3431616/78057703-2944a380-7355-11ea-9471-c1ae6bb74b31.png)

benjiwheeler:
Also, adding `word-break: break-word;` maintains this PR's improvement to the mobile view:

(Chrome, again:)

image

image

benjiwheeler:
Note that `word-break: break-word` is officially deprecated (see https://developer.mozilla.org/en-US/docs/Web/CSS/word-break ), but only very recently, in favor of the new `overflow-wrap: anywhere` (see https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap ); using `overflow-wrap: anywhere` seems like it would require us to bump up our required browser versions for Firefox and Chrome, and to not support Safari at all :( (see comment at https://github.com/LLK/scratch-www/issues/2853#issuecomment-595370314 )

benjiwheeler:
Another example:

production:
image

this PR:
image

benjiwheeler:
@rschamp just a heads up that this is now blocking some prioritized #ScratchAtHome issues

adroitwhiz:
@benjiwheeler It looks like adding `overflow: auto` to the `project-notes` element also fixes overflow without using deprecated CSS properties:

Production | `overflow: auto`
--- | ---
image | image

benjiwheeler:
@adroitwhiz Thanks! I think I tried that, but it's been a little while... do you know if it works on Mac Safari? That was the reason for using the deprecated break-word.

adroitwhiz:
I don't own a Mac so I can't test it out :(

benjiwheeler:
@adroitwhiz It does look like doing that works well in Safari!

Given how much I've tested this PR already, I would advocate for our releasing this change as-is, which seems strictly better than the status quo, and then looking into switching out the `word-break` setting in favor of `overflow: auto` to the `project-notes` element.

PR #2519, opened 2018-12-20
since api proxy of share endpoint returns string, handle string or boolean

benjiwheeler:
### Resolves:

Helps deal with https://github.com/LLK/scratch-api/issues/680 without really resolving the problem

### Changes:

sets `isShared` to true if `is_published` value received from api is the string `"true"`, not just the boolean value `true`.

### Test Coverage:

none


benjiwheeler:
Closing because as @paulkaplan points out, it's probably better to just fix API.

PR #2520, opened 2018-12-20
Add sentry.io for browser usage only

colbygk:
Removes existing sentry.io setup using Raven directly and substitutes usage for in browser, conditionally on if SENTRY_CONFIG is defined or not. I added the browser integration to the start of the template <HEAD> element to capture early errors instead of placing the raven include in the bottom part of <BODY>. Perhaps that is misguided?

### Resolves:

Usage of sentry.io to track errors.

### Changes:

Shifts usage to in browser only. Conditionally included on if SENTRY_CONFIG is defined in the process environment.

### Test Coverage:

N/A.  Coordinated and automated integration testing with sentry.io and browser usage may be possible, but is not feasible before release.


colbygk:
Closing this in favor of using sentry/webpack plugin

colbygk:
Closing in favor of sentry/webpack plugin integration for real.

PR #2526, opened 2018-12-20
Implement new Intro masthead

### LINKED ISSUE #2319: Logged-out Homepage Masthead
## Overview
When launch Scratch 3.0, we will want to update the masthead on the logged-out view of the homepage. Overall the content will be the same, but with a new video and overall look & feel to match 3.0.

### Current Masthead
screen shot 2018-11-15 at 1 39 42 pm

chrisgarrity:
### Resolves:

Fixes #2319 

### Changes:
Changes the logged out masthead to new banner style with new 3.0 intro video. 

### Testing:
- [ ] Launch banner will show instead of new masthead through Jan 18, 2019
- [ ] Launch banner links still work (there were slight localization changes)
- [ ] Masthead is responsive
- [ ] `Start Creating` call to action opens the editor with the getting started tutorial
- [ ] `Join` opens the registration modal
- [ ] Subnavigation buttons all go to the correct places.


benjiwheeler:
Confirmed that current real date shows "3.0 is here!", and 1/20/19 shows this new masthead ("Create stories...")

Wistia embed had a bit of trouble... clicking once gave me this:
image

Seems like ideally, a single click would start video, instead of needing 2 clicks

Couldn't get video to load. Is that expected?

Start Creating button worked

Join button worked

Subnav buttons all worked

Seems like the responsiveness is a bit off:

approx 760 px wide: this looks good
image

approx 1024 px wide: line breaks are a bit unexpected; maybe center aligning text (within its existing area) would be better?
image

approx 1200 px wide: this looks good
image

approx 1800 px wide: the left and right alignment makes the masthead look wrong
image

But, maybe it's best to merge this and then follow up

benjiwheeler:
Two follow-up items:

1. From quick look at Wistia docs, autoplay seems to be a setting that is attached to the file on Wistia's end, not an embed option. (Not sure about this, just from a quick glance.) So we're going to merge this and then follow up re: autoplaying the masthead video so it plays after 1 click, rather than 2.

2. Responsiveness of masthead is ok for now -- we can do another pass later if we think it's necessary.

PR #2535, opened 2018-12-21
Migrate FAQ page to FormattedMessage

### LINKED ISSUE #2521: Get rid of use of FormattedHTMLMessage in FAQ page
The FAQ page (info/faq) makes relatively heavy use of FormattedHtmlMessage.  We should make it use FormattedMessage instead.

Kenny2github:
### Resolves:

Resolves #2521 

### Changes:

Changes all instances of `FormattedHTMLMessage` to `FormattedMessage` and updates l10n to reflect this.


src/views/faq/l10n.json
```diff
@@ -10,11 +10,14 @@
...
+    "faq.makeGameBody":"Check out the {ideasLink} to see lots of ways to get started with Scratch",
+        "faq.ideasLinkText":"Ideas page",
     "faq.whoUsesScratchTitle":"Who uses Scratch?",
-    "faq.whoUsesScratchBody":"Scratch is used by people from all backgrounds, in all countries around the world, in all types of settings -- homes, schools, libraries, museums, and more. Scratch is designed especially for young people ages 8 to 16, but people of all ages create and share with Scratch. Younger children may want to try <a href=\"https://www.scratchjr.org/\">ScratchJr</a>, a simplified version of Scratch designed for ages 5 to 7.",
+    "faq.whoUsesScratchBody":"Scratch is used by people from all backgrounds, in all countries around the world, in all types of settings -- homes, schools, libraries, museums, and more. Scratch is designed especially for young people ages 8 to 16, but people of all ages create and share with Scratch. Younger children may want to try {scratchJrLink}, a simplified version of Scratch designed for ages 5 to 7.",
+        "faq.scratchJrLinkText":"ScratchJr",
```
chrisgarrity: _ScratchJr_ shouldn't get translated, so it doesn't need to be in the l10n file.

src/views/faq/l10n.json
```diff
@@ -27,23 +30,31 @@
...
+    "faq.languagesScratch3Body1":"Yes. To change the language of the programming blocks, click on the “globe” icon in the top navigation bar of the programming editor, then click on the dropdown menu to select a language.",
+    "faq.languagesScratch3Body2":"All of our translations are done by volunteers. The Scratch 3.0 editor has already been translated into 40+ languages. You can view all the languages currently being translated and reviewed on our {transifexLink}. If you want to help with translation or review, please contact {emailLink}."
```
chrisgarrity: missing comma at the end of the line.

src/views/faq/l10n.json
Kenny2github: I anticipated this, but there were a bunch of other strings like "Scratch 2.0" or "Scratch Desktop" that I wasn't sure whether to translate. I left them all in in case you had comments. I'll remove this one, but do let me know if any other strings I added don't actually need to be there.

src/views/faq/l10n.json
Kenny2github: Fixed.

Kenny2github:
Since there's more commits than I anticipated here, once you think it looks okay, could you give me a moment to rebase-squash and force push? If this is really urgent for some reason, then just go ahead and merge when you will.

src/views/faq/l10n.json
```diff
@@ -71,109 +81,118 @@
...
+        "faq.ccLinkText":"Creative Commons search page",
     "faq.whyAccountTitle":"Why is it useful to have a Scratch account?",
     "faq.whyAccountBody":"Even without an account, you can play other people’s projects, read comments and forums, and even create your own projects. But you need an account to save and share projects, write comments and forum posts, and participate in other \"social\" activities in the community (like \"loving\" other people’s projects).",
     "faq.createAccountTitle":"How can I create an account?",
-    "faq.createAccountBody":"Just click <i>Join</i> on the Scratch home page. You’ll need to respond to a few questions, and provide an email address. It takes just a couple minutes, and it’s totally free!",
+    "faq.createAccountBody":"Just click Join on the Scratch home page. You’ll need to respond to a few questions, and provide an email address. It takes just a couple minutes, and it’s totally free!",
```
chrisgarrity: We have a lot of places where we used `<i>` to indicate a command or button. I think it would be good to keep them, but then we need a new placeholder convention for it so that it's easy for translators to know what's going on. How about` {iJoin}`? Looking at other uses, there would be `{iAccountSettings}`, `{iExtensions}` etc. The weird one would be the `{iDeleteAccount}` because it doesn't really make sense to use the whole string. If you have a better idea for naming, please try it!

apple502j: What about `Just click \"Join\" on the`.... ?

Kenny2github: Using italics isn't the normal way to indicate a command or button. More frequently, people use quotes (i.e. click "Join" on the Scratch home page) or (at least in English) simply put it in title case and describe it as a button (i.e. click the Join button on the Scratch home page). Italics is awkward and wastes a lot of overhead, file size, and time. I think (retrospectively) that quotes best delimit a button, especially when one is told to "click" it. I'll switch to quotes.

chrisgarrity:
It all looked fine, but with all the little changes to the text it would be good to have an extra pair of eyes do another QA pass.

PR #2597, opened 2019-01-04
[Master] Fix hashtag search

paulkaplan:
Fixes an issue where the #search would lead to a search that included the `#` in it. 

@rschamp made this against master as per request? Is this right?

/cc @christanbalch @thisandagain


rschamp:
@paulkaplan Almost! Make sure to make the base branch `master`, and then make a second one into `develop` so they can stay in sync (Just did that here https://github.com/LLK/scratch-www/pull/2599)

PR #2624, opened 2019-01-08
Raise z-index of navbar

apple502j:
### Resolves:
Resolves LLK/scratch-gui#4230

### Changes:
Changes the z-index to 99.

According to my research, 99 is the best value for this - navbar should be under modal overlay (like reporting), which has `z-index:100;`.

Confirmed working with Firefox 64's inspector

NOTE: It might not work for Account Settings in editor.

### DIFF
--- a/src/views/preview/project-view.jsx
+++ b/src/views/preview/project-view.jsx
@@ -604,7 +604,7 @@ class Preview extends React.Component {
                             extensions={this.state.extensions}
                             faved={this.state.clientFaved}
                             favoriteCount={this.state.favoriteCount}
-                            isFullScreen={this.state.isFullScreen}
+                            isFullScreen={this.props.fullScreen}
                             isLoggedIn={this.props.isLoggedIn}
                             isNewScratcher={this.props.isNewScratcher}
                             isRemixing={this.state.isRemixing}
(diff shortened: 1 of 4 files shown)


benjiwheeler:
@apple502j I did some testing of your change and played around with it. My concern is that there are really three separate `z-index` contexts: the player itself; the editor, which doesn't appear on the project page at all (and has its own, separate navbar, that only looks similar!); and the rest of the project page.

When the project page navbar appears, we know we want the player --and everything in it-- to appear underneath basically everything except the page's background. So I think an even more fitting way to make this change would be to instead set the `z-index` in the `preview.scss` file, in the `.guiPlayer` class. Maybe to a value of 1, or 0, or 2?

What do you think?

benjiwheeler:
@apple502j @chrisgarrity I updated this PR to follow the approach I suggested above. Note that that involves correcting a seeming bug in `project-view.jsx` that wrongly named the flag it was using to mark fullscreen mode.

src/views/preview/project-view.jsx
```diff
@@ -604,7 +604,7 @@ class Preview extends React.Component {
                             extensions={this.state.extensions}
                             faved={this.state.clientFaved}
                             favoriteCount={this.state.favoriteCount}
-                            isFullScreen={this.state.isFullScreen}
+                            isFullScreen={this.props.fullScreen}
```
benjiwheeler: Note that this prop seems to have been pointing to a variable that no longer exists! I had to correct it so that `preview/presentation.jsx` would be able to use the right value.

PR #2638, opened 2019-01-08
Revert "Check for recommended browser version"

paulkaplan:
Reverts LLK/scratch-www#2630

We are going to take a softer approach on this, @thisandagain will follow up with an issue


thisandagain:
I tried to represent our discussion here: https://github.com/LLK/scratch-gui/issues/4297

PR #2702, opened 2019-01-22
added default censored messages

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-api/issues/693

### Changes:

Added client-side default censored message. This mirrors what we did on scratchr2, and allows the message to be translated.

Tweaked the vertical margins within the banner on the project page. There was not enough spacing around text, and spacing was inappropriately put on the banner button that really belonged on the banner as a whole.

### Test Coverage:

None.

### Screenshots

Temporarily censored:
image

Permanently censored:
image

Appearance of the unshared banner after my CSS tweaks:
image

Appearance of the remixed banner after my CSS tweaks:
image


benjiwheeler:
By markup, do you mean html/jsx? It seems like the right thing to do is refactor the whole banner into a separate file, but I think that would make more sense as a separate PR. I could do that as an immediate follow-on, how does that sound?

Re: the color change -- forgot to push!

Fixed the spacing -- I'm not sure why that happened. Fixed.

src/views/preview/presentation.jsx
```diff
@@ -131,22 +132,22 @@ const PreviewPresentation = ({
...
     let banner;
     if (visibilityInfo.deleted) { // If both censored and deleted, prioritize deleted banner
         banner = (<Banner
             className="banner-danger"
             message={<FormattedMessage id="project.deletedBanner" />}
         />);
     } else if (visibilityInfo.censored) {
+        const censoredMessage = (
+            <CensoredMessage
+                messageHTML={visibilityInfo.messageHTML}
```
paulkaplan: The visibilityInfo has a property called `message`, but it doesn't have one called `messageHTML`

benjiwheeler:
Corrected that error! Need more scenarios to test :(

I think it's good to be explicit that this isn't just a string of text -- it contains HTML that we are going to render (unsafely)

benjiwheeler:
OK, tested with and without a censored message:

image

PR #2703, opened 2019-01-22
Revert "Revert "Add to studio modal should only show studios user can actually change""

benjiwheeler:
Reverts LLK/scratch-www#2693

Since @colbygk and I tracked down the bug causing unexpected 403 errors on the backend, we're ready to give this another try!


benjiwheeler:
Blocking this until we merge https://github.com/LLK/scratch-api/pull/720, https://github.com/LLK/scratch-api/pull/721 and https://github.com/LLK/scratchr2/pull/5269

PR #2714, opened 2019-01-25
[Master] Wrap project inputs to show server validation

rschamp:
Replaces #2712

### DIFF
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -149,7 +149,7 @@ $stage-width: 480px;
         $arrow-border-width: 1rem;
         display: block;
         position: absolute;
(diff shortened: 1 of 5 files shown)


src/views/preview/preview.scss
```diff
@@ -149,7 +149,7 @@ $stage-width: 480px;
         $arrow-border-width: 1rem;
         display: block;
         position: absolute;
-        z-index: 1;
+        z-index: 10;
```
rschamp: `10` is the same value as the nav-bar so should be somewhere between 1 and 10 non-inclusive. So how about `5`.
```suggestion
        z-index: 5;
```

PR #2718, opened 2019-01-25
[develop] Fix project info updates being reflected after "See Inside"

paulkaplan:
This is a hotfix for a small issue with the fix we made earlier that allowed project info fields to show errors. They were not propagating changes up to the top-level store, so it was not seeing any changes after you "see inside".

Fixed by propagating those changes after successful updates.

Tested by modifying a project title, clicking see inside and noticing that the title in the editor is up-to-date.

Another test is modifying the project instructions, clicking "see inside" and then clicking "see project page", you'll notice that the instructions are up-to-date.

/cc @BryceLTaylor


src/views/preview/project-view.jsx
```diff
@@ -947,6 +950,9 @@ const mapDispatchToProps = dispatch => ({
         dispatch(previewActions.resetComments());
         dispatch(previewActions.getTopLevelComments(id, 0, isAdmin, token));
     },
+    handleReduxProjectInfo: info => {
```
rschamp: I think this could be much simpler if we just connected this in `FormsyProjectUpdater`.

PR #2719, opened 2019-01-25
[master] Pass updates up from project updater to top-level redux store

paulkaplan:
See https://github.com/LLK/scratch-www/pull/2718 for description

### DIFF
--- a/src/views/preview/formsy-project-updater.jsx
+++ b/src/views/preview/formsy-project-updater.jsx
@@ -1,10 +1,12 @@
 const PropTypes = require('prop-types');
 const React = require('react');
 const bindAll = require('lodash.bindall');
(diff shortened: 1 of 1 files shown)


src/views/preview/formsy-project-updater.jsx
```diff
@@ -41,6 +43,7 @@ class FormsyProjectUpdater extends React.Component {
         }, (err, body, res) => {
             if (res.statusCode === 200) {
                 this.setState({value: body[this.props.field], error: false});
+                this.props.onUpdate(jsonData);
```
rschamp: Most correct would be to update with body itself, in case we modified the input for some reason. I don't think that applies in any case though, so this is good for now

PR #2720, opened 2019-01-28
Add missing resource to .tx/config

chrisgarrity:
The `download/scratch2` l10n file was missing from the config file, along with the new ideas page.

Also, the `name` field in the routes json needs to match the resource name. So changed the `download-scratch2` route name to match the l10n name.

This is why you would get lots of things missing when making translations. The resources still need to be included in scratchr2_translations for the 'make' step to work properly.

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -104,6 +104,12 @@ source_file = src/views/download/l10n.json
 source_lang = en
 type = KEYVALUEJSON
 
+[scratch-website.download-scratch2-l10njson]
+file_filter = localizations/download/scratch2/<lang>.json
+source_file = src/views/download/scratch2/l10n.json
+source_lang = en
+type = KEYVALUEJSON
+
 [scratch-website.camp-l10njson]
 file_filter = localizations/camp/<lang>.json
 source_file = src/views/camp/l10n.json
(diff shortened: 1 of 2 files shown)


.tx/config
```diff
@@ -175,3 +181,8 @@ type = KEYVALUEJSON
 source_file = src/views/ideas/l10n.json
 source_lang = en
 type = KEYVALUEJSON
+
```
benjiwheeler: These lines below look like they're duplicated unnecessarily

src/routes.json
```diff
@@ -205,7 +205,7 @@
         "title": "Scratch 1.4"
     },
     {
-        "name": "scratch2",
+        "name": "download-scratch2",
```
benjiwheeler: is this renaming just to be clearer? Or does it have some functional effect?

chrisgarrity: This really has a functional effect. The `make translations` script iterates over the non-redirect routes in the `routes.json` to build a translations file for each view. The `name` of the route is what the script uses to look up the translation files. So if the name doesn't match, you get the 'no translations for ...' for every language.

.tx/config
chrisgarrity: Yup, too many copy/pastes. I force pushed to keep the commit history cleaner.

PR #2739, opened 2019-02-01
Wrap ‘need help?’ link when translation is too long

### LINKED ISSUE #2309: Handle translations that don't fit in the fixed width login panel
In some translations, the text does not fit in the fixed width of the login panel (200px). Examples are Thai and Greek (shown at the end)

This is not specifically a Scratch editor issue, the login component is used on every page on the site, and is currently a fixed 200px wide.

### Steps to Reproduce

Switch to Thai or Greek
image4
image1

chrisgarrity:
### Resolves:

Fixes #2309 

### Changes:

The `need help` link for resetting a password was awkwardly positioned when the translation was too bit to fit beside the `submit` button. 

Refactored the submit row of the login component to use Flex Row instead of `float: right` for the `need help` link.  The bulk of the changes are white space, so remember to ignore whitespace for review. The link wraps so that it is under the button if it doesn't fit next to it.

### Test Coverage:
A selection of languages:
screen shot 2019-02-01 at 2 20 43 pm
screen shot 2019-02-01 at 2 14 33 pm
screen shot 2019-02-01 at 2 13 45 pm
screen shot 2019-02-01 at 2 13 19 pm
screen shot 2019-02-01 at 2 12 54 pm

### DIFF
--- a/src/components/login/login.scss
+++ b/src/components/login/login.scss
@@ -28,14 +28,15 @@
         font-weight: bold;
     }
 
-    .right {
-        float: right;
-    }
-
     .spinner {
         margin: 0 .8rem;
         width: 1rem;
         vertical-align: middle;
+    }
+    
+    .submit-row {
+        justify-content: space-between;
+        flex-direction: row; // override 'column' layout at small widths
     }
 
     .submit-button {
@@ -43,7 +44,7 @@
     }
 
     a {
-        margin-top: 15px;
+        margin: auto 0;
         color: $ui-white;
 
         &:link,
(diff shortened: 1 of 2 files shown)


src/components/login/login.scss
```diff
@@ -28,22 +28,22 @@
...
-        float: right;
-    }
-
     .spinner {
         margin: 0 .8rem;
         width: 1rem;
         vertical-align: middle;
     }
+    
+    .submit-row {
```
benjiwheeler: I tested this, and it works great except that at widths under 768px, the `flex-direction` of `FlexRow` components switches to `column`; so to prevent the button from being centered in an ugly way, we need to enforce `row` here, like:

```
    .submit-row {
        justify-content: space-between;
        flex-direction: row; // override centered columns
    }
```

chrisgarrity: Good catch - I'd forgotten about the switch to centered column. I think the `FlexRow` component is poorly named. Another bit of tech debt.

PR #2748, opened 2019-02-05
Migrate Starter Projects

### LINKED ISSUE #2744: Migrate Starter Projects
https://scratch.mit.edu/starter_projects/
(Linked from Ideas)

This should be migrated to scratch-www.

Questions:
1) will it retire? (I don't think so, ST kept it when they made Ideas)
2) should new projects be added?
3) localization of project notes/description?

kyleplo:
### Resolves:

Resolves #2744 

### Changes:

Migrates the starter projects page to scratch-www.

### DIFF
--- a/src/views/starter-projects/starter-projects.json
+++ b/src/views/starter-projects/starter-projects.json
@@ -1,0 +1,284 @@
+{
+ "animation": [
+  {
+   "id": 10128067,
+   "type": "project",
+   "title": "Dance Party",
+   "image": "https://cdn2.scratch.mit.edu/get_image/project/10128067_144x108.png",
+   "author": {"username": "Scratchteam"},
+   "href": "https://scratch.mit.edu/projects/10128067/",
+   "stats": {"loves": 0, "remixes": 0}
+  },
+  {
+   "id": 10015059,
+   "type": "project",
+   "title": "Animate the Crab",
+   "image": "https://cdn2.scratch.mit.edu/get_image/project/10015059_144x108.png",
+   "author": {"username": "Scratchteam"},
+   "href": "https://scratch.mit.edu/projects/10015059/",
+   "stats": {"loves": 0, "remixes": 0}
+  },
+  {
+   "id": 10015060,
+   "type": "project",
+   "title": "The Pico Show: Intro",
+   "image": "https://cdn2.scratch.mit.edu/get_image/project/10015060_144x108.png",
+   "author": {"username": "Scratchteam"},
+   "href": "https://scratch.mit.edu/projects/10015060/",
+   "stats": {"loves": 0, "remixes": 0}
+  },
+  {
+   "id": 11806234,
+   "type": "project",
+   "title": "Greeting Card",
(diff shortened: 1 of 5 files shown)


carljbowman:
@kyleplo - Do mind adding a screenshot of the page? Thanks!

src/views/starter-projects/starter-projects.json
```diff
@@ -0,0 +1,206 @@
+{
+ "animation": [
+  {
+   "id": 10128067,
+   "type": "project",
+   "title": "Dance Party",
+   "thumbnailUrl": "https://cdn2.scratch.mit.edu/get_image/project/10128067_144x108.png",
+   "creator": "",
```
chrisgarrity: The Carousel component requires a `username` for projects. It expects `item.author.username`. For now you can probably just use 'Scratchteam' for all of them:
```
"author": {"username": "Scratchteam"},
```

src/views/starter-projects/starter-projects.json
```diff
@@ -5,31 +5,31 @@
    "type": "project",
    "title": "Dance Party",
    "thumbnailUrl": "https://cdn2.scratch.mit.edu/get_image/project/10128067_144x108.png",
-   "creator": "",
+   "creator": "Scratchteam",
```
chrisgarrity: Sorry, this isn't going to work. The `Thumbnail` component in Carousel is looking for certain properties in the current `item`: 
https://github.com/LLK/scratch-www/blob/9f2ca8428339f660db87f6735360c2f1de457fb8/src/components/carousel/carousel.jsx#L70-L81
On line 71, it is looking for `item.author.username`, so the item must have a property `author`, that consists of an object that includes a property `username`

Note: the `carousel.json` example json corresponds to the data needed for the `LegacyCarousel`, not `Carousel`. Other changes include the `stats` object, and the image should be `item.image`, not `item.thumbnailUrl`

chrisgarrity:
@carljbowman 
Here's what the new page looks like. The layout is reusing the 'carousel' component from the splash page which expects 5 projects not 4. We could create a new variant of the component that works for 4 projects, or we could consider adding another project to each category.
screen shot 2019-02-19 at 8 27 53 am

kyleplo:
@chrisgarrity Also, we could use some space between the rows. I noticed in the carousel component that there is a center option. Would that work?

chrisgarrity:
@kyleplo after discussion with Carl I made a minor update to the styles and now I think it's ready to go. I also changed the style name to use 'kebab-case' as that's the naming convention we use in scratch-www styles. 

There is one last thing that would be helpful if you think you can manage it.  There are a lot of commits on this patch, and it will be easier to read in the history if it's squashed into one commit. You'd use `git rebase -i <branch>`, keep the first commit, and squash all the rest. Then you need to do a force push to your github repo.

kyleplo:
@chrisgarrity Right now, I'm on vacation and I only have access to a Chromebook which doesn't have the command line, so I'm using Github's online editor. I'll have access to a Window computer tomorrow.

chrisgarrity:
@kyleplo sounds good - enjoy the rest of your vacation. I'll look for a rebased version later this week.

Issue #2768, opened 2019-02-11
Mod job posting: "read and write another language" rather than "speak"


seotts:
Under "Qualification" on https://scratch.mit.edu/jobs/moderator, it says, "Not required, but would be cool: Ability to speak another language." But really, we're looking for folks who can read and write languages other than English-- just speaking would not be as useful.

So, I propose we change "Not required, but would be cool: Ability to speak another language." to "Not required, but would be cool: Ability to read and write another language."

Kenny2github:
I think that's a bug, not a suggestion - it definitely *should* say read and write rather than speak. Is this `help wanted`?

jakel181:
I've done it! :) It is linked in this pull request. https://github.com/LLK/scratch-www/pull/2773 I would love to apply for it too, but I would need to wait two more years hahaha. Anyways, I hope it has been done correctly! :)

jakel181:
Well I think this can be closed now! :D @thisandagain https://github.com/LLK/scratch-www/pull/2807

PR #2773, opened 2019-02-13
Update moderator requirements.

jakel181:
### Changes:

_Changed: "Not required, but would be cool: Ability to speak another language." to "Not required, but would be cool: Ability to read and write another language."._
 Based on this issue: https://github.com/LLK/scratch-www/issues/2768

### DIFF
--- a/src/views/jobs/moderator/moderator.jsx
+++ b/src/views/jobs/moderator/moderator.jsx
@@ -60,7 +60,7 @@ const Moderator = () => (
                 </li>
                 <li>
                     Not required, but would be cool: Ability to read and
-                    write another language.
+                    write another language
                 </li>
             </ul>
             <br />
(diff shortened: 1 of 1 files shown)


src/views/jobs/moderator/moderator.jsx
```diff
@@ -59,8 +59,8 @@ const Moderator = () => (
                     Interest in visual arts, programming, or teaching
                 </li>
                 <li>
-                    Not required, but would be cool: Ability to speak another
-                    language
+                    Not required, but would be cool: Ability to read and
+                    write another language.
```
joker314: All the other bullet points do not terminate with a "."

```suggestion
                    write another language
```

jakel181: Sure! I removed the period. Good catch! @joker314

jakel181:
Well I think this can be closed now! (This too) :) @thisandagain #2807

PR #2792, opened 2019-02-18
Add support for "Recent" on search

JuegOStrower:
The Scratch API supports 3 modes, but in the search page there are only 2. This adds support for the missing one (recent).

### Changes:
Add the "Recent" option to the search page.

### Test Coverage:
I didn't

### DIFF
--- a/src/views/search/search.jsx
+++ b/src/views/search/search.jsx
@@ -18,7 +18,7 @@ const Tabs = require('../../components/tabs/tabs.jsx');
 const Page = require('../../components/page/www/page.jsx');
 const render = require('../../lib/render.jsx');
 
-const ACCEPTABLE_MODES = ['trending', 'popular'];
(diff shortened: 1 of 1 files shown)


apple502j:
No support, these may result some inappropriate results. (If someone searched for bad words recent projects that are not reviewed yet will show up, yeah, which is bad.)

Mr-Dave2:
@apple502j I’m not sure that I follow what you’re saying.  The Popular or Trending sort options don’t seem to do any filtering.  I just shared a project with the text “Looking Glass Factory” and it shows up in a search even though it is neither Popular nor Trending (but it is recent).  So I assume if I searched for bad words the projects I’d see would include recent projects even if sorted by the popular or trending key.

thisandagain:
@juegostrower We decided to not show this option out of moderation concerns. In the future please file an issue for discussion prior to posting a PR.

Issue #2761, opened 2019-02-08
Update Scratch Conference website 


me-win:
To support the regional Scratch conferences happening this year, we'll need to switch the Scratch conference website (scratch.mit.edu/conference) to this template: https://scratch.mit.edu/conference/2017  

We're still in the process of collecting content, which can be found in this [Google Doc](https://docs.google.com/document/d/1M_LJqOjAxxYFm3j-D8WbvFtWhZJgPCbZ5aNGSA2KZyc/edit?usp=sharing).

thisandagain:
@benjiwheeler Can you take this on as a task for this month? We'd like to have this deployed by March 1st.

/cc @chrisgarrity

benjiwheeler:
@thisandagain Will do.

benjiwheeler:
blocked pending getting access to the google doc.

me-win:
@benjiwheeler granted you access to the Google Doc!

jakel181:
I would be happy to help as well. :)

IamTregsthedev:
:D

me-win:
*** Due to unforeseen circumstances, there's a possibility that Scratch Africa will be postponed until next year. Organizers are still considering their options and will let us know ASAP.**

*** Update the URL for Scratch Conference Europe.** 
Raspberry Pi folks just got back to us with an updated link for their website: scratchconferenceeurope.raspberrypi.org

*** Update the name and description for Conferencia Scratch al Sur.**
Name: Scratch al Sur Conferencia Chile 2019

Organizers provided us with a new, more detailed description...

_Scratch al Sur Conferencia Chile 2019 is an event aimed at teachers of all educational areas and levels, who seek to innovate in the classroom through creative learning, thus supporting public policies that are promoted through the National Plan of Digital Languages, launched by the Chilean government as of 2019. Various workshops, panels, experiences, stands, a presentation of the new Scratch 3.0, Makey-Makey, and much more will be offered._

These are not high priority updates, and it would be fine to push them mid or late May (or later).

thisandagain:
Thanks @me-win! 

Just to confirm: what action would you like to take re: the Africa conference? Couple options:
- Remove the whole section
- Replace current copy with text provided
- Add text provided to the beginning of the current copy block
- Add text provided to the bottom of the current copy block
- Anything else you would like! :-)

@benjiwheeler Do you mind if I add this to your plate for May?

me-win:
Hi, sorry for the delay!

[...]

PR #2795, opened 2019-02-19
Add 2019 Regional conferences page

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/2761

### Changes:

* Adds 2019 Conferences page to the site, in the general format of the 2017 page, which listed several regional conferences around the world (as opposed to 2018, which focused on the Cambridge MA conference)
* Adds svg icons for use in the 2019 conference page


benjiwheeler:
Here's my tentative design so far. Compare to the 2017 page: https://scratch.mit.edu/conference/2017

Note that the text between the title and the listings is longer and more varied than in 2017. I tried to format it in a straightforward way, but we may need to give it another design pass.

image

image

image

benjiwheeler:
Updated current design:

image

image

src/routes.json
```diff
@@ -72,6 +72,14 @@
         "view": "conference/2018/schedule/schedule",
         "title": "Conference Schedule"
     },
+    {
+        "name": "conference-index-2019",
+        "pattern": "^/conference/2019/?$",
+        "routeAlias": "/conference(?!/201[4-5])",
```
chrisgarrity: I've forgotten why we need this - Ray would remember. It would be good to refresh everyone's knowledge.

src/views/conference/2019/index/index.jsx
```diff
@@ -0,0 +1,348 @@
...
+                        <tr className="conf2019-panel-row">
+                            <td className="conf2019-panel-row-icon">
+                                <img
+                                    alt="Language Icon"
+                                    className="conf2019-panel-row-icon-image"
+                                    src="/svgs/conference/index/language-icon-solid.svg"
+                                />
+                            </td>
+                            <td><FormattedMessage id="conference-2019.language" /></td>
+                            <td>{'Español (Spanish) - simultaneous translation during plenary sessions'}</td>
```
chrisgarrity: This should probably be translated, but we don't want `Español` translated, so I might use a placeholder:
`{espanol} (Spanish) - simultaneous translation during plenary sessions` 

We should also add information to the context in transifex so that people know how to translated it.

Is the translation just to English? Should probably include that if so.

src/routes.json
rschamp: The routeAlias is just a way for us to reduce the size of a giant regex in Fastly. But I think at this point we don't have to worry about that size limit? We should check. In case it still matters, the regex in routeAlias should match the pattern for the route, so this should change like so:
```suggestion
        "routeAlias": "/conference(?!/201[4-9])",
```

And the point of the alias is to group many similar patterns together, so anything that was using the `/conference(?!/201[4-5])` routeAlias should be updated to the new alias.

src/components/footer/conference/2019/footer.jsx
```diff
@@ -0,0 +1,119 @@
...
+                <p className="legal">
+                    <FormattedMessage id="general.copyrightDraft" />
+                </p>
+            </div>
+            <div className="media">
+                <div className="contact-us">
+                    <h4>Contact</h4>
+                    <p>
+                        <a href="mailto:help@scratch.mit.edu">
+                            Email Us
```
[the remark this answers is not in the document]
benjiwheeler: Done

src/components/footer/conference/2019/footer.jsx
```diff
@@ -0,0 +1,119 @@
...
+                        </li>
+                    </FlexRow>
+                </FlexRow>
+                <p className="legal">
+                    <FormattedMessage id="general.copyrightDraft" />
+                </p>
+            </div>
+            <div className="media">
+                <div className="contact-us">
+                    <h4>Contact</h4>
```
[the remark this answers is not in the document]
benjiwheeler: Done

src/components/footer/conference/2019/footer.jsx
```diff
@@ -0,0 +1,119 @@
...
+                    </FlexRow>
+                    <FlexRow
+                        as="ul"
+                        className="column"
+                    >
+                        <li>
+                            <a href="http://day.scratch.mit.edu">Scratch Day</a>
+                        </li>
+                        <li>
+                            <a href="http://sip.scratch.mit.edu">Scratch In Practice</a>
```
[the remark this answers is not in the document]
benjiwheeler: Leaving it for now, asked Lily

src/components/navigation/conference/2019/navigation.jsx
```diff
@@ -0,0 +1,27 @@
...
+                <a
+                    className="logo-a"
+                    href="/conference"
+                >
+                    <img
+                        alt="Scratch Logo"
+                        className="logo-a-image"
+                        src="/images/logo_sm.png"
+                    />
+                    <p className="logo-a-title">Conferences</p>
```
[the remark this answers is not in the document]
benjiwheeler: Done

src/views/conference/2019/index/index.jsx
benjiwheeler: My thinks it's just into English; changing.

I'm going to just use "Spanish" for now, and it will read "Espanol" in Espanol.

src/routes.json
benjiwheeler: Done

benjiwheeler:
1. 

> There are a few missing translations that should get added. I think I marked all the ones I found. Quick way to find missing strings: remove `intl/<view>.intl.js`, then load the page. Look for things in English.

Hm, doesn't seem to work for me, but it seems like it should!

2.  Seems like `Conferencia Scratch al Sur` should stay un-translated. Do you agree? (I'll ask My in the meantime)

src/views/conference/2019/index/index.jsx
```diff
@@ -53,7 +53,7 @@ const ConferenceSplash = () => (
                         src="/svgs/conference/flags/cl.svg"
                     />
                     <div className="conf2019-panel-title-text">
-                        <h3><FormattedMessage id="conference-2019.chileTitle" /></h3>
+                        <h3>Conferencia Scratch al Sur</h3>
```
chrisgarrity: +1

PR #2819, opened 2019-03-05
Update the integration test for the Scratch Conference link in the fo…

BryceLTaylor:
### Resolves:

The integration tests were not passing because we changed the Scratch Conference link in the footer to redirect to /conference/2019

### Changes:

The test now looks for the correct URL

### DIFF
--- a/test/integration/smoke-testing/test_footer_links.js
+++ b/test/integration/smoke-testing/test_footer_links.js
@@ -281,9 +281,9 @@ tap.test('clickScratchDayLink', options, t => {
 // SCRATCH CONFERENCE
 tap.test('clickScratchConferenceLink', options, t => {
     const linkText = 'Scratch Conference';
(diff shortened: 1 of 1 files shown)


test/integration/smoke-testing/test_footer_links.js
```diff
@@ -281,7 +281,7 @@ tap.test('clickScratchDayLink', options, t => {
 // SCRATCH CONFERENCE
 tap.test('clickScratchConferenceLink', options, t => {
     const linkText = 'Scratch Conference';
-    const expectedHref = '/conference';
+    const expectedHref = '/conference/2019';
```
chrisgarrity: This will work fine until next year when we switch it to 2020. Would it be ok to use `t.match` instead of `t.equal` and then just make sure the url contains `'/conference'` or `'/conference/20'` if you want to check that the redirect happened.

PR #2842, opened 2019-03-14
[master] Release 03/14/2019

paulkaplan:
Features
- Slider monitors min/max/step-size can be changed via right-click menu on the monitor. (credit to @apple502j)
- Renaming custom block inputs will automatically update argument reporters in the define stack.
- Performance improvements around lazy evaluation of code.

Bugfixes
- Fix for issue where changing inputs on a custom block would cause detached empty text blocks.
- Fix issues around sending too many commands to hardware extensions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "react-responsive": "3.0.0",
     "react-slick": "0.16.0",
     "react-string-replace": "0.4.1",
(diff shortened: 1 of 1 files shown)


towerofnix:
>Slider monitors min/max/step-size can be changed via right-click menu on the monitor.

Credits to @apple502j!

PR #2859, opened 2019-03-23
add ownerUsername to comment API endpoint URIs

benjiwheeler:
Must be merged at the same time as https://github.com/LLK/scratch-api/pull/774, which changes the same endpoint on the backend.

### Resolves:

Resolves https://github.com/LLK/scratch-api/issues/755

### Changes:

* adds `ownerUsername` to all 3 comment API endpoint URIs
* handles error responses to those requests in a less catastrophic way; before, we would crash the page with an "Oops" message; now, we just fail silently (this came up in development)

### Test Coverage:

None

### DIFF
--- a/src/redux/preview.js
+++ b/src/redux/preview.js
@@ -462,19 +462,19 @@ module.exports.getFavedStatus = (id, username, token) => (dispatch => {
     });
 });
 
-module.exports.getTopLevelComments = (id, offset, isAdmin, token) => (dispatch => {
+module.exports.getTopLevelComments = (id, offset, ownerUsername, isAdmin, token) => (dispatch => {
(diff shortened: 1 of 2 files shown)


src/redux/preview.js
```diff
@@ -767,7 +767,7 @@ module.exports.getProjectStudios = (id, ownerUsername, isAdmin, token) => (dispa
     dispatch(module.exports.setFetchStatus('projectStudios', module.exports.Status.FETCHING));
     api({
         uri: `${isAdmin ? '/admin' : `/users/${ownerUsername}`}/projects/${id}/studios`,
-        authentication: token
+        authentication: token ? token : null
```
benjiwheeler: I'm making the same change as above to here, for consistency

PR #2872, opened 2019-04-02
include comments with visibility set to marked by filter

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-api/issues/701 , along with https://github.com/LLK/scratch-api/pull/775

Should be merged at the same time as https://github.com/LLK/scratch-api/pull/775

### Changes:

Considers both comments visibility values of `'visible'` and `'markedbyfilter'` to indicate that comments should be visible.

### Test Coverage:

None :(

### DIFF
--- a/src/views/preview/comment/comment.jsx
+++ b/src/views/preview/comment/comment.jsx
@@ -111,7 +111,9 @@ class Comment extends React.Component {
             visibility
         } = this.props;
 
-        const visible = visibility === 'visible';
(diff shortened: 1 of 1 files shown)


paulkaplan:
@benjiwheeler as discussed offline, I think we want to make the API change so that there is no front-end implications

benjiwheeler:
Closing because we are pursuing a solution instead that has no front-end implications, per discussion with @paulkaplan

PR #2925, opened 2019-04-26
change studios for project API request to use admin status, owner status and token

benjiwheeler:
### Resolves:

Together with API PR https://github.com/LLK/scratch-api/pull/810, resolves https://github.com/LLK/scratch-api/issues/756

Must be merged at the same time as that API PR , as both change the URL for the same endpoint

### Changes:

Changes handling of the "studios for project" endpoint:

* URI now includes owner of project
* now passes auth token
* waits until after retrieving project info to request studios list, because we need to know ownership of project
* determines whether to hit admin endpoint or regular endpoint by checking if current user is admin

### Test Coverage:

None

### DIFF
--- a/src/redux/preview.js
+++ b/src/redux/preview.js
@@ -763,10 +763,11 @@ module.exports.getRemixes = id => (dispatch => {
     });
 });
 
-module.exports.getProjectStudios = id => (dispatch => {
+module.exports.getProjectStudios = (id, ownerUsername, isAdmin, token) => (dispatch => {
     dispatch(module.exports.setFetchStatus('projectStudios', module.exports.Status.FETCHING));
     api({
-        uri: `/projects/${id}/studios`
+        uri: `${isAdmin ? '/admin' : `/users/${ownerUsername}`}/projects/${id}/studios`,
+        authentication: token
     }, (err, body, res) => {
         if (err) {
             dispatch(module.exports.setFetchStatus('projectStudios', module.exports.Status.ERROR));
--- a/src/redux/preview.js
+++ b/src/redux/preview.js
@@ -763,12 +763,16 @@ module.exports.getRemixes = id => (dispatch => {
     });
 });
 
+
 module.exports.getProjectStudios = (id, ownerUsername, isAdmin, token) => (dispatch => {
     dispatch(module.exports.setFetchStatus('projectStudios', module.exports.Status.FETCHING));
-    api({
-        uri: `${isAdmin ? '/admin' : `/users/${ownerUsername}`}/projects/${id}/studios`,
-        authentication: token
-    }, (err, body, res) => {
+    const opts = {
+        uri: `${isAdmin ? '/admin' : `/users/${ownerUsername}`}/projects/${id}/studios`
+    };
+    if (token) {
+        Object.assign(opts, {authentication: token});
+    }
+    api(opts, (err, body, res) => {
         if (err) {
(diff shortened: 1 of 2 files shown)


src/views/preview/project-view.jsx
```diff
@@ -159,7 +159,10 @@ class Preview extends React.Component {
             if (typeof this.props.projectInfo.id === 'undefined') {
                 this.initCounts(0, 0);
             } else {
+                const token = this.props.user ? this.props.user.token : null;
                 this.initCounts(this.props.projectInfo.stats.favorites, this.props.projectInfo.stats.loves);
+                this.props.getProjectStudios(this.props.projectInfo.id,
```
benjiwheeler: had to move this here because now we need author username to target the correct endpoint

src/redux/preview.js
```diff
@@ -763,10 +763,11 @@ module.exports.getRemixes = id => (dispatch => {
     });
 });
 
-module.exports.getProjectStudios = id => (dispatch => {
+module.exports.getProjectStudios = (id, ownerUsername, isAdmin, token) => (dispatch => {
     dispatch(module.exports.setFetchStatus('projectStudios', module.exports.Status.FETCHING));
     api({
-        uri: `/projects/${id}/studios`
+        uri: `${isAdmin ? '/admin' : `/users/${ownerUsername}`}/projects/${id}/studios`,
```
paulkaplan: Hmmm wouldn't this cause the admin route to just be `/admin` instead of what I assume should be `/admin/users/:name/projects/:id/studios`?

paulkaplan: Oh wait misread the string interpolation. I thought we were doing admin routes just by prefixing the normal route with `admin`, but it looks like here admin route doesn't have the same route (doesn't have /users/:user). Is that right? Is that supposed to be that way?

src/redux/preview.js
```diff
@@ -763,10 +763,11 @@ module.exports.getRemixes = id => (dispatch => {
...
+        authentication: token
```
paulkaplan: In other places that we do this, we specifically check for falsey token and set it to `null`, I think we should do the same here (see e.g. https://github.com/LLK/scratch-www/blob/5899420a27a13311ebff7e2c6f577df947f704c6/src/redux/preview.js#L469)

src/redux/preview.js
benjiwheeler: Yes, admin and non-admin routes are supposed to be different in this way.

The idea with the separate admin routes is that we can skip a bunch of checking about the user's permission to view this particular resource; if the user is indeed an admin, we just let them view it.

We're adding the resource's owner's username to non-admin routes so that we can do a standardized first pass permissions check; the basic logic is "is the user the owner? if so, then proceed. if not, is the resource publicly visible/shared? if so, then proceed."

src/redux/preview.js
benjiwheeler: We want to use the token even if the user is not an admin, but I think you're still right that we should be explicitly using `null`. Will change.

benjiwheeler: OK, I used an approach that's used elsewhere in preview.js, where if token is truthy, I add it to the opts using `Object.assign()`.

benjiwheeler:
Hey, I put up a change that I think addresses your point, take a look!

src/redux/preview.js
paulkaplan: Ah ok i see.

PR #2932, opened 2019-04-30
LEGO BOOST extension landing page

ericrosenbaum:
Landing page for the LEGO BOOST extension at /boost, with instructions, starter projects, troubleshooting info, etc.

Also, the extension chip to display under the project on the project page.

Screen Shot 2019-04-30 at 4 07 41 PM


src/views/boost/l10n.json
```diff
@@ -0,0 +1,34 @@
...
+    "boost.headerText": "The {boostLink} kit brings your LEGO creations to life with powerful motors, a color sensor and more. By combining it with Scratch, you can build your own robotic creatures, tell physical-digital stories, invent new game controllers, or whatever else you can imagine.",
+    "boost.gettingStarted": "Getting Started",
+    "boost.connectingBoost": "Connecting BOOST to Scratch",
+    "boost.powerBoost": "Turn on your sensor by pressing the power button.",
+    "boost.useScratch3": "Use the {scratch3Link} editor.",
+    "boost.addExtension": "Add the BOOST extension.",
+    "boost.thingsToTry": "Things to Try",
+    "boost.makeAMotorMove": "Make a motor move",
+    "boost.findTurnMotorOnForSeconds": "Find the {turnMotorOnForSeconds} block and click on it.",
+    "boost.turnMotorOnForSeconds": "“turn motor A on for 1 seconds”",
```
benjiwheeler: I notice you're using curly quote characters here. That's probably fine, but I wanted to flag it in case there's some reason we don't want to do that. I do see curly quotes used in other l10n strings elsewhere, so I imagine it's fine.

src/views/boost/l10n.json
```diff
@@ -0,0 +1,34 @@
...
+    "boost.connectALegoBeam": "Connect a LEGO beam with an axle to motor A and click the block again to make it spin.",
+    "boost.starterProjects": "Starter Projects",
+    "boost.troubleshootingTitle": "Troubleshooting",
+    "boost.checkOSVersionTitle": "Make sure your operating system is compatible with Scratch Link",
+    "boost.checkOSVersionText": "The minimum operating system versions are listed at the top of this page. See instructions for checking your version of {winOSVersionLink} or {macOSVersionLink}.",
+    "boost.winOSVersionLinkText": "Windows",
+    "boost.macOSVersionLinkText": "Mac OS",
+    "boost.closeScratchCopiesTitle": "Close other copies of Scratch",
```
benjiwheeler: I think ideally we would move some of these common strings -- "Starter Projects", etc. -- into a shared general file, rather than reproducing them for each extension. But it's no worse here than with 5 other extensions' l10n files, so I guess this is fine for now.

src/views/boost/l10n.json
apple502j: Nope, don't use curly quotes. #1238
use normal ones escaped.

carljbowman:
@benjiwheeler - Awesome, thanks for reducing those file sizes!

src/views/boost/l10n.json
ericrosenbaum: Good catch- I'll address this as a followup (and talk to @chrisgarrity about it). thanks!

src/views/boost/l10n.json
ericrosenbaum: Good point! As we keep adding more of these landing pages, it might be time for some cleanup of the strings... I'll file it as a follow up issue.

PR #2952, opened 2019-05-08
revised conference page to postpone kenya, change sur title, eu desc

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/2761

### Changes:

* revise title of chile conference
* revise description of chile conference
* revise link to eu conference page
* postpone kenya conference with new description
* remove kenya location/date/etc. section
* move kenya section to bottom

### DIFF
--- a/src/views/conference/2019/index/l10n.json
+++ b/src/views/conference/2019/index/l10n.json
@@ -23,9 +23,10 @@
     "conference-2019.kenyaTitle": "Scratch2019NBO",
     "conference-2019.kenyaSubTitle": "Waves of Innovation",
     "conference-2019.kenyaDesc": "In recognition of Africa's technological contributions to the world and the potential of the youth of Africa, Scratch2019NBO will be held in Nairobi, Kenya. Join educators from around the world to share lessons, empower young people, and celebrate accomplishments in creative coding.",
+    "conference-2019.kenyaPostpone": "The Scratch2019NBO conference, originally planned for Nairobi, Kenya in July 2019, has been postponed. Information about future plans will be available later this year.",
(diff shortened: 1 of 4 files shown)


src/views/conference/2019/index/l10n.json
```diff
@@ -23,9 +23,10 @@
     "conference-2019.kenyaTitle": "Scratch2019NBO",
     "conference-2019.kenyaSubTitle": "Waves of Innovation",
     "conference-2019.kenyaDesc": "In recognition of Africa's technological contributions to the world and the potential of the youth of Africa, Scratch2019NBO will be held in Nairobi, Kenya. Join educators from around the world to share lessons, empower young people, and celebrate accomplishments in creative coding.",
+    "conference-2019.kenyaPostpone": "The Scratch2019NBO conference, originally planned for Nairobi, Kenya in July 2019, has been postponed. Information about future plans will be available later this year.",
```
chrisgarrity: This will have the same missing string id because it's new, and it won't be in scratch-l10n.

chrisgarrity:
@benjiwheeler I pushed an update to the script that fixes the problem with the script that builds the translations. I figured it out because languages before English in the list were fine, but English and later  would be missing the new l10n string.

benjiwheeler:
Thanks for figuring out the translation issues! And, I think we should merge these conference changes separately from that. I don't think these changes require that translation change (any more than other added strings have).

benjiwheeler:
Moved the `build-locales` change to its own PR: https://github.com/LLK/scratch-www/pull/2969

benjiwheeler:
Merging per conversation with @chrisgarrity

PR #2965, opened 2019-05-09
for admins, handle markedbyfilter comments differently, color them ye…

benjiwheeler:
### Resolves:

Follow on to https://github.com/LLK/scratch-api/pull/775 , further addressing questions raised in https://github.com/LLK/scratch-www/pull/2872 per feedback from mods.

### Changes:

For admins, we already distinguish visually between comments that are totally normal/visible, and comments that were reported; now we also distinguish comments that were markedbyfilter.


benjiwheeler:
What this looks like:

image

benjiwheeler:
> I confirmed that the 'markedbyfilter' comments are only visible to admins and they are highlighted in yellow. The description is a bit confusing, but I'm assuming that marked by filter comments should be hidden (assumed bad) until an admin can look at them.

Sorry, I should have explained this more clearly. You'd need an up to date API running to see that API will now return the `markedbyfilter` comments so that all users can see them -- which is how Scratch 2.0 behaved.

chrisgarrity:
@benjiwheeler 👍 I updated my api, and the flagged comment appears normal to a non-admin user. Thanks for the clarification.

Issue #2831, opened 2019-03-08
Search url escaping does not happen, throws error if user clicks between Projects and Studios


benjiwheeler:
### Steps to Reproduce

1. search for "100% pen"
2. results look good, and URL querystring looks like `?q=100%25%20pen`
3. now click Studios
4. site throws error with "Oops", URL querystring looks like `?q=100%+pen/`

The same problem happens if you first go to `https://scratch.mit.edu/search/studios?q=100%25%20pen`, then click Projects.

Kenny2github:
I believe it's not HTML escaping but rather URL escaping... and the specific problem is that the links to "projects" or "studios" aren't URL-escaped.

439bananas:
Hey, might I bring up that \*%\* causes an error?

MasterOfTheTiger:
> Hey, might I bring up that *%* causes an error?

That is exactly it. The "%" sign is escaped in some situations but not others.

PR #2966, opened 2019-05-11
make search urls use consistent encoding

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/2831

### Changes:

in search urls:
* use %20 for spaces everywhere
* encode search term strings every time we create them
* hide ‘q’ key if it has no value

### Test Coverage:

Added tests


test/integration/search/test_search.js
```diff
@@ -0,0 +1,58 @@
+/*
```
chrisgarrity: Yay - a test! But right now it's not run automatically because it's an 'integration' test.

`make test` runs `unit, functional, localization` (no integration)
`make smoke` only runs `integration/smoke-testing`, so anything that is just in `integration` falls through the cracks. 

I also just noticed that there are no `functional` tests. Perhaps they were removed at some point.

Is this test something that should run locally for developers, or only as part of an automated smoke test?

benjiwheeler: Thanks for pointing this out. I'm moving the test into `smoke-testing`, though it doesn't need the full smoke environment. 

I also filed this issue re: integration tests not running: https://github.com/LLK/scratch-www/issues/2992

Issue #2603, opened 2019-01-04
Remix Credit info has awkward alignment if project title is short


carljbowman:
### Overview
Looks to be a flexbox issue. 

<img width="470" alt="screen shot 2019-01-04 at 10 06 05 am" src="https://user-images.githubusercontent.com/3409578/50694658-7ded5780-1008-11e9-94ae-59b32778a2a9.png">

Ideally the icon and text would be left aligned with in the larger credit container.

### Expected Behavior

With short lines of text, the remix credit should look like this

![image](https://user-images.githubusercontent.com/3431616/57789749-89afff80-7707-11e9-94cf-af1558aa2c26.png)

...or maybe this

![image](https://user-images.githubusercontent.com/3431616/57789758-8e74b380-7707-11e9-8a98-3a0765e6d938.png)

### Actual Behavior

It looks like this 

![image](https://user-images.githubusercontent.com/3431616/57789770-93396780-7707-11e9-9f44-9d4e064654dc.png)

![image](https://media2.giphy.com/media/jquDWJfPUMCiI/200.webp?cid=790b76115cdc35383655505a6f120bd8&rid=200.webp)

PR #2973, opened 2019-05-15
fix alignment of remix credit avatar, text

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/2603

### Changes:

* fixes alignment of remix credit avatar by overriding its vertical alignment, to align to the top. 
* fixes alignment of remix credit text by aligning it to the left.

### DIFF
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -3,7 +3,8 @@
 
 /* stage size constants */
 $player-width: 482px;
-$player-height: 406px;
+$player-height: 362px;
+$player-header: 44px;
 $stage-width: 480px;
 
 /* override view padding for share banner */
@@ -393,11 +394,11 @@ $stage-width: 480px;
 
     .project-notes {
         margin-left: 1rem;
-        height: $player-height;
+        height: calc(#{$player-height} + #{$player-header} - .3125rem);
         align-items: flex-start;
         flex: 1;
         flex-flow: column;
-        padding-top: .25rem;
+        margin-top: .3125rem;
 
         @media #{$medium-and-smaller} {
             margin-top: .5rem;
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -471,8 +471,14 @@ $stage-width: 480px;
     .project-description-form {
         display: flex;
         width: 100%;
+        // surprisingly, without setting height: 100% here, flex-grow causes the
+        // height to be too large
         height: 100%;
+        // surprisingly, just having some min-height actually reduces the height
+        // of the element when it is being vertically crowded by other elements
         min-height: 2rem;
+        // necessary to cause the element to take up the maximum height
+        // available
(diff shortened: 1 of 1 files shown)


benjiwheeler:
Screenshots of this change: 


image

image

image

image

image

image

image

carljbowman:
@benjiwheeler - Everything looks good to me from a layout perspective. Also learning that individual elements within flex-box can have their own flow and wrap styles is pretty amazing. A new magical layout power to add to our tool kit.

Thanks again for investigating!

src/views/preview/preview.scss
```diff
@@ -403,8 +404,12 @@ $stage-width: 480px;
             width: 100%;
         }
 
-        > .description-block:first-child {
-            margin-top: 1rem;
+        > * {
```
chrisgarrity: I think what you've got is probably the best solution. I just wanted to point out that the original `first-child` css was to make sure the top of the Instructions lined up with the top of the stage where there were no credits.
Screen Shot 2019-05-20 at 2 57 37 PM

chrisgarrity: @carljbowman just want to confirm that this is ok for the case where there are no remix credits.

carljbowman: That seems fine with me, thanks for checking. It was nice that those lined up but not critical. 

Chatted with @benjiwheeler quickly on a follow-up question: the player adds a `0.5rem`  padding-top to the of the controls. This padding makes sense for full screen, but in the project page player view could we drop the 0.5rem padding to get some more clean alignment:

benjiwheeler: The only problem with @carljbowman's idea is that it involves a change to scratch-gui's css, for a problem in www.

Here's what I came up with:

Add a .25rem padding-top to the div that wraps the instructions/remix credit column. That way the top item is brought into line with the top of the fullscreen button. 

Result:

image

image

carljbowman: @benjiwheeler - While the addtional .25rem padding-top aligns "Instructions" the fullscreen button, it pushes that entire column and the bottom of the player and the bottom of "Notes and Credits" become misaligned. 

misaglinement

If we had to choose, I'd prefer to keep the bottoms of the player and "Notes and Credits" aligned. They have much stronger horizontal lines for the eye to follow, so the misalignment feels much more apparent.

chrisgarrity: Changing the stage css (that's what this would require) is tricky. It's already handling normal, small stage and fullscreen. FWIW, I would just align the bottoms.

benjiwheeler: I agree, I am aligning the bottoms.

benjiwheeler:
OK, it looks pretty solid now!

image

image

carljbowman:
@benjiwheeler - Layout wise everything looks good to me. Thanks for iterating on this.

PR #3025, opened 2019-06-03
Skip flakey tests

BryceLTaylor:
### Resolves:

Some tests in the footer were flaky since they linked to other sites.  Also we were using an extremely old version of Tap.  Updating it caused other  tests to fail.

### Changes:

Updates TAP to version 14.2.0.  Several tests were skipped so we can revisit them individually later to make them pass.  This also adds a waitUntilGone function that makes sure the loading screen on a project page is gone before clicking anything else.


benjiwheeler:
@BryceLTaylor, could you create an issue or two to track the follow-up test work we'll need to do to resolve the skipped tests?

test/integration/smoke-testing/test-my-stuff.js
```diff
@@ -26,8 +26,6 @@ var password = process.env.SMOKE_PASSWORD;
 var rootUrl = process.env.ROOT_URL || 'https://scratch.ly';
 var url = rootUrl + '/users/' + username;
 
-tap.plan(7);
```
rschamp: Is removing the plan necessary?

test/integration/smoke-testing/test_search.js
```diff
@@ -13,7 +13,7 @@ const tap = require('tap');
 const test = tap.test;
 
 // Set test url through environment variable
-const rootUrl = process.env.ROOT_URL || 'http://localhost:8333';
+const rootUrl = process.env.ROOT_URL || 'https://scratch.ly';
```
rschamp: Is this change intentional?

BryceLTaylor: Yes.  All other tests in this suite point to scratch.ly by default.  This one pointed to localhost by default, meaning that if you don't put in a url, most tests will point to staging and these would fail unless you have a build running locally.

test/integration/smoke-testing/test-my-stuff.js
BryceLTaylor: It probably isn't.  It seems like we should look into how necessary tap.plan is in a separate pr.  I'll update this one.

package.json
```diff
@@ -6,7 +6,7 @@
     "start": "make start",
     "stop": "make stop",
     "test": "make test",
-    "smoke": "tap ./test/integration/smoke-testing/*.js --timeout=3600",
+    "smoke": "tap ./test/integration/smoke-testing/*.js --timeout=3600 --no-coverage -R classic",
```
rschamp: We should probably also add `--no-coverage -R classic` to all the other tap tests to keep the output similar across them.

BryceLTaylor: I added it to smoke-sauce, but I only added the --no-coverage to smoke-verbose since it already used a different report method, -R spec

Issue #3053, opened 2019-06-18
Revise and update user join flow


benjiwheeler:
We should revise and update the user join flow!

This is a big project with many, many parts and stages.

benjiwheeler:
Join flow has launched! It's wild to look at this sea of merged PRs.

There are still outstanding issues, both small fixes and large design ideas. We've been filing those individually with the new "join flow" label.

cc: @picklesrus @kathymakes @thisandagain @carljbowman @BryceLTaylor

PR #3054, opened 2019-06-18
introduce stub components for new scratch3 join modal

benjiwheeler:
### Resolves:

Initial step towards resolving https://github.com/LLK/scratch-www/issues/3053

### Changes:

introduces stub components for new scratch3 join modal


picklesrus:
This looks good to me!  If you're still happy with the approach, I think we should merge.

PR #3058, opened 2019-06-19
Provide valid country choices during registration

rschamp:
Otherwise this data silently falls back to “United States”. In case of other failures during registration, the country validation error will be erroneously displayed.

### Resolves:

Partial fix for https://github.com/LLK/scratchr2/issues/5321, by allowing errors other than this "country" field error to display.  Currently the country error occurs on every registration but is silent.

### Changes:

Update the country choices (and the default choice) during student and teacher registration to valid choices on scratchr2's side so that they will be recorded.

### Test Coverage:

This is hard to test since it's an interaction between scratch-www and the scratchr2 backend. This is also probably why we missed it for so long... Suggestions welcome on how to cover this case and others like it!


rschamp:
I wasn't sure whether to assign @picklesrus or @benjiwheeler — both welcome to review!

PR #3066, opened 2019-06-20
Add Boost and Vernier landing pages to the transifex configuration

chrisgarrity:
The Boost and Vernier pages were never added to the transifex config, and never pushed to transifex for translation. This adds them to the config. I also ran `tx push -s` to push the latest version of everything to transifex.


apple502j:
Hmm, I saw it on Transifex (and also got "new files notification") today. @chrisgarrity

PR #3075, opened 2019-06-24
added formik dependency, join flow steps using it

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

Adds formik dependency, adds join flow steps file using formik

### DIFF
--- a/src/components/join-flow/join-flow.jsx
+++ b/src/components/join-flow/join-flow.jsx
@@ -1,8 +1,12 @@
+const bindAll = require('lodash.bindall');
+const defaults = require('lodash.defaultsdeep');
 const PropTypes = require('prop-types');
 const React = require('react');
 
 const injectIntl = require('../../lib/intl.jsx').injectIntl;
 const intlShape = require('../../lib/intl.jsx').intlShape;
+
+const JoinFlowSteps = require('../registration/join-flow-steps.jsx');
 
 /*
 eslint-disable react/prefer-stateless-function, react/no-unused-prop-types, no-useless-constructor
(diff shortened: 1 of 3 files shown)


src/components/registration/join-flow-steps.jsx
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable react/no-multi-comp */
```
picklesrus: What do you think of just naming this one of the actual steps (e.g. CountryStep or UsernameStep)?  I think it can still be a skeleton though.

src/components/registration/join-flow-steps.jsx
```diff
@@ -0,0 +1,40 @@
...
+        );
+    }
+}
+/* eslint-enable */
+
+ExampleStep.propTypes = {
+};
+
+ExampleStep.defaultProps = {
+    showPassword: false,
```
picklesrus: I don't think these props make sense yet.

src/components/join-flow/join-flow.jsx
```diff
@@ -1,19 +1,35 @@
+const bindAll = require('lodash.bindall');
```
picklesrus: looks like we don't need bindall just yet

benjiwheeler: we're using it below

src/components/registration/join-flow-steps.jsx
benjiwheeler: Sure, done

src/components/registration/join-flow-steps.jsx
benjiwheeler: Removed

PR #3082, opened 2019-06-25
working username step of join flow

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

Adds form code and handlers to make username/password step of join flow work.

### DIFF
--- a/src/components/join-flow/join-flow-steps.jsx
+++ b/src/components/join-flow/join-flow-steps.jsx
@@ -31,57 +31,6 @@ class UsernameStep extends React.Component {
     handleChangeShowPassword () {
         this.setState({showPassword: !this.state.showPassword});
     }
-    // validateUsernameLocally (username, require) {
-    //     let error;
-    //     if (username.length === 0 && !require) { // allow blank username if not submitting yet
-    //         return null;
-    //     }
-    //     if (!username || username === '') {
-    //         error = this.props.intl.formatMessage({id: 'form.validationRequiredNice'});
-    //     } else if (username.length < 3) {
-    //         error = this.props.intl.formatMessage({id: 'registration.validationUsernameMinLength'});
-    //     } else if (username.length > 20) {
-    //         error = this.props.intl.formatMessage({id: 'registration.validationUsernameMaxLength'});
-    //     } else if (!/^[\w-]+$/i.test(username)) {
-    //         error = this.props.intl.formatMessage({id: 'registration.validationUsernameRegexp'});
-    //     }
-    //     return error;
-    // }
-
-    // validateUsernameRemotely (username, require) {
-    //     if (username.length === 0 && !require) { // allow blank username if not submitting yet
-    //         return null;
-    //     }
-    //     return new Promise((resolve, reject) => {
-    //         api({
-    //             uri: `/accounts/checkusername/${username}/`
-    //         }, (err, body, res) => {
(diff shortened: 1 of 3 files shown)


src/components/join-flow/join-flow-steps.jsx
```diff
@@ -22,14 +93,81 @@ class UsernameStep extends React.Component {
...
+                                    )}
+                                </div>
+                                <FormikInput
+                                    className={errors.password ? 'fail' : ''}
+                                    error={errors.password}
+                                    id="password"
+                                    name="password"
+                                    type={this.state.showPassword ? 'text' : 'password'}
+                                    validate={this.validatePasswordIfPresent}
+                                    onBlur={() => validateField('password')} // eslint-disable-line react/jsx-no-bind
```
picklesrus: Why do you need to use an arrow function here (and in the other places in this file)? I know very little about this, but seems like the check is there for a reason.

src/components/join-flow/join-flow-steps.jsx
```diff
@@ -22,14 +93,81 @@ class UsernameStep extends React.Component {
...
+                    } = props;
+                    return (
+                        <JoinFlowStep
+                            description={this.props.intl.formatMessage({id: 'registration.usernameStepDescription'})}
+                            title={this.props.intl.formatMessage({id: 'general.joinScratch'})}
+                            waiting={isSubmitting}
+                            onSubmit={handleSubmit}
+                        >
+                            <div>
+                                <div className="username-label">
```
picklesrus: is the username-label class used yet?

src/components/join-flow/join-flow-steps.jsx
```diff
@@ -22,14 +93,81 @@ class UsernameStep extends React.Component {
...
+                                    validate={this.validateUsernameIfPresent}
+                                    onBlur={() => validateField('username')} // eslint-disable-line react/jsx-no-bind
+                                />
+                                <b>
+                                    {this.props.intl.formatMessage({id: 'general.password'})}
+                                </b>
+                                <div
+                                    onClick={this.handleChangeShowPassword}
+                                >
+                                    {this.state.showPassword ? 'Hide password' : (
```
picklesrus: I'm I'm reading correctly "Hide Password" is just some text we want to show that we don't have a message for yet? Maybe add a TODO to turn it into a message?

src/components/join-flow/join-flow-steps.jsx
benjiwheeler: So, it's a linting rule because when you use these, they create a new function on every render. If you render often, that can be slow.

But there are times there's really no way around them, and I think this is one, because of the need to pass the 'password' param.

Also, if the component won't be rerendering much, it's not a big deal. There's even a lot of disagreement out there about if it's ever a big deal.

src/components/join-flow/join-flow-steps.jsx
benjiwheeler: Good idea!

src/components/join-flow/join-flow-steps.jsx
picklesrus: Cool. Thanks for the explanation!

src/components/join-flow/join-flow-steps.jsx
benjiwheeler: No, good catch, removing

PR #3086, opened 2019-06-26
use html form instead of formik form

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

in join-flow-step.jsx, use html form instead of formik form.


benjiwheeler:
> can you give me a little context on 3086? (the change from to html form) Code looks fine, but would like to understand why :slightly_smiling_face:

So, this is a matter of taste... I tend to prefer having less magic, more that's explicit. I realized that I had no idea where to find the code that watched for the form submit and transmitted that to Formik. When I saw that I could handle this explicitly, I made that change because I think that makes it much clearer what the flow is from our code into, and back out of, Formik.

Issue #3106, opened 2019-07-08
modal title styles are referenced out of scope


benjiwheeler:
In most of the modals, their presentation component references the "content-label" class, even though that class exists only in a higher scope than them. This works ok, but it would be better if there were some kind of ModalTitle component instead, that managed its own styling.

PR #3107, opened 2019-07-08
created ModalTitle component

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/3106

### Changes:

Introduces `ModalTitle` component, which contains its own styling instead of having its styles referenced by other components, out of scope.

### DIFF
--- a/src/components/modal/base/modal-title.jsx
+++ b/src/components/modal/base/modal-title.jsx
@@ -1,17 +1,25 @@
+const classNames = require('classnames');
 const PropTypes = require('prop-types');
 const React = require('react');
 
 require('./modal-title.scss');
 
 const ModalTitle = ({
+    className,
     title
 }) => (
-    <div className="modal-title">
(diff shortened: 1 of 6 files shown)


src/components/modal/base/modal-title.jsx
```diff
@@ -0,0 +1,18 @@
+const PropTypes = require('prop-types');
+const React = require('react');
+
+require('./modal-title.scss');
+
+const ModalTitle = ({
+    title
+}) => (
+    <div className="modal-title">
```
rschamp: Since it's likely we might want to further modify the styles, it would be prudent to include a `className` prop (which would be added in addition to `modal-title`).

benjiwheeler: I agree, and I had that lined up for the next PR, but I'm happy to put it into this one!

Issue #3112, opened 2019-07-10
two ".modal-content" styles are used in redundant and confusing ways


benjiwheeler:
the `src/components/modal/base/modal.scss` file contains two `.modal-content` styles, which are treated differently depending on whether one is inside the other.

In addition, the pattern of using the inner `.modal-content` after the modal's title is redundant across many of the modal types' render methods, and might make sense to refactor.

Finally, the use of that inner `.modal-content` is happening in a scope that does not include the definition of that style. This also might make sense to refactor.

PR #3113, opened 2019-07-10
add ModalInnerContent component, use it in modals

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/3112

### Changes:

* renames the `.modal-content` style when it is used inside another `.modal-content` style; now we'll call that inside one `.modal-inner-content`.
* refactors uses of that inner modal-content style, so they use a component instead of using the style out of scope.

### DIFF
--- a/src/components/modal/addtostudio/presentation.jsx
+++ b/src/components/modal/addtostudio/presentation.jsx
@@ -11,6 +11,7 @@ const Spinner = require('../../spinner/spinner.jsx');
 const FlexRow = require('../../flex-row/flex-row.jsx');
 const StudioButton = require('./studio-button.jsx');
 const ModalTitle = require('../base/modal-title.jsx');
+const ModalInnerContent = require('../base/modal-inner-content.jsx');
 
 require('../../forms/button.scss');
 require('./modal.scss');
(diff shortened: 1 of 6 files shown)


src/components/modal/addtostudio/presentation.jsx
```diff
@@ -11,6 +11,7 @@ const Spinner = require('../../spinner/spinner.jsx');
 const FlexRow = require('../../flex-row/flex-row.jsx');
 const StudioButton = require('./studio-button.jsx');
 const ModalTitle = require('../base/modal-title.jsx');
+const ModalInnerContent = require('../base/modal-inner-content.jsx');
```
rschamp: Curious about why it's specified as InnerContent, rather than just Content?

src/components/modal/addtostudio/presentation.jsx
```diff
@@ -49,7 +50,7 @@ const AddToStudioModalPresentation = ({
             <div className="addToStudio-modal-header modal-header">
```
rschamp: I just noticed we are still wrapping the title in another div — is that necessary or can those be absorbed into the `ModalTitle` component?

src/components/modal/addtostudio/presentation.jsx
benjiwheeler: To distinguish it from the existing "modal-content" css style (which I think isn't the right name), which wraps the entire modal.

src/components/modal/addtostudio/presentation.jsx
benjiwheeler: That's the header bar, which some modals have. I think that would make more sense as a separate, ModalHeader component.

benjiwheeler: There's lots more to refactor, trying to keep focused on what I need to move forward with Join Flow :)

Issue #3115, opened 2019-07-10
Need to correctly position validation errors in join flow


benjiwheeler:
The validation errors should not look like this:

![image](https://user-images.githubusercontent.com/3431616/60992051-b9e7d900-a319-11e9-9a90-9a8bbc91c717.png)

PR #3116, opened 2019-07-10
Format and position validation error messages in join flow

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/3115

### Changes:

* Adds `validationClassName` parameter to `FormikInput` component
* Set formatting of validations in join flow using `validationClassName`

### DIFF
--- a/src/components/modal/base/modal-inner-content.scss
+++ b/src/components/modal/base/modal-inner-content.scss
@@ -6,7 +6,7 @@ necessary to prevent modal-inner-content's margins from
 unexpectedly collapsing with a parent
 */
 .modal-inner-clear {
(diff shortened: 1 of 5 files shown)


src/components/modal/base/modal-inner-content.scss
```diff
@@ -6,7 +6,7 @@ necessary to prevent modal-inner-content's margins from
 unexpectedly collapsing with a parent
 */
 .modal-inner-clear {
-    overflow: auto;
+    float: left;
```
benjiwheeler: better method of clearing margins, which allows validation errors to overflow box

PR #3117, opened 2019-07-10
Format join flow using ModalTitle, ModalInnerContent, NextStepButton components

benjiwheeler:
### Resolves:

### Changes:

* use new ModalTitle, ModalInnerContent components in join flow
* set width, margins to make join flow modal look closer to spec
* use ModalTitle in the "Next" button at bottom

Now looks like:

image

### DIFF
--- a/src/components/join-flow/join-flow-step.jsx
+++ b/src/components/join-flow/join-flow-step.jsx
@@ -2,6 +2,10 @@ const React = require('react');
 const PropTypes = require('prop-types');
 
 const NextStepButton = require('./next-step-button.jsx');
+const ModalTitle = require('../modal/base/modal-title.jsx');
+const ModalInnerContent = require('../modal/base/modal-inner-content.jsx');
+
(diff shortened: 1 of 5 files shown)


benjiwheeler:
This can be reviewed now.

src/components/join-flow/join-flow-step.jsx
```diff
@@ -12,19 +16,22 @@ const JoinFlowStep = ({
...
+            <ModalInnerContent className="join-flow-inner-content">
+                {title && (
+                    <ModalTitle
+                        className="join-flow-title"
+                        title={title}
+                    />
+                )}
+                {description && (
+                    <p>
+                        <span>
```
picklesrus: do we need both the p and the span? Seems like the span would be unnecessary here.

benjiwheeler: Will fix this in a follow-up.

PR #3121, opened 2019-07-11
introduce formik select component

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

Introduce `formik-select.jsx` component, modeled on `formik-input.jsx` but handling an `options` prop

### DIFF
--- a/src/components/formik-forms/formik-select.jsx
+++ b/src/components/formik-forms/formik-select.jsx
@@ -24,7 +24,7 @@ const FormikSelect = ({
         </option>
     ));
     return (
-        <div className="col-sm-9 row row-relative">
+        <div className="col-sm-9 row">
             <Field
                 className={classNames(
                     'select',
@@ -49,13 +49,13 @@ const FormikSelect = ({
 FormikSelect.propTypes = {
     className: PropTypes.string,
     error: PropTypes.string,
-    // expect structure like: [
-    //    {value: 'US', label: 'United States'}
-    //    {value: 'AG', label: 'Angola'}
-    //    ...
-    // ]
-    options: PropTypes.arrayOf(PropTypes.any),
+    options: PropTypes.arrayOf(PropTypes.shape({
+        disabled: PropTypes.bool,
+        label: PropTypes.string.isRequired,
+        value: PropTypes.string.isRequired
+    })).isRequired,
     validationClassName: PropTypes.string,
+    // selected value
     value: PropTypes.oneOfType([PropTypes.number, PropTypes.string])
 };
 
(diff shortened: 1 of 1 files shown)


src/components/formik-forms/formik-select.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+FormikSelect.propTypes = {
+    className: PropTypes.string,
+    error: PropTypes.string,
+    // expect structure like: [
+    //    {value: 'US', label: 'United States'}
+    //    {value: 'AG', label: 'Angola'}
+    //    ...
+    // ]
+    options: PropTypes.arrayOf(PropTypes.any),
```
picklesrus: If we have specific requirements, about the object shape, I think we should write a more specific validator. I tried the code below out and it seems to work as I hoped:
options: PropTypes.arrayOf(PropTypes.shape({
     value: PropTypes.string.isRequired,
     label: PropTypes.string.isRequired,
 })),

src/components/formik-forms/formik-select.jsx
```diff
@@ -0,0 +1,62 @@
...
+    validationClassName: PropTypes.string,
+    value: PropTypes.oneOfType([PropTypes.number, PropTypes.string])
```
picklesrus: Does the number type work? I tried passing 1 and 2 to my list of length 2 and it didn't change what was selected. Perhaps I'm using it wrong though? passing Using your example in comments above "AG" worked fine though.

benjiwheeler: You're right, I never added the code to set the selected value!

The number type shouldn't be the problem, though.

src/components/formik-forms/formik-select.jsx
benjiwheeler: Good idea, thanks!

src/components/formik-forms/formik-select.jsx
```diff
@@ -0,0 +1,62 @@
...
+        </div>
+    );
+};
+
+
+FormikSelect.propTypes = {
+    className: PropTypes.string,
+    error: PropTypes.string,
+    // expect structure like: [
+    //    {value: 'US', label: 'United States'}
```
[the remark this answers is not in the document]
benjiwheeler: Done

src/components/formik-forms/formik-select.jsx
```diff
@@ -0,0 +1,62 @@
...
+    const optionsList = options.map((item, index) => (
+        <option
+            key={index}
+            value={item.value}
+        >
+            {item.label}
+        </option>
+    ));
+    return (
+        <div className="col-sm-9 row row-relative">
```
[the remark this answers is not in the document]
benjiwheeler: You're right, removing

Issue #3129, opened 2019-07-11
comments need unique "key" prop


benjiwheeler:
If you view any project with comments, and look at the js console, you will see the error:

```
Warning: Each child in an array or iterator should have a unique "key" prop.

Check the render method of `Comment`. See https://fb.me/react-warning-keys for more information.
    in EmojiText (created by Comment)
    in Comment (created by TopLevelComment)
    in div (created by FlexRow)
    in FlexRow (created by TopLevelComment)
    in TopLevelComment (created by PreviewPresentation)
    in div (created by FlexRow)
    in FlexRow (created by PreviewPresentation)
    in div (created by PreviewPresentation)
    in div (created by FlexRow)
    in FlexRow (created by PreviewPresentation)
    in div (created by PreviewPresentation)
    in div (created by PreviewPresentation)
    in div (created by PreviewPresentation)
    in PreviewPresentation (created by InjectIntl(PreviewPresentation))
    in InjectIntl(PreviewPresentation) (created by Preview)
    in div (created by Page)
    in div (created by Page)
    in ErrorBoundary (created by Page)
    in Page (created by Preview)
    in Preview (created by Connect(Preview))
    in Connect(Preview)
    in IntlProvider
    in Provider
```

PR #3131, opened 2019-07-11
provide key to comments

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/3129

### Changes:

Provides `key` props for reply comments, and top level comments

### DIFF
--- a/src/views/preview/comment/comment.jsx
+++ b/src/views/preview/comment/comment.jsx
@@ -196,11 +196,12 @@ class Comment extends React.Component {
                           */}
 
                         <span className="comment-content">
-                            {commentText.map(fragment => {
+                            {commentText.map((fragment, index) => {
                                 if (typeof fragment === 'string') {
                                     return (
                                         <EmojiText
                                             as="span"
+                                            key={`comments-${id}-fragment-${index}`}
                                             text={fragment}
                                         />
                                     );


src/views/preview/comment/top-level-comment.jsx
```diff
@@ -96,6 +96,7 @@ class TopLevelComment extends React.Component {
             <FlexRow className="comment-container">
                 <Comment
                     highlighted={highlightedCommentId === id}
+                    key={`top-level-comment-${id}`}
```
benjiwheeler: wasn't sure if there's a good key naming convention to use... these are sorta inconsistent, but maybe that doesn't matter

benjiwheeler:
> @benjiwheeler my understanding of the key property is that it is only needed when you have an array of react components, which I think only applies to src/views/preview/comment/comment.jsx change and not the other one. Do you still see the warning if you remove the key property from top-level-comment?

Hm, you're right! Removed it.

benjiwheeler:
> @benjiwheeler this is fine, but figure it is worth reminding as a PSA: remember that using the `index` as a `key` should really be a last resort if there is no stable identifying information available for each part of an array, since the key is used for faster reordering, the array index does not help because it is not stable to reordering. If there is nothing you can use to identify, or if the order will never change (which appears to be the case here) it is fine to use the index.

Thanks for the reminder. Yeah, I couldn't see any other option here because the `commentText` fragments are not unique and don't have id's.

PR #3132, opened 2019-07-11
add ModalInnerContent component, use it in modals

benjiwheeler:
Redo of https://github.com/LLK/scratch-www/pull/3113, after it was reverted.

### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/3112

### Changes:

* renames the `.modal-content` style when it is used inside another `.modal-content` style; now we'll call that inside one `.modal-inner-content`.
* refactors uses of that inner modal-content style, so they use a component instead of using the style out of scope.

The problem in https://github.com/LLK/scratch-www/pull/3113 was that a `height: 100%` style was not inheriting the same 100% height as before, because I'd introduced a new wrapping `<div>`. Adding adding `height: 100%` to that wrapping `<div>` seemed to fix the problem.

### Testing

Didn't add any tests, but manually tested several modals to confirm they look the same as production:
* add to studio
* copy link
* report project
* tutorial card
* join flow

### DIFF
--- a/src/components/modal/base/modal-inner-content.scss
+++ b/src/components/modal/base/modal-inner-content.scss
@@ -1,0 +1,18 @@
+@import "../../../colors";
+@import "../../../frameless";
+
+/*
+necessary to prevent modal-inner-content's margins from
+unexpectedly collapsing with a parent
+*/
+.modal-inner-clear {
+    overflow: auto;
+    height: 100%;
+}
+
+.modal-inner-content {
+    box-sizing: border-box;
+    display: flex;
+    border-radius: 0;
+    flex-direction: column;
+}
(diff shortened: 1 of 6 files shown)


src/components/modal/base/modal-inner-content.scss
```diff
@@ -0,0 +1,18 @@
+@import "../../../colors";
+@import "../../../frameless";
+
+/*
+necessary to prevent modal-inner-content's margins from
+unexpectedly collapsing with a parent
+*/
+.modal-inner-clear {
+    overflow: auto;
+    height: 100%;
```
benjiwheeler: this is the line I had to add

PR #3133, opened 2019-07-12
Format validation; redo of PR 3116

benjiwheeler:
Retry of https://github.com/LLK/scratch-www/pull/3116, after it was reverted in https://github.com/LLK/scratch-www/pull/3125

### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/3115

### Changes:

* Adds `validationClassName` parameter to `FormikInput` component
* Set formatting of validations in join flow using `validationClassName`

Note that unlike in https://github.com/LLK/scratch-www/pull/3116, I'm _not_ changing the css I use to prevent margins from collapsing. After this PR, I will consider ways to prevent margin collapse without interfering with other modals' formatting.

### DIFF
--- a/src/components/formik-forms/formik-input.jsx
+++ b/src/components/formik-forms/formik-input.jsx
@@ -11,9 +11,10 @@ require('../forms/row.scss');
 const FormikInput = ({
     className,
     error,
+    validationClassName,
     ...props
 }) => (
-    <div className="col-sm-9 row">
+    <div className="col-sm-9 row row-relative">
         <Field
             className={classNames(
                 'input',
@@ -22,7 +23,10 @@ const FormikInput = ({
             {...props}
         />
         {error && (
-            <ValidationMessage message={error} />
+            <ValidationMessage
+                className={validationClassName}
+                message={error}
+            />
         )}
     </div>
 );
@@ -31,7 +35,8 @@ const FormikInput = ({
 FormikInput.propTypes = {
     className: PropTypes.string,
     error: PropTypes.string,
-    type: PropTypes.string
+    type: PropTypes.string,
+    validationClassName: PropTypes.string
 };
 
 module.exports = FormikInput;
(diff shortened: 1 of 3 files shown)


src/components/formik-forms/formik-input.jsx
```diff
@@ -11,9 +11,10 @@ require('../forms/row.scss');
 const FormikInput = ({
     className,
     error,
+    validationClassName,
     ...props
 }) => (
-    <div className="col-sm-9 row">
+    <div className="col-sm-9 row row-relative">
```
rschamp: What requires the row to be relative? Is it because it might have a validation message? Right now it's not clear to me from this why I would need this class or not if I were extending this or re-using it in the future. Can the class name refer to why it's needed (e.g. `has-validation` or something specific like that) so that's self-documenting?

rschamp:
Checking it out on scratch.ly it looks like there are some regressions in the styles related to this.

**Production**
image

**Staging**
image

benjiwheeler:
The overflow/clipping issue was from another PR; a fix for that is at https://github.com/LLK/scratch-www/pull/3140. I'll update this PR after that one goes in.

rschamp:
Sounds good — I still think that class name could be improved since it's not very semantic.

benjiwheeler:
> Sounds good — I still think that class name could be improved since it's not very semantic.

Yes, I'm thinking of `.row-with-tooltip` since the absolutely positioned notice can be a validation error or just a friendly FYI. What do you think?

rschamp:
Sounds good to me!

src/components/formik-forms/formik-input.jsx
benjiwheeler: Renamed -- LMK what you think.

benjiwheeler:
OK, updated and ready for review again. Updated look: no longer clipping.

image

image

PR #3135, opened 2019-07-15
update package-lock.json with newer google-libphonenumber

benjiwheeler:
Follow-on to https://github.com/LLK/scratch-www/pull/3094

Question about PRs like https://github.com/LLK/scratch-www/pull/3094, and this:

Now that we have package-lock.json tracked/checked in, if we make an update to a package.json line via greenkeeper, should we also make an update to package-lock.json?

If so, does that mean that the process of merging in greenkeeper PRs is more complicated now -- check out locally, npm install, maybe test the change, add package-lock.json changes, push to origin, merge? For ones that we don't typically test locally, will we still need to do all this, minus the local testing?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5438,9 +5438,9 @@
       }
     },
     "google-libphonenumber": {
-      "version": "1.0.21",
-      "resolved": "https://registry.npmjs.org/google-libphonenumber/-/google-libphonenumber-1.0.21.tgz",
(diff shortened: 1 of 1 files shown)


rschamp:
FYI here's the change that lets Greenkeeper update package-lock.json: https://github.com/LLK/scratch-www/pull/3134/commits/2b558efb1d1bd5df521542e96b91d970651645a7

benjiwheeler:
Got it, thanks!

PR #3137, opened 2019-07-15
Move Makefile into package.json, remove Docker config

rschamp:
Docker wasn't super usable in its current state and was one of the last things requiring the Makefile.

Since there are a lot more scripts, apply a naming scheme to group test subcommands.


rschamp:
@BryceLTaylor FYI I put this up with a `hotfix/` branch name so we can see if the changes affect (break) deployment.

Makefile
```diff
@@ -1,96 +0,0 @@
-ESLINT=./node_modules/.bin/eslint
-NODE= NODE_OPTIONS=--max_old_space_size=8000 node
-SCRATCH_DOCKER_CONFIG=./node_modules/.bin/docker_config.sh
-S3CMD=s3cmd sync -P --delete-removed --add-header=Cache-Control:no-cache,public,max-age=3600
-TAP=./node_modules/.bin/tap
-WATCH= NODE_OPTIONS=--max_old_space_size=8000 ./node_modules/.bin/watch
-WEBPACK= NODE_OPTIONS=--max_old_space_size=8000 ./node_modules/.bin/webpack
```
paulkaplan: it looks like the custom `max_old_space_size` param was dropped in this migration. /cc @picklesrus

Issue #3141, opened 2019-07-16
select style only applies to descendents


benjiwheeler:
CSS file `src/components/forms/select.scss` defines a `.select` class, with a `select` element selector nested under it. However, a `select` element with a `.select` class set on it should also receive the same style.

benjiwheeler:
Closing because there may be a workaround. But we might want to revisit this and revise many of the forms-related css files, which are unnecessarily specific about when to apply styles, and when not to.

PR #3142, opened 2019-07-16
make form select style apply not only to descendents

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/3141

### Changes:

makes form select style apply not only to `<select>` elements that are descendents of an element with the `.select` class, but also to elements that have the `.select` class applied to them directly.

### DIFF
--- a/src/components/forms/select.scss
+++ b/src/components/forms/select.scss
@@ -1,53 +1,51 @@
 @import "../../colors";
 @import "../../frameless";
 
-.select {
-    label {
-        font-weight: 500;
+.select label {
+    font-weight: 500;
+}
+
+select.select, .select select {
+    transition: border .5s ease;
+    margin-bottom: .75rem;
+    border: 1px solid $active-gray;
+    border-radius: 5px;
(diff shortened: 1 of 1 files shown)


benjiwheeler:
> Is there somewhere you want to use this? It's hard to review without context. 

Yeah, the place I want to use this is not checked in yet. I can wait for that, so you can see where the change will be useful.

> Ideally I think we'd want to use one or the other style.

I could try to structure things so they'll be able to use the existing style, but I think that way of defining styles is generally wrong! Why would I want to define a style that only applies to all children that are a particular element, but not to the element itself if it's not a child? It seems to me that this isn't an additional style, it's just a way of declaring the style that imposes fewer assumptions that amount to tech debt.

benjiwheeler:
Closing this. Might re-open once the join flow code that needs it is merged in.

rschamp:
I agree about having bare elements respond to the parent class. In GUI we're pretty good about having everything applied to a specific class name instead. So if you're refactoring it, I think it'd be good to move to that style for the existing components as well. e.g. `.select`, `.select-label`, `.select-option` etc.

PR #3143, opened 2019-07-16
CommonsChunkPlugin error when i run npm run build on win10

DaveWoo:
Error:  webpack.optimize.CommonsChunkPlugin has been removed, please use config.optimization.splitChunks instead.

Caused: webpack 4 version should be deprecated webpack.optimize.CommonsChunkPlugin, but the Chinese network applied the code, causing an error

### Resolves:
Add segment "optimization" in webpack.common.js or webpack.config.js

### Changes:
Add segment "optimization" in webpack.common.js or webpack.config.js


rschamp:
Hi @DaveWoo, it looks like you've reverted all the changes, so I'm going to close this. If you're having trouble building, please open an issue and maybe we can work out the problem there.

PR #3145, opened 2019-07-17
Add jest

BryceLTaylor:
### Changes:

This adds the Jest testing framework and example tests so we can start moving unit and integration tests into Jest instead of Tap.  It moves all of the old tests into legacy folders and updates the package json so that both sets of tests get run.

This will make it so we can easily start adding new tests using Jest.


package.json
```diff
@@ -4,13 +4,18 @@
   "description": "Standalone WWW client for Scratch",
   "scripts": {
     "start": "node ./dev-server/index.js",
-    "test": "npm run test:lint && npm run build && npm run test:tap",
+    "test": "npm run test:lint && npm run build && npm run test:unit:tap",
```
benjiwheeler: This should be `run test:unit`, right? So it'll run jest tests too?

package.json
```diff
@@ -4,13 +4,18 @@
...
     "test:lint": "eslint . --ext .js,.jsx,.json",
-    "test:smoke": "tap ./test/integration/smoke-testing/*.js --timeout=3600 --no-coverage -R classic",
-    "test:smoke:verbose": "tap ./test/integration/smoke-testing/*.js --timeout=3600 --no-coverage -R spec",
-    "test:smoke:sauce": "SMOKE_REMOTE=true tap ./test/integration/smoke-testing/*.js --timeout=60000 --no-coverage -R classic",
-    "test:tap": "tap ./test/{unit,localization}/*.js --no-coverage -R classic",
-    "test:coverage": "tap ./test/{unit,localization}/*.js --coverage --coverage-report=lcov",
+    "test:integration": "npm run test:integration:jest && npm run test:smoke",
+    "test:integration:jest": "jest ./test/integration/*.test.js",
```
benjiwheeler: I think we should have more general paths for most of these, but that could be a follow-on change by me or you

package.json
```diff
@@ -4,13 +4,18 @@
...
+    "test:integration:remote": "npm run test:integration:jest && npm run test:smoke:sauce",
+    "test:smoke": "tap ./test/integration-legacy/smoke-testing/*.js --timeout=3600 --no-coverage -R classic",
+    "test:smoke:verbose": "tap ./test/integration-legacy/smoke-testing/*.js --timeout=3600 --no-coverage -R spec",
+    "test:smoke:sauce": "SMOKE_REMOTE=true tap ./test/integration-legacy/smoke-testing/*.js --timeout=60000 --no-coverage -R classic",
+    "test:unit": "npm run test:unit:jest && npm run test:unit:tap",
+    "test:unit:jest": "jest ./test/unit/*.test.js && jest ./test/localization/*.test.js",
+    "test:unit:tap": "tap ./test/{unit-legacy,localization-legacy}/*.js --no-coverage -R classic",
+    "test:coverage": "tap ./test/{unit-legacy,localization-legacy}/*.js --coverage --coverage-report=lcov",
```
benjiwheeler: do we also want to run a jest coverage test, or is there no such thing?

package.json
```diff
@@ -115,6 +122,7 @@
     "sass-loader": "6.0.6",
     "scratch-gui": "0.1.0-prerelease.20190711011201",
     "scratch-l10n": "latest",
+    "selenium-webdriver": "3.6.0",
```
benjiwheeler: I know many other dependencies are set to the "bug" level of change (X.X.X), but aren't we basically ok with not stressing about those tiny version changes? Can we do (~3.6.0) instead, since we don't expect breaking changes below the major level (and tilde would be even more careful, since it would still bug us about minor changes)

test/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
benjiwheeler: What does this do? Config eslint for these dirs, right? I'm looking at the GUI one and don't understand the line:
```
    extends: ['scratch/react', 'scratch/es6'],

```
...is this doing something we should do here?

test/integration-legacy/selenium-helpers.js
```diff
@@ -0,0 +1,168 @@
+const webdriver = require('selenium-webdriver');
```
benjiwheeler: Why is this categorized as legacy?

test/integration/README.md
```diff
@@ -1,8 +1,6 @@
 # Requirements
```
benjiwheeler: Do we really want to have a separate integration test README?

package.json
```diff
@@ -60,6 +65,7 @@
     "babel-preset-react": "6.22.0",
     "bowser": "1.9.4",
     "cheerio": "1.0.0-rc.2",
+    "chromedriver": "75.1.0",
```
thisandagain: @paulkaplan I remember us moving this out on purpose at some point in the past. Can you review?

benjiwheeler:
Also, using this PR locally, I was able to add jest unit tests and got them to run!

package.json
BryceLTaylor: Yes.  I changed it briefly for an interstitial commit and forgot to change it back.  I'll add that back in in a moment.

package.json
BryceLTaylor: I think that should come in a later pr that includes tests organized into sub folders.

benjiwheeler: Great. Maybe make an issue to track that?

package.json
BryceLTaylor: In this case I just moved the selenium-webdriver package from the integration folder where it was in a separate package.json.  I didn't modify it at all in the move.  For what it's worth, the last time this package changed was two years ago.   A newer version has been in alpha for about that long and will likely come out as 4.0.0.

We could change it, but I'm not sure how much it'll matter.

benjiwheeler: OK

test/.eslintrc.js
BryceLTaylor: @rschamp worked with me on getting this part to work. I believe the linter needed the `env: {jest: true}` part to allow the test files to pass the linter.  Without it it has a lot of "test was used but never defined" and similar errors since we're using many functions defined in Jest.

 I didn't add the other parts from gui because they weren't necessary. They must pertain to other files in the gui tests.

test/integration-legacy/selenium-helpers.js
BryceLTaylor: There are currently two copies of selenium-helpers, one in the legacy and one in the current integration tests.  They're currently the same, but we may have different needs when using Jest so I allowed them to be separate.  I suspect these files will diverge as work is done on the integration tests.

test/integration/README.md
BryceLTaylor: I'm going to move the README information into the top level README where it will be more useful. I'll also add a little bit about Jest.

BryceLTaylor: I moved it

benjiwheeler:
FWIW, this LGTM!

package.json
BryceLTaylor: I filed an issue here: https://github.com/LLK/scratch-www/issues/3157

package.json
BryceLTaylor: I think this should be added in a follow up PR.  It's turning into a bigger project getting it set up than I originally thought and I'd like to get this in so we can start writing tests.  I filed an issue for this work: https://chromebookdb.com/chromebook/toshiba-chromebook-2-cb35-c3350/

PR #3156, opened 2019-07-18
updated several eslint-related packages, fixed lint errors

benjiwheeler:
### Resolves:

Resolves greenkeeper prs:
Resolves https://github.com/LLK/scratch-www/pull/3052
Resolves https://github.com/LLK/scratch-www/pull/2945
Resolves https://github.com/LLK/scratch-www/pull/2771
Resolves https://github.com/LLK/scratch-www/pull/2790

### Changes:

* updates eslint, babel-eslint, eslint-plugin-json, and eslint-plugin-react
* updates various pieces of code to fix new linter errors


src/components/modal/addtostudio/container.jsx
```diff
@@ -17,7 +17,7 @@ class AddToStudioModal extends React.Component {
         };
     }
 
-    componentWillUpdate () {
+    componentDidUpdate () {
```
benjiwheeler: Need to confirm this works

src/views/teacherwaitingroom/teacherwaitingroom.jsx
```diff
@@ -11,8 +11,8 @@ const render = require('../../lib/render.jsx');
 require('./teacherwaitingroom.scss');
 
 class TeacherWaitingRoom extends React.Component {
-    componentWillReceiveProps (nextProps) {
-        if (nextProps.approved) {
+    componentDidUpdate () {
```
benjiwheeler: need to confirm this works

src/components/modal/addtostudio/container.jsx
benjiwheeler: Tested, works fine

src/views/teacherwaitingroom/teacherwaitingroom.jsx
benjiwheeler: Added similar handling in constructor. I think this should actually be better than before, because we would not expect `componentWillReceiveProps` to ever be called!

benjiwheeler:
@rschamp One note: when I run `npm i`, I get the warnings:
```
npm WARN eslint-config-scratch@5.0.0 requires a peer of babel-eslint@^8.0.1 but none is installed. You must install peer dependencies yourself.
npm WARN eslint-config-scratch@5.0.0 requires a peer of eslint@^4.0 but none is installed. You must install peer dependencies yourself.
```
Seems like there's a version incompatibility. Do we need to change the peer dependencies of `eslint-config-scratch`? Does that need to happen in conjunction with these changes?

src/components/extension-landing/project-card.jsx
```diff
@@ -5,6 +5,7 @@ const ProjectCard = props => (
     <a
         className="project-card"
         href={props.cardUrl}
+        rel="noopener noreferrer"
```
rschamp: I don't think this is necessary in the case of internal links. Did eslint ask you to do this because of the `target="_blank"`?

src/components/modal/addtostudio/container.jsx
rschamp: Seems risky to do all of these lifecycle updates in a lint PR. I agree we should update these so we can upgrade React, but maybe in a separate PR. Instead I think we should set our React version for `eslint-plugin-react`, like gui does: https://github.com/LLK/scratch-gui/blob/8eee001c12d0b40790da684c228bab1b32695100/src/.eslintrc.js#L22

rschamp:
Re: eslint-config-scratch, yes I think we just need to update those peer dependencies there, maybe to say `>` those versions rather than within those major versions. For now we can ignore the warning.

src/components/modal/addtostudio/container.jsx
benjiwheeler: Hm, I tried this, but the setting doesn't seem to be being recognized by the linter.

I made the file look like:
```
module.exports = {
    extends: ['scratch', 'scratch/node'],
    plugins: ['json'],
    settings: {
        react: {
            version: '16.2' // Prevent 16.3 lifecycle method errors
        }
    }
};
```

...but when I run `npm run test:lint`, I see the warning:

```
Warning: React version not specified in eslint-plugin-react settings. See https://github.com/yannickcr/eslint-plugin-react#configuration .
```

benjiwheeler: Aah, I had to set it in `src/eslintrc.js` -- I had been setting it in the top-level `./eslintrc.js`

src/components/extension-landing/project-card.jsx
benjiwheeler: Yes, because of the target="_blank".

Is it easier/better to 
1) disable the rule in these cases; or
2) just use rel="noopener noreferrer" anyway? Does it hurt anything?

rschamp: It doesn't hurt anything unless it's to another of our internal html pages. In that case it would hide information from our own analytics, while not doing anything security-wise.  If they are internal links to PDFs, I don't think there's any reason not to.

So probably 1 in the case of internal links to html pages, 2 for PDFs.

According to https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md, this rule would not be triggered if it could tell that the link was internal/relative. There's an option to never enforce it if the link is dynamic, but I don't think there's any guarantee that we wouldn't supply a dynamic external link.

PR #3161, opened 2019-07-22
Add join flow country step

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

* adds `country-step.jsx` 
* adds several additional functions to `lib/country-data.js` so that we have the ability to put various country options at the top of the list, and the ability to rename some country labels

This is just a first draft, and has placeholder text and a placeholder image.

### Test Coverage:

Added tests of all the country data functions using jest

### DIFF
--- a/src/components/join-flow/country-step.jsx
+++ b/src/components/join-flow/country-step.jsx
@@ -11,21 +11,6 @@ const JoinFlowStep = require('./join-flow-step.jsx');
 
 require('./join-flow-steps.scss');
 
-/**
- * Return a list of country options to give to select
- * @param  {object} reactIntl      react-intl, used to localize strings
- * @return {object}                ordered set of county options select
- */
-const getCountryOptions = reactIntl => {
-    const processedCountryData = countryData.registrationCountryOptions;
-    processedCountryData.unshift({
-        disabled: true,
-        label: reactIntl.formatMessage({id: 'registration.selectCountry'}),
-        value: 'null'
-    });
-    return processedCountryData;
-};
-
 class CountryStep extends React.Component {
     constructor (props) {
         super(props);
@@ -34,6 +19,20 @@ class CountryStep extends React.Component {
             'validateForm',
             'validateSelect'
         ]);
+        this.countryOptions = [];
+    }
+    componentDidMount () {
+        this.setCountryOptions();
+    }
+    setCountryOptions () {
+        if (this.countryOptions.length === 0) {
+            this.countryOptions = [...countryData.registrationCountryOptions];
+            this.countryOptions.unshift({
+                disabled: true,
+                label: this.props.intl.formatMessage({id: 'registration.selectCountry'}),
+                value: 'null'
+            });
+        }
     }
     validateSelect (selection) {
         if (selection === 'null') {
@@ -49,7 +48,7 @@ class CountryStep extends React.Component {
         this.props.onNextStep(formData);
     }
     render () {
-        const countryOptions = getCountryOptions(this.props.intl);
+        this.setCountryOptions();
         return (
             <Formik
                 initialValues={{
@@ -89,7 +88,7 @@ class CountryStep extends React.Component {
                                     error={errors.country}
                                     id="country"
                                     name="country"
-                                    options={countryOptions}
+                                    options={this.countryOptions}
                                     validate={this.validateSelect}
                                     validationClassName="validation-full-width-input"
                                 />
--- a/src/components/join-flow/country-step.jsx
+++ b/src/components/join-flow/country-step.jsx
@@ -1,0 +1,113 @@
+const bindAll = require('lodash.bindall');
+const classNames = require('classnames');
+const React = require('react');
+const PropTypes = require('prop-types');
+import {Formik} from 'formik';
+const {injectIntl, intlShape} = require('react-intl');
+
+const countryData = require('../../lib/country-data');
+const FormikSelect = require('../../components/formik-forms/formik-select.jsx');
+const JoinFlowStep = require('./join-flow-step.jsx');
+
+require('./join-flow-steps.scss');
+
+/**
+ * Return a list of options to give to select
+ * @param  {object} reactIntl      react-intl, used to localize strings
+ * @param  {string} defaultCountry optional string of default country to put at top of list
+ * @return {object}                ordered set of county options formatted for frc select
+ */
+const getCountryOptions = reactIntl => {
+    const processedCountryData = countryData.registrationCountryOptions;
+    processedCountryData.unshift({
+        label: reactIntl.formatMessage({id: 'registration.selectCountry'}),
+        disabled: true,
+        value: 'null'
+    });
+    return processedCountryData;
+};
+
+class CountryStep extends React.Component {
+    constructor (props) {
+        super(props);
+        bindAll(this, [
+            'handleValidSubmit',
+            'validateForm',
+            'validateSelect'
+        ]);
+    }
+    validateSelect (selection) {
+        if (selection === 'null') {
+            return this.props.intl.formatMessage({id: 'form.validationRequired'});
+        }
+        return null;
+    }
+    validateForm () {
+        return {};
+    }
+    handleValidSubmit (formData, formikBag) {
+        formikBag.setSubmitting(false);
+        this.props.onNextStep(formData);
+    }
+    render () {
+        const countryOptions = getCountryOptions(this.props.intl);
+        return (
+            <Formik
+                initialValues={{
+                    country: 'null'
+                }}
+                validate={this.validateForm}
+                validateOnBlur={false}
+                validateOnChange={false}
+                onSubmit={this.handleValidSubmit}
+            >
+                {props => {
+                    const {
+                        errors,
+                        handleSubmit,
+                        isSubmitting
+                    } = props;
+                    return (
+                        <JoinFlowStep
+                            description={this.props.intl.formatMessage({id: 'registration.countryStepDescription'})}
+                            headerImgSrc="/images/hoc/getting-started.jpg"
+                            title={this.props.intl.formatMessage({id: 'general.joinScratch'})}
+                            waiting={isSubmitting}
+                            onSubmit={handleSubmit}
+                        >
+                            <div
+                                className={classNames(
+                                    'col-sm-9',
+                                    'row'
+                                )}
+                            >
+                                <FormikSelect
+                                    className={classNames(
+                                        'join-flow-select',
+                                        'join-flow-select-country',
+                                        {fail: errors.country}
+                                    )}
+                                    error={errors.country}
+                                    id="country"
+                                    name="country"
+                                    options={countryOptions}
+                                    validate={this.validateSelect}
+                                    validationClassName="validation-full-width-input"
+                                />
+                            </div>
+                        </JoinFlowStep>
+                    );
+                }}
+            </Formik>
+        );
+    }
+}
+
+CountryStep.propTypes = {
+    intl: intlShape,
+    onNextStep: PropTypes.func
+};
+
+const IntlCountryStep = injectIntl(CountryStep);
+
+module.exports = IntlCountryStep;
--- a/src/components/join-flow/country-step.jsx
+++ b/src/components/join-flow/country-step.jsx
@@ -12,16 +12,15 @@ const JoinFlowStep = require('./join-flow-step.jsx');
 require('./join-flow-steps.scss');
 
 /**
- * Return a list of options to give to select
+ * Return a list of country options to give to select
  * @param  {object} reactIntl      react-intl, used to localize strings
- * @param  {string} defaultCountry optional string of default country to put at top of list
- * @return {object}                ordered set of county options formatted for frc select
+ * @return {object}                ordered set of county options select
  */
 const getCountryOptions = reactIntl => {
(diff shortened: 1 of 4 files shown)


src/components/join-flow/country-step.jsx
```diff
@@ -0,0 +1,115 @@
...
+        formikBag.setSubmitting(false);
+        this.props.onNextStep(formData);
+    }
+    render () {
+        const countryOptions = getCountryOptions(this.props.intl);
+        return (
+            <Formik
+                initialValues={{
+                    country: 'null',
+                    birth_year: 'null'
```
picklesrus: should birth_year be here?

src/lib/country-data.js
```diff
@@ -12,6 +12,32 @@ module.exports.countryOptions = Object.keys(countries).map(code => ({
         return 1;
     });
 
+const overrideOptions = module.exports.overrideOptions = (countryNameOverrides, thisCountryOptions) => {
```
picklesrus: At first glance, it is hard for me to tell how to use these. i.e. What kind of object/list do I pass as countryOverrides?
Also, What does "thisCountryOptions" mean?  Similar question below about commonCountryCodes and thisCountryOptions. It looks like the thisCountryOptions is optional?  Could you add some documentation?

src/lib/country-data.js
```diff
@@ -12,6 +12,32 @@ module.exports.countryOptions = Object.keys(countries).map(code => ({
...
+    const newCountryOptions = thisCountryOptions.map(item => ({...item}));
+    commonCountryCodes.reverse().forEach(commonCountryCode => {
+        const existing = newCountryOptions
+            .find(countryOption => countryOption.value === commonCountryCode.toLowerCase());
+        if (existing) newCountryOptions.unshift(existing);
+    });
+    return newCountryOptions;
+};
+
+module.exports.processedCountryOptions =
```
picklesrus: Same here. What does "processed" mean? Looks like the code does something very specific, but has a very generic name.

test/unit/lib/country-data.test.js
```diff
@@ -0,0 +1,62 @@
...
+        expect(countryOptionsWithCommon.length).toEqual(251);
+        expect(countryOptionsWithCommon[0]).toEqual({value: 'ca', label: 'Canada'});
+        expect(countryOptionsWithCommon[40]).toEqual({value: 'ca', label: 'Canada'});
+        expect(countryOptionsWithCommon[1]).toEqual({value: 'gb', label: 'United Kingdom'});
+        expect(countryOptionsWithCommon[234]).toEqual({value: 'gb', label: 'United Kingdom'});
+    });
+
+    test('processedCountryOptions', () => {
+        expect(typeof processedCountryOptions).toBe('object');
+        // const processedCountryOptionsObj = processedCountryOptions();
```
picklesrus: remove this line?

test/unit/lib/country-data.test.js
```diff
@@ -0,0 +1,62 @@
...
+        expect(countryOptions[38]).toEqual({value: 'ca', label: 'Canada'});
+        expect(overriddenCountryOptions[38]).toEqual({value: 'ca', label: 'Canadia'});
+        expect(countryOptions[232]).toEqual({value: 'gb', label: 'United Kingdom'});
+        expect(overriddenCountryOptions[232]).toEqual({
+            value: 'gb',
+            label: 'The United Kingdom of Great Britain and Northern Ireland'
+        });
+    });
+
+    test('commonOptions', () => {
```
picklesrus: Might be helpful to have messages on some of these expects to explain what you're testing. 
I know you're trying to get CA and GB to show up at the top and in alphabetical order, but that isn't apparent just from reading the code.

src/components/join-flow/country-step.jsx
```diff
@@ -0,0 +1,115 @@
...
+        ]);
+    }
+    validateSelect (selection) {
+        if (selection === 'null') {
+            return this.props.intl.formatMessage({id: 'form.validationRequired'});
+        }
+        return null;
+    }
+    validateForm () {
+        return {};
```
picklesrus: IIRC, In the current we actually validate the country you choose against the list of acceptable ones. Does that belong here or does that happen server side in scratchr2 as part of create user?

src/components/join-flow/country-step.jsx
picklesrus: Remove birth_year?

src/components/join-flow/country-step.jsx
```diff
@@ -0,0 +1,115 @@
...
+                            headerImgSrc="/images/hoc/getting-started.jpg"
+                            title={this.props.intl.formatMessage({id: 'general.joinScratch'})}
+                            waiting={isSubmitting}
+                            onSubmit={handleSubmit}
+                        >
+                            <div
+                                className={classNames(
+                                    'col-sm-9',
+                                    'row',
+                                    'birthdate-select-row'
```
picklesrus: I think this css is unused (and leftover from the other step).

src/components/join-flow/country-step.jsx
```diff
@@ -0,0 +1,115 @@
...
+                                )}
+                            >
+                                <FormikSelect
+                                    className={classNames(
+                                        'join-flow-select',
```
picklesrus: same here about css

src/components/join-flow/country-step.jsx
benjiwheeler: No!!!

test/unit/lib/country-data.test.js
```diff
@@ -0,0 +1,62 @@
...
+    overrideOptions,
+    commonOptions,
+    processedCountryOptions,
+    subdivisionOptions
+} = require('../../../src/lib/country-data');
+
+describe('unit test lib/country-data.js', () => {
+    test('countryOptions', () => {
+        expect(typeof countryOptions).toBe('object');
+        expect(countryOptions.length).toEqual(249);
```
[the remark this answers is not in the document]
benjiwheeler: I think that's ok -- it changes approximately once every 5 years, and we'll see in that greenkeeper PR that it's changed, and can update the tests.

src/lib/country-data.js
benjiwheeler: i can give them comments explaining what they do.

src/lib/country-data.js
benjiwheeler: yeah, i'm not sure what to call this. it's like, the version of country options that we've messed with so it's what we want. ideas?

test/unit/lib/country-data.test.js
benjiwheeler: Will do.

test/unit/lib/country-data.test.js
benjiwheeler: will do

src/components/join-flow/country-step.jsx
benjiwheeler: hm, not sure what you mean about "in the current". you mean, scratchr2 does that? where does it do that?

src/components/join-flow/country-step.jsx
benjiwheeler: yes, will do!

src/components/join-flow/country-step.jsx
benjiwheeler: ulp, thanks for catching

src/components/join-flow/country-step.jsx
benjiwheeler: Will do

src/components/join-flow/country-step.jsx
benjiwheeler: Removed.

src/lib/country-data.js
benjiwheeler: I'm trying "registrationCountryOptions"

src/lib/country-data.js
benjiwheeler: Done

test/unit/lib/country-data.test.js
benjiwheeler: ok, i clarified these a lot. take a look!

test/unit/lib/country-data.test.js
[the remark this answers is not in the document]
benjiwheeler: i think you're right that it'd be better not to be so specific about matching iso-3166-2 precisely. the bigger issue is, are the lengths at least in the right ballpark, and if we add two entries, is the length two greater than before? I made a bunch of changes reflecting this.

benjiwheeler:
How this looks now:

image

image

image

src/components/join-flow/country-step.jsx
benjiwheeler: If you're talking about checking the value the user provides against this list, maybe that's something we can do later, before submitting?

picklesrus: Yeah. I'm not sure where it is or if we do it, but I got that impression from a conversation we had in a meeting the other day.  How about putting in a TODO to figure this out later?

src/components/join-flow/country-step.jsx
```diff
@@ -0,0 +1,113 @@
...
+            >
+                {props => {
+                    const {
+                        errors,
+                        handleSubmit,
+                        isSubmitting
+                    } = props;
+                    return (
+                        <JoinFlowStep
+                            description={this.props.intl.formatMessage({id: 'registration.countryStepDescription'})}
```
apple502j: l1on.json missing this key

benjiwheeler: Thanks for noticing. Because we aren't activating this code for users yet, it's ok that it has strings that are missing. We'll add them later.

src/lib/country-data.js
```diff
@@ -26,3 +59,66 @@ module.exports.subdivisionOptions = Object.keys(countries).reduce((subByCountry,
...
+ * Take startingCountryOptions, and override any number of its country labels.
+ * E.g., to use the German and Italian names of Germany and Italy, do:
+ *     overrideLabels({de: 'Deutschland', it: 'Italia'})
+ * @param {array} startingCountryOptions array of country objects
+ *     like [{value: 'ca', label: 'Canada'}, ...]
+ * @param {object} countryNameOverrides map of country codes we should override,
+ *     to labels to use instead of default
+ * @returns {array} revised array of country objects, with overridden labels
+ */
+const overrideLabels = module.exports.overrideLabels = (startingCountryOptions, countryNameOverrides) => {
```
picklesrus: I don't see this one used outside the module itself. Maybe it shouldn't be exported?

src/lib/country-data.js
```diff
@@ -26,3 +59,66 @@ module.exports.subdivisionOptions = Object.keys(countries).reduce((subByCountry,
...
+ * takes startingCountryOptions, and duplicates any number of its country labels
+ * at the beginning of the array.
+ * E.g., to have "Germany" and "Italy" at the top of the list, use:
+ *     dupeCommonOptions(['de', 'it')
+ * @param {array} startingCountryOptions array of country objects
+ *     like [{value: 'ca', label: 'Canada'}, ...]
+ * @param {array} commonCountryCodes array of country codes we should duplicate
+ * @returns {array} revised array of country objects, with duplicates inserted at beginning
+ */
+const dupeCommonOptions = module.exports.dupeCommonOptions = (startingCountryOptions, commonCountryCodes) => {
```
picklesrus: Same comment on this one about exporting.

test/unit/lib/country-data.test.js
```diff
@@ -0,0 +1,108 @@
...
+        expect(typeof countryOptions).toBe('object');
+        expect(countryOptions.length > 200).toEqual(true);
+        expect(countryOptions.length < 300).toEqual(true);
+        const sortedCountryOptions = countryOptions.sort((a, b) => {
+            if (a.label < b.label) {
+                return -1;
+            }
+            return 1;
+        });
+        expect(countryOptions.map(option => option.label))
```
picklesrus: With this last test, are you trying to test that re-sorting doesn't change the order to prove that it is already sorted?

src/lib/country-data.js
benjiwheeler: it needs to be exported to be unit tested.

src/lib/country-data.js
benjiwheeler: it needs to be exported to be unit tested.

test/unit/lib/country-data.test.js
benjiwheeler: yes. i'll add a comment to make that clearer.

benjiwheeler:
> Also, sounds like you've looked into country mappings as they go into the DB. Are we good on that front or is there more work to do after this PR? If so, could you add a TODO?

Yeah, I'm still not sure which source we should use for these, and I want to get clearer before I merge this.

picklesrus:
Ben, assigning this one back to you since there's some stuff about the country list to work out.

src/components/join-flow/country-step.jsx
benjiwheeler: I think that's beyond the scope of the front end -- I think any validation that the country the front end provided is valid belongs in the back end. It would be different if we were allowing the user to type out their own country string.

benjiwheeler:
OK, I think everything should be worked out now!

* now, this uses the new country data that ray and I redid
* the "Email address" string is now localized
* I also redid the way we generate the list in the select element, so it only happens once

PR #3178, opened 2019-07-26
Join flow gender step

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

* add gender-step to join flow
* add formik-radio-button component
* revise formik-input component so it can be included inline
* handle values typed in to the input
* default gender choice to "Prefer not to say"
* add relevant strings

### Test Coverage:

Not sure how to test this -- it's just components, without a lot of logic.


src/components/formik-forms/formik-radio-button.jsx
```diff
@@ -0,0 +1,106 @@
...
+const PropTypes = require('prop-types');
+const React = require('react');
+import {Field} from 'formik';
+
+const FormikInput = require('./formik-input.jsx');
+
+require('./formik-radio-button.scss');
+require('../forms/row.scss');
+
+const FormikRadioButtonSubComponent = ({
```
picklesrus: Question for my own understanding of React/Formik: How does this get props?  I see it passed as component=FormikRadioButtonSubComponent on the Field below. Is the formik field doing that work, or is is some other part of the React component system?

src/components/join-flow/gender-step.jsx
```diff
@@ -0,0 +1,145 @@
...
+                    gender: 'null',
+                    other: ''
+                }}
+                onSubmit={this.handleValidSubmit}
+            >
+                {props => {
+                    const {
+                        handleSubmit,
+                        isSubmitting,
+                        setFieldValue,
```
picklesrus: Silly React newbie question: Where is setFieldValue sent in here?  i.e.  below in each <GenderOption>, there's the onSetFieldValue={setFieldValue} line. Where is the actual definition  of that?

benjiwheeler: Short answer: `<Formik>` assigns a `props` object, which we use to unpack whichever values we want to use. The possible values we can use are listed here: https://jaredpalmer.com/formik/docs/api/formik

Long answer: This is one of the magical (I mean that in a bad way) things that js can do, that formik takes advantage of. Basically, when we say `<Formik>{some stuff here}</Formik>`, react passes the `{some stuff here}` to the `Formik` component as the `children` prop. As I (shakily) understand it, `Formik` expects `children` to be a function, and in its render function it calls `{children({props object here})}`. So the `Formik` component actually renders all this `{some stuff here}`, NOT the `GenderStep` component itself.

Confusing!

src/components/forms/row.scss
```diff
@@ -30,3 +30,13 @@
     margin-bottom: .75rem;
     line-height: 1.7rem;
 }
+
+.row-inline {
```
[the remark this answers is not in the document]
benjiwheeler: Yeah, I'm looking forward to later when we'll pull out all the formsy stuff and convert it to use formik, but will still want to use its css. I'd like both sets of forms components to use the same css as much as possible now, because that'll mean less sorting it all out later. So rather than make a separate formik-forms/row.scss file, I'm trying to treat this file like it applies to www forms in general.

Does that make sense? I'm not super confident about this approach, that's just my reasoning so far.

src/components/formik-forms/formik-radio-button.jsx
```diff
@@ -0,0 +1,106 @@
...
+                /* eslint-enable react/jsx-no-bind */
+            />
+        )}
+    </Field>
+);
+
+FormikRadioButton.propTypes = {
+    buttonValue: PropTypes.string,
+    className: PropTypes.string,
+    isOther: PropTypes.bool,
```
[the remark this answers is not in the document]
benjiwheeler: Good idea, will do.

src/components/formik-forms/formik-radio-button.jsx
```diff
@@ -0,0 +1,106 @@
...
+        component={FormikRadioButtonSubComponent}
+        id="radioOption1"
+        label={label}
+        labelClassName={isOther ? 'formik-radio-label-other' : ''}
+        {...props}
+    >
+        {isOther && (
+            <FormikInput
+                className="formik-radio-input"
+                id="other"
```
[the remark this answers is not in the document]
benjiwheeler: `FormikInput` does something special with id -- it maps it to one of the `Formik` form's values that it tracks.

The problem is that we need to specifically name and initialize all the values that it tracks... so i think we're stuck with just one user-input value per formik radio button group.

src/components/formik-forms/formik-radio-button.jsx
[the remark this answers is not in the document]
benjiwheeler: Done, take a look!

src/components/formik-forms/formik-radio-button.jsx
```diff
@@ -0,0 +1,106 @@
...
+    isOther,
+    label,
+    onSetOther,
+    ...props
+}) => (
+    <Field
+        buttonValue={buttonValue}
+        className={className}
+        component={FormikRadioButtonSubComponent}
+        id="radioOption1"
```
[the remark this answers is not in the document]
benjiwheeler: Hm, you're right. I was able to pull out the ids here and in the "other" field, and everything still works! I guess those aren't needed after all!

src/components/formik-forms/formik-radio-button.jsx
benjiwheeler: Yes, the `Formik` component passes a bunch of props to the custom component that you tell it to use for radio buttons. 

From the docs, https://jaredpalmer.com/formik/docs/api/field:

```
Custom React Components will be passed FieldProps which is same render prop parameters of <Field render> plus any other props passed to directly to <Field>.
```

src/components/formik-forms/formik-radio-button.jsx
```diff
@@ -0,0 +1,107 @@
...
+            onChange={field.onChange} /* eslint-disable-line react/jsx-handler-names */
+            {...props}
+        />
+        {label && (
+            <label
+                className={classNames(
+                    'formik-radio-label',
+                    labelClassName
+                )}
+                htmlFor={buttonValue}
```
paulkaplan: @benjiwheeler sorry for the super random comment, and I know this was already merged, but I think this isn't the correct use of the `htmlFor` attribute. `htmlFor` is the react version of the `for` attribute, which only refers to inputs by `id` [(src)](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#Attributes). The result of this (I think) would be that clicking on the label will not toggle the radio button the way it is should. You could probably use `field.name` for the id and htmlFor instead?

benjiwheeler: Not random, and thanks for pointing this out! I confess to some sloppy cut and paste, and should have learned what the heck `htmlFor` does! :) Fixing this in https://github.com/LLK/scratch-www/pull/3250 .

PR #3181, opened 2019-07-26
better validation positioning on birthdate step

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

* use `row-with-tooltip` class in formik-select component, so validations will be correctly positioned
* tweak css transform in validations so they are positioned in a way that will correspond to the select elements they refer to, without blocking user from clicking on the selects
* tweak css classes when window is narrow, so validations are no longer transformed to the right

### Before:

when window is wide:

image

when window is narrow:

image

### After:

image

image

image

### DIFF
--- a/src/components/join-flow/join-flow-steps.scss
+++ b/src/components/join-flow/join-flow-steps.scss
@@ -22,6 +22,23 @@
     transform: translate(21.5625rem, 0);
 }
 
+.validation-birthdate-input {
+    transform: translate(8.75rem, .25rem);
+    width: 7.25rem;
+}
+
+@media #{$intermediate-and-smaller} {
+    .validation-full-width-input {
+        transform: unset;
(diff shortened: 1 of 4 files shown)


src/components/join-flow/join-flow-steps.scss
```diff
@@ -22,6 +22,23 @@
     transform: translate(21.5625rem, 0);
 }
 
+.validation-birthdate-input {
+    transform: translate(8.75rem, .25rem);
```
picklesrus: What are these numbers based on? Is it the length of some text? If so, this won't be right in other languages.

benjiwheeler: That's a great question! It's based on the fixed width of the select form elements, which is actually set below in `select.join-flow-select`. So it should not be language-dependent.

PR #3184, opened 2019-07-30
added stubs for email step, welcome step

benjiwheeler:
### Resolves:

Steps towards resolving #3053

### Changes:

* introduced stub components for join flow's email entry step, welcome step
* pass username to welcome step, so it can use it in title
* add several strings for these steps' language


src/l10n.json
```diff
@@ -152,6 +152,8 @@
     "registration.confirmYourEmail": "Confirm Your Email",
     "registration.confirmYourEmailDescription": "If you haven't already, please click the link in the confirmation email sent to:",
     "registration.createUsername": "Create a username",
+    "registration.emailStepTitle": "What's your email?",
```
apple502j: This is not good, because for children under 16 they need parent's email. In that case this should be `Ask your parents or guardians to fill it`

src/l10n.json
```diff
@@ -152,6 +152,8 @@
...
+    "registration.emailStepDescription": "We need this to finish creating your account. Scratch will always keep this information private.",
```
apple502j: It should also note that
1) the email must be accurate and must not be temporary
2) there is an email confirm before sharing the first project
3) the email is used for password reset

src/l10n.json
thisandagain: Thanks for the feedback! After careful consideration and discussion with our legal team we have decided to use this language for the email field.

src/components/join-flow/email-step.jsx
```diff
@@ -0,0 +1,63 @@
...
+    }
+    handleValidSubmit (formData, formikBag) {
+        formikBag.setSubmitting(false);
+        this.props.onNextStep(formData);
+    }
+    render () {
+        return (
+            <Formik
+                initialValues={{
+                    birth_month: 'null',
```
picklesrus: Remove these (might be why your values are null on the last page in the demo?)

src/components/join-flow/welcome-step.jsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+    handleValidSubmit (formData, formikBag) {
+        formikBag.setSubmitting(false);
+        this.props.onNextStep(formData);
+    }
+    render () {
+        return (
+            <Formik
+                initialValues={{
+                    birth_month: 'null',
```
picklesrus: Remove these too

src/components/join-flow/email-step.jsx
benjiwheeler: Aahhh i see

src/l10n.json
benjiwheeler: These are good suggestions, and we'll consider them in the future when we have the chance to take another look at the language here. Thanks!

PR #3186, opened 2019-07-30
Join flow add images

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

Adds images to birthdate, country, email and welcome steps


benjiwheeler:
> Have the pngs been run through an optimizer (e.g. pngcrush)?

No, great idea! Done.

benjiwheeler:
> Looks like one the test broke :( Maybe because the CSS class changed names? Maybe there's a better element to look for?

Yeah, I moved the class, and needed to update the test to match. Done!

PR #3199, opened 2019-07-31
Consolidate country lists, fix country field in teacher account registration

benjiwheeler:
### Resolves:

* Step towards resolving https://github.com/LLK/scratchr2/issues/5354
* Step towards resolving #3053
* Closes #3058 
* Partially fixes LLK/scratchr2#5321, as explained in #3058

### Changes:

* creates `lib/country-list.js`, which is a static list based on the list we have used in scratchr2 registration since 2013
* adds several functions that enable creation of select options from this country list, using display version of country names as labels, and either country codes or the country names already in the db as values
* adds tests for country-data and country-list

### Test Coverage:

added unit tests for all functions in country-data and country-list


src/components/registration/steps.jsx
```diff
@@ -948,6 +946,7 @@ class AddressStep extends React.Component {
     render () {
         let stateOptions = countryData.subdivisionOptions[this.state.countryChoice];
         stateOptions = [{}].concat(stateOptions);
+        const countryOptions = getCountryOptions(this.props.intl);
```
benjiwheeler: this change isn't really necessary but i think it makes the code more readable and consistent

src/components/registration/steps.jsx
```diff
@@ -33,29 +33,20 @@ require('./steps.scss');
 const DEFAULT_COUNTRY = 'us';
```
benjiwheeler: note that now, the main function of DEFAULT_COUNTRY has moved into the country-data library itself

src/lib/country-data.js
```diff
@@ -1,22 +1,1111 @@
...
+ * E.g., to have "Germany" and "Italy" at the top of the list, use:
+ *     dupeCommonCountries(['de', 'it')
+ * @param {array} startingCountryInfo array of country objects
+ *     like [{value: 'ca', label: 'Canada'}, ...]
+ * @param {array} commonCountryCodes array of country codes we should duplicate
+ * @returns {array} revised array of country objects, with duplicates inserted at beginning
+ */
+const dupeCommonCountries = module.exports.dupeCommonCountries = (startingCountryInfo, commonCountryCodes) => {
+    const newCountryInfo = startingCountryInfo.map(item => ({...item}));
+    commonCountryCodes.reverse().forEach(commonCountryCode => {
```
rschamp: Why does this get reversed?

src/components/registration/steps.jsx
```diff
@@ -445,6 +436,13 @@ class DemographicsStep extends React.Component {
         this.setState({otherDisabled: gender !== 'other'});
     }
     handleValidSubmit (formData) {
+        // look up country name using user's country code selection
+        if (formData.countryCode) {
+            const countryInfo = countryData.lookupCountryInfo(formData.countryCode);
+            if (countryInfo) {
+                formData.user.country = countryInfo.name;
+            }
```
rschamp: If `countryInfo` is not found, then I think we haven't reached a valid state yet — that should be validated somewhere before this point.  If we do get to this point (where there is no `countryInfo`), then we should enter into an error state.

src/components/registration/steps.jsx
```diff
@@ -445,6 +436,13 @@ class DemographicsStep extends React.Component {
         this.setState({otherDisabled: gender !== 'other'});
     }
     handleValidSubmit (formData) {
+        // look up country name using user's country code selection
+        if (formData.countryCode) {
```
rschamp: Similarly, we should validate that there's a `countryCode` before this point, and if it's not supplied, enter into an error state.

src/components/registration/steps.jsx
rschamp: Do you mean because the library supplies a list with several "default" countries at the top?

benjiwheeler: Yes. So we no longer use DEFAULT_COUNTRY to sort a country to the top in this file.

src/components/registration/steps.jsx
benjiwheeler: Yes, this makes sense. Let me think about what sort of error state makes sense...

src/components/registration/steps.jsx
benjiwheeler: OK, now it should pop up a validation error before it gets to this point, and if it doesn't find a country name, it won't call `onNextStep`.

src/components/registration/steps.jsx
benjiwheeler: OK, now it should pop up a validation error if it doesn't find a country name.

src/lib/country-data.js
benjiwheeler: it's just the way i'm getting the resulting array to have the same order as the `commonCountryCodes` -- I reverse the list, then prepend the big array with each "common" country entry, one at a time.

But your question is making me think of a more straightforward way of handling this, like:

```
    const commonCountriesInfo = commonCountryCodes.map(commonCountryCode => (
        startingCountryInfo.find(country => country.code === commonCountryCode.toLowerCase())
    ));
    return [...commonCountriesInfo, ...startingCountryInfo];
```
...simpler, right?

PR #3200, opened 2019-07-31
Add Enzyme to www and add a test for the NextStepButton. This takes o…

picklesrus:
…ur enzyme helpers from gui and copies them here. It also requires a babel plugin to play nicely with jest.

### Test Coverage:
A small one was added, but now there can be more!


package.json
```diff
@@ -61,6 +61,7 @@
     "babel-eslint": "10.0.2",
     "babel-loader": "7.1.0",
     "babel-plugin-transform-object-rest-spread": "6.26.0",
+    "babel-plugin-transform-require-context": "^0.1.1",
```
picklesrus: should I pin this?

benjiwheeler:
This is exciting! I don't get a bunch of the config stuff but I trust it's necessary! :)

test/unit/components/next-step-button.test.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import {mountWithIntl} from '../../helpers/intl-helpers.jsx';
+import NextStepButton from '../../../src/components/join-flow/next-step-button';
```
benjiwheeler: Travis is complaining that React and NextStepButton aren't used. But obviously NextStepButton is used! Not sure what's necessary to get the linter to see this. I was seeing linting errors make a similar mistake before, but it seemed to clear up when the lint versions were updated.

picklesrus: We didn't quite have the right react eslint config and my file was a .js file instead of a .jsx file.... Fixed now though.

package.json
rschamp: Looks like in this repo we pin the babel stuff, so yes — we may as well keep that consistent.  (Since they are usually pretty good about semver and also release super frequently, we usually use `^` in other repos.)
```suggestion
    "babel-plugin-transform-require-context": "0.1.1",
```

PR #3204, opened 2019-08-01
formatted email step, added input

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

* formats email step further
* adds email input
* revises `FormikInput` to take a placeholder class

### Test Coverage:

none


src/components/join-flow/email-step.jsx
```diff
@@ -13,9 +16,18 @@ class EmailStep extends React.Component {
...
             'handleValidSubmit',
+            'validateEmailIfPresent',
             'validateForm'
         ]);
     }
+    validateEmailIfPresent (email) {
+        if (!email) return null; // skip validation if email is blank; null indicates valid
+        const localResult = isEmail.validate(email);
+        if (localResult) {
+            return null; // TODO: validate email address remotely
```
apple502j: Maybe you should check `.gov.xx`, `.go.xx` or `.gov` because no governments want spam mail & there will be no use

src/components/join-flow/email-step.jsx
```diff
@@ -13,9 +16,18 @@ class EmailStep extends React.Component {
         super(props);
         bindAll(this, [
             'handleValidSubmit',
+            'validateEmailIfPresent',
             'validateForm'
         ]);
     }
+    validateEmailIfPresent (email) {
+        if (!email) return null; // skip validation if email is blank; null indicates valid
+        const localResult = isEmail.validate(email);
```
picklesrus: maybe rename?  this variable represents whether it is valid, right?  I think isValid would be a bit more readable.

package.json
```diff
@@ -89,6 +89,7 @@
     "glob": "5.0.15",
     "google-libphonenumber": "3.2.3",
     "html-webpack-plugin": "2.22.0",
+    "isemail": "2.2.1",
```
picklesrus: Curious why you chose this version and not the most recent. Though maybe you'll have to change this because travis is unhappy?

benjiwheeler: Yeah, i was trying to get Travis to be happy, but it turns out it's not Travis's fault! I'm going to try a different package instead.

src/components/join-flow/email-step.jsx
benjiwheeler: Done

package.json
benjiwheeler: Used another package, working better!

src/components/join-flow/email-step.jsx
```diff
@@ -35,17 +47,35 @@ class EmailStep extends React.Component {
...
+                            <FormikInput
+                                className={classNames(
+                                    'join-flow-input',
+                                    'join-flow-input-tall',
+                                    {fail: errors.email}
+                                )}
+                                error={errors.email}
+                                id="email"
+                                name="email"
+                                placeholder="Email address"
```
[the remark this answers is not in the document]
benjiwheeler: The "Email address" string is now localized!

Issue #2550, opened 2018-12-28
Project reports do not include the thumbnail


paulkaplan:
The code handling project reports always creates a blank white thumbnail (source: https://github.com/LLK/scratch-www/blob/develop/src/redux/preview.js#L855-L858)

Warning: this might be a bit complex to implement since the GUI is the one who is in control of the the thumbnail stuff. I also do not fully understand when project reports include thumbnails and when they do not.

paulkaplan:
@rschamp from looking at the default thumbnail data we include in the `reportProject` reducer action [src](https://github.com/LLK/scratch-www/blob/develop/src/redux/preview.js#L964-L967) it seems like maybe the problem is that we are including the `data:image` prefix in the POST instead of just the b64 encoded image. That is, we might be able to fix just by stripping `data:image/png;base64,` from the thumbnail we get back from the GUI.

thisandagain:
Resolved in #3211

paulkaplan:
For future generations, this bug was the focus of **The Bug Report Episode 2: Thumbnail Holy Grail**

PR #3210, opened 2019-08-02
Fix project report thumbnailer

paulkaplan:
Fixes https://github.com/LLK/scratch-www/issues/2550

From that comment:
>  from looking at the default thumbnail data we include in the reportProject reducer action https://github.com/LLK/scratch-www/blob/develop/src/redux/preview.js#L964-L967 it seems like maybe the problem is that we are including the data:image prefix in the POST instead of just the b64 encoded image. That is, we might be able to fix just by stripping data:image/png;base64, from the thumbnail we get back from the GUI.

### DIFF
--- a/src/views/preview/project-view.jsx
+++ b/src/views/preview/project-view.jsx
@@ -388,7 +388,10 @@ class Preview extends React.Component {
         const submit = data => this.props.reportProject(this.state.projectId, data, this.props.user.token);
         if (this.getProjectThumbnail) {
             this.getProjectThumbnail(thumbnail => {
-                const data = Object.assign({}, formData, {thumbnail});
+                const data = Object.assign({}, formData, {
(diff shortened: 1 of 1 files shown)


src/views/preview/project-view.jsx
```diff
@@ -388,7 +388,10 @@ class Preview extends React.Component {
         const submit = data => this.props.reportProject(this.state.projectId, data, this.props.user.token);
         if (this.getProjectThumbnail) {
             this.getProjectThumbnail(thumbnail => {
-                const data = Object.assign({}, formData, {thumbnail});
+                const data = Object.assign({}, formData, {
+                    // Strip the data:image prefix, server just wants the b64 encoded image
+                    thumbnail: thumbnail.replace('data:image/png;base64,', '')
```
rschamp: I think we still need to confirm that we don't have to strip trailing `=` padding.

PR #3216, opened 2019-08-05
Add unittest for JoinFlowStep.

picklesrus:
### Changes:

Now that we have enzyme, add unittest tests for join flow component.

### Test Coverage:

That's all it does :D

### DIFF
--- a/test/unit/components/join-flow-step.test.jsx
+++ b/test/unit/components/join-flow-step.test.jsx
@@ -4,7 +4,7 @@ import JoinFlowStep from '../../../src/components/join-flow/join-flow-step';
 
 describe('JoinFlowStep', () => {
 
-    test('test components exist when props present', () => {
+    test('components exist when props present', () => {
         const props = {
             children: null,
             className: 'join-flow-step-class',
@@ -33,15 +33,27 @@ describe('JoinFlowStep', () => {
         expect(component.find('div.join-flow-description').text()).toEqual(props.description);
 
 
-        component.find('button[type="submit"]').simulate('submit');
-        expect(props.onSubmit).toHaveBeenCalled();
-
         expect(component.find('NextStepButton').prop('waiting')).toEqual(true);
         expect(component.find('NextStepButton').prop('content')).toEqual(props.nextButton);
+
         component.unmount();
     });
 
-    test('test components do not exist when props not present', () => {
+    test('components do not exist when props not present', () => {
+        const component = mountWithIntl(
+            <JoinFlowStep />
+        );
+
+        expect(component.find('div.join-flow-header-image').exists()).toEqual(false);
+        expect(component.find('.join-flow-inner-content').exists()).toEqual(true);
+        expect(component.find('.join-flow-title').exists()).toEqual(false);
+        expect(component.find('div.join-flow-description').exists()).toEqual(false);
+        expect(component.find('NextStepButton').prop('waiting')).toEqual(false);
+
+        component.unmount();
+    });
+
+    test('clicking submit calls passed in function', () => {
         const props = {
             onSubmit: jest.fn()
         };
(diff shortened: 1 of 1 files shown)


test/unit/components/join-flow-step.test.jsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import {mountWithIntl} from '../../helpers/intl-helpers.jsx';
+import JoinFlowStep from '../../../src/components/join-flow/join-flow-step';
+
+describe('JoinFlowStep', () => {
+
+    test('test components exist when props present', () => {
```
benjiwheeler: i suggest removing the word "test" here -- it took me a few readings to understand this. As I read the code, the string should just be an assertion of the state when things are working right, so together you'd have `test('only Jedis allowed')`, not `test('test only Jedis allowed')`.

test/unit/components/join-flow-step.test.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+        component.find('button[type="submit"]').simulate('submit');
+        expect(props.onSubmit).toHaveBeenCalled();
+
+        expect(component.find('NextStepButton').prop('waiting')).toEqual(true);
+        expect(component.find('NextStepButton').prop('content')).toEqual(props.nextButton);
+        component.unmount();
+    });
+
+    test('test components do not exist when props not present', () => {
```
benjiwheeler: same as above

test/unit/components/join-flow-step.test.jsx
```diff
@@ -0,0 +1,69 @@
...
+        expect(component.find('.join-flow-title').exists()).toEqual(true);
+
+        expect(component.find('.join-flow-title').first()
+            .prop('title')).toEqual(props.title);
+
+        expect(component.find('div.join-flow-description').exists()).toEqual(true);
+        expect(component.find('div.join-flow-description').text()).toEqual(props.description);
+
+
+        component.find('button[type="submit"]').simulate('submit');
```
benjiwheeler: I suggest moving this to a separate test -- I try (having learned this from others) to keep my tests as narrowly focused as possible, so that if a test fails, you immediately know the sort of thing that's going wrong

test/unit/components/join-flow-step.test.jsx
```diff
@@ -0,0 +1,69 @@
...
+        );
+        expect(component.find('div.join-flow-header-image').exists()).toEqual(false);
+
+        expect(component.find('.join-flow-inner-content').exists()).toEqual(true);
+        expect(component.find('.join-flow-title').exists()).toEqual(false);
+
+
+        expect(component.find('div.join-flow-description').exists()).toEqual(false);
+
+        component.find('button[type="submit"]').simulate('submit');
```
benjiwheeler: same as above; maybe you only need to test this once

test/unit/components/join-flow-step.test.jsx
picklesrus: Good call. Was thinking the setup was the same, but it isn't :)

test/unit/components/join-flow-step.test.jsx
picklesrus: yep. removed.

test/unit/components/join-flow-step.test.jsx
picklesrus: done!

test/unit/components/join-flow-step.test.jsx
picklesrus: done

test/unit/components/join-flow-step.test.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+        expect(component.find('.join-flow-inner-content').exists()).toEqual(true);
+        expect(component.find('.join-flow-title').exists()).toEqual(true);
+
+        expect(component.find('.join-flow-title').first()
+            .prop('title')).toEqual(props.title);
+
+        expect(component.find('div.join-flow-description').exists()).toEqual(true);
+        expect(component.find('div.join-flow-description').text()).toEqual(props.description);
+
```
benjiwheeler: this is a tiny quibble, but could you reduce the whitespace some? Seems to me like most of these are similar, and could be in one big bunch! (Like they are in the next test)

benjiwheeler:
This looks good! I had one comment about whitespace, but you should merge it, whether you make a change for that first or not!

test/unit/components/join-flow-step.test.jsx
picklesrus: Sure! Done.

PR #3224, opened 2019-08-07
Revert "Merge pull request #3217 from LLK/hotfix/fastly-concurrency"

paulkaplan:
Reverts LLK/scratch-www#3217

This may not have been the root cause of the deploy issues, it looks like travis was having network connectivity issues on their end.

It is hotfix branch just to test deploying, I think we can just merge it into develop if all looks good.


rschamp:
Travis did finally reply to my support request saying that they thought this was due to their network problems.

PR #3227, opened 2019-08-08
[Master] Release 2019-08-07

colbygk:
## Release Notes
- Fix up locale data
- Work towards resolving #3053 
  - makes the "next" button's text into a `PropTypes.oneOf([PropTypes.string, PropTypes.node]) so we can pass an arrow svg
  - sets correct text for most buttons to "Next", rather than "Next Step"
  - sets final step to "Make a Project ->"
- Add enzyme to scratch-www, bringing in enzyme helpers
- Fix for #3058, work towards resolving LLK/scratchr2#5321
  - create `lib/country-list.js`, static list based on list from scratchr2, 2013
  - functions enabling handling country info for labels mapping to db values


apple502j:
Release notes?

colbygk:
Release notes are coming

PR #3233, opened 2019-08-08
Add standalone registration page that just contains the JoinFlowModal.

picklesrus:
Adds a standalone page for the new join flow. 
We'll eventually point entry points from scratchr2 here.


picklesrus:
Closing in favor of #3242

PR #3236, opened 2019-08-10
handle username validation errors states better

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

* Focusing on a field clears the error (until field is blurred again)
* If username and password are the same, password field reports error right away (instead of waiting for user to click "Next")
* Design of inputs with an error that are focused on are no longer a weird combination of blue and orange borders!

### Screenshots

Focusing on a field clears the error (until field is blurred again):

clearerror

If username and password are the same, password field reports error right away (instead of waiting for user to click "Next"):

errorrightaway

### DIFF
--- a/src/components/formik-forms/formik-input.scss
+++ b/src/components/formik-forms/formik-input.scss
@@ -25,9 +25,4 @@
             outline: none;
         }
     }
-
-    /* IE10/11-specific style resets */
-    &::-ms-reveal, &::-ms-clear {
-        display: none;
-    }
 }
--- a/src/components/formik-forms/formik-input.scss
+++ b/src/components/formik-forms/formik-input.scss
@@ -1,0 +1,33 @@
+@import "../../colors";
+
+.input {
+    height: 2.75rem;
+    border-radius: .5rem;
+    background-color: $ui-white;
+    margin-bottom: .5rem;
+    transition: all .5s ease;
+    border: 1px solid $active-gray;
+    padding: 0 1rem;
+    color: $type-gray;
+    font-size: .875rem;
+
+    &:focus {
+        box-shadow: 0 0 0 .25rem $ui-blue-25percent;
+        outline: none;
+        border: 1px solid $ui-blue;
+    }
+
+    &.fail {
+        border: 1px solid $ui-orange;
+
+        &:focus {
+            box-shadow: 0 0 0 .25rem $ui-orange-25percent;
+            outline: none;
+        }
+    }
+
+    /* IE10/11-specific style resets */
+    &::-ms-reveal, &::-ms-clear {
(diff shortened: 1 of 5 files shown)


src/lib/validate.js
```diff
@@ -40,13 +40,21 @@ module.exports.validateUsernameRemotely = username => (
...
+ * @return {object}              {valid: boolean, errMsgId: string}
+ */
+module.exports.validatePassword = (password, username) => {
     if (!password) {
         return {valid: false, errMsgId: 'form.validationRequired'};
     } else if (password.length < 6) {
         return {valid: false, errMsgId: 'registration.validationPasswordLength'};
     } else if (password === 'password') {
         return {valid: false, errMsgId: 'registration.validationPasswordNotEquals'};
+    } else if (username && password === username) {
```
picklesrus: Can you add a test for this case?

src/components/formik-forms/formik-input.scss
```diff
@@ -0,0 +1,33 @@
...
+
+        &:focus {
+            box-shadow: 0 0 0 .25rem $ui-orange-25percent;
+            outline: none;
+        }
+    }
+
+    /* IE10/11-specific style resets */
+    &::-ms-reveal, &::-ms-clear {
+        display: none;
```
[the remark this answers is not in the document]
benjiwheeler: I carried these over when I copied similar code. Ray suggests we try omitting them here, so i'll do that.

src/lib/validate.js
benjiwheeler: Done

benjiwheeler:
Updated this slightly, per feedback from Katny:
1) keep showing validation errors on focus, until first keystroke
2) prioritize vulgarity over other validations

src/components/join-flow/username-step.jsx
```diff
@@ -36,20 +36,26 @@ class UsernameStep extends React.Component {
     // we allow username to be empty on blur, since you might not have typed anything yet
     validateUsernameIfPresent (username) {
         if (!username) return null; // skip validation if username is blank; null indicates valid
+        // if username is not blank, run both local and remote validations
         const localResult = validate.validateUsernameLocally(username);
-        if (localResult.valid) {
-            return validate.validateUsernameRemotely(username).then(
-                remoteResult => {
-                    if (remoteResult.valid) return null;
+        return validate.validateUsernameRemotely(username).then(
```
picklesrus: hmm, but how aren't we making a network request all the time? I'm not sure this is a worthwhile tradeoff.

benjiwheeler: Resolved as ok, per discussion in join flow subcommittee

PR #3241, opened 2019-08-11
Join flow info button

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

* format info button in join flow 
* use info button in birthdate step, gender step
* add strings for the above

### Note:

1. in a follow up, we should replace the css/text-formatted question mark with an svg to make the appearance more consistent
2. note that the placement of the info button can be weird if it happens to make the text wrap right there... not sure what to do about that

image

### DIFF
--- a/src/components/info-button/info-button.jsx
+++ b/src/components/info-button/info-button.jsx
@@ -1,18 +1,46 @@
+const bindAll = require('lodash.bindall');
 const PropTypes = require('prop-types');
 const React = require('react');
 
 require('./info-button.scss');
 
-const InfoButton = ({
-    message
-}) => (
-    <div className="info-button">
-        ?
-    </div>
-);
+class InfoButton extends React.Component {
+    constructor (props) {
+        super(props);
+        bindAll(this, [
+            'handleHideMessage',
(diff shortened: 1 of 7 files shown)


src/components/info-button/info-button.jsx
```diff
@@ -0,0 +1,46 @@
+const bindAll = require('lodash.bindall');
+const PropTypes = require('prop-types');
+const React = require('react');
+
+require('./info-button.scss');
+
+class InfoButton extends React.Component {
```
picklesrus: I think there's enough testing groundwork in place for this to have a test and for it not to take long to write it! (Still working on a Formik example), but https://github.com/LLK/scratch-www/blob/develop/test/unit/components/next-step-button.test.jsx and https://github.com/LLK/scratch-www/pull/3216/files should have enough info in them. Happy to pair on it if that'd be helpful.

benjiwheeler: Added! Take a look!

benjiwheeler:
> Thanks for the test. Curious to hear how much effort it was.

Not much!

PR #3242, opened 2019-08-12
Standalone join flow page.

picklesrus:
Page that contains the Join Flow modal.

Before committing I need to setup a config so this isn't visible in production. I want to double check that we're all in agreement about where this page should live too.  I put it on the agenda for our Tuesday meeting.


rschamp:
Oh except there are conflicts. I guess you need to rebase on develop and run `npm install --package-lock-only`?

picklesrus:
Reverting because the www build is now hanging on travis.

PR #3244, opened 2019-08-12
Allow admins to delete comments without confirmation

paulkaplan:
Requested mod feature, do not show the "Are you sure you want to delete and not report?" modal for project comments if the user is an admin.


rschamp:
Discussed offline, this is a regression!

PR #3250, opened 2019-08-13
Join flow gender custom focus

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

* gets ref to custom gender input
* if custom gender option is clicked on, focuses cursor on input field
* corrects incorrect `htmlFor` value in `formik-radio-button` pointed out by @paulkaplan in https://github.com/LLK/scratch-www/pull/3178#discussion_r310618224

### Test Coverage:

n/a

### DIFF
--- a/src/components/formik-forms/formik-radio-button.jsx
+++ b/src/components/formik-forms/formik-radio-button.jsx
@@ -10,24 +10,26 @@ require('./formik-radio-button.scss');
 require('../forms/row.scss');
 
 const FormikRadioButtonSubComponent = ({
-    buttonValue,
     children,
     className,
-    field,
+    field, // field.value is the current selected value of the entire radio group
+    id,
     label,
     labelClassName,
+    value,
     ...props
 }) => (
     <React.Fragment>
         <input
-            checked={buttonValue === field.value}
+            checked={value === field.value}
             className={classNames(
                 'formik-radio-button',
                 className
             )}
+            id={id}
             name={field.name}
             type="radio"
-            value={buttonValue}
+            value={value}
             onBlur={field.onBlur} /* eslint-disable-line react/jsx-handler-names */
             onChange={field.onChange} /* eslint-disable-line react/jsx-handler-names */
             {...props}
@@ -39,7 +41,6 @@ const FormikRadioButtonSubComponent = ({
                     'formik-radio-label',
                     labelClassName
                 )}
-                htmlFor={buttonValue}
             >
                 {label}
             </label>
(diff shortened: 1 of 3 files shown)


src/components/formik-forms/formik-radio-button.jsx
```diff
@@ -10,24 +10,26 @@ require('./formik-radio-button.scss');
 require('../forms/row.scss');
 
 const FormikRadioButtonSubComponent = ({
-    buttonValue,
```
picklesrus: why did you change the name of buttonValue?  Also, I'm confused again, what's the difference between field.value and value?

benjiwheeler: Yeah, this is one of the most confusing things about formik's handling of radio buttons...

So, when you give the formik `<Field>` component a `subComponent` prop (which you need to do for radio buttons because there's no built-in handling of them, unlike inputs and selects... which is annoying), it automatically passes it a `field` prop. And what's weird is, `field.value` is the value of the _radio button group_, not this individual button. I can't figure out why this is, but that's the `field.value` when I inspect/log it.

Meanwhile, `value` is the value of this individual button. 

So to see if this button should be currently "selected", you need to compare those two kinds of value.

benjiwheeler: oh, and I thought that `value` would be simpler than `buttonValue`. What do you think?

picklesrus: I see (maybe :) ). In this example, are they using a strategic value of id to get around this? https://codesandbox.io/s/pjqp3xxq7q

benjiwheeler: In that example, I think this section of `CheckboxGroup` is doing a lot of work:

```
            return React.cloneElement(child, {
              field: {
                value: value.includes(child.props.id),
                onChange: this.handleChange,
                onBlur: this.handleBlur
              }
            });
```

picklesrus: Sorry, I meant the RadioGroup. In the UI (and I think the code too) the CheckboxGroup is a separate thing.  Sorry to to get stuck on this point. Just want to make sure I have a better understanding of how it works.

benjiwheeler: Sorry, yes, your previous comment is exactly right. In that codesandbox example, they use the `id` prop to set `value`, and they use `field.value` to compare to `id` to determine if the button is selected.

Make sense? I guess formik is acting like all the radio buttons are a single "field", or something? I don't totally understand it, to tell the truth.

Issue #2830, opened 2019-03-07
Add micro:bit cards to the /microbit page


ntlrsk:
Add link to download micro:bit Scratch cards to the microbit extension page. Place it after Starter Projects, before Troubleshooting.

Here is the [PDF](https://www.dropbox.com/s/8ud9bei5do7fzem/microbit-scratch-cards.pdf?dl=0) (Note: PDF may get updated - Natalie and KO will confirm )

/cc @kosiecki17

benjiwheeler:
Seems simple. We could borrow the scratch cards section on the ideas page, or design something more Microbit-specific.

thisandagain:
@ntlrsk I moved this to May due to some scheduling pressure. Is that ok with you?

benjiwheeler:
Labeling this as 'needs discussion' pending @carljbowman clarifying plans.

benjiwheeler:
Update: still waiting on further discussion from the team. Maybe this should be more explicitly put on one person's plate, or backlogged?

thisandagain:
@benjiwheeler I added this to our weekly design meeting agenda. Will follow-up after discussion.

/cc @carljbowman

kathymakes:
Here's one potential design & copytext that we'd like to chat more about with @ntlrsk and @kosiecki17. We'll loop back after our design meeting on 8/15:
 <img width="714" alt="Screen Shot 2019-08-01 at 5 17 22 PM" src="https://user-images.githubusercontent.com/8203939/62328136-4b152000-b480-11e9-965f-1e01f529f1ea.png">

thisandagain:
@kathymakes Looks great! Let's move forward. My only nitpick would be to remove the duplicate trail of "S"s in the illustration, but I'm happy to defer to your judgement. Let's move forward with implementation.

/cc @benjiwheeler

kathymakes:
### Illustrated section
<img width="1317" alt="Screen Shot 2019-08-08 at 2 06 40 PM" src="https://user-images.githubusercontent.com/8203939/62726663-c416f900-b9e5-11e9-9f50-8bd07f0917c8.png">

### In context

![microbit Cards Design](https://user-images.githubusercontent.com/8203939/62726611-a5b0fd80-b9e5-11e9-9bde-116eace503c1.png)

PR #3263, opened 2019-08-16
Microbit add cards

benjiwheeler:
NOTE: must be merged AFTER https://github.com/LLK/scratch-resources/pull/74 , which puts the microbit cards on the resources server

### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/2830

### Changes:

* adds microbit cards section to microbit landing page
* tweaks extensions landing page css so if you have cards section right before faq section, padding is not too big
* adds microbit cards url

### Test Coverage:

n/a


chrisgarrity:
@benjiwheeler this isn't a 'small' change because download links need to be handled differently for localization. I'll follow up with more details.

src/views/microbit/l10n-static.json
```diff
@@ -0,0 +1,3 @@
+{
+    "cards.microbit-cardsLink": "https://resources.scratch.mit.edu/www/cards/en/microbit-cards.pdf"
```
chrisgarrity: Yes, we can't just add a link in the l10n file because a translator doesn't know whether a translated file is available, or how to reference it. The resources server generates `localized-urls.json` based on what's actually available on the server.

benjiwheeler: Thanks. So does this look good now?

benjiwheeler:
Just tested it with the merged resources change, seems to work for me!

PR #3266, opened 2019-08-18
use media queries to make info message responsive

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

* uses media queries in both jsx and sass to handle info button differently in small screen sizes

For small screen sizes, we keep the info message absolutely positioned, but instead of the reference parent element with `position: relative` being the info icon itself, we make a new div, a sibling to the icon, with `position: relative`.

This creates a new row, which is where the info message positions itself.

Since the message is absolutely positioned, it overlaps content below (rather than causing the enclosing modal or other element to have to change form).

### Screenshots

wide window/screen (>= 942px):

image

small window/screen (< 942px):

image

### DIFF
--- a/test/unit/components/info-button.test.jsx
+++ b/test/unit/components/info-button.test.jsx
@@ -11,6 +11,15 @@ describe('InfoButton', () => {
         );
         expect(component.find('div.info-button-message').exists()).toEqual(false);
     });
+    test('mouseOver on info button makes info message visible', () => {
+        const component = mountWithIntl(
+            <InfoButton
+                message="Here is some info about something!"
+            />
+        );
+        component.find('div.info-button').simulate('mouseOver');
+        expect(component.find('div.info-button-message').exists()).toEqual(true);
+    });
     test('clicking on info button makes info message visible', () => {
         const component = mountWithIntl(
             <InfoButton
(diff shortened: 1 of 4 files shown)


test/unit/components/info-button.test.jsx
```diff
@@ -26,7 +35,7 @@ describe('InfoButton', () => {
                 message="Here is some info about something!"
             />
         );
-        component.find('div.info-button').simulate('click');
+        component.find('div.info-button').simulate('mouseOver');
```
picklesrus: why did you change this part of the test?  Looks like the message is still expected to be shown on click, right?

benjiwheeler: Yes, it should be shown on either `click` or `mouseOver`. This test is focused on the behavior of `mouseOut`, and I thought `mouseOver` would be a more common step before `mouseOut` than `click`.

picklesrus: I was just looking at the diff wrong (missed the part that it collapsed and thought this was in the test after). Sorry about that.

picklesrus:
The code looks good, but I'm not sure we've landed at the right behavior for that on desktop vs mobile. Happy for it to be merged though and see how it feels when we have a bunch of devices around to test on.

benjiwheeler:
> The code looks good, but I'm not sure we've landed at the right behavior for that on desktop vs mobile. Happy for it to be merged though and see how it feels when we have a bunch of devices around to test on.

Agree on both points!

PR #3273, opened 2019-08-20
Able to use scratch3 join flow in editor

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

Adds new join flow to editor page, using a flag that can be changed to turn it on

### Test Coverage:

I wrote integration tests, but couldn't get them to run consistently because of the loading screen issue.


src/views/preview/project-view.jsx
```diff
@@ -32,6 +33,8 @@ const frameless = require('../../lib/frameless');
 const GUI = require('scratch-gui');
 const IntlGUI = injectIntl(GUI.default);
 
+const USE_SCRATCH3_REGISTRATION = false;
```
picklesrus: What would it take to use the same flag to enable this for both navigation and the editor?

benjiwheeler: Done!

PR #3276, opened 2019-08-20
Initial work for captcha in new join flow.

picklesrus:
TODOS:
- handle error states
- Setup keys for different environments.
- Make sure remote validators are run before captcha.


rschamp:
@picklesrus If you look at the commit that build is for, it's the result of merging this PR, so it's warning you that once you merge it, the build will fail. So if you merge develop in or rebase this onto develop, you'll see the lint errors.

picklesrus:
Rebased against develop and fixed all the merge conflicts. I'll merge once I know I won't conflict with the use of staging for the release

PR #3282, opened 2019-08-21
fix toolTip error by accepting bool as toolTip type

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

* in formik-input, accept bool as toolTip type in PropTypes, so a value of `false` will be handled correctly


src/components/formik-forms/formik-input.jsx
```diff
@@ -55,7 +55,8 @@ FormikInput.propTypes = {
     className: PropTypes.string,
     error: PropTypes.string,
     setRef: PropTypes.func,
-    toolTip: PropTypes.string,
+    // toolTip can be false, in which case we ignore it
+    toolTip: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
```
picklesrus: It seems inconsistent that error and toolTip are treated differently. If I can not pass an error and the component works (i.e. doesn't display an error), I think I should be able to skip passing toolTip and have that work as well.

benjiwheeler: It actually does what you said -- you can skip passing toolTip and have that work.

The difference is that there are places that formik-input is passed a `false` toolTip value, and there don't happen to be places that a `false` value of error is passed in.

But, it seems like if that ever happens, formik-input should handle it correctly, right? I'll change error to behave the same way.

benjiwheeler: Done

PR #3306, opened 2019-08-28
restyle next button to remove border, add hover color

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

restyles next button to:
* remove border
* add hover color

My only questions are for @kathymakes:
1. what do you think of the hover color? 
2. do you think this merits adding a color to _colors.scss?

Note that in _colors.scss, there are actually two different hue/saturation values both called "orange" -- I'm not sure what the intention was with the 25 percent and 10 percent values.

No hover:

image

Yes hover:

image

### DIFF
--- a/src/_colors.scss
+++ b/src/_colors.scss
@@ -8,7 +8,7 @@ $ui-orange: hsla(38, 100, 55, 1); // #FFAB19 Control Primary
 $ui-orange-high-contrast: hsla(30, 100, 55, 1); // #FFAB19 Control Primary
 $ui-orange-10percent: hsla(35, 90, 55, .1);
 $ui-orange-25percent: hsla(35, 90, 55, .25);
-$ui-orange-75percent: hsla(38, 100, 55, .75);
+$ui-orange-90percent: hsla(38, 100, 55, .9);
 
 $ui-dark-orange: hsla(30, 100, 55, 1); // ##FF8C1A Variables Primary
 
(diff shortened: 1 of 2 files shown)


kathymakes:
Hey Ben, having second thoughts about adding a hover color. Now that I'm seeing this, this particular color makes it looks disabled (although I'm sure it will feel clearer once you are interacting with it.) This might be a useful color to use if we ever have spinners/loaders to indicate that you can't click the button.

After poking around, looks like we don't really have hover states in other parts of the site unless you have multiple buttons next to each other. Maybe we can leave the hover change for now, and just remove the border? Thanks for putting in this change so quickly :)

benjiwheeler:
@kathymakes and I looked at this and revised it, and it's ready for code review!

src/_colors.scss
```diff
@@ -8,6 +8,7 @@ $ui-orange: hsla(38, 100, 55, 1); // #FFAB19 Control Primary
 $ui-orange-high-contrast: hsla(30, 100, 55, 1); // #FFAB19 Control Primary
 $ui-orange-10percent: hsla(35, 90, 55, .1);
 $ui-orange-25percent: hsla(35, 90, 55, .25);
+$ui-orange-90percent: hsla(38, 100, 55, .9);
```
picklesrus: I was expecting there to be consistency between our naming of something as ui-color-YYpercent and the saturation for those colors. Seems like that isn't the case though, right? i.e. the 100 here is not consistent with the 90 for the ui-orange-10percent and ui-orange-25percent.  As long as that inconsistency is expected, LGTM

benjiwheeler: I think the 90 percent thing is confusing here because there are really three kinds of orange that we are treating as if they're two; my 90 percent uses the same base orange as $ui-orange, whereas 10 percent and 25 percent don't. Since 90 percent is so much closer to 100 percent than to 10 and 25 percent, this feels like the most consistent choice.

The "percent" in the other color names refers to brightness/alpha (which is the same thing on a white background).

PR #3307, opened 2019-08-28
in username step, only show one validation at a time

benjiwheeler:
### Resolves:

Step towards resolving #3053

### Changes:

* for error state in password and passwordConfirm inputs, only have error state be active if the previous fields are _not_ in error state.

### DIFF
--- a/src/components/join-flow/username-step.jsx
+++ b/src/components/join-flow/username-step.jsx
@@ -177,10 +177,13 @@ class UsernameStep extends React.Component {
                                     <FormikInput
                                         className={classNames(
                                             'join-flow-input',
-                                            {'join-flow-input-password':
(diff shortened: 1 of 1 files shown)


picklesrus:
Have we talked about this as a group? Happy to be overruled, but I hate when forms tell me what things are wrong 1 by 1 when I submit. I want to fix them all at once and try again.  
In this case there aren't that many possible errors so maybe my concern is misplaced.

benjiwheeler:
This is on hold because we weren't sure we want it, and weren't sure it's worth the added complexity.

benjiwheeler:
Closing this because we decided it wasn't necessary; could revisit someday if we want to reevaluate the usability of the username step.

PR #3322, opened 2019-08-29
Add a unittest for the EmailStep.

picklesrus:
### Test Coverage:
yep!

### DIFF
--- a/test/unit/components/email-step.test.jsx
+++ b/test/unit/components/email-step.test.jsx
@@ -1,7 +1,7 @@
-import React from 'react';
-import {shallowWithIntl} from '../../helpers/intl-helpers.jsx';
-import EmailStep from '../../../src/components/join-flow/email-step.jsx';
-import JoinFlowStep from '../../../src/components/join-flow/join-flow-step.jsx';
+const React = require('react');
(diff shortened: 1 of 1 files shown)


test/unit/components/email-step.test.jsx
```diff
@@ -0,0 +1,134 @@
+import React from 'react';
+import {shallowWithIntl} from '../../helpers/intl-helpers.jsx';
+import EmailStep from '../../../src/components/join-flow/email-step.jsx';
+import JoinFlowStep from '../../../src/components/join-flow/join-flow-step.jsx';
```
rschamp: This file uses a mix of `import` and `require`. I think `require` is the broadly used way in this repo, so we should consolidate on that.

picklesrus: Good call. done!

picklesrus:
Fixed the import/require. I also changed the initial data in one of the objects to be less cute since it was confusing on initial reading of the code. Lastly, Ben pointed out that there were better ways to do the object equality check for arguments to toHaveBeenCalledWith so I changed that too.

rschamp:
Looks good! I agree the updated equality checks look better. Thanks!

PR #3334, opened 2019-09-11
Add captcha token to ther request. It was left out of #3291.

picklesrus:
### Resolves:
Sends capatcha token to create new user request for new join flow


benjiwheeler:
LGTM

PR #3351, opened 2019-09-15
move info button down, and give its message a shadow

benjiwheeler:
### Resolves:

### Changes:

Shifts info button downwards to be more in line with text; extends the shadow that we added to the validation message when it is in "suggestion" mode, to the info rollover window

### DIFF
--- a/src/components/info-button/info-button.jsx
+++ b/src/components/info-button/info-button.jsx
@@ -26,7 +26,7 @@ class InfoButton extends React.Component {
     }
     render () {
         const messageJsx = this.state.visible && (
-            <div className="info-button-message">
+            <div className="info-button-message validation-info">
                 {this.props.message}
             </div>
         );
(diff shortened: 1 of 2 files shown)


src/components/info-button/info-button.jsx
```diff
@@ -26,7 +26,7 @@ class InfoButton extends React.Component {
     }
     render () {
         const messageJsx = this.state.visible && (
-            <div className="info-button-message">
+            <div className="info-button-message validation-info">
```
picklesrus: Are you adding validation info to pick up the styles from https://github.com/LLK/scratch-www/blob/5ab2bf100b59123c41654dca76c615f6b4046002/src/components/forms/validation-message.scss#L82

I don't love that because if I change the validation-info class, the changes I'm going to know that I also need to check the InfoButton are pretty small.  Also, is the css compiler smart enough not to rename them differently given there's no require here?

benjiwheeler: Yeah, you're right, this is not the right way to do it! Even if it works in practice...

Take a look now.

picklesrus:
LGTM if travis is still happy!

PR #3352, opened 2019-09-16
Join flow revise descriptions and privacy messages

benjiwheeler:
### Resolves:

### Changes:

* moves privacy messages out of descriptions, and below inputs
* removes most descriptions entirely
* makes all privacy messages identical
* changes formatting of privacy messages to be smaller and light gray
* changes accompanying info button to be light gray
* formats each join flow title’s padding differently, to match spec
* adjust email step vertical spacing so that it's likely to be constrained to 520px
* redo email step footer
* adds email privacy info string

### Notes

Note that the email privacy info string content is tentative, and should be discussed


src/components/join-flow/email-step.jsx
```diff
@@ -130,7 +131,6 @@ class EmailStep extends React.Component {
                     } = props;
                     return (
                         <JoinFlowStep
-                            description={this.props.intl.formatMessage({id: 'registration.emailStepDescription'})}
```
picklesrus: There's a test that expects this to exist. I think that's why travis is failing.

benjiwheeler: Thanks, fixed!

Issue #2471, opened 2018-12-13
Make project/embed view easy to cache


chrisgarrity:
The new projects `.../embed` view should be able to be cached indefinitely:
- ignore session info
- do not connect to cloud vars server

picklesrus:
We noticed this while the db was loaded. The requests for comments are especially expensive.

PR #3356, opened 2019-09-16
skip loading comments, remixes, studios for embed

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/2471

### Changes:

* gets `props.isEmbed` by checking GUI mode
* sets `props.userPresent` to false if `props.isEmbed` is true
* if `props.isEmbed` is true, skip fetching project's studios, remix info, and comments

### Test Coverage:

None so far

### Note

previously, this relied on https://github.com/LLK/scratch-gui/pull/5171

### DIFF
--- a/src/views/preview/project-view.jsx
+++ b/src/views/preview/project-view.jsx
@@ -160,7 +160,7 @@ class Preview extends React.Component {
         if (this.props.projectInfo.id !== prevProps.projectInfo.id) {
             if (typeof this.props.projectInfo.id === 'undefined') {
                 this.initCounts(0, 0);
-            } else {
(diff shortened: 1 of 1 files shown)


benjiwheeler:
Tentative new plan is to set a variable with embed status in project-view, then use that in place of the isEmbed prop, and pass down a revised hasCloudPermission to GUI.

Longer term, it would be good to create an entirely separate embed route and file, where we only do the minimum we need to present the project.

benjiwheeler:
Trying embed-view approach in https://github.com/LLK/scratch-www/pull/3377 ; closing this for now.

PR #3357, opened 2019-09-16
One approach to hiding the close button in the modal.

picklesrus:
Sets the default value of the showCloseButton prop to true and pass it through as false from the new standalone join flow page. 

### Test Coverage:
Added very basic tests for modal that the close button is hidden/shown appropriately.


src/views/join/join.jsx
```diff
@@ -6,11 +6,13 @@ const ErrorBoundary = require('../../components/errorboundary/errorboundary.jsx'
 const Page = require('../../components/page/www/page.jsx'); // eslint-disable-line no-unused-vars
 
 const openModal = true;
+const showCloseButton = false;
```
benjiwheeler: i think it would be simpler here to just put these values in the render props below directly. They don't need to be top level-scoped variables.

picklesrus: I like that better. Thanks for showing me how to do it :)

PR #3362, opened 2019-09-17
format and update timing of join flow spinner

benjiwheeler:
### Resolves:

### Changes:

* remove unnecessary --webkit prefix lines
* slow down spinner, and set timing with bezier (cleared with @kathymakes and @carljbowman)
* format size of spinner for join flow

Sep-16-2019 14-21-15

Sep-16-2019 14-31-59


picklesrus:
Code looks good. Have you verified the other spots that use Spinner in www look right?

benjiwheeler:
> Code looks good. Have you verified the other spots that use Spinner in www look right?

Yes -- the gif I put up shows Add to Studio, for instance.

> And that this works in Safari?

Confirmed.

PR #3367, opened 2019-09-18
Handle errors of captcha loading by setting error state on JoinFlow.

picklesrus:
### Resolves:

Some TODOs to handle captcha reloading.

### Changes:
Introduces a new function to JoinFlow to handle errors that happen in the middle of registration rather than on an explicit next step taking by the user

### Test Coverage:
Added tests for new join flow function as well as for email-step.

### DIFF
--- a/src/components/join-flow/join-flow.jsx
+++ b/src/components/join-flow/join-flow.jsx
@@ -23,6 +23,7 @@ class JoinFlow extends React.Component {
         super(props);
         bindAll(this, [
             'handleAdvanceStep',
+            'handleMidRegistrationError',
             'handlePrepareToRegister',
             'handleRegistrationResponse',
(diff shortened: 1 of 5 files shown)


src/components/join-flow/join-flow.jsx
```diff
@@ -23,6 +23,7 @@ class JoinFlow extends React.Component {
         super(props);
         bindAll(this, [
             'handleAdvanceStep',
+            'handleMidRegistrationError',
```
benjiwheeler: Could we just call this "handle/onRegistrationError" rather than "MidRegistration"? I know this isn't only used for errors during the final registration request, but I think this could be more broadly applicable.

picklesrus: Yes. Done! :)

Issue #1342, opened 2017-06-07
Clone Fastly active version not most recent on deploy


rschamp:
This will avoid unexpected results if there's a more recent inactive version with changes.

PR #3374, opened 2019-09-19
when looking for latest fastly VCL to clone, use only active VCLs

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/1342

### Changes:

When comparing fastly vcls to find most recent one, prefer one vcl to another if and only if it is both higher-numbered and active. (Previously, we only considered number.)

### Test Coverage:

Added a test.

### DIFF
--- a/bin/lib/fastly-extended.js
+++ b/bin/lib/fastly-extended.js
@@ -37,18 +37,15 @@ module.exports = function (apiKey, serviceId) {
             if (err) {
                 return cb('Failed to fetch versions: ' + err);
             }
-            var latestVersion = versions.reduce(function (lateVersion, version) {
-                if (!lateVersion) return version;
-                // if versions we're comparing are both active, or both inactive, prefer
-                // whichever has a higher version number
-                if (lateVersion.active === version.active) {
-                    if (version.number > lateVersion.number) return version;
-                    return lateVersion;
-                }
-                // if only one of the versions is active, prefer that one
-                if (version.active === true) return version;
-                return lateVersion;
-            });
+            var latestVersion = versions.reduce((latestActiveSoFar, cur) => {
+                // if one of [latestActiveSoFar, cur] is active and the other isn't,
+                // return whichever is active. If both are not active, return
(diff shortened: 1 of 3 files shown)


bin/lib/fastly-extended.js
```diff
@@ -39,7 +39,14 @@ module.exports = function (apiKey, serviceId) {
             }
             var latestVersion = versions.reduce(function (lateVersion, version) {
```
colbygk: I'm surprised our linter didn't warn/error on this. Don't we prefer (lateVersion, version) => { ... } ? Also, the var statements here ought to be const's. But that is pre-existing, so a new issue and not this PR.

test/unit/lib/fastly-extended.test.js
```diff
@@ -0,0 +1,44 @@
...
+                    active: false
+                },
+                {
+                    number: 1,
+                    active: true
+                },
+                {
+                    number: 2,
+                    active: true
+                },
```
colbygk: I don think this is good test data. Two versions cannot be active at the same time for a single service by definition no?

bin/lib/fastly-extended.js
colbygk: Also to avoid confusion, I think the two objects being compared in the reduce() function ought not be named the same as the object being returned via callback at line 52 and defined/getting return from the reduce() at line 40.

colbygk: (sorry, just noticed they're not the same name, just close named)

test/unit/lib/fastly-extended.test.js
benjiwheeler: Good point, changing!

benjiwheeler:
OK, I revised and (I think) simplified the `latestActiveVersionSoFar()` code, made the tests only include one active version, and added more tests for edge cases. I also changed `latestActiveVersionSoFar()` so that if there is no active version, it returns null.

PR #3375, opened 2019-09-20
Include canonical link in project meta tags

paulkaplan:
This will tell crawlers that this page and all its exact duplicates (/fullscreen, /editor, /embed, etc...) should be considered the same and use the project page as the canonical crawled page. We had an issue before where by not selected a canonical, google was considering all the project pages duplicates :(

### Resolves:

Hopefully resolves a problem with public scratch projects not showing up on google.

### Changes:

Add a canonical link to the project page template

/cc @rschamp we do not use a variable for the root url, so this won't be quite right in staging, but that isn't crawled anyway ;)

### DIFF
--- a/src/views/preview/meta.jsx
+++ b/src/views/preview/meta.jsx
@@ -30,8 +30,8 @@ const Meta = props => {
                 property="og:description"
             />
             <link
-                rel="canonical"
                 href={`https://scratch.mit.edu/projects/${id}`}
+                rel="canonical"
             />
         </Helmet>
     );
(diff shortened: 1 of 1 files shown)


src/views/preview/meta.jsx
```diff
@@ -29,6 +29,10 @@ const Meta = props => {
                 content={truncatedInstructions}
                 property="og:description"
             />
+            <link
+                rel="canonical"
+                href={`https://scratch.mit.edu/projects/${id}`}
```
rschamp: Linter is complaining that href should come before rel.

PR #3382, opened 2019-09-20
validate password and passwordConfirm fields to cover more cases

benjiwheeler:
### Resolves:

### Changes:

Tweaks how the username step Fields handle validation, to cover more edge cases. 

For instance, try:
1. type a password
2. type a password confirm, matching password
3. click back on first password
4. add a letter
5. click outside the field to blur it
6. it should show a validation error. Currently, it does not; with this change, it does.

Note one sequence of actions that isn't perfect even with this change:
1. type a password
2. type a password confirm, but make the confirm not match password
3. Then without first blurring the password confirm field, click Next.
4. You will first see a validation appear on password confirm, which will then move to password. This happens because blurring the password confirm shows its error, but submitting the form makes the password field show an error, and we only show one error at a time -- the first one.

However, this is still an improvement.

### DIFF
--- a/src/components/join-flow/username-step.jsx
+++ b/src/components/join-flow/username-step.jsx
@@ -69,14 +69,19 @@ class UsernameStep extends React.Component {
             }
         );
     }
-    validatePasswordIfPresent (password, username) {
+    // need to validate passwordConfirm as well, so we can show password mismatch error
+    // when password field is blurred, not only passwordConfirm field.
+    validatePasswordIfPresent (password, passwordConfirm, username) {
         if (!password) return null; // skip validation if password is blank; null indicates valid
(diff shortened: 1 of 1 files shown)


picklesrus:
I'm not sure if I agree that this is better. I think the second case you outline is going to be far more common than the first. I think the flicker of the first validation is more jarring (though jarring is too strong of a word) than not getting told your passwords don't match until you submit.

benjiwheeler:
Hmm... good point.

What do you think we should show if, say, you hit next and nothing has been entered in any of these fields?

picklesrus:
I like showing required on all of the missing required items even if it is busy!

benjiwheeler:
Closed this as not clearly an improvement.

PR #3383, opened 2019-09-20
scratch3 join flow redirects to /join, outside editor

benjiwheeler:
### Resolves:

### Changes:

instead of popping up new join flow in a modal, this makes most pages redirect to /join. Except for editor.

### DIFF
--- a/src/components/navigation/www/navigation.jsx
+++ b/src/components/navigation/www/navigation.jsx
@@ -18,7 +18,6 @@ const LoginDropdown = require('../../login/login-dropdown.jsx');
 const CanceledDeletionModal = require('../../login/canceled-deletion-modal.jsx');
 const NavigationBox = require('../base/navigation.jsx');
 const Registration = require('../../registration/registration.jsx');
-const Scratch3Registration = require('../../registration/scratch3-registration.jsx');
 const AccountNav = require('./accountnav.jsx');
 
 require('./navigation.scss');
@@ -28,6 +27,7 @@ class Navigation extends React.Component {
         super(props);
         bindAll(this, [
             'getProfileUrl',
+            'handleClickRegistration',
             'handleSearchSubmit'
         ]);
         this.state = {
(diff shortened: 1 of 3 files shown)


test/unit/components/navigation.test.jsx
```diff
@@ -0,0 +1,69 @@
...
+const {shallowWithIntl} = require('../../helpers/intl-helpers.jsx');
+import configureStore from 'redux-mock-store';
+const Navigation = require('../../../src/components/navigation/www/navigation.jsx');
+const sessionActions = require('../../../src/redux/session.js');
+
+describe('Navigation', () => {
+    const mockStore = configureStore();
+    let store;
+
+    beforeEach(() => {
```
picklesrus: should store be reset between tests?

test/unit/components/navigation.test.jsx
```diff
@@ -0,0 +1,69 @@
...
+            .dive() // unwrap redux connect(injectIntl(JoinFlow))
+            .dive(); // unwrap injectIntl(JoinFlow)
+    };
+
+    test('when using old join flow, clicking Join Scratch attemps to open registration', () => {
+        store = mockStore({
+            navigation: {
+                useScratch3Registration: false
+            },
+            session: {
```
picklesrus: Do you need to set session and messageCount in these tests?

src/components/navigation/www/navigation.jsx
```diff
@@ -78,6 +78,13 @@ class Navigation extends React.Component {
         if (!this.props.user) return;
         return `/users/${this.props.user.username}/`;
     }
+    handleClickRegistration (event) {
```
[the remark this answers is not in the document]
benjiwheeler: Done. (I simulate a click, which calls this)

test/unit/components/navigation.test.jsx
benjiwheeler: hm, good idea. doing this

test/unit/components/navigation.test.jsx
benjiwheeler: i think we do need session (there's a conditional in render that checks it), but we don't need messageCount (which I think isn't even spelled right! Ack!)

test/unit/components/navigation.test.jsx
benjiwheeler: Done

PR #3384, opened 2019-09-20
gender step uses blank default, lcase values

benjiwheeler:
### Resolves:

### Changes:

* use "male" and "female" for gender values, not "Male" and "Female". This matches the existing values in the database. (Users will still see capitalized versions.)
* use "non-binary" as value instead of "Non-binary", to better match what users are likely to have typed themselves in the past. (Users will still see capitalized version.)
* use a blank string when user skips gender step, rather than "Prefer not to say"

### DIFF
--- a/src/components/join-flow/gender-step.jsx
+++ b/src/components/join-flow/gender-step.jsx
@@ -67,7 +67,7 @@ class GenderStep extends React.Component {
     handleValidSubmit (formData, formikBag) {
         formikBag.setSubmitting(false);
         if (!formData.gender || formData.gender === 'null') {
-            formData.gender = 'Prefer not to say';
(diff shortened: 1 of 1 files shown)


picklesrus:
This looks good. Is empty string our usual value for something the user chose to skip or is it null?  Might be worth confirming with @rschamp

benjiwheeler:
Yeah, let's ask in the meeting tuesday

benjiwheeler:
Revised the defaults per decisions in 20190924 meeting.

Selecting “Prefer not to say” -> “(Prefer not to say)”
Just clicking Next, without selecting anything -> “(skipped)”
Clicking in the “other” field, but not typing anything -> “(blank)”

PR #3389, opened 2019-09-23
Join flow navigate when finished

benjiwheeler:
### Resolves:

### Changes:

* use Scratch3Registration instead of JoinModal in /join
* pass createProjectOnComplete prop through join flow 
* in editor mode, last step says "Done" and not "Create a project ->"
 
### Screenshot:

(Note that we changed the text/icon in the button in the last step since taking this...)

editorreg2

### DIFF
--- a/src/views/join/join.jsx
+++ b/src/views/join/join.jsx
@@ -1,6 +1,6 @@
 const React = require('react');
 const render = require('../../lib/render.jsx');
-const JoinModal = require('../../components/modal/join/modal.jsx');
+const Scratch3Registration = require('../../components/registration/scratch3-registration.jsx');
 const ErrorBoundary = require('../../components/errorboundary/errorboundary.jsx');
 // Require this even though we don't use it because, without it, webpack runs out of memory...
 const Page = require('../../components/page/www/page.jsx'); // eslint-disable-line no-unused-vars
@@ -20,7 +20,8 @@ const Register = () => (
             </a>
 
         </div>
-        <JoinModal
+        <Scratch3Registration
+            createProjectOnComplete
             isOpen
             key="scratch3registration"
             showCloseButton={false}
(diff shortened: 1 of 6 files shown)


src/views/join/join.jsx
```diff
@@ -1,14 +1,15 @@
...
+const Scratch3Registration = require('../../components/registration/scratch3-registration.jsx');
 const ErrorBoundary = require('../../components/errorboundary/errorboundary.jsx');
 // Require this even though we don't use it because, without it, webpack runs out of memory...
 const Page = require('../../components/page/www/page.jsx'); // eslint-disable-line no-unused-vars
 
 const openModal = true;
 const Register = () => (
     <ErrorBoundary>
-        <JoinModal
+        <Scratch3Registration
```
picklesrus: I'm curious why you switched to using Scratch3Registration here rather than just passing the new prop directly to the JoinModal.  I had a avoiding using Scratch3Registration before just because it had more functionality than I thought we needed, but I'm also still learning how to do things in React.

benjiwheeler: This is a good question! Scratch3Registration is wired up to redux, so we can use the navigation.js library to handle what to do when the user clicks the final button at the very end, rather than adding a copy of this functionality to join.jsx or JoinModal.

picklesrus: Makes sense!  This seems good!

At some point, I'd love to discuss whether it might make sense to send a function or maybe the button itself? as a prop instead of a bool. Then the function or button would just do the right thing when it was clicked and the welcome step wouldn't be concerned with anything.  It feels like that might be a more generic way to do things that keeps the components at the bottom of the tree simpler.  Just thinking out loud and wondering what best practices are w.r.t. React design though.

benjiwheeler:
> At some point, I'd love to discuss whether it might make sense to send a function or maybe the button itself? as a prop instead of a bool. Then the function or button would just do the right thing when it was clicked and the welcome step wouldn't be concerned with anything.

I think that's basically what we're trying to do now, but with the `join-flow-step` being the generic one, and the logic about what to pass to it living above it.

benjiwheeler:
> Watch out for merge conflicts - I recently merged the change to add the logo and remove the x. I think the X change means you're going to need to wire more information through scratch3Registration. That's what prompted my comment below.

Thanks, I didn't realize this at first! Then I wired it through like you said.

PR #3393, opened 2019-09-23
Join flow country placeholder

benjiwheeler:
### Resolves:

### Changes:

* changes placeholder for country to "Select country"

### DIFF
--- a/src/l10n.json
+++ b/src/l10n.json
@@ -185,7 +185,7 @@
     "registration.personalStepDescription": "Your individual responses will not be displayed publicly, and will be kept confidential and secure",
     "registration.private": "We will keep this information private.",
     "registration.receiveEmails": "I'd like to receive emails from the Scratch Team about project ideas, events, and more.",
-    "registration.selectCountry": "select country",
+    "registration.selectCountry": "Select country",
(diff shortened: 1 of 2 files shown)


src/l10n.json
```diff
@@ -18,7 +18,6 @@
     "general.conferences": "Conferences",
     "general.copyright": "Scratch is a project of the Lifelong Kindergarten Group at the MIT Media Lab",
     "general.copyrightDraft": "Scratch is a project of the Scratch Foundation in collaboration with the Lifelong Kindergarten Group at the MIT Media Lab.",
-    "general.country": "Country",
```
picklesrus: Is this still used here? https://github.com/LLK/scratch-www/blob/46de5a23e17c5d3a5226e29e54f09174c3264706/src/components/registration/steps.jsx#L556

benjiwheeler: You're right! I put it back.

apple502j:
Question BTW: Why do you call it "country", when it includes non-countries?

benjiwheeler:
> Question BTW: Why do you call it "country", when it includes non-countries?

That seemed like the clearest name for the category, which uses the [ISO list of country names and codes](https://www.iso.org/iso-3166-country-codes.html). Do you have another suggestion?

PR #3396, opened 2019-09-24
Add Logo to join flow standalone page.

picklesrus:
Logo links back to scratch homepage. 

This will need some positioning css work if we want to match it to teacher registration, especially on small screens. Right now it is set to be 25% of the way across the screen until it gets small when it is set to be centered.

Screen Shot 2019-09-23 at 6 05 03 PM

Screen Shot 2019-09-23 at 6 05 12 PM


benjiwheeler:
The most narrow window widths look a little awkward:

image

(emulating mobile:)

image

But, I think this is strictly better than what we had before, and we could circle back to that in a later design iteration.

picklesrus:
Agreed! It does need some love especially in that smallest case.  I'm not sure how to make the modal move down further in the smallest case. Since the logo isn't in the flow of the page adding spacing around it doesn't work.  I was trying to avoid modifying the modal css, but maybe I have to?  Happy to chat about it !

PR #3397, opened 2019-09-24
Download and initial extension pages for forthcoming ChromeOS app launch

paulkaplan:
This includes the work by @chrisgarrity but adds a feature flag so that we can start testing it on staging. 

All updated pages should be identical if you aren't using the feature flag (`?CHROME_APP_RELEASED=true` while in staging env)

Updates include
- `/download` - Updated language removing Scratch Desktop in favor of Scratch for <platform>
- `/ev3`, `/vernier`, `/microbit`, `/wedo2`, `/boost` all now have the new OS chooser with the Scratch app. 

/cc @rschamp @chrisgarrity


src/components/install-scratch/install-scratch.jsx
```diff
@@ -0,0 +1,146 @@
...
+                    <FormattedMessage
+                        id="installScratch.appHeaderTitle"
+                        values={{operatingsystem: currentOS}}
+                    />
+                ) : (
+                    <React.Fragment>
+                        {isDownloaded(currentOS) && (
+                            <FormattedMessage id="installScratch.desktopHeaderTitle" />
+                        )}
+                        {isFromGooglePlay(currentOS) && (
```
chrisgarrity: If the ChromeApp isn't released, then I don't think the `isFromGooglePlay` can ever be true.

src/components/extension-landing/extension-requirements.jsx
```diff
@@ -12,13 +14,64 @@ const ExtensionRequirements = props => (
             <FormattedMessage id="extensionHeader.requirements" />
         </span>
         <FlexRow className="extension-requirements">
-            {props.children}
+            {props.bluetoothStandard ? (
```
chrisgarrity: removed a lot of duplication - nice!

src/components/install-scratch/install-scratch.jsx
paulkaplan: Ah good point.

PR #3416, opened 2019-09-26
change password default; remove unused state var

benjiwheeler:
### Resolves:

### Changes:

On username step, default to showing password content (rather than masking it).

image

### DIFF
--- a/src/components/join-flow/username-step.jsx
+++ b/src/components/join-flow/username-step.jsx
@@ -20,7 +20,6 @@ class UsernameStep extends React.Component {
     constructor (props) {
         super(props);
         bindAll(this, [
-            'handleChangeShowPassword',
             'handleFocused',
             'handleSetUsernameRef',
             'handleValidSubmit',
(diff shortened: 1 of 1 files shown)


towerofnix:
I imagine there was some design discussion behind this, but I'm curious what the reasons for this change are? I'm always quite startled when I start typing in a password in registration inputs and it's revealed to me-- and to anyone who happens to be watching. Also, for almost all sites I've seen, *including* Scratch as it currently is, by default the password is masked. It seems to me like there ought to be some considerably good reasons to go against that standard. (For errors in typing -- the "retype password" box already prevents nearly all cases of that. And users can always toggle the box and unmask their password themselves, if they prefer.)

benjiwheeler:
@towerofnix I hear your concerns. Our current feeling on this is that this way, we can still protect privacy, and also improve clarity and accessibility. In testing with kids, we've observed that they often struggle with the password creation step, because they aren't sure what they've typed; also, especially when they are already struggling with keyboard input, they sometimes don't realize that they have the ability to unmask their password to check their input. This can sometimes make the account creation process frustrating and even insurmountable for them.

Users have several defenses against people knowing their passwords: they can manually check that no one is looking; they can mask the password, by clicking the checkbox; and they can change their password. I can't speak for everyone, but on balance, I feel that this choice is a better balance of accessibility with privacy than defaulting to masking the password, especially considering the experience of younger users.

That said, this design decision isn't set in stone and we intend to watch rollout and listen to the community to see if it sounds like we're balancing those concerns well.

PR #3423, opened 2019-10-01
make text selectable, so users can copy errors

benjiwheeler:
### Resolves:

### Changes:

Makes text in join-flow modal selectable, so users can copy error messages if necessary.


picklesrus:
I think the result of the discussion was that this should just go on the element containing the error message rather than the whole modal?

benjiwheeler:
After discussion, we agreed to only make error step selectable.

Changed!

PR #3424, opened 2019-10-01
splash page join button works with new join flow

benjiwheeler:
### Resolves:

### Changes:

makes join button on splash page either open existing join flow, or (if flag set to true) navigate user to /join.

### DIFF
--- a/src/components/intro/intro.jsx
+++ b/src/components/intro/intro.jsx
@@ -15,11 +15,19 @@ class Intro extends React.Component {
     constructor (props) {
         super(props);
         bindAll(this, [
+            'handleClickRegistration',
             'handleShowVideo'
         ]);
         this.state = {
             videoOpen: false
         };
+    }
+    handleClickRegistration (event) {
+        if (this.props.useScratch3Registration) {
+            this.props.navigateToRegistration(event);
+        } else {
+            this.props.handleOpenRegistration(event);
+        }
     }
     handleShowVideo () {
         this.setState({videoOpen: true});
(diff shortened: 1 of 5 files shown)


src/components/intro/intro.jsx
```diff
@@ -15,12 +15,20 @@ class Intro extends React.Component {
...
         super(props);
         bindAll(this, [
+            'handleClickRegistration',
             'handleShowVideo'
         ]);
         this.state = {
             videoOpen: false
         };
     }
+    handleClickRegistration (event) {
```
picklesrus: Is there a way to share this and navigateToRegistration between navigation.js and here? It'd be nice if the code were shared so we don't run into this again.

benjiwheeler: Hm, I think you're right, that would be better.

I rewrote the whole thing to refactor that logic into the reducer. take a look.

test/unit/redux/navigation.test.js
```diff
@@ -238,4 +247,28 @@ describe('unit test lib/validate.js', () => {
...
+            registrationOpen: false,
+            useScratch3Registration: true
+        };
+        const action = handleRegistrationRequested();
+        const resultState = navigationReducer(initialState, action);
+        expect(resultState.registrationOpen).toBe(false);
+        expect(global.window.location).toEqual('/join');
+    });
+
+    test('handleRegistrationRequested with useScratch3Registration false does NOT navigate user away, ' +
```
picklesrus: Yay tests!  What about the case where registrationOpen is set to true and so is useScratch3Registration?  I think it is worth making sure we have a test for that case too.

benjiwheeler: Good idea, done!

benjiwheeler:
Tests are performing differently in Travis vs. local... not sure why!

PR #3425, opened 2019-10-01
truncate custom gender, and other inputs

benjiwheeler:
### Resolves:

### Changes:

* when user changes content of the custom gender input, truncate the value to 25 chars -- the limit of the field we store in the database.
* do the same for username (30 chars), password and passwordConfirm (128 chars), and email (254 chars)

these values would be truncated on the backend anyway, but this will reduce the network throughput.

### DIFF
--- a/src/components/join-flow/email-step.jsx
+++ b/src/components/join-flow/email-step.jsx
@@ -174,7 +174,7 @@ class EmailStep extends React.Component {
                                 /* eslint-disable react/jsx-no-bind */
                                 onBlur={() => validateField('email')}
                                 onChange={e => {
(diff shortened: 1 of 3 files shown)


apple502j:
Wasn't the length limit for username 20?

benjiwheeler:
> Wasn't the length limit for username 20?

Yes, and it will continue to be. The reason I'm not using 20 here is that someone might want to have more space to work, to type or paste a longer string and then cut it down. It's somewhat arbitrary, but I'd rather the user see an error message explanation than just hit the limit in the input and be confused. We might revisit this.

PR #3427, opened 2019-10-01
turn off automplete and autocorrect for join-flow inputs

benjiwheeler:
### Resolves:

### Changes:

* adds several attributes to the input fields in the new join-flow. These attributes tell the browser to disable autocomplete and suggestions of past input.

### Screenshots

before:

bad-small

after:

Oct-01-2019 16-36-31

### DIFF
--- a/src/components/join-flow/username-step.jsx
+++ b/src/components/join-flow/username-step.jsx
@@ -143,6 +143,9 @@ class UsernameStep extends React.Component {
                                     {this.props.intl.formatMessage({id: 'registration.createUsername'})}
                                 </div>
                                 <FormikInput
+                                    autoCapitalize="off"
+                                    autoComplete="off"
+                                    autoCorrect="off"
                                     className={classNames(
                                         'join-flow-input'
                                     )}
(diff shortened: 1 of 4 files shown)


src/components/join-flow/username-step.jsx
```diff
@@ -170,6 +174,9 @@ class UsernameStep extends React.Component {
                                         {this.props.intl.formatMessage({id: 'registration.choosePasswordStepTitle'})}
                                     </div>
                                     <FormikInput
+                                        autoCapitalize="off"
+                                        autoComplete={values.showPassword ? 'off' : 'new-password'}
```
picklesrus: Why is there a distinction between new-password and off? Why wouldn't it be new-password regardless of whether the password is visible?

src/components/join-flow/username-step.jsx
```diff
@@ -195,6 +203,9 @@ class UsernameStep extends React.Component {
                                         /* eslint-enable react/jsx-no-bind */
                                     />
                                     <FormikInput
+                                        autoCapitalize="off"
+                                        autoComplete={values.showPassword ? 'off' : 'new-password'}
```
picklesrus: same question about the value of autocomplete here.

src/components/join-flow/username-step.jsx
benjiwheeler: It's because `showPassword` is also used by this input to determine whether it is `type="password"` or `type="input"` (really, just the default). 

AFAI can tell, "new-password" is the value that browsers look for if the input is type="password", and "off" is the value that browsers look for if the input is type="input".

src/components/join-flow/username-step.jsx
benjiwheeler: see above

src/components/join-flow/username-step.jsx
picklesrus: Cool. Thanks for explaining!

PR #3432, opened 2019-10-03
make join flow step scroll on mount

benjiwheeler:
### Resolves:

### Changes:

When a join flow step appears, scroll so that the top of it is in view.

This stops an annoying UI problem in small screens where you can't see the title of the form, because you scrolled down to click Next.

### Screenshots

say you're in a very short window (e.g., keyboard is open in android) and you click next...

image

...before this PR, the next screen would appear at the same scroll position:

image

...after this PR, the next screen will appear scrolled to the top of the modal:

image

...at least for some users!

### DIFF
--- a/src/components/join-flow/join-flow-step.jsx
+++ b/src/components/join-flow/join-flow-step.jsx
@@ -8,72 +8,82 @@ const ModalInnerContent = require('../modal/base/modal-inner-content.jsx');
 
 require('./join-flow-step.scss');
 
-const JoinFlowStep = ({
-    children,
-    innerClassName,
-    description,
-    descriptionClassName,
-    footerContent,
-    headerImgSrc,
-    nextButton,
-    onSubmit,
-    title,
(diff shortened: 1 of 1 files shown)


picklesrus:
I can't repro this on my android phone... scrolling down to click next seems to work okay without this change. Mostly asking because I'm worried about this approach - I don't really like messing with scroll position if we don't have to -  especially on the first step, which I think this code will run on.

Also, the test needs updating - I don't think it renders the full element so componentDidMount is probably throwing an error because it can't find the element that you're attempting to call scrollIntoView on.

benjiwheeler:
Closing this because it doesn't seem necessary. It doesn't seem to be needed on any of the major phones and simulated sizes we test on. Only on screen sizes like the iPhone 4 does it seem to be necessary; such sizes might exist in some places still, but this seems like an edge case that we should leave alone until we hear from a user that it was inconvenient.

PR #3436, opened 2019-10-07
Join flow info button has larger touch target, can be clicked to toggle

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/3527
Resolves https://github.com/LLK/scratch-www/issues/3760

### Changes:

* Makes info message have a toggleable state, that lets you either mouse in and out, or click the info button to make message persist. This addresses a problem with desktop touchscreens that @thisandagain observed.
* adds library lodash.debounce to reduce flickering from rollover states
* revises join flow info button to have invisible padding in its image, so that it has a larger touch target.

toggle info


benjiwheeler:
need to fix tests

benjiwheeler:
It's not clear if this is necessary; only Surface Edge has the problem. This change shouldn't make other platforms worse, but it increases the complexity.

benjiwheeler:
Dropping this PR because we felt it adds unnecessary complexity.

benjiwheeler:
I reopened, updated and resubmitted this for review because https://github.com/LLK/scratch-www/issues/3527 documented an additional case where the toggling behavior was desired: users who use the mouse to read along with text.

picklesrus:
Sorry for taking a long time to come back to this.

Can you say more about where the requests for toggle-ability came from that are mentioned in #3527 that prompted you to reopen it.
I'm still concerned about the complexity we're adding for not much gain.  Do you know how big the debounce library is?  And I'm not quite sure I understand why we need it. Does the pop up flicker? It doesn't seem to now.

benjiwheeler:
The "mouse along with text" use case was something that @kchadha identified in one of our internal playtests.

The debouncing was needed as a result of expanding the size of the touch target in this PR -- it's not needed without it.

I'd be ok with walking away from this PR, but if we do, we should close the related issues with a note that we don't intend to fix.

benjiwheeler:
btw, lodash debounce adds 938 bytes according to bundlephobia: https://bundlephobia.com/result?p=lodash.debounce@4.0.8

Currently, our default project page js for a logged out user is 5.8MB total. So this would add 938/(5.8*2^20), or 0.015% . I think that's negligible enough to be acceptable.

benjiwheeler:
OK, I actually figured out a slightly simpler way to avoid having to do the debouncing! So this is a tiny bit smaller and lower-impact.

benjiwheeler:
updated tests

Issue #3443, opened 2019-10-10
Sentry does not indicate which www ErrorBoundary caught error


benjiwheeler:
When Sentry reports an error caught in a www ErrorBoundary, it does not currently record _which_ www ErrorBoundary caught the error.

It also does not record the fact that the error was caught inside scratch-www code (vs., say, scratch-gui code)

That info would be useful!

PR #3444, opened 2019-10-10
Add Sentry to Join Flow; set Sentry tags on various ErrorBoundaries

benjiwheeler:
### Resolves:

* Resolves https://github.com/LLK/scratch-www/issues/3443
* Step towards resolving https://github.com/LLK/scratch-www/issues/3053

### Changes:

* adds Sentry support to Join Flow, by making the standalone join view initialize Sentry
* makes www's Sentry reporting code report to Sentry:
   * that the error was caught inside scratch-www
   * which ErrorBoundary caught the error

### Test Coverage:

Added some unsurprisingly complex tests :)

### DIFF
--- a/src/components/errorboundary/errorboundary.jsx
+++ b/src/components/errorboundary/errorboundary.jsx
@@ -17,6 +17,9 @@ class ErrorBoundary extends React.Component {
     componentDidCatch (error, errorInfo) {
         // Display fallback UI
         Sentry.withScope(scope => {
+            if (this.props.name) {
+                scope.setTag('errorboundary', this.props.name);
+            }
             Object.keys(errorInfo).forEach(key => {
                 scope.setExtra(key, errorInfo[key]);
             });
@@ -46,7 +49,8 @@ class ErrorBoundary extends React.Component {
     }
 }
 ErrorBoundary.propTypes = {
-    children: PropTypes.node
+    children: PropTypes.node,
+    name: PropTypes.string
 };
 
 module.exports = ErrorBoundary;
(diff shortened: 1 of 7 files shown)


benjiwheeler:
> I'm not sure where "embed" comes from, because it's in a file called embed-view.jsx, a component called EmbedView, and it doesn't wrap anything called embed.

I intended it to basically correspond to the "embed" route... but I agree that that's unclear.

I also think, looking at it now, that I made a mistake and need to add an additional <ErrorBoundary> around the embed view?

benjiwheeler:
Changed:
* moved ErrorBoundary out of embed-view.jsx and into embed.jsx, wrapping entire EmbedView
* changed Sentry tag key string from 'errorboundary' to 'component' per Ray's suggestion
* changed value strings for that tag to refer explicitly to the component that ErrorBoundary is immediately wrapping

src/components/errorboundary/errorboundary.jsx
```diff
@@ -46,7 +50,8 @@ class ErrorBoundary extends React.Component {
     }
 }
 ErrorBoundary.propTypes = {
-    children: PropTypes.node
+    children: PropTypes.node,
+    name: PropTypes.string
```
picklesrus: This is picky, but maybe this should be called componentName (or something like that) to indicate that we expect it to be the name of the component that the error originates from since the sentry tag is "component"?

benjiwheeler: Good idea, doing it

benjiwheeler: Done

benjiwheeler:
I think this is ready to merge. How should we test it once it's in staging?

benjiwheeler:
I'm going to merge this, then monitor it on staging.

benjiwheeler:
Follow-up: I made a hotfix branch with two places where I had it throw a custom error. Both were caught and registered in Sentry, how we would expect. And I was able to filter on the component name to see only those errors:

image


image

Seems like it's working!

PR #3478, opened 2019-10-23
[Develop] Update GUI to add 2 HoC tutorials

ericrosenbaum:


ericrosenbaum:
@rschamp what do you think we should do with this one?

Issue #3182, opened 2019-07-29
teacher registration form has two different font weights


benjiwheeler:
go to [https://scratch.mit.edu/educators/register](https://scratch.mit.edu/educators/register):

"Create a Username" and "Password" have two different font weights!

![image](https://user-images.githubusercontent.com/3431616/62055210-26ebe180-b1e9-11e9-82ef-75a781106dfb.png)

Kenny2github:
Note a few things:
a) The font weights are actually specified differently:
```css
.row label {
    font-weight: 500;
}
b, strong {
    font-weight: bold;
}
```
And the elements are different:
```html
<div class="username-label"><b>Create a Username</b></div>
```
for the username and
```html
<label class="control-label col-sm-3" data-required="true" for="frc-user.password--1310671131">Password<span class="required-symbol"> *</span></label>
```
for the password.
b) The difference is even more pronounced on my system, Chrome 75.0.3770.142 Windows 10 OS Version 1809 (Build 17763.615), where I don't have the site's preferred fonts:
![image](https://user-images.githubusercontent.com/28599280/62264870-da262780-b3ef-11e9-8eae-dc3eceb0e3df.png)
c) Looking at the HTML for the password label, there's a `<span class="required-symbol"> *</span>` there... doesn't show up at all. Should this be in a separate issue?
d) Description paragraph before the form is missing a space:
![image](https://user-images.githubusercontent.com/28599280/62264932-16598800-b3f0-11e9-87df-3e37366758f2.png)
This is because there's literally no space here in the HTML:
```html
...one day.</span><b><span>Please...
```
Fix by adding a nbsp. Should this be in a separate issue?

benjiwheeler:
@Kenny2github wrote:

>  Should this be in a separate issue?

Yes, feel free to create it!

benjiwheeler:
and thanks for the research, this will help fix it! Feel free to submit fixes for these -- seems like it should be fairly straightforward, now that you've looked into it!

benjiwheeler:
@ascii I'm taking the liberty of tagging this "help wanted" because it's a small change, and @Kenny2github added helpful context. (Please LMK if that's not the right thing to do!)

towerofnix:
@benjiwheeler I take it you meant to ping @thisandagain? :P

benjiwheeler:
> @benjiwheeler I take it you meant to ping @thisandagain? :P

Oops, yes, wrong username for this platform! cc: @thisandagain

[...]

PR #3482, opened 2019-10-27
fixed the issue "Create a Username" has different font weight

nandedamana:
### Resolves:

Fixes the issue #3182, Teacher Registration page having different font weights for username and password labels.

### Changes:

Uses label tage instead of div-b combination for 'Create a Username'. Also updated the classname of the parent element to match the bottom margin.

### Test Coverage:

Opened the page http://localhost:8333/educators/register to confirm the font weight is normal for the username label. No automatic test scripts needed.

### DIFF
--- a/src/components/registration/steps.jsx
+++ b/src/components/registration/steps.jsx
@@ -195,8 +195,8 @@ class UsernameStep extends React.Component {
                         }}
                         onValidSubmit={this.handleValidSubmit}
                     >
-                        <div>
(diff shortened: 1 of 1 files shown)


benjiwheeler:
Finally getting to this -- sorry for the delay!

This looks great. I also noticed a few other places that need to change (see https://github.com/LLK/scratch-www/issues/3589 ) -- any desire to fix those as well? :)

PR #3483, opened 2019-10-29
revise join flow username validation tests

benjiwheeler:
### Resolves:

### Changes:

revise, add to join flow username validation tests

### Tests:

added test for underscore; revised username local tests to make them clearer

### DIFF
--- a/src/lib/validate.js
+++ b/src/lib/validate.js
@@ -9,7 +9,7 @@ module.exports.validateUsernameLocally = username => {
         return {valid: false, errMsgId: 'registration.validationUsernameMinLength'};
     } else if (username.length > 20) {
         return {valid: false, errMsgId: 'registration.validationUsernameMaxLength'};
-    } else if (!/^[\w-]+$/i.test(username)) {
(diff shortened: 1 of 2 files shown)


src/lib/validate.js
```diff
@@ -9,7 +9,7 @@ module.exports.validateUsernameLocally = username => {
         return {valid: false, errMsgId: 'registration.validationUsernameMinLength'};
     } else if (username.length > 20) {
         return {valid: false, errMsgId: 'registration.validationUsernameMaxLength'};
-    } else if (!/^[\w-]+$/i.test(username)) {
+    } else if (!/^[\w_-]+$/i.test(username)) {
```
picklesrus: I thought \w was supposed to include underscores? I just created dinopickles_test through the new join flow...

benjiwheeler: aaaahhhhh you're right, OK, that makes it not so surprising that we left it like this!!!!!!

Took that change out, but including the test additions/revisions because I think those are good anyway

PR #3484, opened 2019-10-29
count unicode characters as single characters

benjiwheeler:
### Resolves:

### Changes:

count unicode characters as single characters

### Tests:

added tests for unicode; split tests up to be clearer

### DIFF
--- a/src/lib/validate.js
+++ b/src/lib/validate.js
@@ -50,7 +50,9 @@ module.exports.validateUsernameRemotely = username => (
 module.exports.validatePassword = (password, username) => {
     if (!password) {
         return {valid: false, errMsgId: 'general.required'};
-    } else if (password.length < 6) {
+    // get length of password, considering unicode symbols as single chars.
+    // see discussion at https://stackoverflow.com/a/54370584/2308190
+    } else if (Array.from(password).length < 6) {
         return {valid: false, errMsgId: 'registration.validationPasswordLength'};
     } else if (password === 'password') {
         return {valid: false, errMsgId: 'registration.validationPasswordNotEquals'};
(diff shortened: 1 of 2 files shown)


src/lib/validate.js
```diff
@@ -50,7 +50,9 @@ module.exports.validateUsernameRemotely = username => (
 module.exports.validatePassword = (password, username) => {
     if (!password) {
         return {valid: false, errMsgId: 'general.required'};
-    } else if (password.length < 6) {
+    // get length of password, considering unicode symbols as single chars.
+    // see discussion at https://stackoverflow.com/a/54370584/2308190
+    } else if (Array.from(password).length < 6) {
```
picklesrus: My reading of https://blog.jonnew.com/posts/poo-dot-length-equals-two and some tests with some emoji makes me think this isn't actually particularly helpful. It solves cow, but adds more mystery to some of the others.  e.g. 👩‍🚀.length is 5 and Array.from('👩‍🚀').length is 3.  Then if you change the skin tone, it changes even more.

benjiwheeler: Question: is this change a step forward, at least?

I did some more research and thought about it some more. It seems like there's no simple way to count all unicode combinations as single characters. However, at least it looks like we can distinguish these compound characters from the simple unicode symbols such as cow and most foreign language characters; it might be that this change serves the intention of the password length requirement, which is to require password complexity. In a sense, maybe "woman astronaut with light peach skintone" has similar complexity to three characters than to one, anyway.

benjiwheeler: Revised this comment...

Some characters that are treated differently by `Array.from(string).length` and `string.length`:

🍎
👨🏿‍🚀
🐮

Some characters which are treated the same by both:
counted as 1 character:
é
말
ጬ


...counted as 2 characters:
â
ē

benjiwheeler:
Coming back to this, I think this is at least a small step forward!

PR #3485, opened 2019-10-29
username step and email step caches ignore api failures

benjiwheeler:
### Resolves:

### Changes:

username step and email step caches ignore api failures

### Test Coverage:

no change to tests, as these validations check external API functionality that is not testable here

### DIFF
--- a/src/lib/validate.js
+++ b/src/lib/validate.js
@@ -50,9 +50,7 @@ module.exports.validateUsernameRemotely = username => (
 module.exports.validatePassword = (password, username) => {
     if (!password) {
         return {valid: false, errMsgId: 'general.required'};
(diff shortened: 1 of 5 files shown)


src/lib/validate.js
```diff
@@ -50,7 +50,9 @@ module.exports.validateUsernameRemotely = username => (
 module.exports.validatePassword = (password, username) => {
     if (!password) {
         return {valid: false, errMsgId: 'general.required'};
-    } else if (password.length < 6) {
+    // get length of password, considering unicode symbols as single chars.
```
benjiwheeler: Need to remove this

benjiwheeler: Done

PR #3491, opened 2019-10-30
[Master] Release 2019-03-07

kchadha:
# Release Notes
## Features
- Show all whitespace in monitors (thanks @towerofnix!)
- Correctly show translations for the Vernier extension landing page
- Update all editor translations 
## Bugfixes
- Fix an issue where the paint editor would crash using bitmap tools
- Fix an issue where deleting all of a costume would not update on the stage
- Fix an issue where empty costumes would not show the correct size
- Fix issues with changing the project title
- Fix line height in extension library descriptions
- Fix an issue where sometimes thumbnails in the backpack wouldn’t show up


apple502j:
~~don't trick the title here I gave sweets it's not March 7th~~

PR #3503, opened 2019-11-04
Add analytics logging to join flow.  

picklesrus:
Each step will log a separate page view.

The email step does not log in standalone mode because we already get a page view for loading the /join.

### DIFF
--- a/src/components/join-flow/join-flow.jsx
+++ b/src/components/join-flow/join-flow.jsx
@@ -215,6 +215,18 @@ class JoinFlow extends React.Component {
     resetState () {
         this.setState(this.initialState);
     }
+    sendAnalytics (path) {
+        const gaID = window.GA_ID;
+        if (!window.ga) {
+            return;
+        }
+        window.ga('send', {
+            hitType: 'pageview',
+            page: path,
+            tid: gaID
+        });
+    }
+
     render () {
         return (
             <React.Fragment>
@@ -228,11 +240,26 @@ class JoinFlow extends React.Component {
                     />
                 ) : (
                     <Progression step={this.state.step}>
-                        <UsernameStep onNextStep={this.handleAdvanceStep} />
-                        <CountryStep onNextStep={this.handleAdvanceStep} />
-                        <BirthDateStep onNextStep={this.handleAdvanceStep} />
-                        <GenderStep onNextStep={this.handleAdvanceStep} />
+                        <UsernameStep
+                            sendAnalytics={this.sendAnalytics}
+                            onNextStep={this.handleAdvanceStep}
+                        />
+                        <CountryStep
+                            sendAnalytics={this.sendAnalytics}
+                            onNextStep={this.handleAdvanceStep}
+                        />
+                        <BirthDateStep
+                            sendAnalytics={this.sendAnalytics}
+                            onNextStep={this.handleAdvanceStep}
+                        />
+
+                        <GenderStep
+                            sendAnalytics={this.sendAnalytics}
+                            onNextStep={this.handleAdvanceStep}
+                        />
+
                         <EmailStep
+                            sendAnalytics={this.sendAnalytics}
                             waiting={this.state.waiting}
                             onCaptchaError={this.handleCaptchaError}
                             onNextStep={this.handlePrepareToRegister}
@@ -240,6 +267,7 @@ class JoinFlow extends React.Component {
                         <WelcomeStep
                             createProjectOnComplete={this.props.createProjectOnComplete}
                             email={this.state.formData.email}
+                            sendAnalytics={this.sendAnalytics}
                             username={this.state.formData.username}
                             onNextStep={this.props.onCompleteRegistration}
                         />
--- a/src/components/join-flow/join-flow.jsx
+++ b/src/components/join-flow/join-flow.jsx
@@ -233,12 +233,8 @@ class JoinFlow extends React.Component {
                 {this.state.registrationError ? (
                     <RegistrationErrorStep
                         canTryAgain={this.canTryAgain()}
-<<<<<<< HEAD
                         errorMsg={this.state.registrationError.errorMsg}
-=======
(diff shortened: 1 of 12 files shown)


src/components/join-flow/join-flow.jsx
```diff
@@ -171,18 +183,34 @@ class JoinFlow extends React.Component {
                     />
```
benjiwheeler: Should we pass sendAnalytics to RegistrationErrorStep too?

src/components/join-flow/username-step.jsx
```diff
@@ -37,6 +37,12 @@ class UsernameStep extends React.Component {
         this.usernameRemoteCache = {};
     }
     componentDidMount () {
+        // Send info to analytics when we aren't on the standalone page.
```
benjiwheeler: I still feel confused about this... is there any way to keep from fragmenting our analytics this way?

src/components/join-flow/registration-error-step.jsx
```diff
@@ -14,6 +14,9 @@ class RegistrationErrorStep extends React.Component {
             'handleSubmit'
         ]);
     }
+    componentDidMount () {
+        this.props.sendAnalytics('join-error');
```
benjiwheeler: wrap this with `if (this.props.sendAnalytics) {...}` ?

test/unit/components/email-step.test.jsx
```diff
@@ -21,12 +22,17 @@ jest.mock('../../../src/lib/validate.js', () => (
 const EmailStep = require('../../../src/components/join-flow/email-step.jsx');
 
 describe('EmailStep test', () => {
+    const defaultProps = () => ({
```
benjiwheeler: I'm not clear on why you need this. You are passing a different sendAnalytics function when you actually test it, right?

src/components/join-flow/username-step.jsx
```diff
@@ -279,7 +285,8 @@ class UsernameStep extends React.Component {
 
 UsernameStep.propTypes = {
     intl: intlShape,
-    onNextStep: PropTypes.func
+    onNextStep: PropTypes.func,
+    sendAnalytics: PropTypes.func
```
benjiwheeler: General question that applies to all of these individual steps: we probably want to use `isRequired` on the sendAnalytics prop, right?

test/unit/components/email-step.test.jsx
```diff
@@ -170,33 +187,50 @@ describe('EmailStep test', () => {
...
+            <EmailStep
+                sendAnalytics={sendAnalyticsFn}
+            />);
+        expect(sendAnalyticsFn).toHaveBeenCalledWith('join-email');
+    });
+
     test('validateEmail test email empty', () => {
         const wrapper = shallowWithIntl(
-            <EmailStep />);
+            <EmailStep
```
benjiwheeler: do you need to use defaultProps() in these cases?

src/components/join-flow/registration-error-step.jsx
benjiwheeler: (also applies to most other individual steps)

src/components/join-flow/username-step.jsx
benjiwheeler: (also applies to most other individual steps)

benjiwheeler:
This is looking great! A few questions and one or two spots that need changing.

I also could use some help in testing this...

src/components/join-flow/username-step.jsx
picklesrus: Whoops. I didn't give enough context.  Here are the options I was considering. Option 1 is what this PR currently does.

1. Have standalone join flow start one page view (e.g. join) and modal join flow start with a different page view (e.g. join-email or join-modal is better?).  That's what this PR currently does.
2. Have standalone join flow and join flow in the modal log the same thing. In that case I'd change this PR to send 'join' here instead of 'join-email'
3. Have standalone join flow log an pageview for the loading of the /join page since analytics does this by default and also log a join-email pageview.  In this case, analytics gives a warning that the events are too close together since they're essentially the same page load.

src/components/join-flow/registration-error-step.jsx
picklesrus: If I make sendAnalytics required like you suggested below, would you still suggest adding the if?  I think you're right that it should be required.

test/unit/components/email-step.test.jsx
picklesrus: Nope. In the test that actually tests it, I'm just testing that the prop function gets called with the right parameters.  The defaultProps is here because I didn't guard the call in componentDidMount with a check for whether the prop exists so I have to make it exist.

picklesrus:
Good questions! I'll bring up the question of how we want to log page views at the meeting tomorrow unless we chat about it before then.  
For the other pieces, I'll make code changes once we figure out the isRequired piece.

src/components/join-flow/username-step.jsx
```diff
@@ -37,6 +37,12 @@ class UsernameStep extends React.Component {
         this.usernameRemoteCache = {};
     }
     componentDidMount () {
+        // Send info to analytics when we aren't on the standalone page.
+        // If we are on the standalone join page, the page load will take care of this.
+        if (!window.location.pathname.includes('/join')) {
+            this.props.sendAnalytics('join-email');
```
benjiwheeler: why is this string 'join-email'? What is the email connection?

test/unit/components/email-step.test.jsx
benjiwheeler: OK

test/unit/components/email-step.test.jsx
benjiwheeler: OK

picklesrus:
That was some ugly merge conflict resolution, but I think this is *finally* ready for another look.

benjiwheeler:
LGTM!

PR #3505, opened 2019-11-05
Revert "Revert "fixed country options to use full country name string as option value""

### LINKED ISSUE #3467: Correct country name string values
https://github.com/LLK/scratch-www/pull/3199 introduced a bug: we are sending country codes, like "br", instead of country names, like "Brazil".

So far, this only affects teacher accounts, which have all been defaulting to "United States". But this needs to be fixed ASAP for both teacher accounts and the forthcoming new join flow.

benjiwheeler:
See original PR https://github.com/LLK/scratch-www/pull/3468 and revert https://github.com/LLK/scratch-www/pull/3495

From original:

### Resolves:
Resolves #3467

### Changes:
* for new join flow, sends country name strings, rather than country codes, when user registers
* for existing www registrations (teacher, student), leaves them unchanged
* renames and adds country data functions to be specific about whether they use country code, or country name, as option value

### Test Coverage:
Added test to verify that options values are either country codes, or full country name strings

### DIFF
--- a/src/components/join-flow/country-step.jsx
+++ b/src/components/join-flow/country-step.jsx
@@ -27,7 +27,7 @@ class CountryStep extends React.Component {
     }
     setCountryOptions () {
         if (this.countryOptions.length === 0) {
-            this.countryOptions = [...countryData.registrationCountryOptions];
+            this.countryOptions = [...countryData.registrationCountryNameOptions];
             this.countryOptions.unshift({ // add placeholder as first option
                 disabled: true,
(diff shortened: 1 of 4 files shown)


benjiwheeler:
Post-merge testing:

I was able to create 2 teacher accounts, one with country USA, one with Brazil:

image

benjiwheeler:
Was also able to create two user accounts in new join flow with country Brazil and country USA:

image

PR #3512, opened 2019-11-12
remove truncation of password, passwordConfirm

benjiwheeler:
### Resolves:

Step towards resolving https://github.com/LLK/scratch-www/issues/3053

Resolves https://github.com/LLK/scratch-www/issues/3511

### Changes:

Removes truncation of password and passwordConfirm input values, effectively allowing them to be any arbitrary length

### Test Coverage:

None, this area did not have tests written

PR #3516, opened 2019-11-13
Put the right value in the password field. This got mixed up in #3512.

picklesrus:
Fix of something broken in #3512


benjiwheeler:
LGTM!

PR #3541, opened 2019-11-15
[Master] Hotfix/join flow 

picklesrus:
Enables new join flow
Points link at the end of the flow to the homepage


apple502j:
Question: announcement in the forum or no?

picklesrus:
@apple502j We aren't currently planning on making a forum post. Existing users won't notice since they already have accounts.

PR #3545, opened 2019-11-18
Goodbye jobs page

apple502j:
### Resolves:
Resolves LLK/scratch-gui#5302

### Changes:
Removes Jobs page and redirect it to Foundation's up-to-date page. Also link & test changes

### Test Coverage:
Manual testing:
1. Footer links to new site
2. `/jobs` and `/jobs/moderator` redirects
3. Developer Page has link to new site


src/routes.json
```diff
@@ -148,18 +148,16 @@
...
+        "name": "jobs-redirect",
         "pattern": "^/jobs/?$",
         "routeAlias": "/jobs(/moderator)?/?$",
-        "view": "jobs/jobs",
-        "title": "Jobs"
+        "redirect": "https://www.scratchfoundation.org/opportunities/"
     },
     {
-        "name": "jobs-moderator",
+        "name": "jobs-redirect2",
```
paulkaplan: You can remove this whole block, since we were only linking to it internally and now will only be linking to `/jobs`

src/routes.json
```diff
@@ -148,18 +148,16 @@
         "title": "Ideas"
     },
     {
-        "name": "jobs",
+        "name": "jobs-redirect",
         "pattern": "^/jobs/?$",
         "routeAlias": "/jobs(/moderator)?/?$",
```
paulkaplan: You can remove the `routeAlias` key, we no longer need since we need to remove the whole jobs-moderator block (see comment below)

src/routes.json
apple502j: How do you know that? Someone would have linked that page and it's not good to make it 404.

src/routes.json
```diff
@@ -148,18 +148,16 @@
         "title": "Ideas"
     },
     {
-        "name": "jobs",
+        "name": "jobs-redirect",
         "pattern": "^/jobs/?$",
```
paulkaplan: While we are updating this, one thing about this kind of pattern is that it won't match URLs generated when shared on social media or through email campaigns, which often add `?bunch_of_trackers=1l23kj12l3kj...` query parameters to URLs automatically. We started adding this for simple routes like these that get linked to frequently in email/social media. Could you add the pattern that you see e.g. at the end of the `/parents` route on line 193 to the end of here?

src/routes.json
apple502j: Ah! Makes sense, you mean merging it to jobs?

src/routes.json
apple502j: ✔️ Done

src/routes.json
apple502j: ✔️ Done

src/routes.json
apple502j: ✔️ Done

PR #3547, opened 2019-11-18
Use indexOf instead of includes because of old browsers.

picklesrus:


benjiwheeler:
Tested locally, it did -- and didn't -- try to send to GA where I expected it too!

PR #3564, opened 2019-11-22
[Master] Work around replication lag problems on project creation

rschamp:
### Resolves:

_What Github issue does this resolve (please include link)? Please do not submit PRs that only partially implement an issue. Please do not submit PRs that are not associated with a Github issue, or that only partially implement an issue._

In most cases, resolves the error screen seen when creating, remixing, or copying projects when our databases are experiencing replication lag.  If the lag is longer than 7.5s, the behavior remains the same (appears broken).

### Changes:

_Describe what this Pull Request does. Be sure to include a brief description of the issue the Pull Requests solves too._

Makes sure api can get the project before considering the project create to be complete. We ought to refactor gui and www so that we can use the project data we get back from the project create instead.

### Test Coverage:

_Please show how you have added tests to cover your changes or describe how you have tested the changes (include a screenshot if possible)._

:(

PR #3565, opened 2019-11-22
[Develop] Work around replication lag problems on project creation

rschamp:
See #3564


src/views/preview/project-view.jsx
```diff
@@ -219,6 +223,84 @@ class Preview extends React.Component {
...
+                }, (err, projectInfo, response) => {
+                    if (err) {
+                        log.error(`Could not fetch project after creating: ${err}`);
+                        return resolve(body);
+                    }
+                    if (typeof body === 'undefined' || response.statusCode === 404) {
+                        // Retry after 500ms, 1.5s, 3.5s, 7.5s and then stop.
+                        if (count > 4) {
+                            return resolve(body);
+                        }
```
benjiwheeler: What does the line below do? Does it really return what we want it to return?

picklesrus: Technically it is returning the id from setTimeout, but we don't actually care about the return value. We want it to stop the function execution. What really matters is the resolve a few lines below.

paulkaplan:
@rschamp @picklesrus now that the code from save-project-to-server in GUI is totally unused, can you file an issue to remove it?

Issue #3539, opened 2019-11-15
Replication lag can prevent new user from appearing to be logged in, including a new student


benjiwheeler:
In Join Flow, replication lag can prevent new user from appearing to be logged in.

This is a problem with both the existing Join Flow, and the new Join Flow launching in November 2019.

One proposed partial solution is to have the new Join Flow, after hearing back that the user was registered, repeatedly retry refreshing the session until it sees the logged in sessionid for the user; only after seeing this should it show the user the final, "Welcome" step.

jakel181:
I have experienced this issue while creating student accounts from a student sign up class registration link from a teacher. When I clicked "next step" on the second step nothing happen and I clicked "next step" again and it said that "this username has already been taken" I proceed to try to make two more accounts without know it was actually creating the accounts, thinking that each username I tried to use was already taken. It wasn't until after the third time and checked, and saw the account were being created.

benjiwheeler:
@jakel181 Thanks so much for bringing this to our attention.

So to make sure I understand you, are you saying that this happens:

1. the teacher creates a registration link and gives it to the class
1. a student goes to that URL, and chooses a username and password
1. the student clicks "next step" and it seems like nothing happens
1. the student clicks "next step" again, and it says "this username has already been taken"

Does that match what you're seeing?

jakel181:
Yes sir! It happens when you click "next step" on the personal information page.
So it's more like this:
1.The teacher creates a registration link and gives it to the class
2. a student goes to that URL, and chooses a username and password
3. The student clicks "next step" and moves on to the personal information page
4. the student fills out the forum with birth month, year etc.
5. the student clicks "next step" and it seems like nothing happens
6. the student clicks "next step" again, and it says "this username has already been taken"

benjiwheeler:
It would be good to confirm that this issue has been fixed by #3640.

PR #3576, opened 2019-12-04
Retry fetching session until it succeeds, when registering using join flow

benjiwheeler:
### Resolves:

Alleviates the problem of https://github.com/LLK/scratch-www/issues/3539 , even if it doesn't really solve the underlying problem.

### Changes:

* introduces lib/session.js, which has a `requestSessionWithRetry()` function which retries fetching session until it either gets one with a registered user, or runs out of tries
* adds a `refreshSessionWithRetry()` function to /redux/session.js, which replicates functionality of `refreshSession()` but calls `requestSessionWithRetry()` instead
* changes `join-flow.jsx` to use `refreshSessionWithRetry()` instead of `refreshSession()`; this means only setting the `state.waiting` state variable to `false` after the refreshSessionWithRetry promise resolves
* changes redux/navigation.js's `handleCompleteRegistration()` function to use `refreshSessionWithRetry()` instead of `refreshSession()`; this provides a second chance to obtain a valid session before dismissing the registration modal, when user joins from within the editor.

### Testing locally:

A good way to test is to set lib/api.js to return an empty session for most, but not all, session requests.

These changes make that happen, 70% of the time:

image

The result is that, _for this experiment_, 84% (1 - .7 ^ 5) of attempts to register succeed in retrieving the session after registration, but it often takes a second or two.

### Test Coverage:

No added or changed tests. Should I try testing lib/session.js by mocking lib/api.js?

### DIFF
--- a/src/components/join-flow/join-flow.jsx
+++ b/src/components/join-flow/join-flow.jsx
@@ -121,13 +121,16 @@ class JoinFlow extends React.Component {
     //   * "birth_month": ["Ensure this value is greater than or equal to 1."]
     handleRegistrationResponse (err, body, res) {
         this.setState({
-            numAttempts: this.state.numAttempts + 1,
-            waiting: false
+            numAttempts: this.state.numAttempts + 1
         }, () => {
             const success = this.registrationIsSuccessful(err, body, res);
             if (success) {
-                this.props.refreshSession();
-                this.setState({step: this.state.step + 1});
+                this.props.refreshSessionWithRetry().then(() => {
+                    this.setState({
+                        step: this.state.step + 1,
+                        waiting: false
+                    });
+                });
                 return;
             }
             // now we know something went wrong -- either an actual error (client-side
@@ -135,7 +138,10 @@ class JoinFlow extends React.Component {
 
             // if an actual error, prompt user to try again.
             if (err || res.statusCode !== 200) {
-                this.setState({registrationError: {errorAllowsTryAgain: true}});
+                this.setState({
+                    registrationError: {errorAllowsTryAgain: true},
+                    waiting: false
+                });
                 return;
             }
 
@@ -164,7 +170,8 @@ class JoinFlow extends React.Component {
                 registrationError: {
                     errorAllowsTryAgain: false,
                     errorMsg: errorMsg
-                }
+                },
+                waiting: false
             });
         });
     }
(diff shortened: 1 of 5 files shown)


src/redux/session.js
```diff
@@ -103,3 +104,46 @@ module.exports.refreshSession = () => (dispatch => {
         return;
     });
 });
+
+module.exports.refreshSessionWithRetry = () => (dispatch => {
```
benjiwheeler: I left `refreshSession()` intact, as opposed to refactoring the shared code out, because I wanted this PR to not be too complicated. If we go in this direction, we could refactor in a follow-up PR, perhaps by moving both calls to `api()` into a single function in lib/session.js, and perhaps by moving the various redirects into a separate function in lib/session.js or somewhere else.

src/components/join-flow/join-flow.jsx
```diff
@@ -121,21 +121,27 @@ class JoinFlow extends React.Component {
     //   * "birth_month": ["Ensure this value is greater than or equal to 1."]
     handleRegistrationResponse (err, body, res) {
         this.setState({
-            numAttempts: this.state.numAttempts + 1,
-            waiting: false
+            numAttempts: this.state.numAttempts + 1
```
benjiwheeler: we must not set waiting to false yet -- not until we know the session retries are over.

src/components/join-flow/join-flow.jsx
```diff
@@ -121,21 +121,27 @@ class JoinFlow extends React.Component {
...
         }, () => {
             const success = this.registrationIsSuccessful(err, body, res);
             if (success) {
-                this.props.refreshSession();
-                this.setState({step: this.state.step + 1});
+                this.props.refreshSessionWithRetry().then(() => {
```
benjiwheeler: Note that -- at least for now -- we are not specifically checking to see if the user session did eventually come through. If it didn't, here we're not bailing on the registration -- we're still showing the user the Welcome step, with their username (stored locally from before).

Alternatively, we could do something else... but I'm not sure what.

src/components/join-flow/join-flow.jsx
```diff
@@ -121,21 +121,27 @@ class JoinFlow extends React.Component {
...
             }
             // now we know something went wrong -- either an actual error (client-side
             // or server-side), or just a problem with the registration content.
 
             // if an actual error, prompt user to try again.
             if (err || res.statusCode !== 200) {
-                this.setState({registrationError: {errorAllowsTryAgain: true}});
+                this.setState({
+                    registrationError: {errorAllowsTryAgain: true},
+                    waiting: false
```
benjiwheeler: Since I took out the `setState({...waiting:false})` above, I have to be sure to set it in all these cases, or the user will be stuck with a spinner forever...

PR #3583, opened 2019-12-04
Add recognition text and links to the bottom of the page.

picklesrus:
Adds a prop to the Page component so each page can say whether or not it 
wants them to appear.


picklesrus:
@rschamp  reviewed offline so committing.

Issue #2103, opened 2018-09-25
Update error messages for splash page


benjiwheeler:
`src/redux/splash.js` includes lots of error messages that complain about a missing session, but these are not session-related errors! Those messages are probably just copy-and-paste remnants.

PR #3590, opened 2019-12-06
corrected error messages that wrongly mentioned session

benjiwheeler:
### Resolves:

https://github.com/LLK/scratch-www/issues/2103

### Changes:

* fixes error messages that erroneously mentioned session, to instead describe more accurately what's going on!

Note that these still don't seem to be actually logged anywhere.

### DIFF
--- a/src/redux/messages.js
+++ b/src/redux/messages.js
@@ -211,7 +211,7 @@ module.exports.getMessages = (username, token, opts) => {
             }
             if (typeof body === 'undefined') {
                 dispatch(module.exports.setStatus('MESSAGE_STATUS', module.exports.Status.MESSAGES_ERROR));
-                dispatch(module.exports.setMessagesError('No session content'));
+                dispatch(module.exports.setMessagesError('getMessages: response had no content'));
                 return;
             }
             dispatch(module.exports.setStatus('MESSAGE_STATUS', module.exports.Status.FETCHED));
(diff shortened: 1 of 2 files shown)


src/redux/messages.js
```diff
@@ -211,7 +211,7 @@ module.exports.getMessages = (username, token, opts) => {
             }
             if (typeof body === 'undefined') {
                 dispatch(module.exports.setStatus('MESSAGE_STATUS', module.exports.Status.MESSAGES_ERROR));
-                dispatch(module.exports.setMessagesError('No session content'));
+                dispatch(module.exports.setMessagesError('getMessages: response had no content'));
```
kchadha: Is this a new style of error message or do we use messages like this elsewhere (e.g. the `getENDPOINT` structure...?

benjiwheeler:
> Also could you clarify what you mean by: “Note that these still don’t seem to be actually shown anywhere...“?
> Do we not know where those messages are logged to?

By "actually shown anywhere" I meant, brought to the user's attention, or to our attention.
I don't _think_ they're logged to anywhere persistent; I think they're only logged to the user's browser console.

src/redux/messages.js
benjiwheeler: I think of the existing style as being structureless -- just a text string that tries to communicate what's going on. In my change, I'm not meaning to impose or propose a new systematic style, just to more accurately and helpfully communicate what's going on, which is that a particular request's response had no content.

Issue #2853, opened 2019-03-21
Text overflow in project Notes and Credits


Kenny2github:
Text overflows with really long links:
![image](https://user-images.githubusercontent.com/28599280/54734318-aa614a80-4bd9-11e9-8151-021f4c80c017.png)

### Steps to Reproduce

1. Go to https://scratch.mit.edu/projects/10135908/
2. Observe

### Operating System and Browser

Windows 10 OS Build 17763.437, Chrome 73.0.3683.103

Smrman:
Just add the line 
```css
.project-description {
    word-break: break-word;
}
```
to https://github.com/LLK/scratch-www/blob/develop/src/views/preview/preview.scss#L466

_Before_
![image](https://user-images.githubusercontent.com/17651422/68261294-ec8fad00-003f-11ea-9344-6d772702e183.png)
_After_
![image](https://user-images.githubusercontent.com/17651422/68261262-cff37500-003f-11ea-9675-3e7a992734a9.png)

I agree it might not be the prettiest solution, but at least it works and the layout is fixed...
Also, what are the odds for an user to type a 55+ characters long URL?

benjiwheeler:
Looks like https://github.com/LLK/scratch-www/pull/3599 didn't get it right!

This is https://scratch.mit.edu/projects/361718885/ :
![image](https://user-images.githubusercontent.com/3431616/75998748-4d84a000-5ecf-11ea-8645-d567e68fc5c1.png)

Yikes! Also see ModTools' user verification project, https://scratch.mit.edu/projects/10135908/ :
![image](https://user-images.githubusercontent.com/3431616/75998841-6e4cf580-5ecf-11ea-9890-66340557b69e.png)

The main problem is that we should have used `overflow-wrap: anywhere`
Seems like the right combination to use is actually:

```
overflow-wrap: anywhere;
word-break: keep-all;
```

Adding those, the problem seems fixed:

![image](https://user-images.githubusercontent.com/3431616/75999479-44e09980-5ed0-11ea-8326-17436a324ec9.png)

![image](https://user-images.githubusercontent.com/3431616/75999502-4c07a780-5ed0-11ea-9fc0-842e4ab65f39.png)

## Why `keep-all`?

`keep-all` is better than `normal` (or just omitting `word-break:` because it handles Chinese and other Asian languages better. With `normal`:

[...]

PR #3599, opened 2019-12-12
break word in project description

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/2853

### Changes:

Implements the fix that @Smrman proposed in https://github.com/LLK/scratch-www/issues/2853#issuecomment-550104414 , which is to apply the `word-break: break-word` style to the `project-description` class.

### Screenshots:

Chrome:

Before:

image

After:

image

Firefox:

Before: (narrow window size):

image

After: (narrow window size):

image

Safari:

Before:

image

After:

image

### DIFF
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -475,6 +475,7 @@ $stage-width: 480px;
         font-size: 1rem;
         line-height: 1.5rem;
         flex: 1;
+        word-break: break-word;
     }
 
     .project-description:last-of-type {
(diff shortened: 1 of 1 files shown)


benjiwheeler:
Looks like it's `word-wrap` that is deprecated in favor of `overflow-wrap` -- the former is currently an alias for the latter.

`work-break` is something different, apparently.

Seems from some research ( https://stackoverflow.com/questions/1795109/what-is-the-difference-between-word-break-break-all-versus-word-wrap-break ) like it would make sense to include both.

benjiwheeler:
> I disagree, the word-break: break-word option is specifically deprecated in favor of overflow wrap.

Oh, I see, you're right. Removing!

PR #3606, opened 2019-12-17
[MASTER] Hotfix/render bugfix

picklesrus:
### Resolves:
Fixes bug in scratch-svg-renderer


apple502j:
It's during HoC, LLK/scratch-gui#5319 (renderer bump) was reverted because "We want to be conservative about deploys before HOC", the fix was kind of performance fix therefore not critical, why did you choose to release it now, as hotfix?

PR #3610, opened 2019-12-19
Update translation config

chrisgarrity:
Add 2020 conference view to `.tx/config` so that conference strings are pushed to transifex.
Jobs view has been removed, so remove from the config to prevent errors.


chrisgarrity:
Closing because https://github.com/LLK/scratch-www/pull/3643 will make it unnecessary

PR #3627, opened 2020-01-09
fix for session promise issue

benjiwheeler:
### Resolves:

Fixes a problem with https://github.com/LLK/scratch-www/pull/3618 where it was treating a non-promise, undefined object like a promise

### Changes:

Instead of calling `then` on a function call that returns the undefined object, calls `then` on the Promise that wraps the function call.


benjiwheeler:
closing this, in favor of https://github.com/LLK/scratch-www/pull/3640

PR #3640, opened 2020-01-15
Retry of the change that retries fetching session

benjiwheeler:
### Resolves:

Partially addresses https://github.com/LLK/scratch-www/issues/3539 , even if it doesn't really solve the underlying problem.

### Changes:

* introduces lib/session.js, which has a `requestSessionWithRetry()` function which retries fetching session until it either gets one with a registered user, or runs out of tries
* adds a `refreshSessionWithRetry()` function to /redux/session.js, which replicates functionality of `refreshSession()` but calls `requestSessionWithRetry()` instead
* changes `join-flow.jsx` to use `refreshSessionWithRetry()` instead of `refreshSession()`; this means only setting the `state.waiting` state variable to `false` after the refreshSessionWithRetry promise resolves
* changes redux/navigation.js's `handleCompleteRegistration()` function to use `refreshSessionWithRetry()` instead of `refreshSession()`; this provides a second chance to obtain a valid session before dismissing the registration modal, when user joins from within the editor.

### Testing locally:

A good way to test is to set lib/api.js to return an empty session for most, but not all, session requests.

These changes make that happen, 70% of the time:

image

The result is that, _for this experiment_, 84% (1 - .7 ^ 5) of attempts to register succeed in retrieving the session after registration, but it often takes a second or two.

### Test Coverage:

Added tests of new session functionality, old functionality, and registration use of session calling

### DIFF
--- a/src/redux/session.js
+++ b/src/redux/session.js
@@ -1,7 +1,7 @@
 const keyMirror = require('keymirror');
 const defaults = require('lodash.defaults');
 
-const api = require('../lib/api');
+const {requestSession, requestSessionWithRetry} = require('../lib/session');
 const messageCountActions = require('./message-count.js');
 const permissionsActions = require('./permissions.js');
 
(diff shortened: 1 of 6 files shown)


benjiwheeler:
@picklesrus @rschamp Note that the only change from the previous hotfix branch is the "fix for session promise issue" commit above

src/redux/session.js
```diff
@@ -99,20 +99,20 @@ const handleSessionResponse = (dispatch, body) => {
...
 });
 
 module.exports.refreshSessionWithRetry = () => (dispatch => {
     dispatch(module.exports.setStatus(module.exports.Status.FETCHING));
-    return new Promise((resolve, reject) => (
-        requestSessionWithRetry(resolve, reject, 4, 7500)
-    )).then(body => {
+    return new Promise((resolve, reject) => {
+        requestSessionWithRetry(resolve, reject, 4, 7500);
+    }).then(body => {
```
picklesrus: What was the error you were getting before? Was it  both here and above? In this case you were returning the call to requestSessionWithRetry as an object literal? At least that's what I understand the () syntax to mean vs {}.  Are they functionally any different right here? I like this better, I just want to make sure I understand the root of the actual problem.

benjiwheeler:
Reviewed this with @picklesrus, and tested locally with both regular session fetches (on page view/render, on login) and retry fetches (on join/register, both after email step and after completing registration).

src/redux/session.js
benjiwheeler: We spoke about this in person. @picklesrus pointed out that these lines' change from () to {} don't change functionality; they only clarify that the return value of `requestSessionWithRetry` doesn't matter.

picklesrus:
LGTM

Issue #3215, opened 2019-08-05
Make pushing strings to transifex easier


benjiwheeler:
Currently, if someone wants to push strings from current l10n files up to Transifex, they must have the `tx` commandline program installed. But this is inconvenient.

@chrisgarrity writes:
> we should add the transifex npm package and write a script that will do the tx push for the l10n files. Then all anyone needs is a TX_TOKEN, and if it's not someone who's on transifex in any case they can just use the shared one.

> Essentially we need the www version of https://github.com/LLK/scratch-l10n/blob/master/scripts/tx-push-src.js

chrisgarrity:
Actually I was just looking and https://github.com/LLK/scratch-l10n/blob/master/scripts/tx-push-src.js has the basis of what's needed. But it's designed for just one file,

i.e. `tx-push-src scratch-website general-l10njson ./src/l10n.json`

We need something that does it for all the l10n files.

benjiwheeler:
Current thinking: let's tie the update to when developers merge to the develop branch.

This will allow translators the earliest reasonable access to the strings, so that many translations can be in place already when we deploy.

We know there is some string thrashing, but in most cases there is none or it settles down quickly.

We also think this will be better than tying the update to a cron job, because it will be in clearer sync with the version we are preparing to deploy, and there will not be a separate, unique point of failure.

benjiwheeler:
Described some of our longer-term intentions to consolidate route and transifex resource strings here: https://github.com/LLK/scratch-www/issues/3642

benjiwheeler:
Turns out https://github.com/LLK/scratch-www/pull/3643 did not correctly connect Travis cron to the relevant new translations push job.

PR #3643, opened 2020-01-16
Introduce new tx-push-www script to automate string push to transifex

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/3215

### Changes:

* introduces new tx-push-www script to automate string push to transifex
* sets travis to call tx-push-www on cron job running
* removes tips and preview-faq views which are obsolete
* removes old transifex tx push config file

### Test Coverage:

Not directly testable...

We ran the script and confirmed that it seems to correctly push to transifex!


bin/tx-push-www
```diff
@@ -0,0 +1,81 @@
...
+const execSync = require('child_process').execSync;
+
+// determine if this is a dry run, or if we're really pushing to transifex
+let execute = false;
+const args = process.argv.slice(2);
+if (args[0] === '--execute') {
+    process.stdout.write('pushing to transifex...\n');
+    execute = true;
+} else {
+    process.stdout.write('Dry run: pass "execute" as a parameter to add --execute switch to commands\n');
```
chrisgarrity: One little nit, the message should be something like: `'Dry run: pass "--execute" as a parameter to execute push files to transifex\n'`

benjiwheeler: Changed!

PR #3651, opened 2020-01-23
attempt to get travis cron job to only run translations

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/3215 , to correct a mistake in https://github.com/LLK/scratch-www/pull/3643 .

### Changes:

Second attempt to configure .travis.yml file to cause Travis's develop branch cron job to trigger our translation push script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -122,7 +122,6 @@ jobs:
       script: npm run deploy
       on:
         repo: LLK/scratch-www
-        condition: $TRAVIS_EVENT_TYPE != cron
         branch:
         - develop
         - hotfix/*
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -135,13 +136,10 @@ jobs:
         - master
   - stage: smoke
     script: npm run test:integration:remote
+  - stage: update translations
```
chrisgarrity: After re-reading the [conditional builds, jobs, and stages doc](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/) I think we're mixing up jobs and stages here.
If we keep update translations as a stage, then to be consistent, this part should only define the script, and the condition should be in the stages section below.

.travis.yml
```diff
@@ -135,13 +136,10 @@ jobs:
...
+    script: npm run i18n:push
+    if: branch == develop AND type == cron
 stages:
 - test
 - name: smoke
   if: type != pull_request
```
chrisgarrity: I think this is going to run the smoke test on every `cron` job unless we change it to `type NOT IN (cron, pull_request`

benjiwheeler: OK, take a look now

.travis.yml
benjiwheeler: OK, take a look now

PR #3667, opened 2020-01-30
Really delete the cookie

chrisgarrity:
It appears that setting the domain on the cookie prevents it from actually deleting the cookie.


picklesrus:
LGTM

PR #3676, opened 2020-02-05
use api instead of scratchr2 to validate usernames

benjiwheeler:
### Resolves:

* Reverts the core change of https://github.com/LLK/scratch-www/pull/3507
* Sets up https://github.com/LLK/scratch-api/pull/884 to be tested/merged, after this is deployed

### Changes:

* uses api endpoint, rather than scratchr2, to validate usernames.

Feb-05-2020 15-47-27

### DIFF
--- a/src/lib/validate.js
+++ b/src/lib/validate.js
@@ -20,16 +20,14 @@ module.exports.validateUsernameLocally = username => {
 module.exports.validateUsernameRemotely = username => (
     new Promise(resolve => {
         api({
-            host: '', // not handled by API; use existing infrastructure
-            uri: `/accounts/check_username/${username}/`
+            uri: `/accounts/checkusername/${username}/`
         }, (err, body, res) => {
             if (err || res.statusCode !== 200) {
                 resolve({requestSucceeded: false, valid: false, errMsgId: 'general.error'});
             }
             let msg = '';
-            if (body && body[0]) {
-                msg = body[0].msg;
-            }
+            if (body && body.msg) msg = body.msg;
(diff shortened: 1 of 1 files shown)


picklesrus:
Code looks good. Do you have a testing plan for this one? Might be too much work, but my ideal thought would be to pull a day (or maybe  a week's) worth of requests to the current prod endpoint, then throw them at this new api endpoint and see if there are any differences.

benjiwheeler:
> Code looks good. Do you have a testing plan for this one? Might be too much work, but my ideal thought would be to pull a day (or maybe a week's) worth of requests to the current prod endpoint, then throw them at this new api endpoint and see if there are any differences.

I like that! I think it should be possible by exporting from logentries. Let me take a look...

benjiwheeler:
OK! It took quite a bit of work, but I took 20 hours of check_username URL requests from logentries, wrote a script to request both the scratchr2 and api versions of the endpoints, and compared the results.

In a nutshell, everything looks good.

There was only one source of difference: scratchr2 applies badword filtering and checks for existing matches before basic checking like string length.

So, for example:
* scratchr2 replies that "Amirascratchwonderful" is "bad username", while api instead calls it "invalid username"
* the single letter "A" is given "username exists" by scratchr2, but "invalid username" by api.

I did see one confusing instance, out of 70k -- "poop224" seemed to be categorized by api as "valid username", according to my script. But when I reran it, api called it "bad username", successfully. I suspect accessing Cleanspeak failed momentarily?

picklesrus:
LGTM! Thank you for doing all of that testing work!

Issue #308, opened 2016-01-07
Migrate - Contact Us


kaschm:
[Description added by @benjiwheeler in Oct 2019]

The Contact Us page looks like this:

![image](https://user-images.githubusercontent.com/3431616/67578127-d284cf00-f70f-11e9-9e65-342de3c2c326.png)

It should be updated for Scratch 3.0, moved from scratchr2 to this repo (scratch-www), and revised in consultation with the Scratch Team!

St19Galla:
I can take this one off of @technoboy10 's hands if he would like!

Any particular page template you want me to go off of?

thisandagain:
@St19Galla Thanks! I think we should hold off on this one until we have a plan for how to handle the form in our new API.

benjiwheeler:
Some updated collected thoughts, from across the team:

* what text should the contact us page have?
* what options appear in that dropdown?
* are there things we can do with the form that are more complex? (i.e. add options to the dropdown which, if selected, would provide answers to some of the more simple questions people submit instead of sending us an email)

apple502j:
Here are some opinions:
- l10n for forum link. Some Japanese people created several QaS topics in Japanese.
- Optional email for logged-ins (the response comes as alerts or messages) This is quite useful when users themselves don't own emails.
- Dropdown options:
-- Inappropriate contents
lets user type "where" and "who"
-- Impersonation
lets user type who impersonated who
-- Login Problems
let them reset password (not forced)
-- Email Problems
-- Bugs (including Security)
this is forwarded to Dev Team if possible to do so?
-- Licensing and Branding
Kind of "can I make Scratch Cat mug" or something
-- DMCA
Redirects to DMCA page
-- Jobs
-- Press
-- Account Suspension Request

apple502j:
@chrisgarrity this can be closed

PR #3719, opened 2020-02-28
Add new Contact Us information page

chrisgarrity:
### Resolves:

Moves the contact-us form from the legacy scratchr2 system to www. 

### Changes:
Slight update to the wording to make it a bit more friendly. Also uses the Freshdesk formwidget to submit directly to Freshdesk instead of sending an email (will allow us to remove one email path from scratchr2)

* adds new component HelpForm for the Freshdesk component
* adds new contact-us view
* modifies Information page css with style for a header on the right hand nav section
* adds new route for the contact-us page (is identical to the old scratchr2 URL)

_Note: Information Page template is NOT responsive, and that is out of scope for this change._
### Test Coverage:
Screen Shot 2020-02-28 at 9 23 01 AM

**Testing**
- [ ] Text displays (no message ids - check other languages, but don't expect translation)
- [ ] Form displays (only be English for now)
- [ ] Can submit form
- [ ] if logged in the additional info (scroll the form to view) includes Scratch username and browser info
- [ ] if not logged in, additional info has blank username, browser is still filled in
- [ ] browser info is correct (browser, operating system etc. should match `window.navigator.userAgent` in the console)
- [ ] Verify that form is accessible and can be submitted from China


src/components/helpform/helpform.jsx
```diff
@@ -0,0 +1,57 @@
...
+    user: PropTypes.shape({
+        classroomId: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+        thumbnailUrl: PropTypes.string,
+        username: PropTypes.string
+    })
+};
+
+HelpForm.defaultProps = {
+    title: '',
+    user: {usernname: ''}
```
apple502j:
```suggestion
    user: {username: ''}
```
typo here

src/views/contact-us/l10n.json
```diff
@@ -0,0 +1,16 @@
...
+    "contactUs.forumsInfo":"You can also look through the Scratch {forumsLink}.",
+    "contactUs.forumsLinkText":"Discussion forums",
+    "contactUs.forumsAsk":"If you can't find an answer, post a new topic on the {forumsLink} (see links below). There are many experienced Scratchers who can help with the topics below and more:",
+    "contactUs.questionsForum":"Questions About Scratch",
+    "contactUs.scriptsForum":"Help with Scripts",
+	"contactUs.bugsForum":"Bugs and Glitches",
+    "contactUs.formIntro":"If you still need to contact us, please fill out the form below with as much detail as you can. If you have any screenshots, attachments or links that help to explain your problem, please include them. We get a lot of mail, so we may not be able to respond to your message.",
+    "contactUs.feedbackForm":"Feedback Form",
+    "contactUs.findHelp":"Where to find help:",
+    "contactUs.knowledgeBase":"Knowledge Base",
```
apple502j: This is unused; will it be in the future?

src/views/contact-us/l10n.json
```diff
@@ -0,0 +1,16 @@
+{
+    "contactUs.title":"Contact Us",
+    "contactUs.intro":"You can find answers to most questions about Scratch on our {faqLink} page.",
+    "contactUs.faqLinkText":"Frequently Asked Questions",
+    "contactUs.forumsInfo":"You can also look through the Scratch {forumsLink}.",
+    "contactUs.forumsLinkText":"Discussion forums",
+    "contactUs.forumsAsk":"If you can't find an answer, post a new topic on the {forumsLink} (see links below). There are many experienced Scratchers who can help with the topics below and more:",
+    "contactUs.questionsForum":"Questions About Scratch",
```
apple502j: Those forum names should not be translated because the forum itself isn't. Rather it helps to use customized URL for each language (+ translations).

src/views/contact-us/l10n.json
```diff
@@ -0,0 +1,16 @@
...
+    "contactUs.scriptsForum":"Help with Scripts",
+	"contactUs.bugsForum":"Bugs and Glitches",
+    "contactUs.formIntro":"If you still need to contact us, please fill out the form below with as much detail as you can. If you have any screenshots, attachments or links that help to explain your problem, please include them. We get a lot of mail, so we may not be able to respond to your message.",
```
apple502j: It should mention when you *must* contact us (rather than forum topics); for example, reporting or account issues.

src/views/contact-us/l10n.json
```diff
@@ -0,0 +1,16 @@
+{
+    "contactUs.title":"Contact Us",
+    "contactUs.intro":"You can find answers to most questions about Scratch on our {faqLink} page.",
+    "contactUs.faqLinkText":"Frequently Asked Questions",
+    "contactUs.forumsInfo":"You can also look through the Scratch {forumsLink}.",
+    "contactUs.forumsLinkText":"Discussion forums",
+    "contactUs.forumsAsk":"If you can't find an answer, post a new topic on the {forumsLink} (see links below). There are many experienced Scratchers who can help with the topics below and more:",
```
apple502j: Some localized forums don't have topics for each questions, but have one topic for all questions. Japanese forum is a good example. See also ⬇️

src/views/contact-us/l10n.json
```diff
@@ -0,0 +1,16 @@
...
+    "contactUs.questionsForum":"Questions About Scratch",
+    "contactUs.scriptsForum":"Help with Scripts",
+	"contactUs.bugsForum":"Bugs and Glitches",
+    "contactUs.formIntro":"If you still need to contact us, please fill out the form below with as much detail as you can. If you have any screenshots, attachments or links that help to explain your problem, please include them. We get a lot of mail, so we may not be able to respond to your message.",
+    "contactUs.feedbackForm":"Feedback Form",
```
apple502j: This seems unused?

src/components/helpform/helpform.jsx
```diff
@@ -0,0 +1,57 @@
+const connect = require('react-redux').connect;
+const PropTypes = require('prop-types');
+const React = require('react');
+const FormattedMessage = require('react-intl').FormattedMessage;
+
+const HelpForm = props => {
+    const prefix = 'https://mitscratch.freshdesk.com/widgets/feedback_widget/new?&widgetType=embedded&widgetView=yes&screenshot=No&searchArea=No&captcha=yes';
```
apple502j: Just in case: is this accessible in China?

apple502j:
Questions:
1) Can email be optional somehow for logged-ins? Most Scratchers would use parents' emails, and the response should be readable easily without asking their parents. I have some ideas but the easiest would be "Reply by Alert".
2) Is it possible to translate the form in the future?
3) As I said above, forum link l10n would be a nice feature to have. Not all languages have their own forum and question topics, so it may be hard but I think it's worth.
4) We should have more dropdown options: see #308 
5) What would happen if the iframe/js fails to load?

src/components/helpform/helpform.jsx
chrisgarrity: Good question. They do support both traditional and simplified Chinese, and it looks like there are websites in China that use Freshdesk, so I think the answer is yes, but we'll definitely check.

src/components/helpform/helpform.jsx
chrisgarrity: Good catch.

src/views/contact-us/l10n.json
chrisgarrity: @apple502j right now we're just trying to move the contact us from the legacy Scratch website to www. Changing what Forums show up based on language is out of scope for this change.

src/views/contact-us/l10n.json
chrisgarrity: Actually, we'd much rather that people report inappropriate projects etc. via the report flows not through email. Again, this is just trying to move what we currently have on scratchr2 to www.

src/views/contact-us/l10n.json
chrisgarrity: Yes, these are for some follow on work.

src/views/contact-us/l10n.json
chrisgarrity: Another good point, but this matches the current functionality so I'll keep it as is. I'll add some context in transifex for this version and it's something we can thing about for the future.

chrisgarrity:
@apple502j good questions.
This is just the beginning of a larger overhaul of the way that people contact us, so it's not going to do everything. For the particular questions:
1. We could possibly pre-fill the email with whatever email is associated with their account, but it cannot be optional.
2. Yes it will be able to be translated in the future
3. Interesting idea, but probably out of scope for what we're doing right now.
4. Next phase will be more dropdowns and required fields
5. Right now it would just be a blank iframe if freshdesk is unavailable.

src/components/helpform/helpform.jsx
```diff
@@ -0,0 +1,63 @@
...
+    const username = `helpdesk_ticket[custom_field][cf_scratch_name_40167]=${props.user.username || ''}`;
+    const agentText = encodeURI(window.navigator.userAgent.replace(';', ' -'));
+    const browser = `helpdesk_ticket[custom_field][cf_browser_40167]=${agentText}`;
+    const formSubject = `helpdesk_ticket[subject]=${props.subject}`;
+    const formDescription = `helpdesk_ticket[description]=${props.body}`;
+    return (
+        <div>
+            <script
+                src="https://s3.amazonaws.com/assets.freshdesk.com/widget/freshwidget.js"
+                type="text/javascript"
```
rschamp: To prevent this script from from holding up rendering in the case that the domain is blocked/throttled, we can add `async`. And add `defer` just in case there are any `onload` handlers in that script (still unclear what this is here for, but there are errors in the console when we remove it).
```suggestion
                src="https://s3.amazonaws.com/assets.freshdesk.com/widget/freshwidget.js"
                type="text/javascript"
                async defer 
```

chrisgarrity: I added them (not the suggestion) because our eslint requires them to come first. ;)

src/views/contact-us/contact-us.jsx
```diff
@@ -10,44 +10,64 @@ const HelpForm = require('../../components/helpform/helpform.jsx');
...
+    const params = url.split('?')[1];
+    if (typeof (params) !== 'undefined' && params.indexOf('studio') !== -1) {
+        subject = `Inappropriate content reported in studio ${params.split('=')[1]}`;
+        body = `https://scratch.mit.edu/studios/${params.split('=')[1]}`;
+    } else if (typeof (params) !== 'undefined' && params.indexOf('profile') !== -1) {
+        subject = `Inappropriate content reported in profile ${params.split('=')[1]}`;
+        body = `https://scratch.mit.edu/users/${params.split('=')[1]}`;
+    } else if (typeof (params) !== 'undefined' && params.indexOf('confirmation') !== -1) {
+        subject = 'Problem with email confirmation';
+    }
```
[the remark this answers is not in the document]
chrisgarrity: Looking at other pages in www that use the search string the better approach looked like creating a constructor, and putting the values in the state.

src/components/helpform/helpform.jsx
```diff
@@ -0,0 +1,65 @@
...
+            <style
+                media="screen, projection"
+                type="text/css"
+            >
+                @import url(https://s3.amazonaws.com/assets.freshdesk.com/widget/freshwidget.css);
+            </style>
+            <iframe
+                className="freshwidget-embedded-form"
+                frameBorder="0"
+                height="744px"
```
rschamp: I didn't know you could include `px` for the `height` attribute — maybe React is stripping it out?

rschamp:
We should also add to the test plan that we're going to look at availability from China.

src/components/helpform/helpform.jsx
```diff
@@ -0,0 +1,65 @@
...
+                id="freshwidget-embedded-form"
+                scrolling="no"
+                src={`${prefix}&${title}&${username}&${browser}&${formSubject}&${formDescription}`}
+                title={<FormattedMessage id="contactUs.questionsForum" />}
```
rschamp: It looks like this is being rendered as `[object Object]` — I think we need to change it to use `formatMessage` (although I'm not sure what this attribute is here for).

PR #3736, opened 2020-03-05
Scratch Desktop 3.9.0

cwillisf:
### Changes:

Release scratch-desktop 3.9.0 for Windows direct download. Also, for convenience, move the download URLs into a `const` near the top of the file.

The macOS build is not updated due to a problem with accessing the camera and microphone: LLK/scratch-desktop#109. Hopefully that fix will not take long and we'll get version 3.9.1 out soon, but we'll see.

If either of you would prefer not to move the URLs into a `const` like that, let me know and I can back that part out so we can discuss without blocking this release. Thanks!

### DIFF
--- a/src/components/install-scratch/install-scratch.jsx
+++ b/src/components/install-scratch/install-scratch.jsx
@@ -15,7 +15,7 @@ require('./install-scratch.scss');
 
 const downloadUrls = {
     mac: 'https://downloads.scratch.mit.edu/desktop/Scratch%20Desktop-3.6.0.dmg',
-    win: 'https://downloads.scratch.mit.edu/desktop/Scratch%20Desktop%20Setup%203.6.0.exe',
+    win: 'https://downloads.scratch.mit.edu/desktop/Scratch%20Desktop%20Setup%203.9.0.exe',
     googlePlayStore: 'https://play.google.com/store/apps/details?id=org.scratch',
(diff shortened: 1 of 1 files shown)


src/components/install-scratch/install-scratch.jsx
```diff
@@ -13,6 +13,14 @@ const Step = require('../../components/steps/step.jsx');
...
 require('./install-scratch.scss');
 
+const downloadUrls = {
+    mac: 'https://downloads.scratch.mit.edu/desktop/Scratch%20Desktop-3.6.0.dmg',
+    win: 'https://downloads.scratch.mit.edu/desktop/Scratch%20Desktop%20Setup%203.9.0.exe',
+    googlePlayStore: 'https://play.google.com/store/apps/details?id=org.scratch',
+    microsoftStore: 'https://www.microsoft.com/store/apps/9pfgj25jl6x3?cid=storebadge&ocid=badge',
+    macAppStore: 'https://apps.apple.com/us/app/scratch-desktop/id1446785996?mt=12'
+};
+
```
chrisgarrity: 👍 on moving into a const like this. In the future this could be a file we load externally with the data and we wouldn't need to deploy www to change the version of the downloads.

cwillisf: Yep, exactly! :)

cwillisf:
Sorry about not following the correct process... I have two new PRs that should be correct now, I think...

PR #3738, opened 2020-03-05
[Master] hotfix/scratch desktop 3.9.0

cwillisf:
(Replaces #3736)

### Changes:

Release scratch-desktop 3.9.0 for Windows direct download. Also, for convenience, move the download URLs into a `const` near the top of the file.

The macOS build is not updated due to a problem with accessing the camera and microphone: LLK/scratch-desktop#109. Hopefully that fix will not take long and we'll get version 3.9.1 out soon, but we'll see.

If either of you would prefer not to move the URLs into a `const` like that, let me know and I can back that part out so we can discuss without blocking this release. Thanks!


chrisgarrity:
Merged the [Develop] version into the weekly release branch since we didn't already land this.

PR #3747, opened 2020-03-13
[Hotfix/Master] Split project report flow into two steps

paulkaplan:
This code splits the project report action into two separate panels, separated by a "next" button. It does not change the categories shown or anything else. This lays the groundwork for future changes to report flow, including adding subcategories and adding informational steps to help scratchers understand our community guidelines.

Screen Shot 2020-03-13 at 2 27 54 PM
Screen Shot 2020-03-13 at 1 56 16 PM

The major components of work are:
- Separated out the single `form` element into a reusable `FormStep` that can be given child elements to render and have submit handled. This allows us to build a multi-step form that accumulates data. This is inspired by the `Progression`/form usage in the join flow. 
- Separate the "report options" object and use PropTypes to make sure its structure is strictly managed. The reason for this is that I want to add subcategories to that structure, and having it separate and well typed will make this less error prone.
- Use that `FormStep` to build a select step, a textarea step and a confirmation step. 

Future work will be adding additional sub-category steps, and informational steps that prevent submitting reports. 

For manual testing the existing validations/behavior should be tried, e.g.
- submitting a report should show up in the moderation queue
- you can't advance to next step without choosing a report category
- you can't submit a report without meeting min length (50 chars)  and max length (200 chars)
- you cannot submit a report more than once without refreshing the page

### DIFF
--- a/src/components/modal/report/form-step.jsx
+++ b/src/components/modal/report/form-step.jsx
@@ -34,8 +34,8 @@ class FormStep extends React.Component {
     }
     render () {
         const {onNext, children, isWaiting, nextLabel} = this.props;
-        // Allow submit to be force-enabled by submitEnabled prop. Otherwise use form validation,
-        // with default being not-submittable.
+        // Submit button is enabled if form isn't already submitting, and either the form passes validation,
+        // or the submitEnabled prop is true. This lets submitEnabled prop override validation.
         const submitEnabled = (this.props.submitEnabled || this.state.valid) && !isWaiting;
         const submitDisabledParam = submitEnabled ? {} : {disabled: 'disabled'};
         return (
--- a/src/components/modal/report/form-step.jsx
+++ b/src/components/modal/report/form-step.jsx
@@ -45,7 +45,7 @@ class FormStep extends React.Component {
                 onValidSubmit={onNext}
             >
                 {children}
-                <FlexRow className="action-buttons">
+                <FlexRow className={classNames('action-buttons', 'action-buttons-no-inset')}>
                     <Button
                         className={classNames(
                             'action-button',
--- a/src/components/modal/report/form-step.jsx
+++ b/src/components/modal/report/form-step.jsx
@@ -1,0 +1,89 @@
+const bindAll = require('lodash.bindall');
+const PropTypes = require('prop-types');
+const React = require('react');
+const FormattedMessage = require('react-intl').FormattedMessage;
+const classNames = require('classnames');
+
+const Form = require('../../forms/form.jsx');
+const Button = require('../../forms/button.jsx');
+const Spinner = require('../../spinner/spinner.jsx');
+const FlexRow = require('../../flex-row/flex-row.jsx');
+require('../../forms/button.scss');
+
+/**
+ * Step to be used in a form progression. Provides wrapping form element,
+ * renders children input elements, then provides a next button row
+ * that responds to form validation and submission spinner.
+ */
+class FormStep extends React.Component {
+    constructor (props) {
+        super(props);
+        this.state = {
+            valid: false
+        };
+        bindAll(this, [
+            'handleValid',
+            'handleInvalid'
+        ]);
+    }
+    handleValid () {
+        this.setState({valid: true});
+    }
+    handleInvalid () {
+        this.setState({valid: false});
+    }
+    render () {
+        const {onNext, children, isWaiting, nextLabel} = this.props;
+        // Allow submit to be force-enabled by submitEnabled prop. Otherwise use form validation,
+        // with default being not-submittable.
+        const submitEnabled = (this.props.submitEnabled || this.state.valid) && !isWaiting;
+        const submitDisabledParam = submitEnabled ? {} : {disabled: 'disabled'};
+        return (
+            <Form
+                onInvalid={this.handleInvalid}
+                onValid={this.handleValid}
+                onValidSubmit={onNext}
+            >
+                {children}
+                <FlexRow className="action-buttons">
+                    <Button
+                        className={classNames(
+                            'action-button',
+                            'submit-button',
+                            {disabled: !submitEnabled}
+                        )}
+                        {...submitDisabledParam}
+                        key="submitButton"
+                        type="submit"
+                    >
+                        {isWaiting ? (
+                            <div className="action-button-text">
+                                <Spinner />
+                                <FormattedMessage id="report.sending" />
+                            </div>
+                        ) : (
+                            <div className="action-button-text">
+                                <FormattedMessage {...nextLabel} />
+                            </div>
+                        )}
+                    </Button>
+                </FlexRow>
+            </Form>
+        );
+    }
+}
+
+FormStep.propTypes = {
+    children: PropTypes.node.isRequired,
+    isWaiting: PropTypes.bool,
+    nextLabel: PropTypes.shape({id: PropTypes.string.isRequired}).isRequired,
+    onNext: PropTypes.func.isRequired,
+    submitEnabled: PropTypes.bool
+};
+
+FormStep.defaultProps = {
+    isWaiting: false,
+    submitEnabled: false
+};
+
+module.exports = FormStep;
--- a/src/components/modal/report/modal.jsx
+++ b/src/components/modal/report/modal.jsx
@@ -19,8 +19,8 @@ const {reportOptionsShape, REPORT_OPTIONS} = require('./report-options.js');
 require('../../forms/button.scss');
 require('./modal.scss');
 
-// Progression component only addresses steps by number, but this flow
-// may skip steps so the code is easier to read with a map.
+// The Progression component uses numbers to track which step it's on, but that's
+// hard to read. Make the code easier to read by giving each step number a label.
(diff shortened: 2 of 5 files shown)


benjiwheeler:
One initial thought -- seeing the initial report modal come up, I was immediately struck that the button's right margin was out of alignment with the select dropdown.

src/components/modal/report/form-step.jsx
```diff
@@ -0,0 +1,89 @@
...
+    }
+    handleValid () {
+        this.setState({valid: true});
+    }
+    handleInvalid () {
+        this.setState({valid: false});
+    }
+    render () {
+        const {onNext, children, isWaiting, nextLabel} = this.props;
+        // Allow submit to be force-enabled by submitEnabled prop. Otherwise use form validation,
```
benjiwheeler: I find this comment hard to follow.

How about something like:

// Submit button is enabled if form isn't already submitting, and either the form passes validation, or the submitEnabled prop is true.
// This lets submitEnabled prop override validation.

src/components/modal/report/form-step.jsx
```diff
@@ -0,0 +1,89 @@
+const bindAll = require('lodash.bindall');
+const PropTypes = require('prop-types');
+const React = require('react');
+const FormattedMessage = require('react-intl').FormattedMessage;
+const classNames = require('classnames');
+
+const Form = require('../../forms/form.jsx');
```
benjiwheeler: Ideally, we'd take this opportunity to implement what we need using Formik (the new forms system we're using instead of Formsy, which is passing into obsolescence).

An example is src/components/join-flow/country-step.jsx . We don't have an equivalent to the Formsy `Form` component, because we didn't think we needed one for Join Flow. 

We could make that change here, and I'd be happy to help. But I understand if you want to move faster than that would require. Just know that it's adding tech debt.

src/components/modal/report/modal.jsx
```diff
@@ -6,95 +6,50 @@ const FormattedMessage = require('react-intl').FormattedMessage;
...
-        label: {id: 'report.reasonPersonal'},
-        prompt: {id: 'report.promptPersonal'}
-    },
-    {
-        value: '6',
-        label: {id: 'general.other'},
-        prompt: {id: 'report.promptGuidelines'}
-    }
-];
+// Progression component only addresses steps by number, but this flow
```
benjiwheeler: A little hard to read this comment. How about:

// The Progression component uses numbers to track which step it's on, but that's
// hard to read. Make the code easier to read by giving each step number a readable label.

src/components/modal/report/modal.jsx
```diff
@@ -124,120 +79,88 @@ class ReportModal extends React.Component {
...
-                                        validations={{
-                                            maxLength: 500,
-                                            minLength: 20
+                    <ModalInnerContent className="report-modal-content">
+                        {isError && (
+                            <div className="error-text">
+                                <FormattedMessage id="report.error" />
+                            </div>
+                        )}
+                        <Progression step={isConfirmed ? STEPS.confirmation : this.state.step}>
```
benjiwheeler: It's too bad `state.step` itself isn't actually changing to become `STEPS.confirmation`. That would isolate the logic more cleanly. There are a few ways that could be done -- the one that comes to mind is to use `componentDidUpdate`, and have it check like:

```
if (props.isConfirmed true && !prevProps.isConfirmed) {
    this.setState({step: STEPS.confirmation});
}
```

I'm not sure that's so much better, but I'd lean towards doing something like that.

benjiwheeler:
Tested locally several times, including in Espanol. Confirmed data shows up in db (didn't carefully check all fields, just that the report was there).

src/components/modal/report/modal.jsx
```diff
@@ -124,120 +79,88 @@ class ReportModal extends React.Component {
...
+                                        value: option.value,
+                                        label: intl.formatMessage(option.label),
+                                        key: option.value
+                                    }))}
+                                    validationErrors={{
+                                        isDefaultRequiredValue: categoryRequiredMessage
+                                    }}
+                                />
+                            </FormStep>
+                            <FormStep
```
benjiwheeler: Could you put a comment before this line, making clear that it constitutes the content of the `textInput` step?

src/components/modal/report/modal.jsx
```diff
@@ -124,120 +79,88 @@ class ReportModal extends React.Component {
...
-                                            maxLength: 500,
-                                            minLength: 20
+                    <ModalInnerContent className="report-modal-content">
+                        {isError && (
+                            <div className="error-text">
+                                <FormattedMessage id="report.error" />
+                            </div>
+                        )}
+                        <Progression step={isConfirmed ? STEPS.confirmation : this.state.step}>
+                            <FormStep
```
benjiwheeler: Could you put a comment before this line, making clear that it constitutes the content of the `category` step?

src/components/modal/report/modal.jsx
```diff
@@ -124,120 +79,88 @@ class ReportModal extends React.Component {
...
+                                        maxLength: intl.formatMessage({id: 'report.tooLongError'}),
+                                        minLength: intl.formatMessage({id: 'report.tooShortError'})
+                                    }}
+                                    validations={{
+                                        maxLength: 500,
+                                        minLength: 20
+                                    }}
+                                />
+                            </FormStep>
+                            <FormStep
```
benjiwheeler: Could you put a comment before this line, making clear that it constitutes the content of the `confirmation` step?

paulkaplan:
@benjiwheeler thanks for the quick review. Took your suggestions about clarifying comments, and added a comment re: not using componentDidUpdate, there is a bit of functionality that the override simplifies that the componentDidUpdate would not preserve. 

I would be happy to pair on converting this over to Formik in a separate PR.

benjiwheeler:
@paulkaplan Thanks, will re-review. It would be helpful if you could respond to my individual comments saying what the current state of them is -- "Did this", "Left as is, see comment below" etc.

src/components/modal/report/form-step.jsx
benjiwheeler: Status: @paulkaplan and I will consider coming back to this to convert to Formik later

src/components/modal/report/form-step.jsx
benjiwheeler: Done

src/components/modal/report/modal.jsx
benjiwheeler: Done

src/components/modal/report/modal.jsx
benjiwheeler: Done

src/components/modal/report/modal.jsx
benjiwheeler: Done

src/components/modal/report/modal.jsx
benjiwheeler: Done

src/components/modal/report/modal.jsx
```diff
@@ -124,120 +84,93 @@ class ReportModal extends React.Component {
...
+                                        maxLength: intl.formatMessage({id: 'report.tooLongError'}),
+                                        minLength: intl.formatMessage({id: 'report.tooShortError'})
+                                    }}
+                                    validations={{
+                                        maxLength: 500,
+                                        minLength: 20
+                                    }}
+                                />
+                            </FormStep>
+
```
benjiwheeler: Tiny quibble -- The other steps don't have an empty line separating them

benjiwheeler:
> added a comment re: not using componentDidUpdate

I don't see that comment.

benjiwheeler:
Did you respond to this?

> One initial thought -- seeing the initial report modal come up, I was immediately struck that the button's right margin was out of alignment with the select dropdown.

I'm not saying that this must be addressed in this PR, but it might make sense to consider whether this needs to be an action button row, or to file it as a follow-up issue, since the button positioning is a change that's introduced by this PR.

paulkaplan:
> > added a comment re: not using componentDidUpdate
> 
> I don't see that comment.

Sorry not sure how to link to a line in a PR (line highlight linking doesn't work in split view?) It is the comment about the `step` pulled out in https://github.com/LLK/scratch-www/pull/3747/commits/327bdc32be629b7268d4d652c73608b434b0703b

src/components/modal/report/modal.jsx
paulkaplan: Hmmm they all have an empty line I believe...

paulkaplan:
@benjiwheeler I added a modifier class to handle the "margin handled elsewhere" case, bringing the buttons in line with the other elements in the form, which seems right.

benjiwheeler:
> Sorry not sure how to link to a line in a PR (line highlight linking doesn't work in split view?) It is the comment about the step pulled out in 327bdc3

AAAAhhhhhh.... I was thinking "Github PR comment" not "// code comment". Thanks for clarifying -- the comment you added is great, thanks!

benjiwheeler:
LGTM. Thanks for finding a solution for the alignment -- that seems not too hacky :)

image

image

PR #3750, opened 2020-03-16
revert gui to previous version

chrisgarrity:
We decided not to release the gui changes from this week. Removing the scratch-gui update so that the contact us page can be updated.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,475 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "@babel/cli": {
-      "version": "7.8.4",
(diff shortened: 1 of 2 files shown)


picklesrus:
I’m having a hard time following versions. The live gui version (at least the one on master now) is 20200304215547. Should that be what we go back to?

I think 20200310205953 (the one you are setting it to still includes the vm changes for the decimal thingy? Though I’m reading that from greenkeeper’s comments on https://github.com/LLK/scratch-www/pull/3739 and I’m not sure if I’m reading it right (edited)

chrisgarrity:
Good catch - I probably just reverted the translations and nothing else!

PR #3765, opened 2020-03-24
Move reCaptcha code to a component

picklesrus:
It'll get used in another spot later.

### DIFF
--- a/src/components/captcha/captcha.jsx
+++ b/src/components/captcha/captcha.jsx
@@ -69,8 +69,8 @@ class Captcha extends React.Component {
     }
 }
 Captcha.propTypes = {
-    onCaptchaError: PropTypes.func,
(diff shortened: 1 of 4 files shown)


src/components/captcha/captcha.jsx
```diff
@@ -0,0 +1,76 @@
...
+                data-badge="bottomright"
+                data-sitekey={process.env.RECAPTCHA_SITE_KEY}
+                data-size="invisible"
+                ref={this.setCaptchaRef}
+            />
+        );
+    }
+}
+Captcha.propTypes = {
+    onCaptchaError: PropTypes.func,
```
chrisgarrity: These props should probably be required as it won't work without them.

picklesrus: Yes! Done!

PR #3785, opened 2020-04-02
[Develop] Add nested categories to project report module

seotts:
### Changes:

This adds the ability to nest report categories in the project reporting module according to this design:
project-report-mock-final

As a first round of changes, we have added nesting to only the Other option. The intended copy is over here: https://docs.google.com/document/d/13w9MbBrzb4NiztA9MUjiVEP4xdgUA9jz0zzisLIsnhQ/edit

When you report a project, all report categories will take you to a text field except for Other, which will take you to another dropdown.

The Back button will come in a later PR .

### Testing:

We will test these changes with people in the Sustainable Moderation project using this test plan: 

- Go to any project you do not own and report it. All categories you select should work as before except “Other”

- If you select “Other,” you should be prompted to make another selection. Some of those selections should result in a deadend; others will prompt you to input text

- Check that all the copy makes sense, is formatted correctly, and matches the copy written by the team

- Log into an admin account. Check that reports (that do not deadend) create tickets and that the text in the tickets makes sense for the new selections. Reports that do deadend should not create tickets.

cc @paulkaplan


src/components/modal/report/modal.jsx
```diff
@@ -2,6 +2,7 @@ const bindAll = require('lodash.bindall');
 const PropTypes = require('prop-types');
 const React = require('react');
 const connect = require('react-redux').connect;
+const FormattedHTMLMessage = require('react-intl').FormattedHTMLMessage;
```
apple502j, line 5: looks unused

src/l10n.json
```diff
@@ -269,6 +278,19 @@
...
+    "report.tipsSpecific": "Try to be specific with your feedback. For instance: The controls to move the character did not work.",
+    "report.promptDoesntWork": "A Scratch project, like any other application, may contain a few bugs. That is expected and completely okay!",
+    "report.promptDoesntWorkTips": "We encourage you to share any issues you discover directly with the creator of the project. It's also helpful to provide suggestions on how they may improve their project, if possible.",
+    "report.promptTooHard": "If you feel a project could be easier, we encourage you to share that feedback directly with the creator of the project. Or remix it yourself and make it as easy or hard as you like!",
+    "report.promptMisleading": "Tell us more about how it's tricking or misleading people",
+    "report.promptFaceReveal": "Scratch allows people to use pictures of their face in creative projects like games, stories, or animations. However, projects that are just a picture of their face, or which focus entirely on their physical appearance, aren't allowed on Scratch. If this is a 'face reveal', or the project focuses on the person's appearance, please click report.",
```
adroitwhiz, line 290: The button says "Send", not "report" (in English, at least). Is there a way in `react-intl` to reference messages inside other messeages so that `report.send` can be substituted in?

src/components/modal/report/modal.jsx
seotts: thanks!

src/l10n.json
seotts: Good point! I made a change to substitute in `report.send`.

src/l10n.json
```diff
@@ -269,6 +278,19 @@
...
+    "report.promptDoesntWorkTips": "We encourage you to share any issues you discover directly with the creator of the project. It's also helpful to provide suggestions on how they may improve their project, if possible.",
+    "report.promptTooHard": "If you feel a project could be easier, we encourage you to share that feedback directly with the creator of the project. Or remix it yourself and make it as easy or hard as you like!",
+    "report.promptMisleading": "Tell us more about how it's tricking or misleading people",
+    "report.promptFaceReveal": "Scratch allows people to use pictures of their face in creative projects like games, stories, or animations. However, projects that are just a picture of their face, or which focus entirely on their physical appearance, aren't allowed on Scratch. If this is a \"face reveal,\" or the project focuses on the person's appearance, please click \"{send}.\"",
+    "report.promptNoRemixingAllowed": "Please let us know where the project says it is not okay to remix — such as in the Notes & Credits, project title, etc.  ",
```
apple502j, line 291: It has trailing space
```suggestion
    "report.promptNoRemixingAllowed": "Please let us know where the project says it is not okay to remix — such as in the Notes & Credits, project title, etc.",
```

PR #3786, opened 2020-04-03
Split follow splash into FollowUser and FollowStudio

### LINKED ISSUE #3780: Have separate messages for following people and studios
We have only one string defined for a user following another user or studio.
https://github.com/LLK/scratch-www/blob/7f9ffada5fe1167721669b9bc3e2d5d83cc40692/src/views/splash/l10n.json#L16

In some languages you would not use the same translation for both. _The initial report came in for Korean, but it's quite possible for other languages._

In addition to adding two strings to the `l10n` file, we'd need to refactor [FollowMessage](https://github.com/LLK/scratch-www/blob/develop/src/views/splash/activity-rows/follow.jsx), probably into two separate components `FollowStudio` and `FollowUser`.

### Steps to Reproduce

Have an account you're following follow another user and a studio. The same message will be used for both in all languages.

apple502j:
### Resolves:
Resolves #3780 

### Changes:
Splits the follow splash into 2 components, FollowUser and FollowStudio.

- FollowUser uses existing translation; should we change it?
- Removed quirks for supporting both user and studio in one component, including some indent changes


src/views/splash/l10n.json
```diff
@@ -14,6 +14,7 @@
     "messages.becomeManagerText": "{username} was promoted to manager of {studio}",
     "messages.favoriteText": "{profileLink} favorited {projectLink}",
     "messages.followText": "{profileLink} is now following {followeeLink}",
```
chrisgarrity, line 16: Let's rename this `messages.followProfileText`, but keep the actual string the same. In transifex the translation memory should automatically translate the new string if it 100% matches an existing string.

src/views/splash/activity-rows/follow-studio.jsx
```diff
@@ -0,0 +1,43 @@
...
+                )
+            }}
+        />
+    </SocialMessage>
+);
+
+FollowStudioMessage.propTypes = {
+    className: PropTypes.string,
+    followDateTime: PropTypes.string.isRequired,
+    followeeId: PropTypes.string.isRequired,
```
chrisgarrity, line 38: Change to `studioId` and `studtioTitle` (see below).

src/views/splash/presentation.jsx
```diff
@@ -66,9 +66,9 @@ class ActivityList extends React.Component {
                 <FollowStudioMessage
                     followDateTime={message.datetime_created}
                     followeeId={message.gallery_id}
-                    followeeTitle={message.title}
-                    followerUsername={message.actor_username}
                     key={key}
+                    studioTitle={message.title}
+                    stufioUsername={message.actor_username}
```
adroitwhiz, line 71:
```suggestion
                    studioUsername={message.actor_username}
```

PR #3799, opened 2020-04-10
Fix editor title when not logged in

### LINKED ISSUE #3798: The tab title of the editor is nothing when you are signed out.
The title of the tab of the editor is nothing when you are signed out.

### Steps to Reproduce

Sign out and [go to the editor](https://scratch.mit.edu/projects/editor/). The title of the tab is nothing. (depends on your browser, like on chrome it shows up as the URL of the tab, others as unnamed)

Windows NT 10.0, Chrome 81.0.4044.92

apple502j:
### Resolves:
Resolves #3798 

### Changes:
Previously, if `projectInfo.author` is not available, there were no meta-tags. It was not a problem for logged-in users, because when they create new project, a new project is created immediately and author is passed. However, when not logged in, projects are not "created", so there is no author. In this case, check `userPresent` and if it doesn't exist, returns a helmet with `Scratch - Imagine, Program, Share` title. Other meta tags are omitted.

### Test Coverage:
Manually tested

### DIFF
--- a/src/views/preview/meta.jsx
+++ b/src/views/preview/meta.jsx
@@ -1,14 +1,23 @@
 const React = require('react');
 const Helmet = require('react-helmet').default;
+const PropTypes = require('prop-types');
 
 const projectShape = require('./projectshape.jsx').projectShape;
 
 const Meta = props => {
     const {id, title, instructions, author} = props.projectInfo;
 
-    // Do not want to render any meta tags unless all the info is loaded
-    // Check only author (object) because it is ok to have empty string instructions
-    if (!author) return null;
+    if (!author) {
+        // Project info is not ready. It's either fetching state, or logged-out users creating project.
+        if (!props.userPresent) {
+            return (
+                <Helmet>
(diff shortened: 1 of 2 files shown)


src/views/preview/meta.jsx
```diff
@@ -1,14 +1,23 @@
...
 
-    // Do not want to render any meta tags unless all the info is loaded
-    // Check only author (object) because it is ok to have empty string instructions
-    if (!author) return null;
+    if (!author) {
+        // Project info is not ready. It's either fetching state, or logged-out users creating project.
+        if (!props.userPresent) {
+            return (
+                <Helmet>
+                    <title>Scratch - Imagine, Program, Share</title>
```
adroitwhiz, line 15: Should this be localized?

apple502j: Currently, titles are not localized. It can be, but let's make another issue for that.

benjiwheeler:
Thanks @apple502j!

PR #3805, opened 2020-04-15
[Master] Hotfix: Remove last names from credits page

kchadha:
## Release Notes
- Remove last names from people grid on credits page


Accio1:
Since the last names have been removed, wouldn’t it make sense to reorder the page so that it’s alphabetical by first name instead of last name?

PR #3807, opened 2020-04-15
Change up the messaging in student registration. 

picklesrus:
Make the "don't use your real name" a blue info message and give some suggestions on how to choose a username.
image

Old version: 
image

### DIFF
--- a/src/views/studentregistration/studentregistration.jsx
+++ b/src/views/studentregistration/studentregistration.jsx
@@ -126,7 +126,7 @@ class StudentRegistration extends React.Component {
                                 id: 'registration.studentUsernameStepTooltip'
                             })}
                             usernameHelp={this.props.intl.formatMessage({
-                                id: 'registration.studentUsernameFieldHelpText'
+                                id: 'registration.studentUsernameSuggestion'
                             })}
                             waiting={this.state.waiting}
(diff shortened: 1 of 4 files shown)


benjiwheeler:
A screenshot would be helpful here

picklesrus:
Whoops. Sorry. Thought I included one! Will add to PR description too
image

src/views/studentregistration/studentregistration.jsx
```diff
@@ -126,7 +126,7 @@ class StudentRegistration extends React.Component {
                                 id: 'registration.studentUsernameStepTooltip'
                             })}
                             usernameHelp={this.props.intl.formatMessage({
-                                id: 'registration.studentUsernameFieldHelpText'
+                                id: 'registration.studentUsernameSuggestion'
```
benjiwheeler, line 129: Only thought is that now `studentUsernameFieldHelpText` is an orphan. Should we remove it? Or just replace the content and use it instead of `studentUsernameSuggestion`?

benjiwheeler:
Mostly LGTM, one question

src/views/studentregistration/studentregistration.jsx
picklesrus: Good point. I removed the unused message.

PR #3800, opened 2020-04-10
Freshdesk widget

chrisgarrity:
Initial version of the Freshdesk widget that replaces the form with a pop-up widget that includes the ability to search the knowledge base for help. 

Copy is not final.

### Manual Testing
* current Contact Us with form continues to display
* redundant browser field has been removed
* redirects from scratchr2 will display the form with subject and body pre-filled

on staging server
add `?CONTACT_US_POPUP=true` to the Url to see the new contact-us page
* Get Help button and link open the widget
* redirects from scratchr2 will open widget already on the form with the fields pre-filled

_Note: running locally, run your server with `SCRATCH_ENV=staging`_

When this launches consider adding an integration test that clicks the Get Help button and confirms that the widget launches.

Issue #1102, opened 2016-12-10
Missing translations: Duplicate strings in the same l10n.json file


chrisgarrity:
While working on the transifex import I discovered that the scripts that convert po files to intl/* still have a problem if there are duplicate strings in the same l10n file. Some examples:

In src/l10n.json:
"registration.studentUsernameStepTitle": "Create a Scratch Account",
"registration.usernameStepTitleScratcher": "Create a Scratch Account",

In teacherregistration/l10n.json
"teacherRegistration.phoneStepTitle": "Phone Number",
"teacherRegistration.phoneNumber": "Phone Number",
"teacherRegistration.orgStepTitle": "Organization",
"teacherRegistration.organization": "Organization",

Do we really want/need to have the same string defined multiple times in the same view?

Fortunately, in the long run transifex will handle the same text with different keys just fine. However, right now this causes a bug in the scripts that convert files - the earlier fix I made handles the same string in different l10n files, but not in the same one.

chrisgarrity:
@rschamp @mewtaylor I think we decided that there should not be duplicate strings in the same l10n.json file - agree?

mewtaylor:
Fixed in https://github.com/LLK/scratch-www/pull/1150

PR #3817, opened 2020-04-22
Helpwidget update

chrisgarrity:
### Resolves:

Addressing followup comments from: https://github.com/LLK/scratch-www/pull/3800 and meeting discussion.

### Changes:
* delays loading the freshdesk script. _Note: it does not maintain a reference because there are two instances of the widget on the page. If the script gets loaded twice, the launcher button for one of them always appears._
* wording changes to body copy
* moves the profile or studio link out of the description into a new custom field (will now require the sender to fill in something in the description to say why).

### Test Coverage:
* contact us on staging:
  * without feature-flag should show current page with the form
  * with feature-flag `CONTACT_US_POPUP=true` shows new copy and opens widget when the Get Help link or button is clicked.
* if linked to from the inappropriate profile or studio contact us the link will be prefilled into the new link field not the description


src/views/contact-us/contact-us.jsx
```diff
@@ -8,7 +8,7 @@ const render = require('../../lib/render.jsx');
 
 const HelpForm = require('../../components/helpform/helpform.jsx');
 const HelpWidget = require('../../components/helpwidget/helpwidget.jsx');
-const {CONTACT_US_POPUP} = require('../../lib/feature-flags.js');
+const {CONTACT_US_POPUP} = true;
```
apple502j, line 11:
```suggestion
const CONTACT_US_POPUP = true;
```

It will be `undefined` otherwise because `true` does not have property `CONTACT_US_POPUP`

chrisgarrity:
@rschamp the helpwidget PR is failing the test for duplicate strings because I added the helpwidget 'Send' string to the general l10n file, and there is already a report 'Send' string. I think the right thing to do now is remove the test, not the string. The test was added because duplicate strings broke the conversion to/from PO files before we switched www (https://github.com/LLK/scratch-www/issues/1102). That's no longer relevant. /cc @BryceLTaylor

PR #3829, opened 2020-04-26
Fix invisible actor on FollowStudio

apple502j:
I made a mistake on previous PR and now I made a bug that causes actor to be invisible and the studio link to be wrong. This fixes the issue.


adroitwhiz:
This fails lint because props need to be sorted alphabetically

PR #3847, opened 2020-05-01
Educator page mini update

picklesrus:
Adds link to teacher account setup guide on the educator page where you sign up for an educator account.
I will double check the link destination is the final one before committing. 
image

### DIFF
--- a/src/views/teachers/landing/landing.jsx
+++ b/src/views/teachers/landing/landing.jsx
@@ -342,7 +342,7 @@ const Landing = props => (
                             id="teacherlanding.accountsRequestInfo"
                             values={{
                                 setupGuideLink: (
-                                    <a href="" >
+                                    <a href="https://docs.google.com/document/d/1Qb8Lyeiivr-oB49p5Bo17iXU5qxGpBJHuFa_KR5aW-o/view" >
                                         <FormattedMessage id="teacherlanding.accountsSetupGuide" />
                                     </a>
                                 ),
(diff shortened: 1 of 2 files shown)


src/views/teachers/landing/landing.jsx
```diff
@@ -338,7 +338,21 @@ const Landing = props => (
                         <FormattedMessage id="teacherlanding.accountsTitle" />
                     </h2>
                     <p>
-                        <FormattedHTMLMessage id="teacherlanding.accountsDescription" />
+                        <FormattedMessage
```
apple502j, line 341: That was the only place where FormattedHTMLMessage was used - so you can remove it

apple502j:
also why Google Documents? We had problems with "blank 3.0 FAQ" in the past and are we doing it again?

src/views/teachers/landing/landing.jsx
picklesrus: Thanks! That probably explains the lint error too!

picklesrus:
Yes, we're experimenting with using a Google Doc instead here based on some feedback from folks expressing preference to this format for variety of reasons.

benjiwheeler:
Tested, looks great, links work!

benjiwheeler:
@apple502j Can you remind me what the problem was in the past? Was it that Google Docs is blocked in some places?

apple502j:
@benjiwheeler It's mostly from the mainland China - they reported that 3faq is a blank page.

PR #3856, opened 2020-05-04
default student's country to educator's during reg

cwillisf:
### Changes:

When a student is registering with a class, pre-fill the country selection drop-down with that of the educator attached to the class.

There are a few things I would like feedback on, including:
- I stopped short of removing the country selection drop-down entirely, which was suggested in the Trello card. I imagine there are a few situations where the student might legitimately want to register a different country compared to the educator, but I suspect they're very rare. On the other hand it doesn't seem harmful to have the drop-down visible, especially if it's usually pre-filled with the correct option. Thoughts?
- The first time the `StudentRegistration` component renders, it doesn't yet have the classroom's data; that's the reason for `this.state.classroom &&` when passing the `countryName` to the `DemographicsStep`. I'm not sure if it's possible for any other items in that chain (`this.state.classroom.educator.profile.country`) to be null or undefined... do I need to add more checks?
- The classroom info returned by the API includes the educator's country name ('United States') but not the country code ('us') or label ('United States of America'). The `DemographicsStep` really would like the country code, so I added a function to convert a country name to a country code. I suspect that the API has ready access to the country code; is it worth changing the API to include that?

### Test Coverage:

Manually tested by registering an educator, generating a classroom link, and viewing the student registration steps resulting from that link. Both US and Afghanistan work: Afghanistan covers the case where there is no country label so the name is used as the label, and the US option covers the case where the country label is present and differs from the country name.

### DIFF
--- a/src/views/studentregistration/studentregistration.jsx
+++ b/src/views/studentregistration/studentregistration.jsx
@@ -132,6 +132,7 @@ class StudentRegistration extends React.Component {
                             onNextStep={this.handleAdvanceStep}
                         />
                         <Steps.DemographicsStep
+                            countryName={this.state.classroom && this.state.classroom.educator.profile.country}
                             description={this.props.intl.formatMessage({
                                 id: 'registration.studentPersonalStepDescription'
                             })}
--- a/src/views/studentregistration/studentregistration.jsx
+++ b/src/views/studentregistration/studentregistration.jsx
@@ -132,7 +132,8 @@ class StudentRegistration extends React.Component {
                             onNextStep={this.handleAdvanceStep}
                         />
                         <Steps.DemographicsStep
-                            countryName={this.state.classroom && this.state.classroom.educator.profile.country}
+                            countryName={this.state.classroom && this.state.classroom.educator &&
+                                this.state.classroom.educator.profile && this.state.classroom.educator.profile.country}
                             description={this.props.intl.formatMessage({
                                 id: 'registration.studentPersonalStepDescription'
                             })}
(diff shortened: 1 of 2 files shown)


picklesrus:
@benjiwheeler has dealt with country/country code conversion stuff a bunch (although maybe for a different backend?) so I'm going to skip commenting on that part :) 
For the first question about whether to leave the question in: 
Ideally we'd be able to remove this step of registration altogether, but since we're not there, I like defaulting it to the the data from the classroom.

benjiwheeler:
Sorry to take so long to get to this, taking a look ASAP!

benjiwheeler:
This LGTM -- works locally.

src/views/studentregistration/studentregistration.jsx
```diff
@@ -132,6 +132,7 @@ class StudentRegistration extends React.Component {
                             onNextStep={this.handleAdvanceStep}
                         />
                         <Steps.DemographicsStep
+                            countryName={this.state.classroom && this.state.classroom.educator.profile.country}
```
benjiwheeler, line 135: Ideally this would be:

```suggestion
                            countryName={this.state.classroom && this.state.classroom.educator && this.state.classroom.educator.profile && this.state.classroom.educator.profile.country}
```

benjiwheeler:
Response to points in description:
1. Let's do this first, then revisit the question
2. Added suggestion
3. Unfortunately, the db stores the country string, not code! So your approach makes sense. Ideally we'd refactor to put that function in the country code lib...

PR #3857, opened 2020-05-04
Second, smaller round of updates to the FAQ page.

picklesrus:
2nd round of updates to the educator faq
- removes old questions
- add links to educator resources

Screenshots of new/changed content: 
image
image


benjiwheeler:
A few small comments... and apologies if I missed an opportunity to provide feedback earlier!

I find this confusing: "Does this feature exist, and if not, can you please add it?"

I know what we mean -- we mean "I have feature suggestion _________________"... but I think the wording is confusing, it makes it seem like "this" refers to something already being discussed on the page.

Also, I think the following wording goes a bit too far: "However, Scratch is a small organization and our resources are limited, so it may take us a very long time before we can implement any of these changes."

I don't think that's way off base, but how about "it may take a while"?

benjiwheeler:
LGTM, with the comments above submitted for possible consideration!

PR #3867, opened 2020-05-07
Put new help widget back behind feature flag

chrisgarrity:


apple502j:
@chrisgarrity You can still use helpwidget in the production using feature flags - intended?

kchadha:
@apple502j yes, that's intended

PR #3871, opened 2020-05-08
Enable Help widget

chrisgarrity:
Enable Helpwidget, but keep showing the form for iOS 10 and lower.

### Resolves:

Issue found during bug hunt where the help widget was not working on iOS 10

### Changes:

Checks the operating system and continues to show the form on the contact-us page if the OS is 'iOS' and version is <11.

### Test Coverage:

I am not able to test this because I don't have and iOS 10 device. I was able to see it working by switching the condition to `iOS < 13` (since I have iOS 12 on my iPad)

### DIFF
--- a/src/views/contact-us/contact-us.jsx
+++ b/src/views/contact-us/contact-us.jsx
@@ -8,7 +8,7 @@ const render = require('../../lib/render.jsx');
 
 const HelpForm = require('../../components/helpform/helpform.jsx');
 const HelpWidget = require('../../components/helpwidget/helpwidget.jsx');
-const {CONTACT_US_POPUP} = require('../../lib/feature-flags.js');
+const bowser = require('bowser');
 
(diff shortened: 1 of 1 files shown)


src/views/contact-us/contact-us.jsx
```diff
@@ -34,93 +35,62 @@ class ContactUs extends React.Component {
         } else if (query.indexOf('confirmation=') !== -1) {
             this.state.subject = 'Problem with email confirmation';
         }
+        // older iPads (iOS 10 or lower) cannot show help widget, show the form instead
+        if (bowser.osname === 'iOS' && parseInt(bowser.osversion, 10) < 11) {
```
rschamp, line 39: Nit: Since this won't change, you could make `showForm` a prop or even a constant to simplify rendering a bit.

PR #3973, opened 2020-05-13
Add disrespectful report option, reorder options

seotts:
Adds the disrespectful report option and orders options as laid out here https://docs.google.com/document/d/1WS5IK4u8D3ZwlYjNWdIcsfSzdTeEj5rdAhHoAIACVw4/edit#

### Testing

We should check that options are in the order we want and copy is right


paulkaplan:
@seotts remember to merge this!

PR #3974, opened 2020-05-13
move country name lookup to library

benjiwheeler:
Small refactor as follow-up to https://github.com/LLK/scratch-www/pull/3856

Moves country code lookup logic into the country data library


benjiwheeler:
Thanks for the suggestion! I revised this.

PR #4002, opened 2020-05-20
Change the way usernames are handled

### LINKED ISSUE #3850: Username "hasOwnProperty" crashes Join page
If you put `hasOwnProperty` in the username slot of the Join page, and types other words, it will crash. This is probably caused at https://github.com/LLK/scratch-www/blob/develop/src/components/join-flow/username-step.jsx#L61

R4356th:
### Resolves:

Resolves #3850 by changing the way new users' usernames are handled.

### Changes:

The way new users' usernames are currently handled changes.

### DIFF
--- a/src/components/join-flow/username-step.jsx
+++ b/src/components/join-flow/username-step.jsx
@@ -58,7 +58,7 @@ class UsernameStep extends React.Component {
     }
     // simple function to memoize remote requests for usernames
     validateUsernameRemotelyWithCache (username) {
-        if (username in this.usernameRemoteCache this.usernameRemoteCache.hasOwnProperty[username] === 'Object') {
+        if (typeof this.usernameRemoteCache.hasOwnProperty[username] === 'object') {
             return Promise.resolve(this.usernameRemoteCache[username]);
         }
         // username is not in our cache
--- a/src/components/join-flow/username-step.jsx
+++ b/src/components/join-flow/username-step.jsx
@@ -34,7 +34,7 @@ class UsernameStep extends React.Component {
         };
         // simple object to memoize remote requests for usernames.
         // keeps us from submitting multiple requests for same data.
-        this.usernameRemoteCache = {};
+        this.usernameRemoteCache = Object.create(null);
     }
     componentDidMount () {
         // Send info to analytics when we aren't on the standalone page.
--- a/src/components/join-flow/username-step.jsx
+++ b/src/components/join-flow/username-step.jsx
@@ -58,7 +58,7 @@ class UsernameStep extends React.Component {
     }
     // simple function to memoize remote requests for usernames
     validateUsernameRemotelyWithCache (username) {
-        if (typeof this.usernameRemoteCache.hasOwnProperty[username] === 'Object') {
+        if (username in this.usernameRemoteCache this.usernameRemoteCache.hasOwnProperty[username] === 'Object') {
             return Promise.resolve(this.usernameRemoteCache[username]);
         }
         // username is not in our cache
--- a/src/components/join-flow/username-step.jsx
+++ b/src/components/join-flow/username-step.jsx
@@ -58,7 +58,7 @@ class UsernameStep extends React.Component {
     }
     // simple function to memoize remote requests for usernames
     validateUsernameRemotelyWithCache (username) {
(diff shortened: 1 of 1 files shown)


apple502j:
@R4356th hmm that JS doesn't make sense...

R4356th:
@apple502j?

src/components/join-flow/username-step.jsx
```diff
@@ -58,7 +58,7 @@ class UsernameStep extends React.Component {
     }
     // simple function to memoize remote requests for usernames
     validateUsernameRemotelyWithCache (username) {
-        if (this.usernameRemoteCache.hasOwnProperty(username)) {
+        if (typeof this.usernameRemoteCache.hasOwnProperty[username] === 'Object') {
```
towerofnix, line 61: Maybe you meant to get rid of `hasOwnProperty`? Seems like you were looking to access `this.usernameRemoteCache[username]`. `'Object'` should also be `'object'` -- typeof returns lowercase strings.

But this solution wouldn't be foolproof either - where before we would error because we were *overwriting* `hasOwnProperty` with an object (and then trying to access it as though it were still a function), with this change we would get a false-positive for properties which already exist! One example is `__proto__`: `typeof this.usernameRemoteCache['__proto__']` is `'object'` so we would be treating it like a cache value, even though it is not. (Examples like this are why `hasOwnProperty` was used in the first place: it does not return true for "implicit" properties like that.)

That said, I believe this solution *would* work if line 37 were changed to initialize `this.usernameRemoteCache` as `Object.create(null)` instead of `{}`, because then there would be no implicit properties to begin with. (Aside: With that change, it would be enough to just do `username in this.usernameRemoteCache` instead of explicitly checking the `typeof` value.)

It's worth noting that this PR should be contrasted with #4003, which opts to instead use the always-available `Object.prototype.hasOwnProperty`. I think the two PRs would be equally functional; @apple502j's would be a simpler change, but @R4356th's (this one, with changes above) would be making arguably better use of JS's features.

R4356th: I will test it once I get on my computer.

R4356th: @towerofnix, how does it look to you now?

towerofnix: @R4356th Mostly good! You changed the initialization line to `Object.create(null)`, which looks good -- but you missed the changes to the line I commented on, and those are necessary too to make the code work too. :P

R4356th: Alright, @towerofnix. Will fix as soon as I get on my computer.

R4356th: @towerofnix, changed now, I believe. I did not change that line because it looked unnecessary to me before, but the Travis build failed. So it is probably worth fixing.

towerofnix: aa it's a syntax error now! :P here's what the full line as it should be:

```
         if (username in this.usernameRemoteCache) { 
```

the second condition (the `this.usernameRemoteCache.hasOwnProperty[username] === 'Object'` you left there) isn't necessary -- if you kept it (and changed it to `typeof this.usernameRemoteCache[username] === 'object'`), it would be a sufficient check on its own. but `username in this.usernameRemoteCache` is *also* a sufficient check on its own, and it's simpler code, so that's what i recommend keeping in the end.

R4356th: @towerofnix, to me, changing <code>'Object</code> to <code>'object'</code> looks way more simpler (no offense meant), so I edited the file once again and changed it that way.

towerofnix: you forgot to remove `.hasOwnProperty` too, sorry. then it should work :P

R4356th: @towerofnix, my bad. I was on a rush yesterday.I have fixed it now.

towerofnix: @R4356th yeah no worries! looks like it is :'o

R4356th: Yay! Thank you so much for your help!

R4356th:
@benjiwheeler

benjiwheeler:
Thanks @R4356th

PR #4003, opened 2020-05-20
Fix hasOwnProperty username on join flow

### LINKED ISSUE #3850: Username "hasOwnProperty" crashes Join page

apple502j:
### Resolves:
Resolves #3850 

### Changes:
Instead of calling hasOwnProperty, call Object.prototype.hasOwnProperty.call

### DIFF
--- a/src/components/join-flow/username-step.jsx
+++ b/src/components/join-flow/username-step.jsx
@@ -58,7 +58,7 @@ class UsernameStep extends React.Component {
     }
     // simple function to memoize remote requests for usernames
     validateUsernameRemotelyWithCache (username) {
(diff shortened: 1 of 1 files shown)


benjiwheeler:
@apple502j Thanks for this -- I somewhat arbitrarily decided to merge https://github.com/LLK/scratch-www/pull/4002 instead. You taught me something new about JS with this :)

PR #4078, opened 2020-06-09
Add noscript tag

### LINKED ISSUE #3759: Pages with the Scratch 3.0 Theme are blank if JavaScript is disabled
If you attempt to view any pages which have the Scratch 3.0 theme (the home page, project page, etc.) while JavaScript is disabled, the page is blank without any explanation. There should be some kind of message which informs the user that they need to enable JavaScript to view the page.

### Steps to Reproduce

1. Block JavaScript for scratch.mit.edu in your browser settings
2. Visit a page which has the 3.0 theme (homepage, project page, project editor, etc.)

My browser / operating system: ChromeOS 12739.94.0, Chrome 80.0.3987.137, No Flash version detected

apple502j:
### Resolves:
Resolves #3759 

### Changes:
Adds "JavaScript is required to display this page. Please enable JavaScript in order to continue." message.

### DIFF
--- a/src/template.ejs
+++ b/src/template.ejs
@@ -54,7 +54,7 @@
 
     <body>
         <noscript>
-            <p>JavaScript is required to display this page. Please enable JavaScript in order to continue.</p>
+            <p>Your browser has Javascript disabled. Please go to your browser preferences and enable Javascript in order to use Scratch.</p>
         </noscript>
         <div id="app"></div>
 
--- a/src/template.ejs
+++ b/src/template.ejs
@@ -53,6 +53,9 @@
     </head>
 
     <body>
+        <noscript>
+            <p>JavaScript is required to display this page. Please enable JavaScript in order to continue.</p>
+        </noscript>
         <div id="app"></div>
 
         <!-- Vendor & Initialize (Session & Localization)-->


src/template.ejs
```diff
@@ -53,6 +53,9 @@
     </head>
 
     <body>
+        <noscript>
+            <p>JavaScript is required to display this page. Please enable JavaScript in order to continue.</p>
+        </noscript>
```
GrahamSH-LLK, line 58: Looks good! I think we would have to style it, though. Is it styled?

Accio1:
Shouldn't this be stylized using CSS?

GrahamSH-LLK:
> Shouldn't this be stylized using CSS?

Ha! We both said the same thing. I think @benjiwheeler said to give him basic outline, and the ST would style it.

benjiwheeler:
Thanks. The next step is for me to take this to the Design team and get their feedback.

BoomerScratch:
@R4356th Maybe tell the user this?
> JavaScript is required to display this page. Please [enable JavaScript](https://enablejavascript.io) in order to continue. If your browser does not support JavaScript, read about common workarounds.

(link needed for the "read about common workarounds")
We sadly can't add a link to the offline editor because all download pages require JS!

R4356th:
No, @BoomerScratch. That is the download link for 2.0. I am talking about these- https://downloads.scratch.mit.edu/desktop/Scratch%20Desktop%20Setup%203.11.1.exe, https://downloads.scratch.mit.edu/desktop/Scratch%20Desktop-3.11.1.dmg for Windows and MacOS respectively. Now, as for ChromeOS and Android, there should be links to the Google Play Store. All of them cannot stay there as that will take up space. So, I think there should be an only-HTML-and-CSS page containing these links.

GrahamSH-LLK:
@R4356th What if we made a non-react generated copy of [scratch.mit.edu/download](https://scratch.mit.edu/download)? We would have to do some fancy html, but we could probably get it to work.

BoomerScratch:
@R4356th Yes, but these links have to be updated over time. The link on maintenance mode also takes you to the 2.0 download:

I agree about the only-HTML-and-CSS one.

benjiwheeler:
FWIW, I think because this is at least an improvement over what we currently have, we should be able to use it, and then come back to doing a more complete design.

benjiwheeler:
blocked on discussing with design team

benjiwheeler:
Spoke with design team and agreed that this is an improvement! Per their feedback I changed the wording slightly.

Thanks, @apple502j!

PR #4095, opened 2020-06-11
remove email opt-in from join flow

benjiwheeler:
Resolves https://github.com/LLK/scratchr2/issues/5723

### Changes:

* removes email opt-in checkbox from join flow
* omits `subscribe` parameter, which will default to False in scratchr2
* removes email opt in string
* removes tests that no longer apply

### Test Coverage:

_Please show how you have added tests to cover your changes or describe how you have tested the changes (include a screenshot if possible)._

PR #4096, opened 2020-06-12
[Master] remove email opt-in from join flow

benjiwheeler:
hotfix version of PR https://github.com/LLK/scratch-www/pull/4095 which resolves https://github.com/LLK/scratchr2/issues/5723

Test plan:

Already tested on staging:
1. create new account
2. confirm that opt in field does not appear in form
3. confirm that account creation works
4. confirm that email confirmation works
5. confirm that sharing a project works

For production, we should repeat these steps.

We should also confirm that the email address was not added to our Mailchimp list.


benjiwheeler:
Update: tested in production using the steps listed above. Confirmed that the new user is not in the Mailchimp user list.

PR #4114, opened 2020-06-17
Set wiki link per user's selected language

R4356th:
### Resolves:

Resolves #815.

### Changes:

Selects Scratch Wiki link according to the user's selected language.

PR #4121, opened 2020-06-18
Set wiki link per user's selected language

### LINKED ISSUE #815: Update Wiki Footer Link to use User's Language

apple502j:
### Resolves:
Resolves #815 

### Changes:
This is similar to #4114 except two things:
- `getScratchWikiLink` is a library.
- `xx-yy`  fallbacks to `xx` first, then `en`

### Test Coverage:
Added a unit test


R4356th:
So you added a library? I believe the least possible amount of changes should be made. 😛

benjiwheeler:
I made a few changes, just to make the link into a prop so that react won't re-render it unnecessarily.

Issue #4038, opened 2020-05-27
As a new scratcher, there are old images


GrahamSH-LLK:
When you are a new scratcher, you see this 
![image](https://user-images.githubusercontent.com/64214252/83068208-f86faa80-a035-11ea-8639-670a51bb2504.png) box. It shows 2.0 blocks. This would be an easy fix, as it is just an image. This could be confusing for new scratchers.  I suggest using this image because the new block for the drum is too long.
![image](https://user-images.githubusercontent.com/64214252/83075699-49859b80-a042-11ea-8eb1-98c4a5a4f431.png)

### Steps to Reproduce

Create a new account. 
See 2.0 blocks

### Operating System and Browser

N/A

GrahamSH-LLK:
It would also have to be fixed here: 
![image](https://user-images.githubusercontent.com/64214252/83272774-97f97e00-a199-11ea-9ada-041ab431ef96.png), but this may be scratchr2. A mockup looks like this: 
![image](https://user-images.githubusercontent.com/64214252/83273103-16eeb680-a19a-11ea-9efa-2521a19b2d7a.png)

Explosion-Scratch:
Yes, this could be confusing and should be changed. Should I suggest it on the scratch forums to aee what others think?

GrahamSH-LLK:
I think we'll be fine, but you can if you want to.

GrahamSH-LLK:
@Explosion-Scratch Do you know who is in charge of images like this?

Explosion-Scratch:
@GrahamSH-LK I'm sorry, I'm not part of the ST nor do I help develop the site, I just comment and suggest stuff :)

GrahamSH-LLK:
@Explosion-Scratch I know. I just have no idea who is in charge, and I was hoping you would know.

Accio1:
It looks to me like replacing https://github.com/LLK/scratch-www/blob/develop/static/images/welcome-learn.png with a new image would solve this problem.

GrahamSH-LLK:
@Accio1  Yes. That should work. Do _you_ have any idea who is in charge of this on the ST?

GrahamSH-LLK:
@rschamp It says that you created this image, so do you have any thoughts on this?

benjiwheeler:
I will bring this to the attention of the Scratch Team's Resources team.

GrahamSH-LLK:
Great!

benjiwheeler:
[...]

PR #4144, opened 2020-06-25
updated blocks new scratcher sees

benjiwheeler:
Resolves https://github.com/LLK/scratch-www/issues/4038

Adds updated blocks image that new scratcher will see

revises css a bit to make the more narrow code block look even

Before:

image

After:

image


kathymakes:
Adding an updated svg file — to solve for pixelated but more importantly having the intended font — the one in this image is using our system fall back Arial. Thanks so much for working on this!

[homepage_blocks.svg.zip](https://github.com/LLK/scratch-www/files/4833923/homepage_blocks.svg.zip)

benjiwheeler:
I couldn't get that svg below 14kb, but I tried converting it to png and got it down to 4kb.

svg:
image

png:
image

PR #4146, opened 2020-06-25
[Master] use 'evergreen' URLs for Scratch Desktop downloads

cwillisf:
### Changes:

Future Scratch Desktop releases will use redirects to point a stable URL for each platform to the latest version of that platform's installer. This means we will no longer need to make a `scratch-www` change for each Scratch Desktop release. This change is also the first of many steps toward renaming "Scratch Desktop" to "Scratch".

See also #4145

### Test Coverage:

URLs tested by hand.

### DIFF
--- a/src/components/install-scratch/install-scratch.jsx
+++ b/src/components/install-scratch/install-scratch.jsx
@@ -13,8 +13,8 @@ const Step = require('../../components/steps/step.jsx');
 require('./install-scratch.scss');
 
 const downloadUrls = {
-    mac: 'https://downloads.scratch.mit.edu/desktop/Scratch%20Desktop-3.11.1.dmg',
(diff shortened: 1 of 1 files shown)


apple502j:
Will it be possible to download 3.11.1 using current link?

cwillisf:
All previous URLs still work, yeah. In fact, these new links just redirect to the same kind of URL we've used in the past. If you want (too much) more information on the redirects, you can download one of these new links using `curl --verbose --location --remote-name` (or `curl -vLO` for short).

Also: yep, 3.12 is indeed just a GUI bump :)

At some point in the future I hope to offer a way to download previous versions. They're still on the download server, so we might as well provide a way to get to them in case there's any compatibility or testing reason, or even just for historical interest.

Issue #4079, opened 2020-06-09
Poll for notification count updates using decaying time


benjiwheeler:
We currently poll for notification counts on a regular basis, in both www and scratchr2.

We should instead check for updates using decaying times -- 5 minutes, then 10 minutes, then 20 minutes, etc.

The endpoint accounts for about 1/3 of our api backend time.

apple502j:
Is it problem for www? I think scratchr2 is slower

picklesrus:
Fixed by https://github.com/LLK/scratch-www/pull/4293

PR #4169, opened 2020-07-06
Exponentially back off the time between message polling instead of do…

picklesrus:
…ing it every two minutes.

### Resolves:

https://github.com/LLK/scratch-www/issues/4079

### Changes:

Back off how often we poll for new messages

### Test Coverage:

There are unit tests for the new setupMessagePolling function.

### DIFF
--- a/src/components/navigation/www/navigation.jsx
+++ b/src/components/navigation/www/navigation.jsx
@@ -37,14 +37,16 @@ class Navigation extends React.Component {
     componentDidMount () {
         if (this.props.user) {
             // Setup polling for messages to start in 2 minutes.
-            setTimeout(this.pollForMessages.bind(this, 2), 2 * 60 * 1000);
+            const twoMinInMs = 2 * 60 * 1000;
+            setTimeout(this.pollForMessages.bind(this, twoMinInMs), twoMinInMs);
         }
     }
     componentDidUpdate (prevProps) {
         if (prevProps.user !== this.props.user) {
             this.props.handleCloseAccountNav();
             if (this.props.user) {
-                setTimeout(this.pollForMessages.bind(this, 2), 2 * 60 * 1000);
+                const twoMinInMs = 2 * 60 * 1000;
+                setTimeout(this.pollForMessages.bind(this, twoMinInMs), twoMinInMs);
             } else {
                 // clear message count check, and set to default id.
                 clearTimeout(this.state.messageCountIntervalId);
@@ -70,16 +72,16 @@ class Navigation extends React.Component {
         return `/users/${this.props.user.username}/`;
     }
 
-    pollForMessages (minutes) {
+    pollForMessages (ms) {
         this.props.getMessageCount(this.props.user.username);
         // We only poll if it has been less than 30 minutes.
         // Chances of someone actively using the page for that long without
         // a navigation is low.
-        if (minutes < 32) {
-            const nextFetch = minutes * 2;
+        if (ms < 32 * 60 * 1000) { // 32 minutes
+            const nextFetch = ms * 2; // exponentially back off next fetch time.
             const timeoutId = setTimeout(() => {
                 this.pollForMessages(nextFetch);
-            }, nextFetch * 60000);
+            }, nextFetch);
             this.setState({ // eslint-disable-line react/no-did-mount-set-state
                 messageCountIntervalId: timeoutId
             });
--- a/src/components/navigation/www/navigation.jsx
+++ b/src/components/navigation/www/navigation.jsx
@@ -69,7 +69,7 @@ class Navigation extends React.Component {
 
     pollForMessages (ms) {
         this.props.getMessageCount(this.props.user.username);
-        // We only poll if it has been less than 30 minutes.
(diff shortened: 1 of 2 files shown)


src/components/navigation/www/navigation.jsx
```diff
@@ -77,6 +69,23 @@ class Navigation extends React.Component {
         if (!this.props.user) return;
         return `/users/${this.props.user.username}/`;
     }
+
+    setupMessagePolling (minutes) {
```
benjiwheeler, line 73: if it's not too much trouble, it might make more sense to use seconds as the unit -- that seems to be more of a convention than minutes.

src/components/navigation/www/navigation.jsx
```diff
@@ -77,6 +69,23 @@ class Navigation extends React.Component {
...
+        this.props.getMessageCount(this.props.user.username);
+        // We only poll if it has been less than 30 minutes.
+        // Chances of someone actively using the page for that long without
+        // a navigation is low.
+        if (minutes < 32) {
+            const nextFetch = minutes * 2;
+            const timeoutId = setTimeout(() => {
+                this.setupMessagePolling(nextFetch);
+            }, nextFetch * 60000);
```
benjiwheeler, line 82: The convention you used earlier, on line 47, where you showed the multiplication that got you this number ("2 * 60 * 1000") might also be good here

src/components/navigation/www/navigation.jsx
```diff
@@ -27,35 +27,27 @@ class Navigation extends React.Component {
         super(props);
         bindAll(this, [
             'getProfileUrl',
-            'handleSearchSubmit'
+            'handleSearchSubmit',
+            'setupMessagePolling'
         ]);
         this.state = {
             messageCountIntervalId: -1 // javascript method interval id for getting messsage count.
```
benjiwheeler, line 34: this is maybe going beyond the scope of this change, but I don't think that messageCountIntervalId belongs inside state. There's no need for the component to re-render just because messageCountIntervalId changed. 

There are a few examples of how you might just store the response from `setTimeout()` in a regular member variable: in www, src/components/modal/social/container.jsx does this, and in gui, src/components/action-menu/action-menu.jsx does this.

But I totally understand if you don't want to make that change as part of this change!

benjiwheeler:
this looks good, just a few suggestions! Way to go getting those tests working

src/components/navigation/www/navigation.jsx
picklesrus: Good point! This name is left over from a previous iteration. Renamed to pollForMessages since that's actually what it is doing.

src/components/navigation/www/navigation.jsx
picklesrus: I changed everything over to being in ms as you suggested. Since I did that, nextFetch is now in millis so I don't need the multiplication anymore.

src/components/navigation/www/navigation.jsx
picklesrus: I was swamped this week so didn't have time to look into this. If you don't have time to review before you head out of office, maybe I'll take a look next week. Otherwise, I suggest wee keep it for a follow up.

picklesrus:
I moved the intervalId tracking out of the state object to a member variable. Also added some tests -- glad I did because they caught a bug that would have been hard to find otherwise :)

src/components/navigation/www/navigation.jsx
```diff
@@ -27,56 +27,60 @@ class Navigation extends React.Component {
...
         if (!this.props.user) return;
         return `/users/${this.props.user.username}/`;
     }
+
+    pollForMessages (ms) {
+        this.props.getMessageCount(this.props.user.username);
+        // We only poll if it has been less than 30 minutes.
+        // Chances of someone actively using the page for that long without
+        // a navigation is low.
+        if (ms < 32 * 60 * 1000) { // 32 minutes
```
benjiwheeler, line 75: This could be 30 instead of 32 to match the comment, but it's not a big deal

src/components/navigation/www/navigation.jsx
```diff
@@ -27,56 +27,60 @@ class Navigation extends React.Component {
...
-                }, 120000); // check for new messages every 2 mins.
-                this.setState({ // eslint-disable-line react/no-did-update-set-state
-                    messageCountIntervalId: intervalId
-                });
+                const twoMinInMs = 2 * 60 * 1000;
+                this.messageCountTimeoutId = setTimeout(this.pollForMessages.bind(this, twoMinInMs), twoMinInMs);
             } else {
                 // clear message count check, and set to default id.
-                clearInterval(this.state.messageCountIntervalId);
+                clearTimeout(this.messageCountTimeoutId);
```
benjiwheeler, line 51: I think ideally we'd test messageCountTimeoutId to see if it's !== -1 like on line 59. But it's not a big deal because I'm pretty sure JS will just ignore it if we call clearTimeout(-1) or call it with a stale function id

benjiwheeler:
LGTM! Two small notes, but this can go in without my needing to review again!

Issue #4024, opened 2020-05-27
Wrong State info for Bangladesh for teacher registration


R4356th:
Kind of continuing from #4019.

Per the current information Scratch has about Bangladesh, it has about a hundred states; while in reality, the number is simple zero. The states which Bangladesh has according to Scratch are all Bangladesh's 64 districts appearing repeatedly while teacher registration. This makes it confusing for teachers while registering because they get asked for the state they live in while they do not live in any theoretically. 
A screenshot can be found here- https://user-images.githubusercontent.com/61620631/82918678-e0951980-9f96-11ea-94c4-e26cbad69d30.png .
### Steps to Reproduce

Select Bangladesh as your country during teacher registration and complete all the necessary steps to reach the address step.

### Operating System and Browser

Windows 10 Pro, Chrome 83.0.4103.61 (Official Build) (64-bit), Android 6, Chrome 80 (64 bit) (Official build)

R4356th:
https://github.com/LLK/scratch-www/blob/develop/test/unit/lib/country-data.test.js#L120 says that there are 71 sub divisions in Bangladesh. If you really want to include states, then there are eight divisions in Bangladesh, which (even though are not so different in terms of administration and laws, tax etc. are the same for every division) might be safe to consider as states.

apple502j:
The data is from `iso-3166-2` - maybe bump that?

R4356th:
> The data is from `iso-3166-2` - maybe bump that?

@apple502j, I searched in Wikipedia for what it contained and they have got the right information- 64 districts and 8 divisions. However, as I have stated in my previous comment the info about my country (Bangladesh) having 71 sub-divisions is correct, too; but not the way how the country's administrative works are conducted.

benjiwheeler:
@R4356th what is tricky is that each country has different names for what we might call "states" -- and, as you say, some countries have no real similar concept. 

We could just not require a state to be selected -- that would be the easiest change. Or we could add country-specific settings, so that only Divisions of Bangladesh, and not Districts, would show up in that list.

R4356th:
[...]

PR #4175, opened 2020-07-08
make state optional for non-US, limit states to certain kinds of region

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/4024

### Changes:

* makes the "State" select optional for countries besides the US
* change the "State" string to "State / Region"
* adds "majorRegionType" optional field to our country-by-country data library, with code to only include major regions in the object that our subdivisionOptions function returns.

### Test Coverage:


R4356th:
Just a note that it does not solve the wrong division name issue though. Chittagong and Barisal are spelled as Chattogram and Barishal now.

benjiwheeler:
@R4356th Unfortunately, that's an issue with the quality of the underlying ISO-3166 (IIRC) data.

R4356th:
So this does not resolve #4024, right?

ljy-002:
Taiwan and Hong Kong should be deleted because they belong to China

apple502j:
@ljy-002 This will not be fixed - see #4317

benjiwheeler:
> So this does not resolve #4024, right?

Correct -- this just makes #4024 not matter as much

> Taiwan and Hong Kong should be deleted because they belong to China

We are just using the ISO-3166 standard data, we're not sure about the details of what's in there.

benjiwheeler:
I'm removing the milestone because the underlying issue is not currently scheduled and has no priority level.

ljy-002:
Hard!
This is a distressing question!
We try to discuss with the Chinese government
Come on!
@ Taiwanese, come on, too

PR #4249, opened 2020-07-29
Remove tap.plan() from footer links test

BryceLTaylor:
### Resolves:

The tests were failing on Travis, this should make them work smoothly.

### Changes:

Just removes the tap.plan() option.  After removing some tests, it didn't like that the number didn't match.


benjiwheeler:
LGTM, merging

Issue #3028, opened 2019-06-04
Need to address skipped My Stuff tests 


BryceLTaylor:
Several tests are being skipped from this pr: https://github.com/LLK/scratch-www/pull/3025, 
 in order to reduce the flakiness of the integration tests.  

A test in my-stuff that tests whether the +New studio button takes you to the new studio was failing, so I skipped it.  It needs investigation.

PR #4261, opened 2020-08-03
Move my stuff tests to jest

BryceLTaylor:
### Resolves:

resolves https://github.com/LLK/scratch-www/issues/3028. 

### Changes:

This PR moves all of the My-Stuff tests into Jest, modifying them to get them to pass.  

Since the Jest tests run in parallel and this is the second test suite added, I have written the new My-Stuff tests to use a separate username from the other tests.  This should make it so signing in and out doesn't cause session cookies to expire.  The my-stuff tests take the SMOKE_USERNAME environment variable and add a 1 to the end of it.  The accounts should have the same password.  So the username test would become test1 and use the same password for both accounts.

In order to get the new jest tests to pass, I updated to chromedriver version 84.  This caused some other tests to fail, so I had to add a longer timeout and a few waits to get them to pass.


BryceLTaylor:
Chromedriver 84 requires node 10 or 12 to use, which WWW isn't set up for on travis (it seems).  @rschamp  We should discuss the ramifications of this.

apple502j:
@BryceLTaylor I think node-sass is causing problems with upgrading (currently 4.6.1, node10 support added in 4.9.0)

Also *why are you testing scratchr2 stuff in scratch-www*

apple502j:
@BryceLTaylor It's not possible to bump nodejs without updating node-sass. Updating node-sass requires changes to most (if not all) of the SCSS files.

PR #4264, opened 2020-08-03
[Master] finish renaming 'Scratch Desktop' to 'Scratch'

cwillisf:
**Do not merge this until we are ready to release the new version of the Scratch app!**

We're working on some "paperwork" which may take some time. Meanwhile, I'd like to get this PR ready to merge to reduce the delay once we're fully ready for the release.

### Changes:

This change includes the last steps for renaming "Scratch Desktop" to "Scratch" including:
- adjust macOS installation instructions to mention "Scratch" instead of "Scratch Desktop"
- edit the DMG screenshot to fix the app name, DMG title, and DMG icon

### Test Coverage:

Tested manually on staging:

image


apple502j:
@chrisgarrity WWW does have default text (iirc), it's l10n.json

cwillisf:
@chrisgarrity I think the name will probably be settled this week, so let's talk some time between then and the next Scratch Desktop release to make sure I do this right. Also I suspect it'll end up not being a hotfix after all, but it's hard to say for sure yet...

chrisgarrity:
@apple502j in www the uses of `<FormattedMessage ... />` do not include the `defaultMessage` parameter. So if the message id is missing from the translations, you get the id, not the message. If www gets out of sync with respect to scratch-l10n and there are new messages you can get message ids in all languages, even English. It's avoidable, but something we need to be careful of when deploying a hotfix with new strings.

chrisgarrity (approved):
If necessary, I can manually update the download-l10n.json file on transifex before we merge this to make sure the new strings are available. We might also need to force an l10n update.

cwillisf:
Alright, I've updated this for the fact that we ended up calling the app "Scratch 3" instead of "Scratch"

@chrisgarrity if testing goes well for Scratch 3.16.1, we'll probably want to get this in some time next week (October 5-9). We could even do it as part of the normal release rather than a hotfix if that makes things easier. I'm not sure what's best for Transifex purposes, so I'm open to whatever you advise. I'm also happy to make adjustments to this PR if appropriate.

chrisgarrity:
@cwillisf It would be easier if this went in as part of the normal weekly release, on Monday or Tuesday. There is a daily travis build that pushes any new www strings on the develop branch to transifex. Late Tuesday night there is a job that pulls new translations, and new strings into l10n in preparation for the weekly build. 

The alternative is:
* manually push a translations update (grab a TX_TOKEN out of 1 password to use as credentials, then `npm run i18n:push`)
* manually release scratch-l10n (see https://github.com/LLK/scratch-l10n/wiki/Developer-Guide)
scratch-www depends on l10n 'latest', so you don't need to update the package.json.

cwillisf:
Sounds good - I'll close this (since we're not hotfixing it) and merge the `develop` one in preparation for the regular release and the Travis job.

Issue #4025, opened 2020-05-27
Build failure with Node 12.16.2


southwolf:
![image](https://user-images.githubusercontent.com/150648/83043902-06ddb680-a076-11ea-84f1-0c010c3de598.png)

### Steps to Reproduce

```
git pull
npm i
```

According to https://github.com/sass/node-sass/issues/2632

https://github.com/sass/node-sass/releases/tag/v4.12.0

Node-sass needs upgrading to 4.12.0 (or higher) to support NodeJS 12

### Operating System and Browser

MacOS 10.15.4
NodeJS 12.16.2
Node-gyp 6.1.0

benjiwheeler:
It looks like https://github.com/LLK/scratch-www/pull/4026 caused a problem with our SASS media queries and/or SASS variables. 

You can see this when the screen gets narrow:

Production:
![image](https://user-images.githubusercontent.com/3431616/85721372-4cb28c80-b6bf-11ea-83bc-82ddae7a18e1.png)

With https://github.com/LLK/scratch-www/pull/4026:
![image](https://user-images.githubusercontent.com/3431616/85721381-4e7c5000-b6bf-11ea-82e3-10068bb10226.png)

Some relevant issues and posts:
* https://sass-lang.com/blog/sass-35-release-candidate
* https://github.com/sass/node-sass/issues/2358
* https://github.com/sass/node-sass/issues/1886
* https://github.com/sass/node-sass/issues/2889
* https://github.com/sass/node-sass/issues/2762
* https://github.com/LLK/scratch-www/pull/4026/files

benjiwheeler:
@southwolf Please take a look and let us know if you want to keep working on this.

rschamp:
Resolved by #4274

PR #4265, opened 2020-08-03
Update node version used by travis and node-sass

BryceLTaylor:
### Resolves:

Updates node version used by travis, which is necessary for this pr: https://github.com/LLK/scratch-www/pull/4261

### Changes:

Also updates the version of node-sass, which is required after updating node.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- '8'
+- '12'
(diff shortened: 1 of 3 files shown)


apple502j:
@BryceLTaylor Are you sure it doesn't mess with the "breaking changes"?
See also #4025 - also mentioning @benjiwheeler

PR #4274, opened 2020-08-04
Fix bad variable interpolation/calculation in frameless scss

### LINKED ISSUE #4168: What should we do about frameless?
To solve #4025 frameless.scss needs to be updated. However, frameless is no longer maintained. We can fix frameless.scss, or find another library that acts like frameless.

rschamp:
This used to work, even though it shouldn’t have. Upgrading node-sass caused our breakpoints to … break.  Without this, the compiler outputs invalid CSS including the "- 1", which causes the media query definitions to be ignored.

### Resolves:

Resolves #4168 

### Changes:

Update syntax for operations in frameless.scss

### Test Coverage:

Tests for CSS are hard. @BryceLTaylor and I looked at scratch.ly to visually compare to production.

### DIFF
--- a/src/_frameless.scss
+++ b/src/_frameless.scss
@@ -48,20 +48,20 @@ $mobile: 480px;
 *          ... intermediate-and-smaller |
 */
 
-$small: "only screen and (max-width : #{$mobile}-1)";
-$medium: "only screen and (min-width : #{$mobile}) and (max-width : #{$tabletPortrait}-1)";
-$intermediate: "only screen and (min-width : #{$tabletPortrait}) and (max-width : #{$desktop}-1)";
+$small: "only screen and (max-width : #{$mobile - 1})";
+$medium: "only screen and (min-width : #{$mobile}) and (max-width : #{$tabletPortrait - 1})";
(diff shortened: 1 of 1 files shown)


src/_frameless.scss
```diff
@@ -48,20 +48,20 @@ $mobile: 480px;
 *          ... intermediate-and-smaller |
 */
 
-$small: "only screen and (max-width : #{$mobile}-1)";
-$medium: "only screen and (min-width : #{$mobile}) and (max-width : #{$tabletPortrait}-1)";
-$intermediate: "only screen and (min-width : #{$tabletPortrait}) and (max-width : #{$desktop}-1)";
+$small: "only screen and (max-width : #{$mobile - 1})";
+$medium: "only screen and (min-width : #{$mobile}) and (max-width : #{$tabletPortrait - 1})";
+$intermediate: "only screen and (min-width : #{$tabletPortrait}) and (max-width : #{$desktop - 1})";
```
rschamp, line 53: E.g. this was outputting `only screen and (max-width : 480px-1)` but now it outputs `only screen and (max-width : 479px)`

PR #4283, opened 2020-08-06
Localize email banner on splash

### LINKED ISSUE #4251: Translate email confirmation banner strings?
In the src/views/splash/presentation.jsx file, there are a few strings that are not translated.

* "Confirm your email" (a link) and "to enable sharing" -- these may be difficult to translate due to the different structures of sentences in different languages. Perhaps make the whole sentence a link?
* "Having trouble?"

apple502j:
### Resolves:
Resolves #4251

### Changes:
Localizes email banner on splash page, with 3 strings added to splash-l10njson

### Test Coverage:
Cannot test locally, but the code is same as one I sent on the issue

### DIFF
--- a/src/views/splash/l10n.json
+++ b/src/views/splash/l10n.json
@@ -44,7 +44,11 @@
     "welcome.learn": "Learn how to make a project in Scratch",
     "welcome.tryOut": "Try out starter projects",
     "welcome.connect": "Connect with other Scratchers",
-    
+
     "activity.seeUpdates": "This is where you will see updates from Scratchers you follow",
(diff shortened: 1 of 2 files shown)


benjiwheeler:
This looks good to me at first glance -- I'll test it soon so we can merge it! It's currently scheduled to be tested in September.

paulkaplan:
@apple502j thanks for this. @benjiwheeler I'll take over this PR, I'm actually going to bring in the commit from @apple502j to a slightly different PR since we are going to update the text on this banner for the new email confirmation modal. I don't want to merge this with the existing strings because the new one will only have one link, don't want to double-translate. See https://github.com/LLK/scratch-www/issues/4390.

PR #4316, opened 2020-08-14
Change country to location

### LINKED ISSUE #4313: Wrong text in the Join Scratch menu
When you select where you live, it says “country” instead of “location” as the profile and account settings display.

apple502j:
### Resolves:
Resolves #4313 
Resolves? #4315 

### Changes:
Changes "country" to "location" in l10n files. Does not change file names or class names though.

Hopefully this fixes China censorship.


benjiwheeler:
closing per https://github.com/LLK/scratch-www/issues/4313#issuecomment-673646924

PR #4352, opened 2020-08-25
Fix my stuff tests

BryceLTaylor:
### Resolves:

The My Stuff tests were failing on Saucelabs after I moved them out of Tap and into Jest.

### Changes:

It seemed that the next test was starting before the final expect line was being run.  This adds a short wait at the end of several tests to make sure the tests run properly in order.

This also updates the version of Chrome used by Saucelabs to match the recent Chromedriver update.

### DIFF
--- a/test/integration/my-stuff.test.js
+++ b/test/integration/my-stuff.test.js
@@ -41,29 +41,41 @@ describe('www-integration my_stuff', () => {
 
     test('verify My Stuff structure (tabs, title)', async () => {
         await driver.get(myStuffURL);
-        await findByXpath('//div[@class="box-head"]/h2');
-        await findByXpath('//li[@data-tab="projects"]/a');
-        await findByXpath('//li[@data-tab="shared"]/a');
-        await findByXpath('//li[@data-tab="unshared"]/a');
-        await findByXpath('//li[@data-tab="galleries"]/a');
-        await findByXpath('//li[@data-tab="trash"]/a');
+        let header = await findByXpath('//div[@class="box-head"]/h2');
+        let headerVisible = await header.isDisplayed();
+        await expect(headerVisible).toBe(true);
+        let allTab = await findByXpath('//li[@data-tab="projects"]/a');
+        let allTabVisible = await allTab.isDisplayed();
+        await expect(allTabVisible).toBe(true);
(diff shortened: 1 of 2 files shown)


test/integration/selenium-helpers.js
```diff
@@ -68,7 +68,7 @@ class SeleniumHelper {
         let driverConfig = {
             browserName: 'chrome',
             platform: 'macOS 10.14',
-            version: '76.0'
```
rschamp, line 71: Maybe we should pull this from package.json?

test/integration/my-stuff.test.js
```diff
@@ -55,13 +55,15 @@ describe('www-integration my_stuff', () => {
         await driver.sleep(6000);
         let gui = await findByXpath('//div[@class="guiPlayer"]');
         await expect(gui.isDisplayed());
+        await driver.sleep(1500);
```
rschamp, line 58: Why do these need to wait if the test is already over?

BryceLTaylor: We shouldn't, but for some reason when running on saucelabs it seems to start the next test before the expect() line returns.  The tests pass with the driver.sleep and fail without them.

test/integration/selenium-helpers.js
BryceLTaylor: Oh, that's a good idea.

test/integration/my-stuff.test.js
rschamp: Is `gui.isDisplayed` an async function? I assume it is since it has to ask saucelabs for the answer. If that's what we're trying to wait on with the `await`, then I don't think `await expect(gui.isDisplayed())` will work, since it's going to await the `expect`, and `expect` isn't async. So basically if `gui.isDisplayed` is what we're waiting for, we should separately `await` that, and then run the `expect` on the result.  Otherwise `expect` will immediately return, and the next test would start.

rschamp: And I should say, it would be better to await gui.isDisplayed than to await the flat 1.5s because it's possible gui.isDisplayed will finish sooner than 1.5s, and then our tests can finish faster (e.g. when we run tests locally).  And in the case that it takes longer than 1.5s to return, the test will still pass.

PR #4420, opened 2020-09-14
Pin chrome version for saucelabs to chromedriver version

BryceLTaylor:
### Resolves:

Occasionally newer versions of chromedriver will not work with older versions of Chrome and vice versa.  Remote tests will then fail due to a mismatch between Chrome and Chromedriver.   

### Changes:

This change will cause the automated tests that run through Saucelabs to always use a version of chrome based on the chromedriver version.  Chromedriver versions are numbered to match the chrome versions they support, so it extracts the first two version number from the chromedriver package number to get the chrome version.  

This way we only need to update the version in one place.

### DIFF
--- a/test/integration/selenium-helpers.js
+++ b/test/integration/selenium-helpers.js
@@ -1,6 +1,7 @@
 const webdriver = require('selenium-webdriver');
 const bindAll = require('lodash.bindall');
 require('chromedriver');
+let chromedriverVersion = require('chromedriver').version;
 
 const headless = process.env.SMOKE_HEADLESS || false;
 const remote = process.env.SMOKE_REMOTE || false;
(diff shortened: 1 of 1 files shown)


test/integration/selenium-helpers.js
```diff
@@ -62,13 +63,21 @@ class SeleniumHelper {
...
 
+    getChromeVersionNumber () {
+        let chromedriverVersion = pkg.devDependencies.chromedriver;
+        let versionFinder = /\d+\.\d+/;
+        let versionArray = versionFinder.exec(chromedriverVersion);
+        return versionArray[0];
+    }
+
     getSauceDriver (username, accessKey, name) {
+        let chromeVersion = this.getChromeVersionNumber();
```
rschamp, line 74: I think we should handle the case where this comes back `undefined` if e.g. we move `chromedriver` from `devDependencies` to `dependencies`. I think throwing a nice error message about how you couldn't find the`chromedriver` version from `package.json` would be nice.

Now that I write that — can we just import and inspect `chromedriver` for the version rather than looking at the `package.json`?

rschamp (changes requested):

test/integration/selenium-helpers.js
BryceLTaylor: Pulled the version of chromedriver from the package and have removed the reference to the package.json.

rschamp (approved):
Looks good, thanks!

PR #4435, opened 2020-09-17
Stop running smoke tests unless the branch pushes to prod or staging.

picklesrus:
Motivation: stop running them on dependapot branches/PRs since they don't make sense.

Dependabot makes PRs from branches on LLK/scratch-www and travis was running the smoke tests there needlessly.

We're still running builds twice on these PRs... but this is progress.


rschamp (approved):
Thanks!

PR #4456, opened 2020-09-25
Only run integration/smoke tests on specific branches

picklesrus:
This keeps the integration tests from running on PRs that are made from branches on LLK/www (e.g. all dependabot PRs)

It is a redo of #4450 that's regex had messed up slashes. In this one, /'s are escaped in the appropriate places (I think at least!) 
Thanks to @BryceLTaylor for pointing at the / problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -141,6 +141,6 @@ stages:
 - name: test
   if: type != cron
 - name: smoke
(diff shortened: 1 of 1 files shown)


BryceLTaylor (commented):
Looks good.  I think that should work this time.

Issue #4308, opened 2020-08-12
missing space in teacher account request form


JeffaloBob:
there is a missing space between the bold text and the regular text in the teacher sign up request form.

![image](https://user-images.githubusercontent.com/40470736/90045126-e1ecfd00-dcce-11ea-9d19-c50195077c8a.png)

### Steps to Reproduce

1. visit https://scratch.mit.edu/educators/register
2. notice the missing space

### Operating System and Browser
Chrome on Windows 10

Nambaseking01:
Huh, that's the second grammar mistake this week - definitely interesting how many there have been popping up...

JeffaloBob:
this hasn't been set as help wanted yet but i think the issue is here https://github.com/LLK/scratch-www/blob/develop/src/components/registration/steps.jsx#L183

benjiwheeler:
Thanks, @JeffaloBob and @Nambaseking01. This could probably be fixed by adding a `&nbsp;` to https://github.com/LLK/scratch-www/blob/develop/src/components/registration/steps.jsx#L183 .

(Or maybe there's a better solution, we're open to other possibilities)

benjiwheeler:
Fixed by #4474

PR #4474, opened 2020-10-02
add missing space to teacher request form

JeffaloBob:
### Resolves:

#4308 

### Changes:

I added a non breaking space, so the teacher account request form name warning should now have a space (assuming I did it correctly)

### Test Coverage:

unfortunately, I do not have a setup where I can easily test scratch-www, but this is a small change and my fingers are crossed that I didn't mess up


benjiwheeler:
Looks good, thanks @JeffaloBob !

image

PR #4479, opened 2020-10-02
Move footer integration tests to jest

BryceLTaylor:
### Resolves:

moves test out of Tap and into Jest.

### Changes:

Moves the footer integration tests out of TAP and into jest.  Also removes two tests that pointed to external sites.


test/integration/footer-links.test.js
```diff
@@ -0,0 +1,142 @@
...
+    });
+
+    afterAll(async () => await driver.quit());
+
+    // ==== About Scratch column ====
+
+    test('click About Scratch link', async () => {
+        await clickText('About Scratch');
+        let url = await driver.getCurrentUrl();
+        expect(url).toMatch(/about\/?$/);
```
rschamp, line 35: We could be stricter here, right now we only ensure the URL ends in `about`.
```suggestion
        let url = await driver.getCurrentUrl();
        let pathname = (new URL(url)).pathname;
        expect(pathname).toMatch(/^\/about\/?$/);
```

rschamp (changes requested):
This looks good overall, I have one suggestion about how to test the URLs.

test/integration/footer-links.test.js
BryceLTaylor: Okay, I updated it for all of the tests.

rschamp (approved):

Issue #4447, opened 2020-09-23
Improve communication when users get muted


chrisgarrity:
When users post unconstructive or disrespectful comments they can end up "muted", unable to comment, for some amount of time. Currently there is very little explanation of how they ended up muted, and no redirection towards positive engagement.

### Design

Final design copy and graphical elements TBD /cc @kathymakes @zoebentley 

Flow is a non-dismissible three step modal:
1. General info modal with next button
2. Redirect towards positive engagement with 'back' and 'next' buttons
3. Closing message with 'back' and 'close' buttons

kathymakes:
Thanks Chris!

Here's the most up to date **[invision file](https://invis.io/62YSG3GVF5X)** – @picklesrus we can work together to finesse the exact spacing, e.g. making sure everything is a multiple of 4px.

And here's the **[copytext doc](https://docs.google.com/document/d/1FNEB8LN-VrlzcKwRATRjYjqT45buHDQ6bBEBErFJ6OA/edit#)** which is a WIP!

Strings and illustrations are still not final yet

chrisgarrity:
Clarifying when educational modal is shown:
* on first mutable offense, on subsequent mutable offenses we will just show the comment disabled component (#4459)
* if the user clicks 'learn more' link in the comment disabled component

PR #4482, opened 2020-10-05
Initial skeleton of mute modal. 

picklesrus:
Adds a 2 step modal for giving information about what it means to be muted. 
This is a step towards #4447. 

This is just the skeleton. The CSS is pretty solid, but there are still lots of things to do including:
- Wire up the modal to show at the right time
- Update to reflect final content & put the text in messages so it can be translated
- Add final images
- Correctly inject information about about time muted

### DIFF
--- a/src/components/modal/mute/modal.jsx
+++ b/src/components/modal/mute/modal.jsx
@@ -1,0 +1,115 @@
+const bindAll = require('lodash.bindall');
+const PropTypes = require('prop-types');
+const React = require('react');
+const Modal = require('../base/modal.jsx');
+const ModalInnerContent = require('../base/modal-inner-content.jsx');
+const Button = require('../../forms/button.jsx');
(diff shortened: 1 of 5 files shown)


picklesrus:
Here are some screenshots with placeholder text and images.
 
image
image

chrisgarrity (approved):
This looks fine for the first iteration. When we connect this to the actual errors there are a few things we'll need to consider:
* Does the project page do the translation, and then pass the strings down, or does it just pass a type: `uncontructive` `disrespectful` and the modal will handle what to show including translation? Probably the latter.
* check that things are still fine in responsive layouts
* whether it will be easier to use intl.formatMessage on the header attribute, or just move the header section into the `children`
* when adding the number of minutes left consider using https://github.com/formatjs/formatjs/blob/main/website/docs/react-intl/components.md#formattedrelativetime

Issue #4459, opened 2020-09-25
Inform users when they are muted on a project page


chrisgarrity:
Using the same component/styling as #4396, let a user know that they cannot comment (hide the comment input).
![Screen Shot 2020-09-25 at 11 21 44 AM](https://user-images.githubusercontent.com/399209/94285150-54353c00-ff21-11ea-80a3-d9abe171d984.png)

Copy is not final, elements are:
* if the user was just muted and we know the type of offense, use a specific message related to the offense. 
* amount of time remaining in the mute
* link to educational modal (#4447) to get more information

Details:
* amount of time is set when page is loaded (or the mute happens), it does not count down

chrisgarrity:
@picklesrus I think this one is done. We can open a new issue if there are changes to the copy etc.

Issue #4396, opened 2020-09-02
Prompt to confirm email in comments section of the project page


paulkaplan:
Similar to the "logged out" prompt (https://github.com/LLK/scratch-www/issues/4395) we should prompt users who have not confirmed their email that they can comment if they first confirm their email. Clicking should bring up the modal with advice about confirming their email. 

This will need slightly different content depending on whether the project page is _your project_ or _someone else's project_
<img width="836" alt="Screen Shot 2020-09-02 at 3 19 43 PM" src="https://user-images.githubusercontent.com/654102/92026904-bdb19880-ed2f-11ea-9a4e-9d4935dd4704.png">
<img width="790" alt="Screen Shot 2020-09-02 at 3 20 01 PM" src="https://user-images.githubusercontent.com/654102/92026920-c7d39700-ed2f-11ea-952f-3b1674b69d4d.png">

Issue #4395, opened 2020-09-02
Prompt sign in or join in comment section of project page


paulkaplan:
Logged out users should be prompted to sign in or join to comment in the comment section of the project page. This will depend on the introduction of a "sign in modal", also needed in https://github.com/LLK/scratch-www/issues/4394, so this issue may depend on that issue. 

<img width="895" alt="Screen Shot 2020-09-02 at 3 17 09 PM" src="https://user-images.githubusercontent.com/654102/92026706-627fa600-ed2f-11ea-8851-c5406c95c863.png">

PR #4495, opened 2020-10-08
Commenting status component.

picklesrus:
Add a component to use to communicating a users's commenting status. 

This should be usable for the following:
https://github.com/LLK/scratch-www/issues/4459
https://github.com/LLK/scratch-www/issues/4396
https://github.com/LLK/scratch-www/issues/4395

Here's what it looks like when I started using it for #4459 : 
image


paulkaplan (approved):
LGTM

Issue #4508, opened 2020-10-13
Update Address on DMCA Page


dietbacon:
On the page https://scratch.mit.edu/DMCA we need to update the address for Scratch's office

OLD TEXT:

Copyright Agent / Mitchel Resnick
MIT Media Laboratory
77 Massachusetts Ave
Room E14-445A
Cambridge, MA 02139
Tel: (617) 253-9783

NEW TEXT

Copyright Agent / Mitchel Resnick
Scratch Foundation
201 South St. Suite 1-102
Boston, MA 02111
Tel: 857-233-5422

Purple-Hacker:
Do you need help with this? I can make a PR

PR #4524, opened 2020-10-15
Changed address on DMCA page

Purple-Hacker:
### Resolves: #4508 
Address used to be incorrect and I have updated it to the correct address.

### Changes:
Changed the address on the DMCA page so it is now correct.

Changed from:

Copyright Agent / Mitchel Resnick
MIT Media Laboratory
77 Massachusetts Ave
Room E14-445A
Cambridge, MA 02139
Tel: (617) 253-9783

And corrected it to:

Copyright Agent / Mitchel Resnick
Scratch Foundation
201 South St. Suite 1-102
Boston, MA 02111
Tel: 857-233-5422

### DIFF
--- a/src/views/dmca/dmca.jsx
+++ b/src/views/dmca/dmca.jsx
@@ -13,12 +13,11 @@ const Dmca = () => (
 
             <p><FormattedMessage id="dmca.intro" /></p>
             <p>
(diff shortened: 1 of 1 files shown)


Accio1 (changes requested):
It needs to be changed both times on the page. You have only changed it for the first time that it appears on the page.

PR #4526, opened 2020-10-16
Updated addresses on DMCA

Purple-Hacker:
### Resolves: #4508 

Addresses on this page were not updated. 

### Changes:
I have updated both of the addresses so they are correct.


benjiwheeler:
Thanks, @Purple-Hacker!

PR #4540, opened 2020-10-19
Move Navbar integration tests from Tap to Jest

BryceLTaylor:
### Resolves:

Continues to move all integration tests from Tap to Jest.

### Changes:

Moves the navbar link tests from Tap to Jest.  The Sign In and Create Account functionality is tested in the Sign In and Out and Join tests respectively.  This also adds tests to check the text of the navbar links.


test/integration/navbar.test.js
```diff
@@ -0,0 +1,106 @@
...
+        await expect(joinText).toEqual('Join Scratch');
+
+        let signIn = await findByXpath('//li[@class="link right login-item"]/a');
+        let signInText = await signIn.getText();
+        await expect(signInText).toEqual('Sign in');
+    });
+
+    test('create when signed out', async () => {
+        await clickXpath('//li[@class="link create"]');
+        let infoArea = await findByXpath('//div[@class="sprite-info_sprite-info_3EyZh box_box_2jjDp"]');
```
rschamp, line 59: This might break a lot as GUI changes. Instead, maybe we can be super specific about the XPath to the GUI component to exclude the project page view and pass on the editor mode view.

test/integration/navbar.test.js
```diff
@@ -0,0 +1,106 @@
...
+
+    test('About link when signed out', async () => {
+        await clickXpath('//li[@class="link about"]');
+        let aboutPage = await findByXpath('//div[@class="inner about"]');
+        let aboutPageVisible = await aboutPage.isDisplayed();
+        await expect(aboutPageVisible).toBe(true);
+    });
+
+    test('Search Bar', async () => {
+        let searchBar = await findByXpath('//input[@id="frc-q-1088"]');
```
rschamp, line 86: This id is also likely to break at some point, so we should add an explicit class for this.

rschamp (changes requested):

test/integration/navbar.test.js
rschamp: Actually it looks like we already differentiate between editor mode and player mode GUIs by giving editor mode GUI the class `gui` while player mode gui gets a class called `guiPlayer`.  So we could just look for those classes to determine what the page is displaying.

test/integration/navbar.test.js
rschamp: Here's the place to add the className prop: https://github.com/LLK/scratch-www/blob/9e4036029493a44fd45da783968bb74570a66f0a/src/components/navigation/www/navigation.jsx#L136

test/integration/navbar.test.js
BryceLTaylor: Okay.  I updated the test to look for the gui class for this test.  I think changing the other tests to look for the guiPlayer should happen in another pr.

test/integration/navbar.test.js
BryceLTaylor: I added the class search-wrapper to the outer input component.  It was not straightforward to add it to the actual input since it involved interacting with the formsy component and this strategy is working.

test/integration/navbar.test.js
```diff
@@ -0,0 +1,106 @@
...
+        await clickXpath('//li[@class="link about"]');
+        let aboutPage = await findByXpath('//div[@class="inner about"]');
+        let aboutPageVisible = await aboutPage.isDisplayed();
+        await expect(aboutPageVisible).toBe(true);
+    });
+
+    test('Search Bar', async () => {
+        let searchBar = await findByXpath('//div[contains(@class, "search-wrapper")]/div/input');
+        await searchBar.sendKeys('cat');
+        await driver.sleep(500); // without it sends an empty string on submit
```
rschamp, line 88: I'm wary of this kind of thing when we could wait on some state change instead. But probably fine for now.

rschamp (approved):

PR #4545, opened 2020-10-20
Move Project Rows tests from Tap to Jest

BryceLTaylor:
### Resolves:

Moving tests from Tap tests to Jest tests.

### Changes:

Moves the Project Rows tests into Jest.  Also re-wrote some of the xpaths to find the same elements.  The tests that check the links now find elements on the page to confirm its the right page rather than using the url.


test/integration/project-rows.test.js
```diff
@@ -0,0 +1,59 @@
...
+    test('Featured Projects row title', async () => {
+        let projects = await findByXpath('//div[@class="box"]/div[@class="box-header"]/h4');
+        let projectsText = await projects.getText();
+        await expect(projectsText).toEqual('Featured Projects');
+    });
+
+    test('Featured Project link', async () => {
+        await clickXpath('//div[@class="box"][descendant::text()="Featured Projects"]' +
+            '//div[contains(@class, "thumbnail")][1]/a[@class="thumbnail-image"]');
+        let gfOverlay = await findByXpath('//div[@class="stage-wrapper_stage-wrapper_2bejr box_box_2jjDp"]');
```
rschamp, line 41: When we talked in person I believe we decided to detect via the GUI class name instead of the calculated hash class name, which could change.

rschamp (changes requested):

test/integration/project-rows.test.js
BryceLTaylor: I addressed this by searching for the div with id=guiPlayer instead of the green flag overlay.  I also changed the name of the file to make it slightly more accurate.

I updated the version of chromedriver and increased the timeouts for some tests which made them pass locally.

test/integration/homepage-rows.test.js
```diff
@@ -0,0 +1,59 @@
...
+} else {
+    jest.setTimeout(20000);
+}
+
+let driver;
+
+describe('www-integration project rows', () => {
+    beforeAll(async () => {
+        driver = await buildDriver('www-integration project rows');
+        // driver.get(rootUrl);
```
rschamp, line 23: Cruft?

test/integration/homepage-rows.test.js
```diff
@@ -0,0 +1,59 @@
...
+    });
+
+    beforeEach(async () => {
+        await driver.get(rootUrl);
+    });
+
+    afterAll(async () => await driver.quit());
+
+    test('Featured Projects row title', async () => {
+        let projects = await findByXpath('//div[@class="box"]/div[@class="box-header"]/h4');
```
rschamp, line 33: As a follow-up, I think it would be justifiable to add classes to find these, since this is fragile.

rschamp (approved):
Looks good for now (moving from tap to jest), but I think we could follow-up with some improvements. Some of these selectors are pretty convoluted, so probably easy to break. Maybe in a follow-up we can add classes to the elements so they're easier to find? Also, there's some mixed concerns here between making sure that the rows link to something and expecting the linked page to be a certain way. E.g., the featured studio test makes sure that studio pages have descriptions. I think we more want to test that there's a link to a studio page on the homepage, rather than anything about studio pages?

But I don't think those things are the point of this PR, which is just to get these over to tap, so I think it's fine to merge!

PR #4548, opened 2020-10-21
updated gui to latest release

benjiwheeler:
Dependabot seems to not be producing the PR to update the gui version to the latest release. This PR does that manually.


benjiwheeler:
merged because Travis is taking forever today!

PR #4580, opened 2020-10-30
A few small upates to the muting modal

picklesrus:
- Add images and update styles to match.  
- Add timeMuted prop so it can display how long the user is muted for. 
- Text updates

This still needs localization.

This is a step toward #4447


picklesrus:
Since it isn't hooked up to anything yet, here are some screenshots: 
image

picklesrus:
image

src/components/modal/mute/modal.scss
```diff
@@ -31,7 +32,7 @@
         padding-top: 32px;
     }
     .bottom-img {
-        width: 100%;
+        width: 380px;
```
chrisgarrity, line 35: Is there a way not to have an absolute pixel value here - it will may cause issues with responsive/small screens.

chrisgarrity (changes requested):
Just one question, and if it really does need to be an absolute pixel value you can go ahead and merge. Everything else looks fine. 😀

src/components/modal/mute/modal.scss
picklesrus: I was hoping I could keep it at 100%, but, the way the html is currently structured, it ends up being too small. 
I haven't done a responsiveness pass on this modal at all yet yet. What do you think doing that in a follow up pass once it is easier to test on staging?

chrisgarrity: Yes, let's do this as a follow up!

chrisgarrity (approved):
I agree with handling responsiveness as a second pass. I'll open an issue to make sure we don't forget

PR #4616, opened 2020-11-09
Skeleton of code to show mute modal and comment status.

picklesrus:
This is more progress toward #4447

Still todo:
- l10n
- Time needs formatting
- Compose box needs to be shown/formatted correctly based on mute status
- Blue comment status box need to be sticky


chrisgarrity (approved):
As we discussed, modal images are handled in a different PR (broken in this one). Otherwise things look good to land.

PR #4617, opened 2020-11-09
Annual Report - Break out supporters list into separate JSON

kchadha:
Follow up work for the annual report. Separate the list of supporters from the jsx file and break it out into a json file. Dynamically lay out these list of strings into the two column div.

The supporters list section should look exactly the same as before, e.g:

image


seotts:
lgtm!

PR #4637, opened 2020-11-13
Convert mute timestamp to always show in minutes.

picklesrus:
This is minor step towards correct time formatting. More coming later!

Work towards #4447


chrisgarrity (approved):
For a minute there I was going to point out the edge case where `convertToMinutesFromNow` could end up showing two different times in the modal. But then I remembered this is just temporary until we get localization :)

PR #4642, opened 2020-11-16
Circleci project setup

BryceLTaylor:
### Resolves:

Makes progress to moving from Travis to CirlceCI for our CI.

### Changes:

Adds a circleCI configuration that will build and deploy for develop, hotfix, and release branches.  Will not deploy production.


BryceLTaylor:
The build job is failing, most likely because it's running out of memory.  We need to either use less memory or get more of it before we can make the switch.

BryceLTaylor:
Memory issues resolved.  Also switched to using `npm install` instead of `npm ci` because ci was causing the build to fail when dependencies used `BundledDependencies`.  Later we can look into switching back to npm ci.

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+  build-test-deploy:
+    jobs:
+      - build-staging:
+          context:
+            - scratch-www-all
+            - scratch-www-staging
+          filters:
+            branches:
+              only:
+                - circleci-project-setup # remove this before releasing
```
rschamp, line 107: We should do this before merging.

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+                - develop
+                - /^hotfix\/.*/
+                - /^release\/.*/
```
rschamp, line 110: We can check this syntax by pushing this branch to a hotfix/ and/or release/ branch name.

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+      - deploy-staging:
+          context:
+            - scratch-www-all
+            - scratch-www-staging
+          requires:
+            - build-staging
+          filters:
+            branches:
+              only:
+                - circleci-project-setup # remove this before releasing
```
rschamp, line 120: We should do this before merging.

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+                - develop
+                - /^hotfix\/.*/ # make sure this syntax works.  may need a regex
+                - /^release\/.*/
```
rschamp, line 123: Check syntax as above

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+            - scratch-www-staging
+          requires:
+            - deploy-staging
+          filters:
+            branches:
+              only:
+                - circleci-project-setup # remove this before releasing
+                - develop
+                - /^hotfix\/.*/
+                - /^release\/.*/\
```
rschamp, line 136: Check syntax

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+      keys:
+        - v1-build-{{ .Revision }}
+        - v1-build-
+  - &build
+    <<: *defaults
+    resource_class: medium+
+    steps:
+      - *restore_git_cache
+      - checkout
+      # - *restore_npm_cache
```
rschamp, line 48: Now that we're using install I think this becomes useful again.

rschamp (changes requested):
Just some cleanup needed before merging. Nice work!

.circleci/config.yml
BryceLTaylor: Tested this out, both hotfix and release branches run the whole workflow

.circleci/config.yml
BryceLTaylor: Tested this out, both hotfix and release branches run the whole workflow

.circleci/config.yml
BryceLTaylor: Tested this out, both hotfix and release branches run the whole workflow

.circleci/config.yml
BryceLTaylor: done

.circleci/config.yml
BryceLTaylor: done

.circleci/config.yml
BryceLTaylor: removed

rschamp (approved):

PR #4644, opened 2020-11-16
remove version number from micro:bit firmware URL

cwillisf:
### Changes:

The firmware download URL on the micro:bit page no longer contains a version number. The download bucket has been configured to treat this new URL as a redirect to the actual file, including the version number. This way we'll be able to deploy firmware updates without needing to change any content in this repository.

### Test Coverage:

Tested that the new URL redirects correctly. Right now it should redirect to `scratch-microbit-1.1.0.hex.zip`, the same file that was explicitly listed before this change.


ericrosenbaum (approved):
LGTM!

PR #4682, opened 2020-11-23
Add loading spinner in Annual Report

seotts:
Resolves:
Previously, when you clicked "watch video" the thumbnail disappeared while the video iframe was loading, leading to a period of time where it felt broken. 

Changes:
This adds a loading spinner, which shows until the video begins to play.

Test Coverage:
We should test that the videos in the Annual Report now have a loading spinner. We should also test other videos on www to make sure they work the same as before.


src/components/video-preview/video-preview.jsx
```diff
@@ -3,41 +3,60 @@ const PropTypes = require('prop-types');
...
+                                width={this.props.videoWidth}
+                                onLoad={this.handleVideoLoaded}
+                            />
+                        </div>
                     ) : (
                         <div
                             className="video-thumbnail"
                             onClick={this.handleShowVideo}
                         >
+                            <img
```
kchadha, line 56: We should include a comment for why we include this image

src/components/video/video.jsx
```diff
@@ -1,28 +1,69 @@
...
+                    // allowFullScreen is legacy, can we start using allow='fullscreen'?
+                    // allow="fullscreen"
+                    className={classNames('wistia_embed', `wistia_async_${this.props.videoId}`, loadedClass)}
+                    frameBorder="0" // deprecated attribute
+                    height={this.props.height}
+                    scrolling="no" // deprecated attribute
+                    src={`https://fast.wistia.net/embed/iframe/${this.props.videoId}?seo=false&videoFoam=true`}
+                    title={this.props.title}
+                    width={this.props.width}
+                    // onLoad={this.props.onLoad}
```
kchadha, line 61: Cruft?

src/components/video/video.jsx
```diff
@@ -1,28 +1,69 @@
...
+            wistiaScript.src = 'https://fast.wistia.com/assets/external/E-v1.js';
+            wistiaScript.async = false;
+            document.body.appendChild(wistiaScript);
+        }
+
+        window._wq = window._wq || [];
+        window._wq.push({
+            id: this.props.videoId,
+            onReady: video => {
+                video.bind('secondchange', s => {
```
kchadha, line 35: Would be good to have a comment here to describe what we're doing.

Also it looks like this would be firing all the time every second the video is playing.

kchadha (approved):
LGTM!

PR #4709, opened 2020-12-01
Update the SeC page wording

paulkaplan:
Replace the phrase "historically underrepresented in computing” with “ historically excluded from computing.”

Change requested from KO.

/cc @chrisgarrity I asked about the l10n update process about this on slack, but my understanding is that this will clear out existing translations, which I think is appropriate for this situation, since it is a meaningful change and not just a typo.


chrisgarrity (approved):
Looks good. Once this lands on the develop branch, it will be updated automatically on transifex by the daily cron job.

PR #4721, opened 2020-12-03
Show the comment status box on project page loads if the user is muted.

picklesrus:
Don't show the compose box if they're muted, but do show it, but greyed 
out, if they just posted a comment that resulted in a mute.

Still TODO: handle reply link. 

More progress toward https://github.com/LLK/scratch-www/issues/4459

### Test Coverage:
The unittests I wrote for the last pass of this work were already in place. I just had to change the values of what is shown and not to reflect the updated behavior.

### DIFF
--- a/src/views/preview/comment/compose-comment.jsx
+++ b/src/views/preview/comment/compose-comment.jsx
@@ -48,7 +48,7 @@ class ComposeComment extends React.Component {
             error: null,
             appealId: null,
             muteOpen: false,
-            muteExpiresAt: this.props.muteStatus.muteExpiresAt
+            muteExpiresAt: this.props.muteStatus.muteExpiresAt * 1000 // convert to ms
         };
     }
     handleInput (event) {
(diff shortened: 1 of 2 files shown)


src/views/preview/comment/compose-comment.jsx
```diff
@@ -46,7 +46,8 @@ class ComposeComment extends React.Component {
             status: ComposeStatus.EDITING,
             error: null,
             appealId: null,
-            muteOpen: false
+            muteOpen: false,
+            muteExpiresAt: this.props.muteStatus.muteExpiresAt
```
picklesrus, line 50: Thinking about dependencies on backend code a bit more, I think this depends on the session changes. For more resiliency, perhaps I should change this line to this: 
`this.props.muteStatus ? this.props.muteStatus.muteExpiresAt : 0`

or I could change the mapStateToProps to check whether muteStatus exists and then assign a default object. e.g. 
```
mute_status {
  muteExpiresAt: 0
  offenses: []
}
```

Thoughts?

chrisgarrity: I think I like the second option better, but it would be fine to leave for a later PR if we're going to need updates based on upcoming backend changes.

chrisgarrity (approved):
I slightly prefer having a default mute_status object instead of only handling the muteExpires case. Overall, looks good. 👍

src/views/preview/comment/compose-comment.jsx
picklesrus: Done!

PR #4736, opened 2020-12-07
World map fixes - Border thickness, Eswatini, remove unused css

seotts:
### Changes:
* Decreases the thickness of country borders on the map so small countries and islands are more visible (see Hawaii and Indonesia for examples)
* Show data for the country Eswatini. This required changing the way we display country name in the map so that the display name can be different than the country name Plotly uses, as Plotly only knows Eswatini by its old name (Swaziland)
* Remove untested CSS for RTL

### To test:
* Compare the country border thickness to production. It should be thinner. You should be able to see more of Hawaii and Indonesia.
* Eswatini (east of South Africa) should have data and should be called Eswatini.

### DIFF
--- a/src/components/world-map/world-map.jsx
+++ b/src/components/world-map/world-map.jsx
@@ -20,9 +20,7 @@ const WorldMap = props => (
                 locations: props.countryNames,
                 z: props.colorIndex,
                 text: props.countryData,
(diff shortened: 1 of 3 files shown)


kchadha (approved):
LGTM

PR #4751, opened 2020-12-11
Fix isMuted so it doesn't do an extra unit conversion.

picklesrus:
Rename state's muteExpiresAt to muteExpiresAtMs so units are explicit.

### DIFF
--- a/src/views/preview/comment/compose-comment.jsx
+++ b/src/views/preview/comment/compose-comment.jsx
@@ -48,7 +48,7 @@ class ComposeComment extends React.Component {
             error: null,
             appealId: null,
             muteOpen: false,
-            muteExpiresAt: this.props.muteStatus.muteExpiresAt * 1000 // convert to ms
+            muteExpiresAtMs: this.props.muteStatus.muteExpiresAt * 1000 // convert to ms
         };
     }
     handleInput (event) {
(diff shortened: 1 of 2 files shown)


src/views/preview/comment/compose-comment.jsx
```diff
@@ -118,7 +118,7 @@ class ComposeComment extends React.Component {
     }
 
     isMuted () {
-        return this.state.muteExpiresAt * 1000 > Date.now();
```
picklesrus, line 121: muteExpiresAt in the state object is already being saved in ms in the constructor and in the comment's post response so I'm pretty sure this one is a bug.

chrisgarrity (commented):
It would be good to figure out a test. It seems as though there are two things we're counting on, redis returns expiration in seconds - that's the value in mute_status. Date wants everything in milliseconds.  Could we set up a test to ensure that `state.muteExpiresAtMs == props.mute_status.muteExpiresAt * 1000`.

It would need a second mockStore that set up the mute_status value.

picklesrus:
Good idea. Test added!

chrisgarrity (approved):
Looks good!

picklesrus:
Tests failed after I rebased because of where now mounted (vs shallow mounted) component was looking for the `store`.  Fixed it up in a separate commit. If that passes, I'll merge this.

PR #4752, opened 2020-12-11
Localize mute

### LINKED ISSUE #4691: Internationalize the mute messaging modal
Refactor the mute messaging modal to allow for localization of the messages.

This should be done for the first set of messaging, before we have multiple types of mute messages.

chrisgarrity:
### Resolves:

resolves #4691 

### Changes:
Wraps the MuteModal in the Intl HOC and moves all strings into l10n files.

The strings about muting are in the general `l10n.json` because they will be used in both the commenting status and mute modal components. `next`, `back`, and `close` were already in the general strings, so just reused them.

Strings needed for the compose-comment component are only on the project page (preview view), so they are in the view's `l10n.json` file.

### Test Coverage:

Updated the mute modal tests to handle being wrapped in the intl HOC. For the tests that actually mount the component, it meant finding and checking the `MuteModal` component within the wrapper instead of testing the wrapper component.

### DIFF
--- a/src/components/modal/mute/modal.jsx
+++ b/src/components/modal/mute/modal.jsx
@@ -1,6 +1,9 @@
 const bindAll = require('lodash.bindall');
 const PropTypes = require('prop-types');
 const React = require('react');
+const FormattedMessage = require('react-intl').FormattedMessage;
+const injectIntl = require('react-intl').injectIntl;
+const intlShape = require('react-intl').intlShape;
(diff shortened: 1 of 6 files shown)


picklesrus:
Looks good, but I think the comment-compose test needs to be updated similarly to the mute modal one.

chrisgarrity:
@picklesrus whoops, I totally forgot to check the compose-comment test.

PR #4753, opened 2020-12-11
Fix ?mode=trending

### LINKED ISSUE #3696: Searching nothing and choosing "Trending" causes you to search for "?mode=trending"
If you search for an empty string and then choose to sort by trending it shows your search string as "?mode=trending" instead of an empty string and shows results as if you have searched for that string.  It should not show any text in the search bar.

### Steps to Reproduce

On the home page click the magnifying glass in the search bar.
Note: it takes you to the search page with no results
Change "Popular" to "Trending"
Note: the search bar changes to show the string "?mode=trending" and items appear on the search page.  It appears to have searched for that string.

Mac Chrome

mxmou:
### Resolves:

Resolves #3696 

### Changes:

Clicking the search icon without searching anything and setting mode to "Trending" now works correctly instead of searching for `?mode=trending`.

I don't know if the extra `canReturnEmpty` parameter is necessary or not because I don't understand what the second call to `stripQueryValue()` does.

### Test Coverage:

Manually tested.
image

### DIFF
--- a/src/views/search/search.jsx
+++ b/src/views/search/search.jsx
@@ -62,17 +62,17 @@ class Search extends React.Component {
         const query = decodeURIComponent(window.location.search.split('+').join('%20'));
         let term = query;
 
-        const stripQueryValue = function (queryTerm) {
+        const stripQueryValue = function (queryTerm, canReturnEmpty) {
             const queryIndex = query.indexOf('q=');
             if (queryIndex !== -1) {
                 queryTerm = query.substring(queryIndex + 2, query.length).toLowerCase();
             }
-            return queryTerm;
+            return canReturnEmpty ? '' : queryTerm;
         };
         // Strip off the initial "?q="
-        term = stripQueryValue(term);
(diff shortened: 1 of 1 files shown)


src/views/search/search.jsx
```diff
@@ -62,17 +62,17 @@ class Search extends React.Component {
...
             const queryIndex = query.indexOf('q=');
             if (queryIndex !== -1) {
                 queryTerm = query.substring(queryIndex + 2, query.length).toLowerCase();
             }
-            return queryTerm;
+            return canReturnEmpty ? '' : queryTerm;
         };
         // Strip off the initial "?q="
-        term = stripQueryValue(term);
+        term = stripQueryValue(term, true);
```
benjiwheeler, line 73: I think this will delete the search term, no matter what it is.

benjiwheeler:
@mxmou I think this area of the code needs a bit more attention -- the addition of unit tests, for example. Thanks for this PR -- I'm going to take my own crack at it, based on your code.

PR #4764, opened 2020-12-15
Add a timeout to stop displaying the mute status when a mute expires.

### LINKED ISSUE #4740: Commenting status box should disappear at expiration time
If a user is sitting on a project page when their mute expires, the status box should disappear.

This should be pretty straightforward to do by setting up a setTimeout that fires at the mute expiration time.

picklesrus:
Resolves #4740


seotts:
This is looking good! I noticed that it doesn't set state.status back to EDITING when the mute times out; the effect of this is that the mute message goes away but the comment box is still disabled.

picklesrus:
This should be ready for another look.  I changed the setState line to include resetting the compose status, message, and error and it seems to be working now.

I had some merge conflicts that took a few tries to fix properly (silly syntax & lint errors), but I *think* travis is going to pass this time.

seotts:
lgtm!

PR #4774, opened 2020-12-17
Show warning message when user has been muted several times.

### LINKED ISSUE #4494: Moderation warning communication
When the number of mutable offenses by a user crosses a particular threshold (determined by the back end), show a warning modal in addition to the commenting blocked box.

See #4447 for current design link. Since it's based on the same multi-step modal, it's blocked until we finalize the imagery and copy.

picklesrus:
Adds this step to the modal when backend says it is appropriate
image

Resolves #4494


src/views/preview/comment/compose-comment.jsx
```diff
@@ -80,20 +81,23 @@ class ComposeComment extends React.Component {
                 let muteOpen = false;
                 let muteExpiresAtMs = 0;
                 let rejectedStatus = ComposeStatus.REJECTED;
+                let showWarning = false;
                 if (body.status && body.status.mute_status) {
                     muteExpiresAtMs = body.status.mute_status.muteExpiresAt * 1000; // convert to ms
                     rejectedStatus = ComposeStatus.REJECTED_MUTE;
                     if (this.shouldShowMuteModal(body.status.mute_status.offenses)) {
```
chrisgarrity, line 88: In general we're not showing the full modal if you get muted multiple times within a 7-day period. However, if you get the warning we should probably show the modal again.
```suggestion
                    if (body.status.mute_status.showWarning || this.shouldShowMuteModal(body.status.mute_status.offenses)) {
```

chrisgarrity (changes requested):
One minor change so that we show the modal when showWarning is true. 

Also, while we didn't pop-up the modal for every offense after the first one, I think that we **should** pop it up every time there's an offense with the score over 10. That's probably something to run by the team as I don't think we specified that precisely, but I think every time is what we should propose.

src/views/preview/comment/compose-comment.jsx
picklesrus: Sounds good. I'll build this into "shouldShowMuteModal" so all the logic about whether to show it lives (and is tested) through that.

picklesrus:
Made the change a slightly different way that you suggested, but it should accomplish the same thing. Let me know what you think. If you'd prefer to keep this logic out of shouldShowMuteModal, we can!

chrisgarrity (approved):
Looks good - it was definitely better to keep all the logic in shouldShowMuteModal.

PR #4828, opened 2021-01-07
Add back new automod messaging

seotts:
This adds back the reverted change with additional automod messaging. It fixes the issue that resulted in an error about 'currentMessageType.'

To test when variable muting is not enable, check that you can comment regular comments and that you get a regular orange error when you try to comment vulgarity.

When variable muting is enabled, test that different types of inappropriate comments result in different messaging.


picklesrus (approved):
Looks good to me!

The infinite post problem is gone :)

PR #4832, opened 2021-01-08
Remove use of FormattedHTMLMessage in Annual Report

seotts:
Removes all usage of FormattedHTMLMessage in Annual Report and replaces them with FormattedMessage. All of the uses were in the Supporters section and had no actual HTML in the strings. There should be no visible changes to that or any section.

**In bughunt...**
* Test that the Supporters section text is the same in production and staging when language is English
* Test that production and staging are the same amount translated when language is not English (Italian, French, and Korean have a lot of translations)


kchadha (approved):
LGTM

Issue #4869, opened 2021-01-19
Translation id is shown instead of text


chrisgarrity:
`comment.vulgar.content2` is shown instead of the correct text in the modal that is shown when the user posted bad words.

chrisgarrity:
Fixed in #4880

PR #4880, opened 2021-01-20
Show correct mute messaging instead of always the general message

### LINKED ISSUE #4868: First message on inappropriate comments is always the "generic" message
We have specific messages for several kinds of bad behavior, and one fallback for all the others. On projects you get the generic message when you post the bad message. If you re-open the modal on subsequent projects while muted you see the more specific message.

seotts:
Fixes #4868 and #4869

Fixes the issue where the "general" mute message is shown before refresh. Also fixes the translation id issue in the vulgarity message. 

When you get muted for vulgarity, unconstructive language, or PII, you should see the correct messaging.


picklesrus (approved):
LGTM

PR #4895, opened 2021-01-25
Add @formatjs/relativetimeformat-polyfill

chrisgarrity:
Safari < 14 does not support Intl.RelativeTimeFormat which is used in the moderation messaging. Adding this polyfill to support it.

### Testing
make sure to test on a variety of browsers and versions. Specific ones to test:
- [ ] Safari 13.x
- [ ] Safari 14.x
- [ ] Chrome
- [ ] Firefox
- [ ] Edge

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,9 @@
     "scratch-storage": "^0.5.1"
   },
   "devDependencies": {
+    "@formatjs/intl-locale": "^2.4.13",
+    "@formatjs/intl-pluralrules": "^4.0.6",
+    "@formatjs/intl-relativetimeformat": "^8.0.3",
     "ajv": "6.4.0",
     "async": "3.1.0",
     "autoprefixer": "6.3.6",
@@ -91,6 +94,7 @@
     "glob": "5.0.15",
     "google-libphonenumber": "3.2.6",
     "html-webpack-plugin": "2.22.0",
+    "i": "^0.3.6",
     "iso-3166-2": "0.4.0",
     "jest": "^23.6.0",
     "json-loader": "0.5.2",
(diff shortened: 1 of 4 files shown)


picklesrus:
I have no experience with pollyfilling , so if this is the usual way to do things, definitely happy to defer to that.  I have a couple questions though, mostly trying to sort out what gets loaded when

Should the require happen in the format time library instead? It would be nice not to have to know to do this from each spot  we use that library.

Also, I don't really understand what the implications of guarding the require with an if are.  Aren't we shipping the whole bundle of javascript around so every user would have to download all of it anyway. Or do we serve different bundles of javascript to different browsers? What about languages? e.g. Do we serve a different bundle for users with "en" vs  "fr" as their langauge?

apple502j:
We usually add polyfills via `/static/js/polyfill.min.js` - and I'm sure we ship everything (including translations of all languages or polyfills for IE8).

picklesrus (approved):

package.json
```diff
@@ -108,6 +112,7 @@
     "minilog": "2.0.8",
     "node-dir": "0.1.16",
     "node-sass": "4.14.1",
+    "npm": "^6.14.11",
```
rschamp, line 115: Was this intentional?

package.json
```diff
@@ -91,6 +94,7 @@
     "glob": "5.0.15",
     "google-libphonenumber": "3.2.6",
     "html-webpack-plugin": "2.22.0",
+    "i": "^0.3.6",
```
rschamp, line 97: Was this intentional? It looks l10n related but not related to this polyfill?

Issue #4747, opened 2020-12-10
Moderation Feedback modal


chrisgarrity:
Provide a way for users to submit feedback on automatic moderation. 
- [x] Add "Think this was a mistake? Let us know." as a new paragraph after the community guidelines link paragraph. This link only appears when the user just got muted
- [x] Add feedback modal with a single textbox. Text box cannot be empty and has a maximum length of 500 characters.
Header: Think this was a mistake?
Body: Sometimes the filter catches things by accident. Your feedback will help prevent this from happening in future.
- [x] On submit replace the content of the modal with:
Header: Thanks for letting us know!
Body: Your feedback will help us to improve our filtering system.
- [x] save feedback, include:
  - [x] content of the text box
  - [x] text of the original comment
  - [x] userid of the user submitting feedback
  - [x] language of the user
  - [x] type of message the user was shown (e.g., unconstructive, pii etc.)

<img width="520" alt="Screen Shot 2021-01-14 at 3 33 41 PM" src="https://user-images.githubusercontent.com/399209/104649679-15b7be80-5683-11eb-8b82-74b67f66f61b.png">

[invision link](https://invis.io/62YSG3GVF5X#/432618254_1-12_--_B)

PR #4901, opened 2021-01-26
Automod feedback modal steps

seotts:
This addresses the first 3 steps of #4747:

- Add "Think this was a mistake? Let us know." as a new paragraph after the community guidelines link paragraph. This link only appears when the user just got muted
- Add feedback modal with a single textbox. Text box cannot be empty and has a maximum length of 500 characters.
Header: Think this was a mistake?
Body: Sometimes the filter catches things by accident. Your feedback will help prevent this from happening in future.
- On submit replace the content of the modal with:
Header: Thanks for letting us know!
Body: Your feedback will help us to improve our filtering system.

Right now, it does nothing with the input other than outputting it to the console. (Can also remove the console log if that would be weird to merge!)

Something seems off about the diff being shown here. I think it is just showing the diff from the last commit, but this PR should include all of the changes.


picklesrus (commented):
Just about to do some testing, but code generally looks good with a couple minor comments.  Also, one of the tests looks like it might have broken during a refactor

picklesrus:
I noticed on my version that the button messages show up as messages Ids. The rest of the messages are filled in though so maybe these ids aren't quite right?
image

Also, is the image for the feedback step new?  It is showing up as missing for me:
image

seotts:
I believe I fixed the things you noticed, including the tests, the image, and the localization ids.

picklesrus (approved):

Issue #4870, opened 2021-01-19
Disable the comment buttons when the user is muted and cannot comment


chrisgarrity:
The Post and Cancel buttons are still clickable when the user us muted. They should be disabled.

picklesrus:
Fixed by #4919

PR #4918, opened 2021-01-28
Disable the comment buttons when the user is muted

picklesrus:
The buttons already had a disabled style applied, but weren't actually disabled so you could still click them. 

### Resolves:

#4870


picklesrus:
Whoops. pushed wrong branch. Will replace with another PR in a moment

PR #4931, opened 2021-01-29
[Master] Hotfix SEC Text Update

kchadha:
Release the hotfix for the SEC page updates.


WorldLanguages:
Note from a native Spanish speaker: the correct word to use here no matter who you ask is "Latina", because "people" is a feminine noun (la gente, las personas). The sentence in Spanish would also end up using feminine gendered versions of "Black" and "Indigenous" because of this same reason. "Latinx" isn't used a lot outside of the United States and can't be properly read by most screen readers.

Issue #4886, opened 2021-01-22
Keep mute time remaining at 1 min when less than a minute to go.


chrisgarrity:
If you update the page when you're muted, but there is less than a minute to go it says "You will be able to comment again in 0 minutes", this is misleading.

It would be better to just keep it at 1 minute until it expires.

lovecodeabc:
Maybe like “You can comment again in ___ seconds”

picklesrus:
Also done in scratchr2 so closing. https://github.com/LLK/scratchr2/pull/5961

PR #4938, opened 2021-02-01
Change time library to default to 1 minute when the time remaining is less than 30 seconds.

picklesrus:
### Resolves:
#4886 

### Changes:

Round to the nearest minute, except for when there is less than 1 minute. In that case pin to 1.

### Test Coverage:

Unit test added for 15sec remaining.

### DIFF
--- a/src/lib/format-time.js
+++ b/src/lib/format-time.js
@@ -1,3 +1,6 @@
+// IMPORTANT: any changes to the time algorithm also need to be made in the corresponding
+// scratchr2 file 'lib/format-time.js'
+
 /**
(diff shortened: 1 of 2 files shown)


chrisgarrity (approved):
Looks good - there's one thing that I would probably change. This demonstrates the fragility of the system as the same code change has to be made here and in scratchr2. In scratchr2 we refer to this file as the original, it would be good to add a comment to the top of this file that mentions the scratchr2 file. e.g.
```
// IMPORTANT: any changes to the time algorithm also need to be made in the corresponding scratchr2 file 'lib/format-time.js'

Issue #4741, opened 2020-12-08
When a user is muted, the "Reply" button on comments should open the modal


picklesrus:
When the reply button is clicked, the modal should open to the card/screen that tells you how long you're muted for.

PR #4943, opened 2021-02-02
Open mute modal from reply button if you're muted.

picklesrus:
### Resolves:

https://github.com/LLK/scratch-www/issues/4741

### Changes:
There are a few cases this tries to handle:

- User is already muted and clicks to reply to a comment:  The mute modal should open to the page that tells them how long they are muted for. The blue box with comment info should no longer show under the reply since it should already be at the top of the list of comments. 
- User replies to a comment with content that gets them muted: The user should get the mute modal opened on the first page and show the blue box since it won't be shown at the top of the list of comments.

### Test Coverage:

There are unittests for this new logic

### DIFF
--- a/src/views/preview/comment/compose-comment.jsx
+++ b/src/views/preview/comment/compose-comment.jsx
@@ -50,7 +50,7 @@ class ComposeComment extends React.Component {
             status: ComposeStatus.EDITING,
             error: null,
             appealId: null,
-            muteOpen: false,
+            muteOpen: muteExpiresAtMs > Date.now() && this.props.isReply,
             muteExpiresAtMs: muteExpiresAtMs,
             muteType: this.props.muteStatus.currentMessageType,
             showWarning: this.props.muteStatus.showWarning ? this.props.muteStatus.showWarning : false
@@ -232,7 +232,7 @@ class ComposeComment extends React.Component {
     render () {
         return (
             <React.Fragment>
-                {this.isMuted() ? (
+                {(this.isMuted() && !this.props.isReply) ? (
                     <FlexRow className="comment">
                         <CommentingStatus>
                             <p><FormattedMessage id={this.getMuteMessageInfo().commentType} /></p>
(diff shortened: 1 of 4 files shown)


chrisgarrity:
@picklesrus I'm still reviewing, but I have found one issue. If you post a comment that gets you muted, you can still open the reply input and post a comment. It still comes back that you're muted, but it seems like it would be better to open the modal.

picklesrus:
> @picklesrus I'm still reviewing, but I have found one issue. If you post a comment that gets you muted, you can still open the reply input and post a comment. It still comes back that you're muted, but it seems like it would be better to open the modal.

Yeah, I noticed that too -- I think that is the case on the current page too so I figured this was a step forward. I'm not sure how to solve that problem at the moment. The compose-comment component only knows at construction time if the user is muted based on the session request. I think to fix this we'd have to  refreshing the session or put something in redux so all components have access to it?

picklesrus:
Added a fix for the reply button issue we found on Friday.

src/views/preview/comment/compose-comment.jsx
```diff
@@ -147,6 +147,13 @@ class ComposeComment extends React.Component {
         this.setState({
             muteOpen: false
         });
+
+        // Cancel (i.e. complete) the reply action if the user clicked on the reply button while
+        // alreay muted. This "closes" the reply.  If they just got muted, we want to leave it open
+        // so the blue CommentingStatus box shows.
+        if (this.props.isReply && this.state.status !== ComposeStatus.REJECTED_MUTE) {
```
chrisgarrity, line 154: One really subtle thing that is NOT worth fixing. When we show the modal, we still add some padding that would have surrounded the input, and then remove it when we close the modal. Unless you're looking at what's behind the modal, you'd never notice.

chrisgarrity (approved):
I'm coming around to thinking that compose comment should be using redux so that we can catch things on the page where they're muted. However, I think this is generally better than always opening the blue box for replies.

PR #4980, opened 2021-02-11
Remove unintentional dependency additions.

picklesrus:
They were added in #4895


apple502j:
Still unsure where other dependencies are used, seems like they never remove dependencies:
- `compression`
- `cheerio` - used in now-removed Preview FAQ. We don't do server-side rendering.
- `create-react-class`
- `exenv`
- `json2po-stream` - leftover from `.po` file age?
- `lodash.clone` 
- `lodash.isarray`
- `lodash.range`
- `node-dir`
- `po2icu` - leftover from `.po` file age?
- `watch`

picklesrus:
Travis won't allow me to re-run the tests so I pushed my changes to an upstream branch (where I can rerun the build) instead. Closing this in favor of that PR https://github.com/LLK/scratch-www/pull/4981.

picklesrus:
Thanks @apple502j! At the moment I'm just removing the ones that were added recently, but we do need to do a full audit of our deps at some point.

PR #4981, opened 2021-02-11
Remove unused deps

picklesrus:
They were added in #4895 unintentionally


picklesrus:
This is failing tests in the same way last week's deploy did. That is under separate investigation so I'm going to wait to see what the outcome is before I spend more time digging in.

apple502j:
That's jsdom/jsdom#2040

picklesrus:
Closing in favor of a change bundled with some travis updates that also fix the build.

PR #5013, opened 2021-02-17
Update to react 16.14

paulkaplan:
This will update www to use react and react-dom 16.14, the highest version before 17

Why not 17?
- Scratch GUI and Paint require peer dependency of 16, pending resolution of breaking issues on 17 (https://github.com/LLK/scratch-gui/issues/6721#issuecomment-780384044)

~~This PR is blocked on https://github.com/LLK/scratch-gui/pull/6754, which ensures that any package using scratch-gui (like this one) is able to provide its own version of React, and not have GUI bundle its own. That is important because using different version of react can cause hard to find bugs.~~ That was merged.

The addition of the resolve: {symlinks: false} is a feature of all our webpack configs (gui/desktop/paint) that expect to use `npm link` for dependencies that have dependencies. This is needed now that scratch-gui expects to be able to `require('react')` without having a copy of it. This is only needed to get local development with npm link working.

TODO
- [x] Mark deprecated lifecycle methods "unsafe" in order to avoid avalanche of console warnings in dev mode

---

In terms of testing, I've been looking at this updated version of react on GUI and Paint and don't see anything immediate, but it will be good to have a fair bit of time devoted next week to going through the project page, gui and paint editor. 

---

Note: there will be console warnings about deprecated lifecycle methods, even though I've moved over all of the components in www to use UNSAFE_. That is because some 3rd party packages use React and still use those. They will need to be updated. Notably react-redux, react-responsive, formsy-react-components and formsy-react. That can be done next, to reduce the testing burden all at once. Just a heads up that in developer mode the console will get a fair number of those warnings.


package.json
```diff
@@ -117,8 +117,8 @@
     "postcss-loader": "2.0.10",
     "prop-types": "15.6.0",
     "query-string": "^5.1.1",
-    "react": "16.2.0",
-    "react-dom": "16.2.1",
+    "react": "^16.14.0",
```
kchadha, line 120: Do we need the `^` since the next version is 17 and it's already out?

paulkaplan: I think its common to keep updating older versions with security patches, etc

kchadha (approved):
LGTM

Issue #4913, opened 2021-01-27
1.4 Download page should say Mac OS version 10.4-10.14


benjiwheeler:
1.4 Download page should say Mac OS version "10.4 through 10.14"

Currently it says "10.4 or later"

RoBlockHead:
working on it!

BryceLTaylor:
This information appears in multiple places on the page.  I believe this issue is specifically referring to here:
![Screen Shot 2021-01-28 at 3 36 21 PM](https://user-images.githubusercontent.com/1088291/106195767-9c3dc700-617e-11eb-9855-214e43a3d0c2.png)

JoelGritter:
@RoBlockHead still working on it? I could also make a PR for this issue.

RoBlockHead:
> @RoBlockHead still working on it? I could also make a PR for this issue.

Unfortunately not, had some stuff interrupt my work. It's all yours.

PR #5017, opened 2021-02-18
Fix gh-4913: MacOS version string update

JoelGritter:
### Resolves: 

#4913 

### Changes: 

l10n string changed from "Compatible with Mac OSX 10.4 or later" to "Compatible with Mac OSX 10.4 through 10.14".

I also checked to make sure that there weren't any other instances of "Compatible with Mac OSX 10.4 or later" left in the repo, seems that there was only this one instance.

(Though notably, the "through 10.14" substring was already present in the same l10n file, on line 25: https://github.com/LLK/scratch-www/blob/develop/src/views/scratch_1.4/l10n.json#L25 )

### Test Coverage:

No tests were conducted since this is just a simple l10n string change.


benjiwheeler:
Thanks, @JoelGritter!

PR #5019, opened 2021-02-18
Circle ci configure tests

BryceLTaylor:
### Changes:

Gets tests to run and store artifacts/test results in a way that circleCI can track.  Also turns off the deploy and integration jobs for staging by commenting them out.

Splits the the npm run test step into its individual pieces that run in different steps in the build job so that it is easier to see which part fails.

Changes the docker image used by all jobs to a newer version that comes with python3.

Updates to use python3 and uses newer versions of pip and s3cmd for the deploy job. 

Splits the integration tests into a TAP job and a JEST job so they can run in parallel.


.circleci/config.yml
```diff
@@ -45,57 +44,101 @@ aliases:
...
+            npm run test:unit:jest
+            npm run test:unit:tap:ci
+            npm run test:unit:convertReportToXunit
       - *save_npm_cache
       - *save_git_cache
       - *save_build_cache
+      - store_test_results:
+          path: test/results
+      - store_artifacts:
+          path: test/results
```
rschamp, line 72: We want this to be the build directory. And since we will have uploaded the build at this point, we won't need `store_artifacts` in any other job.

.circleci/config.yml
```diff
@@ -45,57 +44,101 @@ aliases:
...
       - run:
-          name: "integration tests"
+          name: "integration tests with Jest"
           command: |
-            npm run test:integration:remote
+            npm run test:integration:jest:remote
+      - store_test_results:
+          path: test/results
+      - store_artifacts:
+          path: test/results
```
rschamp, line 103: Can remove `store_artifacts` here.

.circleci/config.yml
```diff
@@ -45,57 +44,101 @@ aliases:
...
+      - run:
+          name: "integration tests with Tap"
+          command: |
+            mkdir ./test/results
+            npm run test:smoke:sauce
+            npm run test:smoke:convertReportToXunit
+      - store_test_results:
+          path: test/results
+      - store_artifacts:
+          path: test/results
```
rschamp, line 119: Remove `store_artifacts`

package.json
```diff
@@ -6,16 +6,20 @@
...
     "test": "npm run test:lint && npm run build && npm run test:unit",
     "test:lint": "eslint . --ext .js,.jsx,.json",
+    "test:lint:ci": "eslint . --ext .js,.jsx,.json --format junit -o ./test/results/lint-results.xml",
     "test:integration": "npm run test:integration:jest && npm run test:smoke",
     "test:integration:jest": "jest ./test/integration/*.test.js",
     "test:integration:remote": "npm run test:integration:jest:remote && npm run test:smoke:sauce",
-    "test:integration:jest:remote": "SMOKE_REMOTE=true jest ./test/integration/*.test.js",
-    "test:smoke": "tap ./test/integration-legacy/smoke-testing/*.js --timeout=3600 --no-coverage -R classic",
+    "test:integration:jest:remote": "SMOKE_REMOTE=true JEST_JUNIT_OUTPUT_NAME=integration-jest-results.xml jest ./test/integration/*.test.js",
+    "test:smoke": "tap ./test/integration-legacy/smoke-testing/*.js --timeout=3600 --no-coverage -R classic --output-file ./test/results/integration-raw-tap.tap",
```
rschamp, line 14: Will these changes make it harder for us to run these things locally? If we can configure this in the CI config, I think that would be better, and leave the configuration here as we would want it for local testing.

rschamp (changes requested):

package.json
BryceLTaylor: They shouldn't.  They both output normally to the console.  You may need to create an /integration directory in ./test.  I have tested it locally, and both of these tests output useful to the terminal.

I will have to make a separate PR to update the version of Chromedriver, which I needed to do to run locally, but that is independent of the changes in this PR.

.circleci/config.yml
BryceLTaylor: I removed these lines

.circleci/config.yml
BryceLTaylor: I removed these lines

.circleci/config.yml
BryceLTaylor: I removed these lines

package.json
BryceLTaylor: I've updated the readme to include information about needing a ./test/results directory.

.circleci/config.yml
```diff
@@ -68,8 +68,6 @@ aliases:
       - *save_build_cache
       - store_test_results:
           path: test/results
-      - store_artifacts:
-          path: test/results
```
rschamp: I think we still want this here, we just want the path to be stored as `build`, so we could download and inspect what was built/deployed.

package.json
rschamp: If `test/results` will get created when you run tests, then we should probably ignore it in `.gitignore` with this PR. Otherwise it will start showing up in people's working copies and then it's just a matter of time before it gets committed, which we don't want.

rschamp (changes requested):

.circleci/config.yml
rschamp: And to be clear, I just mean this one — the other two should stay removed like you did.

package.json
BryceLTaylor: It's already in there

.circleci/config.yml
BryceLTaylor: Okay, updated.  It now stores the build folder.

BryceLTaylor:
I'm not sure why it lost my comment, but I already put the test/results directory in the gitignore.

rschamp (approved):

PR #5036, opened 2021-02-22
Update to Webpack 4

paulkaplan:
_Note: this is a draft, I have a bit of work left to get this updated to avoid merge conflicts and do a bit more testing_

This PR updates the build system to use Webpack 4. Staying more closely up-to-date will allow us to more easily share webpack configs between our repos (GUI, Paint and Desktop use webpack4) and take advantage of new features and performance improvements. My primary goal here was to match the behavior of the webpack 2 config. 

---
## Update webpack package and related packages
I chose to match the version used by scratch-gui as a starting point. This includes html-webpack-plugin, copy-webpack-plugin. Also had to include webpack-cli since that was split from webpack between 2 and 4. 

## Replace CommonsChunkPlugin
In webpack 2 we used the `CommonsChunkPlugin`, together with an entry called `common` to include common libraries and an `init.js` file on every page, as well as remove those chunks from the page-specific bundle.

In webpack 4 that is replaced by the more general purpose `optimization.splitChunks` section, which allows you very fine-grain control over how to the code to share between entrypoints. A couple tricky bits:
- I removed the `common` entrypoint, since it is not needed when using splitChunks, and will cause problems.
- I changed the entrypoint for every page to explicitly include the init.js file, i.e. `entry: [init.js, [route].js]` instead of just `entry: [route].js`. This is beacuse the `init.js` file is not imported anywhere, and needs to be added as a dependency to each entrypoint to recover the "include this script on every page"-behavior we had with webpack2
- I added the `optimization.splitChunks` section to recover the same behavior as CommonsChunk, similar to [this config](https://webpack.js.org/plugins/split-chunks-plugin/#split-chunks-example-1) but making sure to set `minChunks` so that only modules shared by _all_ entrypoints are extracted.

## Replace UglifyJSPlugin
Webpack 4 automatically minifies when the `mode` is set to production, using the `TerserPlugin`. We do not need to manually specify uglify, and terser seems to have slightly better performance. There is no need to specify sourceMaps on the plugin, since that is taken care of by the `devtool` webpack config line.

## Add the Bundle Visualizer as an option

The `WebpackBundleAnalyzer` package was extremely helpful in reconstructing and matching the chunk splitting behavior. Instead of every time removing and re-adding it, I just kept it in under a process variable, so you can run `ANALYZE_BUNDLE=true npm run build` to get the visualizer open, it looks like this
Screen Shot 2021-02-19 at 12 50 41 PM

## Testing

I tested a few things to make sure I was on the right track. I used the webpack bundle analyzer to ensure that the same packages are being included in the `common.bundle.js` as before (even though i removed `react`, `react-intl`, etc from the common chunk, since those packages are common to all pages and therefore get extracted correctly without any additional work). I also checked the resulting bundle sizes, and they were within 1kb or so, which seems reasonable for using a totally different JS compressor. 

----

Other notes:
- I didn't upgrade the `HTMLWebpackPlugin` all the way to the most recent version, but noticed that the next major version up includes the ability to use any loader to load and execute the html template, so we could conceivably use `jsx` and actually have the html template for each page be a pre-rendered version of each page (using `ReactDOMServer.renderToString` [docs](https://reactjs.org/docs/react-dom-server.html), and use `ReactDOM.hydrate` [docs](https://reactjs.org/docs/react-dom.html#hydrate) on the client to get the interactive bits. There would be more to figure out (like changing the way we include css, etc.) but could be a way to get rendered pages to people faster.
- We'll need to upgrade babel, probably soon, as the version we are using may not play great with webpack4. Upgrading babel is next on my list.


paulkaplan:
@rschamp My remaining question is about how the `common` entry explicitly required some vendor libraries like react, react-dom, redux. I removed those and made the only explicit dependency the `init.js` file because it didn't seem to matter to specify the 3rd party libs, they are common to every page so get extracted into common with or without being explicitly listed. I saw some mention of avoiding some blank screen problems related to caching in the commit message, but I didn't understand how forcing those libraries into common would impact that.

rschamp (commented):
Looks great! I think the comment about the WSOD re:vendor libraries is outdated. We hit that issue a couple of times since that commit, so I doubt your changes will affect anything w/r/t. I think it might be good to put this up on a hotfix branch so we can see if we hit the WSOD issue on staging before merging it though.

rschamp:
Put this as changes requested pending the package-lock conflict and doing at least one "pre-staging" test before we merge (since updating dependencies back and forth can be a pain).

paulkaplan:
@rschamp Tested on staging, everything seems to be working ok. I'm going to merge and put a note into the testing doc about things to look out for.

PR #5058, opened 2021-02-26
Don't require a big phone number validation library on every page

paulkaplan:
Paired on this with @picklesrus 

### Resolves:

This fixes an issue (unfiled) where the `libphonenumber` dependency was being loaded on every page, even though only one of our pages actually uses phone number validation (teacher registration). 

### Changes:

There are two parts: 
1. da2ca40 moves the `isPhone` formsy validation out of the common validations registered from `forms/form.jsx` out into the `forms/phone-input.jsx` component itself so that the `isPhone` validation is only registered when you require the `PhoneInput` component. This allows forms that do not use that validation (e.g. login, language switcher) to not load the 500k+ libphonenumber dependency. This change removed that library from the `common.bundle.js` and made it only get required from `teacherregistration.html` but also `teacher waiting room` and `student registration`.
2. 0353135 splits the `PhoneNumberStep` out of `registration/steps.jsx` into its own file. Previously any view that used any of the steps from `registration/steps` would load the phone-input validation. Now you have to load the `PhoneStep` separately, confining its dependencies to that specific view. 

### Tests

The splitting of the bundles was tested by using the `webpack-bundle-analyzer`. In combination with the libphonenumber package update, we should manually verify that the phone input in teacher registration is still working, and that the other step-wise registration flows are still working as well (since those files are touched as well). /cc @BryceLTaylor

### DIFF
--- a/src/components/forms/phone-input.jsx
+++ b/src/components/forms/phone-input.jsx
@@ -17,6 +17,21 @@ const validationHOCFactory = require('./validations.jsx').validationHOCFactory;
 require('./row.scss');
 require('./phone-input.scss');
 
+const Formsy = require('formsy-react');
+const libphonenumber = require('google-libphonenumber');
+const phoneNumberUtil = libphonenumber.PhoneNumberUtil.getInstance();
+
+Formsy.addValidationRule('isPhone', (values, value) => {
+    if (typeof value === 'undefined') return true;
+    if (value && value.national_number === '+') return true;
+    try {
+        const parsed = phoneNumberUtil.parse(value.national_number, value.country_code.iso2);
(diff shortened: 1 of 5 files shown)


paulkaplan:
Woops @picklesrus sorry a few lint/test issues I'll revisit this in a minute

paulkaplan:
Ok @picklesrus fixed, just missed one shared component when splitting the steps

paulkaplan:
Also I ran the production build before/after this change, it reduces the size of the `common.bundle.js` from 1224kb to 684kb, a 540kb (45%) savings! Proportionally the same after gzipping, 260kb -> 150kb.

picklesrus (approved):
LGTM!

PR #5075, opened 2021-03-01
Create Plotly bundle to load only choropleth

seotts:
This reduced the size of the library that we are loading for WorldMap, by creating a Plotly bundle that only supports the choropleth visualization type.

To test, we should verify that this decreases the size of Plotly in www. We should also check that the world map continues to show up on the annual report.


paulkaplan (approved):
LGTM. The expected savings is almost 4mb uncompressed.

PR #5077, opened 2021-03-02
Bump chromedriver version to 88, increase jest integration timeouts

BryceLTaylor:
### Resolves:

Allows integration tests to pass locally while running chrome version 88

### Changes:

Updates chromedriver to 88.
Increases the timeouts for certain jest tests from 10 seconds to 20 seconds when running locally.  The new version of chromedriver was causing these tests to timeout, but increasing the timeout allowed them to pass.


benjiwheeler:
There was a package-lock conflict, so I moved the package-lock and package.json changes to a separate commit, and rebased on develop. Unfortunately I couldn't get `npm ci` to run (it kept stalling forever on one particular step), so I just did `npm install`.

PR #5100, opened 2021-03-08
Extract comments reducer from preview reducer

paulkaplan:
Right now all comment state lives within the project reducer. However, a similar state tree will be used in the future for studios, user profiles, etc. The first step is extracting comment data from within the project reducer, and promote it to its own slice. I moved all the relevant data, action creators into `redux/comments.js`. I moved the thunks (the API calls specialized for project comments) to their own file, `redux/project-comment-actions.js`. I also split off the test file (leaving it in place using tap, hopefully the diff will be easier to read that way!) and updated the preview `mapStateToProps` and `mapDispatchToProps` to get data/actions from this new file. 

Additionally, I added a test to cover the setFetchStatus action, and changed the packages.json to make sure the tests in unit-legacy were not being run, fixed this in 8b2bb5f

Testing
- Unit tests are now being run, and added an additional one.
- Manual testing as a group should cover the following comment actions: posting a comment/reply, deleting a reply, reporting a comment/reply, loading multiple pages of comments/replies, restoring a comment/reply (admin)


paulkaplan:
@kchadha we had talked briefly about how to make the action creators for project comments "generalize" to studio comments. In thinking about the approach on the backend, I think the right answer here is to be a bit repetitive and move all the "thunk" action creators that use the API into a separate file, and plan to make a similar file for studio comment actions. This will help us not need to fight ourselves with an incorrect abstraction, and make changes to different comment type actions as needed. The data aspect of the reducer, how comments are managed in the client side state, is all in the "common" comment reducer. Like ray mentioned in https://github.com/LLK/scratch-www/pull/5149, if the way the actual reducer works diverges between studio and project comments, we can just split them up further. 

Other possibilities I considered were:
- A single set of comment actions, have the consumer pass in the comment type and "object id", do everything inline. This is the tightest possible coupling, and means the backend responses have to be identical. 
- A single set of comment actions, but the consumer passes in an "api" object to call, which would let us not duplicate the "get replies for top level comments as soon as they are loaded" logic, for example. I think this could work, but is introducing an abstraction before we have many examples, which I want to avoid.
- Two separate sets of reducers/actions - The actual reducer functionality is battle tested in production, so it would be nice to maintain. Furthermore, the UIs for project/studio comments should be the same, so there shouldn't be any new operations one can take on studio comments that would impact the reducer (an example would be something like re-ordering, which would need to be supported at the reducer level).

src/views/preview/preview.jsx
```diff
@@ -5,6 +5,7 @@ const Page = require('../../components/page/www/page.jsx');
 const render = require('../../lib/render.jsx');
 
 const previewActions = require('../../redux/preview.js');
+const commentsActions = require('../../redux/comments.js');
```
paulkaplan, line 8: Change to `commentActions`

src/views/preview/project-view.jsx
```diff
@@ -29,6 +29,8 @@ const Meta = require('./meta.jsx');
 const sessionActions = require('../../redux/session.js');
 const navigationActions = require('../../redux/navigation.js');
 const previewActions = require('../../redux/preview.js');
+const commentsActions = require('../../redux/comments.js');
```
paulkaplan, line 32: Remove this and move `resetComments` and `addComment` to the projectCommentActions file. This will simplify how the consumer of a type of comment has to interact with the actions

package.json
```diff
@@ -15,8 +15,8 @@
     "test:smoke:sauce": "SMOKE_REMOTE=true tap ./test/integration-legacy/smoke-testing/*.js --timeout=60000 --no-coverage -R classic",
     "test:unit": "npm run test:unit:jest && npm run test:unit:tap",
     "test:unit:jest": "jest ./test/unit/ && jest ./test/localization/*.test.js",
-    "test:unit:tap": "tap ./test/{unit-legacy,localization-legacy}/*.js --no-coverage -R classic",
-    "test:coverage": "tap ./test/{unit-legacy,localization-legacy}/*.js --coverage --coverage-report=lcov",
+    "test:unit:tap": "tap ./test/{unit-legacy,localization-legacy} --no-coverage -R classic",
```
paulkaplan, line 18: There is a merge conflict here, the change is still needed though

paulkaplan: Done d02ddef

src/views/preview/project-view.jsx
paulkaplan: Done 8a33d1c /cc @kchadha

src/views/preview/preview.jsx
paulkaplan: Done 8a33d1c

kchadha (approved):
LGTM, thanks for making those changes!

PR #5101, opened 2021-03-08
Removed unused dependencies

picklesrus:
These were found with @kchadha and @BryceLTaylor and depcheck. 

Note: eslint-plugin-cypress was used by the cypress tests, but we weren't using the tests so we removed them along with the dependency.


kchadha (approved):
LGTM :shipit:

PR #5117, opened 2021-03-09
Add dev route with very bare studio playground

paulkaplan:
🙌 Studios work is starting  🙌

It'll be a while until they are ready for primetime, but this adds an initial skeleton of the page, including a new routing system `react-router` to handle tying together the tab navigation and the URL routing. 

It is also the first appearance of react hooks in our codebase :) using the custom hooks from the react-router package to allow easy retrieval of route parameters anywhere in the component tree.

Finally, this PR updates the webpack config to publish routes in the `routes-dev.json` folder (previously just the defunct component library which I'd like to resurrect, now the `/studios-playground/...` route as well) 

/cc @rschamp want to check that it is ok to do that webpack config change for doing dev routes on staging. Seems like it should be ok, just wanted to check. 

By the way the reason I started with the URL routing / tab nav is because I want to default to lazy loading all the non-active tab feeds, just to get an early handle on data requests. 

image


src/routes-dev.json
```diff
@@ -4,5 +4,12 @@
...
         "title": "Components"
+    },
+    {
+        "name": "studio",
+        "pattern": "^/studios-playground/\\d+(/projects|/curators|/activity|/comments)?/?(\\?.*)?$",
+        "routeAlias": "/studios-playground/?$",
+        "view": "studio/studio",
+        "title": "Studio Playground"
     }
 ]
```
kchadha, line 15: Missing newline at EOF

paulkaplan: Fixed, thanks

paulkaplan:
@kchadha I removed the part of this that would push the dev routes to staging, pending conversation based on rays feedback offline. We can add it back once we talk about it a bit more.

src/views/studio/studio-tab-nav.jsx
```diff
@@ -0,0 +1,41 @@
...
+                exact
+            >
+                Projects
+            </NavLink>
+            &nbsp;|&nbsp;
+            <NavLink
+                activeStyle={{textDecoration: 'underline'}}
+                to={`${match.url}/curators`}
+            >
+                Curators
```
kchadha, line 21: Just noticed that the order of the tabs is different from the current studios page (not that it really matters for this skeleton)

kchadha (approved):
LGTM

PR #5121, opened 2021-03-10
Made changes to conf 2020 page (fixed dependencies)

zoebentley:
Update the Conference page to look like this doc:

https://docs.google.com/document/d/1-veTXYDlckOry1gKH8RzTp4dPkNrqmThUC27ST1kE1A/edit?ts=60419837#heading=h.3qwdbj2wz8u9

### DIFF
--- a/src/views/conference/2020/index/index.jsx
+++ b/src/views/conference/2020/index/index.jsx
@@ -3,7 +3,6 @@ const FormattedMessage = require('react-intl').FormattedMessage;
 const React = require('react');
 const render = require('../../../../lib/render.jsx');
 
-const FlexRow = require('../../../../components/flex-row/flex-row.jsx');
 const Page = require('../../../../components/page/conference/2020/page.jsx');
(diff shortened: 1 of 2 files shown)


src/views/conference/2020/index/index.jsx
```diff
@@ -32,9 +32,6 @@ const ConferenceSplash = () => (
                     <strong><FormattedMessage id="conference-2020.desc1a" /></strong>
```
seotts, line 32: I notice in the google doc, this is on a new line, but isn't here. Not sure if that's intentional.

seotts (changes requested):
This looks good to me, aside from one question that I raised. Also, looks like tests are failing, but I have no idea why. I'll try to investigate-- it likely doesn't have anything to do with your changes.

seotts:
Ah, just found the problem with the tests-- since the section that used FlexRow got removed, we had to remove the import.

seotts (approved):
Looks good!

PR #5139, opened 2021-03-15
[Master] Hotfix - automod everyone

### LINKED ISSUE #5091: Prevent moderation feedback on comments that did not trigger the mod event
The intent of the moderation feedback form is to gather data when the user gets muted for a particular comment. If a user has multiple tabs or windows open it is possible to get muted in one window and then try to submit a comment in another window that already had the comment input field available. In this case the new comment is rejected because the user is already muted, not because of the comment.

When users then follow through and submit feedback on this comment, it's confusing to the moderators because the comment that's included with the feedback is often fine as it's not what triggered the mute. We should detect this situation (this comment did not trigger the mute), and not provide the feedback link. 

### Steps to Reproduce

See https://github.com/LLK/scratch-www/issues/5045

chrisgarrity:
Fixes #5091 
Prior to enabling auto-mod for everyone we need to make sure that the feedback system isn't sending confusing data.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21440,9 +21440,9 @@
       }
     },
     "scratch-l10n": {
-      "version": "3.11.20210311031457",
-      "resolved": "https://registry.npmjs.org/scratch-l10n/-/scratch-l10n-3.11.20210311031457.tgz",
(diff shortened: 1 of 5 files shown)


chrisgarrity:
Re-doing the branch to remove extra dependabot commits and avoid conflicts.

PR #5164, opened 2021-03-19
Fetch studio info and roles in redux

paulkaplan:
Create a reducer for the studio info and studio roles data. Nothing too interesting, changed the `useState` call that was previously there to use redux instead. The roles information is based on an upcoming API for getting the logged in users relationship to the studio. 

With the role information in redux, we can start making selectors to slice permissions information out of the store, like whether the current user can add/remove projects, edit the studio description, invite/promote curators, etc... That is next up.

### DIFF
--- a/src/views/studio/studio-info.jsx
+++ b/src/views/studio/studio-info.jsx
@@ -1,35 +1,53 @@
-import React, {useState, useEffect} from 'react';
+import React, {useEffect} from 'react';
+import PropTypes from 'prop-types';
 import {useParams} from 'react-router-dom';
-import {infoFetcher} from './lib/fetchers';
+import {connect} from 'react-redux';
 import Debug from './debug.jsx';
+import {Status as SessionStatus} from '../../redux/session';
+import {getInfo, getRoles} from '../../redux/studio';
 
-const StudioInfo = () => {
+const StudioInfo = ({username, studio, token, onLoadInfo, onLoadRoles}) => {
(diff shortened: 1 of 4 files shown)


src/views/studio/studio-info.jsx
```diff
@@ -1,35 +1,54 @@
...
-            .then(data => setState({loading: false, error: null, data}))
-            .catch(error => setState({loading: false, error, data: null}));
+    
+    useEffect(() => { // Load studio info after first render
+        if (studioId) onLoadInfo(studioId);
     }, [studioId]);
 
+    useEffect(() => { // Load roles info once the username is available
+        if (studioId && username && token) onLoadRoles(studioId, username, token);
+    }, [studioId, username, token]);
```
rschamp, line 17: Just curious — would it continue working fine if you only gave it `[studioId, username]`? And this causes it to rerender when you log in/out?

rschamp (approved):

src/views/studio/studio-info.jsx
paulkaplan: > would it continue working fine if you only gave it [studioId, username]?
Since username/token are updated together, yeah I think it would, but I thought that knowledge isn't really represented in the code so I should make the dependency explicit. 

> this causes it to rerender when you log in/out?
I think right now our logout redirects to the homepage, but if it was done without refresh it would cause onLoadRoles to be called again.

Issue #5045, opened 2021-02-24
Improve user moderation messaging when muted in another tab


chrisgarrity:
This depends on the user having multiple tabs or windows open. If you add a comment on project in a tab that was opened before getting muted. It looks like the current comment (which may be fine) is the one that resulted in the mute.

Ideally we would know that the user is currently muted, and that this comment was not the one that caused the mute. We should not imply that the current comment is the one that caused the problem. The messages should have a version that makes it clear that a previous comment was responsible for the mute.

We also don't want the user to provide feedback in this situation (https://github.com/LLK/scratch-www/issues/5091)

### Steps to Reproduce
1. Log in and open two tabs. In one have a project page, in the other it can be a project, studio or profile.
2. Add a comment that would get you muted in the second tab
3. Go back to the first tab, the comment input is still visible
4. Try to submit a good comment
5. You get the notification that you're muted (if you just get the blue status box, click 'learn more')
6. Navigate to the second page, there will be a 'let us know' link to provide feedback about mistakes.

chrisgarrity:
Fixed in #5166

PR #5166, opened 2021-03-19
When a user is already muted, show a different message in the blue box

### LINKED ISSUE #4988: Moderation Feedback is misleading when muted by a moderator
When a user is muted manually by a moderator, they can still submit moderation feedback if they think they were wrongly muted. 

There is no comment data submitted with the feedback because the mute was not in response to a specific comment. We should not show the moderation feedback form if muted by an admin.

The text on the feedback modal makes it sound like the filter caused the mute when it was a moderator action. Usually the moderator will be following up with a message about the situation that led to the mute action, but it might be better if it was clear that it was a moderator. This will be followed up in a separate issue.

### Steps to Reproduce

1. As a moderator, mute a user
2. As the muted user, visit a project page.
3. If you don't get the modal already opened, click the learn more link to open the modal and navigate to the page that includes the Think we made a mistake? 'Let us know' link.
4. Click the 'Let us know' link, and submit feedback.

seotts:
This fixes #4988 and #5045 by changing the messaging in the blue box to something somewhat more vague when you are already muted (rather than when you just got muted.)

To accomplish this, we added another ComposeState, COMPOSE_DISALLOWED to indicate when you are already muted. I modified some tests and added a couple to reflect this.

Things we should manually test:
- When you get muted, the text in the blue message refers to the comment you just made
- When you get muted and then navigate to a new project, the text in the blue message refers "one of your recent comments."
- These cases should also refer to "one of recent comments":
    - You got muted by an admin
    - You had a project tab open, got muted in another tab, and then returned to the project and attempted to make a comment
- Replies should work roughly the same as before
    - When you get muted by making a reply, you should see the modal and the blue box and the message in the blue box should refer to the comment you just made
    - When you get muted _and then_ try to make a reply (on the same page or a new page), you should just get the modal.
         - When you are on the same page, it lets you try to compose and submit, and then shows the the modal. When you are on a new page, it shows the modal immediately when you click "reply"

This PR does not change the modal messages. There are a couple of message types that refer to the comment you just made in the modal; the plan is to change that in a future PR.


picklesrus (commented):
Code looks good!  I have one question though.  In the testing notes it says:

> When you get muted and then try to make a reply (on the same page or a new page), you should just get the modal.

When I do this on the same page, I do get the modal when I try and post, but it gives me the chance to compose and post. I *think* this is expected and just wanted to double check.  It might be worth being explicit about that in testing notes since it is kind of confusing.  As long as this is expected, LGTM!

PR #5171, opened 2021-03-22
Fix components page

ericrosenbaum:
The components page is a dev route meant to display a library of reusable UI components. It currently has errors that prevent it from loading. This change fixes it just enough so that it loads.


paulkaplan (approved):
LGTM, thanks for getting this fixed so quick!

PR #5175, opened 2021-03-23
Show different modal message when already muted

seotts:
Makes it so we show slightly different messages on the modal for unconstructive and vulgarity if the user is already muted and goes to a new project page.

I noticed that content1.past and comment.type.past are identical to each other for unconstructive and vulgarity, so I switched to just using comment.type.past.


test/unit/components/compose-comment.test.jsx
```diff
@@ -555,20 +555,46 @@ describe('Compose Comment test', () => {
         global.Date.now = realDateNow;
     });
 
-    test('getMuteMessageInfo: muteType set', () => {
+    test('getMuteMessageInfo: muteType set and just got muted', () => {
         const commentInstance = getComposeCommentWrapper({}).instance();
         commentInstance.setState({muteType: 'unconstructive'});
-        expect(commentInstance.getMuteMessageInfo().commentType).toBe('comment.type.unconstructive');
+        expect(commentInstance.getMuteMessageInfo(true).commentType).toBe('comment.type.unconstructive');
```
chrisgarrity, line 561: Super minor, and not necessary to land. I just realized that it would be a little more self documenting if it declared a const for the boolean:
```suggestion
        const justMuted = true;
        expect(commentInstance.getMuteMessageInfo(justMuted).commentType).toBe('comment.type.unconstructive');
```

src/views/preview/comment/compose-comment.jsx
```diff
@@ -291,11 +292,7 @@ class ComposeComment extends React.Component {
                         <CommentingStatus>
                             <p>
                                 <FormattedMessage
-                                    id={
-                                        this.state.status === ComposeStatus.REJECTED_MUTE ?
-                                            this.getMuteMessageInfo().commentType :
-                                            this.getMuteMessageInfo().commentTypePast
-                                    }
+                                    id={this.getMuteMessageInfo(this.state.status === ComposeStatus.REJECTED_MUTE).commentType}
```
chrisgarrity, line 295: This line gets a lint error - too long

chrisgarrity (changes requested):
The lint error needs to be fixed so that the tests pass. The other comment is more of a suggestion, definitely not necessary.

PR #5176, opened 2021-03-23
Add initial selectors for studio permissions

paulkaplan:
This PR adds selectors to make it easier to test the permissions associated with studios. Initially i've just added `selectCanEditInfo` and `selectCanAddProjects`, but we'll add the other project related permissions and the curator management permissions in time. The test setup attempts to be pretty exhaustive, so when someone asks "who can add projects" we have a really simple thing to read off to confirm the behavior. 

This is a departure from the project page, where the permissions are "computed" in the mapStateToProps function. Separating out into pure "selector" functions that consume the top-level state makes it easier to test. I think in the future this pattern may be good to adopt as a way to clean up the very large `project-view.jsx` file.

This PR also correctly plucks the "owner" field from the studio info response. 

I also include the permissions in the relevant areas using the `Debug` component, this is what it looks like

image


src/redux/studio.js
```diff
@@ -111,10 +120,32 @@ const getRoles = (studioId, username, token) => (dispatch => {
...
 });
 
+// Selectors
+
+// Fine-grain selector helpers - not exported, use the higher level selectors below
+const isCreator = state =>
+    state.studio.owner !== null && // Never try matching if owner has not been set
+    selectUserId(state) === state.studio.owner;
+const isCurator = state => state.studio.curator;
+const isManager = state => state.studio.manager || isCreator(state);
```
apple502j, line 130: Logic seems to be flipped - curators aren't managers?

apple502j (changes requested):

src/redux/studio.js
paulkaplan: Hmmm maybe you misread, that reads "manager or creator", not curator?

apple502j: Why does isManager return whether someone is "manager or creator"? isManager should only return true for managers, otherwise code blow would break.

apple502j: Whoops, creators :P Might need to take a dyslexia test

src/redux/permissions.js
```diff
@@ -54,3 +54,7 @@ module.exports.setPermissionsError = error => ({
     type: Types.SET_PERMISSIONS_ERROR,
     error: error
 });
+
+// Selectors - Being extra cautious with stict truthiness
```
fsih, line 58: spelling strict

src/redux/studio.js
```diff
@@ -72,6 +78,8 @@ const setRoles = roles => ({
     roles: roles
 });
 
+// Thunks
```
fsih, line 81: What is a thunk?

src/redux/studio.js
```diff
@@ -111,10 +120,32 @@ const getRoles = (studioId, username, token) => (dispatch => {
     });
 });
 
+// Selectors
+
+// Fine-grain selector helpers - not exported, use the higher level selectors below
+const isCreator = state =>
```
fsih, line 126: Does it make sense to name these to start with select, or remove the "select" part of "selectIsAdmin, selectIsSocial" for naming consistency?

test/unit/redux/permissions.test.js
```diff
@@ -0,0 +1,31 @@
...
+        expect(selectIsAdmin(state)).toBe(true);
+    });
+
+    test('isSocial', () => {
+        let state = {
+            permissions: {}
+        };
+        const newPermissions = {social: true};
+        state.permissions = permissionsReducer(state.session, setPermissions(newPermissions));
+        expect(selectIsSocial(state)).toBe(true);
```
fsih, line 29: Maybe test that the other permission hasn't changed

test/unit/redux/studio.test.js
```diff
@@ -0,0 +1,98 @@
...
+            expect(selectCanEditInfo(state)).toBe(true);
+
+            state.studio = fixtures.studio.creator1;
+            state.session = fixtures.session.user1;
+            expect(selectCanEditInfo(state)).toBe(true);
+        });
+        test('is not editable by curators', () => {
+            state.studio = fixtures.studio.isCurator;
+            state.session = fixtures.session.user1;
+            expect(selectCanEditInfo(state)).toBe(false);
```
fsih, line 56: Maybe test having more than one of these permissions

fsih (commented):
Just some small suggestions

src/redux/studio.js
apple502j: iirc, asynchronous action (see redux-thunk)

src/redux/studio.js
paulkaplan: Yeah I didn't name them in the convention just because they are helpers and I wanted to avoid the temptation to export them.

PR #5217, opened 2021-03-31
Studio Comments Fetching & Initial UI

kchadha:
### Changes:

This PR starts some of the studio comments work, fetching the top level comments and displaying them on the comments tab of the studios playground.
- Create a studio comment actions creator using the one from project comment actions as a model.
- Hook up the comments reducer and actions to the studio comments tab
- Generalize the comment components (still currently located with the Project page code) so that they can also be used with studios. The main piece of work here was just to pull out the comment POST URI out as a component prop.
- Update project page to pass in the `postURI` prop instead of the `projectId` prop
- Add the UI for composing a comment and the list of comments to the studio comments tab
- Start to hook up some initial state controlled rendering (only displaying the comment compose UI and 'reply comment' links if a user is logged in)
- Fix a minor bug in the ComposeComment code where we were referencing an object property before it was defined

### Test Coverage:

Tested locally in the studios playground:
- Adding new comments and comment replies on the scratchr2 studios page, seeing them in the studios playground
- Adding new comments on the project page
- Adding new comments and replies in the studios playground

### DIFF
--- a/src/redux/studio-comment-actions.js
+++ b/src/redux/studio-comment-actions.js
@@ -1,0 +1,180 @@
+const eachLimit = require('async/eachLimit');
+
+const api = require('../lib/api');
+const log = require('../lib/log');
+
+const COMMENT_LIMIT = 20;
+
+const {
+    addNewComment,
+    resetComments,
+    Status,
+    setFetchStatus,
+    setCommentDeleted,
+    setCommentReported,
+    setCommentRestored,
+    setMoreCommentsToLoad,
+    setComments,
+    setError,
+    setReplies,
+    setRepliesDeleted,
+    setRepliesRestored
+} = require('../redux/comments.js');
+
+const getReplies = (studioId, commentIds, offset, isAdmin, token) => (dispatch => {
+    dispatch(setFetchStatus('replies', Status.FETCHING));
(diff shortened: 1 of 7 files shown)


paulkaplan:
@kchadha this looks great! I added a couple of commits that are mostly cleanup:
1. 1ad53ae adds back the postURI to TopLevelComment in studio-comments, I think you had it there while we were testing but maybe got removed on final commits?
2. I changed the anonymous function to instead use `useCallback`. The reason for that lint rule is to prevent the re-renders that would happen by always passing a new function reference to the child component. It only works if the child is set up to actually prevent rerenders if the props haven't changed (`shouldComponentUpdate` or the new `React.memo`), so it isn't actually useful here. But the linter prefers it and it is better especially when passing functions to possibly long chains of child components (which will happen when we start passing in callbacks for deleting/reporting/etc). 
3. I cleaned up the code a bit by renaming some props and listing them more explicitly.

paulkaplan (approved):
I made a few updates, things that occured to me after we paired. I left a comment detailing them, mostly minor, but you might want to review them (its a lot of commits but only for readability-sake).

kchadha:
Thanks @paulkaplan! The changes you made make sense to me.

PR #5218, opened 2021-04-01
Move the studio route from routes-dev.json to routes.json

BryceLTaylor:
### Resolves:

Allows the in progress www studios page to be viewable on staging and production.

### Changes:

Moves the studios route from routes-dev.json to routes.json

These routes are being blocked by fastly if you are not coming from certain specific ip addresses.


paulkaplan (approved):
🎉  lets make triple sure checking the visibility of this is on the bughunt/deploy checklist before release 😬

PR #5219, opened 2021-04-01
Include onAddComment for studio comments

paulkaplan:
@kchadha looks like we forgot to include the `onAddComment` prop for the studio comments (I think we must have missed a last commit, postURI was also missing but I caught that before the previous PR)


kchadha (approved):
LGTM thanks for catching that!

PR #5220, opened 2021-04-01
Add privacy policy and terms of use to allowlist

### LINKED ISSUE #3698: Banned users cannot see the Terms of Use or Privacy Policy
If you are banned, you cannot see Terms of Use or Privacy Policy, which may be a problem because violation of Terms of Use can ban users but then they can't review it.

apple502j:
### Resolves:
Resolves #3698 

### Changes:
Adds privacy policy and terms of use to allowlist. (Should the variable names be changed?)

### DIFF
--- a/src/redux/session.js
+++ b/src/redux/session.js
@@ -12,12 +12,9 @@ const Types = keyMirror({
 });
 
 const banWhitelistPaths = [
-    '/accounts/banned-response/',
-    '/community_guidelines/',
+    '/accounts/banned-response',
     '/community_guidelines',
-    '/privacy_policy/',
     '/privacy_policy',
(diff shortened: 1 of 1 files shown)


Naleksuh:
I'm not too familiar with React so pardon if this is a novice question but is it a good idea to define every path twice? Might it be better to change how the whitelist is being used?

Naleksuh:
Actually, I'm looking at the code and it seems that the path only needs to *contain* the string, not match it-- meaning the entries with trailing slashes are likely redundant/removable.

benjiwheeler:
Updated per @Naleksuh's comment. Thanks, @apple502j and @Naleksuh !

PR #5231, opened 2021-04-05
Use getState in thunks instead of passing props around

paulkaplan:
This is a proposal of sorts, to start using a capability we don't use anywhere else in our codebase.

### Problem
We often extract data from the redux store, give it to a component via props, not so that it can be rendered, but so it can be used as the arguments to a callback that calls dispatch ([see this for example](https://github.com/LLK/scratch-www/blob/develop/src/views/studio/studio-info.jsx#L47-L58)). This clutters our components with props that aren't used for rendering. It also isn't necessary: the state is available to redux thunks via the `getState` function. The redux-thunk readme shows how this works: https://github.com/reduxjs/redux-thunk#motivation

### Proposal
Use `getState` inside of thunks to retrieve the data we need (often username/token/permissions) instead of passing that data out of the store into components via mapStateToProps and back in through mapDispatchToProps. This clean up the components, removing data that isn't relevant to rendering, as well as reducing the need for `useCallback`/`memo`/`shouldComponentUpdate` because we no longer have to bind the dispatch functions together with state data in the component. 

https://github.com/LLK/scratch-www/commit/c6b0493e084d9721b74f6ebaa46feaed88af4291 Shows how much we can reduce repetition by using the data already present in the store (username/token/studio id) instead of passing it around explicitly through props.

https://github.com/LLK/scratch-www/commit/bd279fe5573b2e7956e835a98c23b5c546b1fdb1 Shows reducing the repetition _and_ removing the need for `useCallback` because the function is completely static, it doesn't need to be combined with other state data in order to be called.

/cc @rschamp what do you think about this? I think it has the potential to greatly reduce error-prone boilerplate. I did not fully remove all possible arguments from the files, I wanted to provide 2 simple examples before going whole hog

### DIFF
--- a/src/redux/comments.js
+++ b/src/redux/comments.js
@@ -186,3 +186,6 @@ module.exports.setMoreCommentsToLoad = moreCommentsToLoad => ({
 module.exports.resetComments = () => ({
     type: 'RESET_COMMENTS'
 });
+
+// Selectors
+module.exports.selectCommentCount = state => state.comments.comments.length;
--- a/src/redux/studio-comment-actions.js
+++ b/src/redux/studio-comment-actions.js
@@ -18,8 +18,18 @@ const {
     setError,
     setReplies,
     setRepliesDeleted,
-    setRepliesRestored
+    setRepliesRestored,
+    selectCommentCount
 } = require('../redux/comments.js');
+
+const {
+    selectIsAdmin,
+    selectToken
+} = require('./session');
+
+const {
+    selectStudioId
+} = require('./studio');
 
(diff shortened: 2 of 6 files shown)


paulkaplan:
Maybe @kchadha do you want to take a look at this for the code-review part, maybe could just get a 👍  or 👎  from @rschamp about the use of `getState` in thunks, curious if you've done that elsewhere/have thoughts.

rschamp (approved):
Looks better to me. I think initially we were trying to leave the reducers decoupled from each other, which may have been why we didn't use getState in the reducers as much? Or we just didn't think to use it 🤷

PR #5269, opened 2021-04-21
Move deploy and integration tests to circle

BryceLTaylor:
### Changes:

Adds the deploy-production and integration jobs to the circleci config.
Removes the deploy step from travis entirely.
Removes the smoke job from travis entirely.

### Test Coverage:

Includes running integration tests against production after the deploy

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,15 +123,15 @@ jobs:
   deploy-staging:
     <<: *deploy
   deploy-production:
-    # <<: *deploy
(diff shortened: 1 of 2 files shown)


rschamp (approved):
This looks reasonable to me.

If Circle will now be doing the bulk of the work, it would be a good time to re-enable it on the repo as a check, so we can make sure PRs pass with Circle consistently before removing Travis.  Sorry I forgot we hadn't done that yet.

PR #5274, opened 2021-04-21
Add client-side editing of studio info

paulkaplan:
This PR introduces a new reducer, studio-mutations, which includes data about client-side editing of studio information like following, title, description, etc. This PR also includes 3 new wireframed components for editing studio title, description and follow status. 
image

I found it made more sense to keep the loading/error states for mutating data in a separate place from where we are initially fetching and storing that data. The main studio reducer listens for `COMPLETE_STUDIO_MUTATION` to update the actual data in redux, and the studio-mutations reducer just tracks an in-progress flag (per field) and a mutation error property (per field). 

This PR also includes an additional permission selector (canFollowStudio) which right now is just "isLoggedIn", but can be modified as needed in the future. 

A note on terminology. I'm going to try to avoid ever using the word "loading" in this part of the code, since it just invites so much confusion. There are now at least 2 senses in which data can be "loading", it could be being fetched from the server, or it could be being modified by the user, waiting for server confirmation. So I'm using "fetching" and "mutating" explicitly from now on :) 

This PR is best reviewed commit-by-commit, since there is some refactoring.

/cc @rschamp leaning into redux, actually enjoying it :)

### DIFF
--- a/src/views/studio/studio-image.jsx
+++ b/src/views/studio/studio-image.jsx
@@ -1,0 +1,69 @@
+/* eslint-disable react/jsx-no-bind */
+import React from 'react';
+import PropTypes from 'prop-types';
+import {connect} from 'react-redux';
+
+import {selectStudioImage, selectIsLoadingInfo} from '../../redux/studio';
+import {selectCanEditInfo} from '../../redux/studio-permissions';
+import {
+    mutateStudioImage, selectIsMutatingImage, selectImageMutationError
+} from '../../redux/studio-mutations';
+import Spinner from '../../components/spinner/spinner.jsx';
+
+const StudioImage = ({
+    imageError, isLoading, isMutating, image, canEditInfo, handleUpdate
+}) => (
+    <div>
+        <h3>Image</h3>
+        {isLoading ? (
+            <h4>Loading...</h4>
+        ) : (
+            <div>
+                <div style={{width: '200px', height: '150px', border: '1px solid green'}}>
+                    {isMutating ?
+                        <Spinner color="blue" /> :
+                        <img
+                            style={{objectFit: 'contain'}}
+                            src={image}
+                        />}
+                </div>
+                {canEditInfo &&
+                    <label>
+                        <input
+                            disabled={isMutating}
+                            type="file"
+                            accept="image/*"
+                            onChange={e => {
+                                handleUpdate(e.target);
+                                e.target.value = '';
+                            }}
+                        />
+                        {imageError && <div>Error mutating image: {imageError}</div>}
+                    </label>
+                }
+            </div>
+        )}
+    </div>
+);
+
+StudioImage.propTypes = {
+    imageError: PropTypes.string,
+    canEditInfo: PropTypes.bool,
+    isLoading: PropTypes.bool,
+    isMutating: PropTypes.bool,
+    image: PropTypes.string,
+    handleUpdate: PropTypes.func
+};
+
+export default connect(
+    state => ({
+        image: selectStudioImage(state),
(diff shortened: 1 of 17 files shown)


paulkaplan:
@seotts an update, I pushed a commit to include the same feature but for updating the studio image.

src/views/studio/studio-info.jsx
```diff
@@ -36,10 +36,7 @@ const StudioInfo = ({
 
 StudioInfo.propTypes = {
     isLoggedIn: PropTypes.bool,
-    studio: PropTypes.shape({
-        title: PropTypes.string,
-        description: PropTypes.description
-    }),
+    studio: PropTypes.shape({}), // TODO remove, just for <Debug />
```
seotts, line 41: Looks like you meant to remove this

src/views/studio/studio-image.jsx
```diff
@@ -0,0 +1,69 @@
...
+    mutateStudioImage, selectIsMutatingImage, selectImageMutationError
+} from '../../redux/studio-mutations';
+import Spinner from '../../components/spinner/spinner.jsx';
+
+const StudioImage = ({
+    imageError, isLoading, isMutating, image, canEditInfo, handleUpdate
+}) => (
+    <div>
+        <h3>Image</h3>
+        {isLoading ? (
```
seotts, line 18: Just because you specifically called out the language around "loading"-- I noticed here and in the other views the use of "loading" where "fetching" might be more precise. Not sure that needs to be changed in this PR, necessarily.

seotts (changes requested):
While testing this manually, I noticed a couple of unhandled error types: `"editable-text-too-long"` and `"This field is required."` on the title and description. I get the latter by trying to set either field to be completely empty, but I'm guessing that at least the description _should_ be able to be empty. Maybe that one is a problem on the api side?

Also, adding an image appears to work, but when I refresh, the image is gone. Is that expected at this point?

Because of those things, and the minor comments I made, I'm going to set this to "changes requested." But since this won't be in front of Scratchers yet, I'm cool with you merging this if it is at the place you intend it to be at this point!

seotts:
Upon looking at this a little further, I think the problem with the image is with the fetching/displaying rather than with the mutation of it. After changing the image, and refreshing, I can see it on the `/studios` page but not `/studios-playground`. It seems like they are trying to access different versions of the image (200x130 vs 170x100.)

paulkaplan:
@seotts yeah it looks like the api is returning an image with a full cdn2.scratch.mit.edu/... url, which would be right for production, but isn't working locally. The response that comes from legacy backend, including after you update the thumbnail and when you view the old studio page, uses the local version. So seems like a local API issue. It will also interfere with testing on staging, so i might look into the API issue.

seotts:
Also, I'm getting a server error when I toggle comments allowed-- "Only post requests are allowed for this url."

Toggling "open to all" works though!

paulkaplan:
@seotts wow strange thanks for noticing that. I was just adding a new permission selector because the current behavior is that only studio creator can toggle comments (not managers), so I'll also look into that.

paulkaplan:
Yeah should have been a POST not a PUT. Not sure why it worked for one way and not the other though... 

I added a separate permission  selector to make sure comment enabling only is visible for people it currently functions for (admins and creators)

src/views/studio/studio-info.jsx
paulkaplan: Oh yeah in the future, its still useful to have that debug information visible at all times in the playground

seotts:
Your changes LGTM!

PR #5279, opened 2021-04-22
Move comment mute strings to general l10n file

paulkaplan:
### Resolves:

Fixes an issue where the studio page did not contain the strings for specifics about mute status. Thanks @picklesrus and @BryceLTaylor pointing this out. 

### Changes:

Moves the strings related to the mute status out of the project-page-specific l10n file and into the general l10n file, so they work for the new studio page as well.

### Test Coverage:

No added tests. It would be nice if we had a test that checked if you used any formatted message IDs that weren't defined in the files... not sure how to do that though.


apple502j:
The best way IIRC is to just open the page with the component (after `npm run translate`) and checking the console.

Also, I hope this does not mean translators need to copy-paste everything...

paulkaplan:
@apple502j yeah, the issue is the component is deeply nested and conditionally visible (the mute modal), so it wasn't throwing errors normally because it isn't normally rendered.

And @chrisgarrity can confirm, but I believe since we are maintaining the same IDs, we don't need to retranslate. The different files just split up the strings for the website, but they get smushed all together for transifex (i think...)

chrisgarrity:
@apple502j the project settings for scratch-website are set to automatically translate when there is an exact match in the translation memory. So the strings should still be translated even though they have moved from one resource to another, even if they didn't keep the same key. (Just FYI @paulkaplan the different files are maintained as different resources in transifex - they have different priorities for translation, e.g. translating the project page is much more important than translating the develop page.)

chrisgarrity (approved):
LG and I confirmed that the settings in transifex shouldn't require folks to retranslate.

PR #5280, opened 2021-04-22
Reset studio comments when admin logs in

paulkaplan:
### Resolves:

Resolves an issue we found in testing where an admin user might navigate directly to /studios-playground/123/comments and would see the comments as if they weren't an admin. The reason is that comments do not wait to load for the user session to return, because for almost all users it doesn't matter. But admins see reported comments, so the comment list either has to wait for the user session to return (slowing down the experience for all users) or reset the comment list when the session returns as an admin (possibly causing a layout shift only for admins as their logged out comment list gets replaced by the admin one). I chose the second option, since it will provide the fastest experience for far more people.

### Changes:

Instead of only loading the initial comment list on first render, I added `comments.length === 0` to the effect hook dependency list, so it will re-load comments any time that number changes to 0 (or initially). 

Then I added another effect hook that uses `useRef` to track the previous value of `isAdmin`, and compares it against the new value (this is like `prevProps.X !== props.X` in componentDidUpdate). If the user "becomes" an admin, the comments are reset, which then gets the comments reloaded. 

### Test Coverage:

Added a unit test for the `studio-comment` component (and exported the unconnected version for simpler testing). It tests the situations when the initial comment load is called, and when the admin reset is called.

For manual testing, the repro is:
- Log in as an admin
- Go to a studio with studios-playground that has reported comments
- Click the "comments" tab -> see the reported comments in red.
- Reload the page, which will load the comment tab directly with url `/comments`
- You don't see the reported comments. (this is fixed with this PR).

### DIFF
--- a/src/views/studio/studio-comments.jsx
+++ b/src/views/studio/studio-comments.jsx
@@ -1,4 +1,4 @@
-import React, {useEffect} from 'react';
+import React, {useEffect, useRef} from 'react';
 import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
 import {FormattedMessage} from 'react-intl';
@@ -9,6 +9,7 @@ import TopLevelComment from '../preview/comment/top-level-comment.jsx';
 import studioCommentActions from '../../redux/studio-comment-actions.js';
 import StudioCommentsAllowed from './studio-comments-allowed.jsx';
 
+import {selectIsAdmin} from '../../redux/session';
 import {
     selectShowCommentComposer,
     selectCanDeleteComment,
@@ -22,6 +23,7 @@ import {selectStudioCommentsAllowed} from '../../redux/studio.js';
 const StudioComments = ({
     comments,
     commentsAllowed,
+    isAdmin,
     handleLoadMoreComments,
     handleNewComment,
     moreCommentsToLoad,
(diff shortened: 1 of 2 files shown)


src/views/studio/studio-comments.jsx
```diff
@@ -30,12 +31,22 @@ const StudioComments = ({
...
     handleLoadMoreReplies
 }) => {
     useEffect(() => {
         if (comments.length === 0) handleLoadMoreComments();
-    }, []); // Only runs once after the first render
+    }, [comments.length === 0]);
+
+    // The comments you see depend on your admin status
+    // so reset them if isAdmin changes.
+    const adminRef = useRef(isAdmin);
```
paulkaplan, line 44: refs are often used to hold on to DOM nodes, but they actually can contain any mutable value (they return an object with a `.current` property, you can assign whatever you want to it). The first argument to `useRef` is the initial current value, if the ref hasn't already been created. https://reactjs.org/docs/hooks-reference.html#useref

src/views/studio/studio-comments.jsx
```diff
@@ -30,12 +31,22 @@ const StudioComments = ({
...
+    useEffect(() => {
+        const wasAdmin = adminRef.current;
+        adminRef.current = isAdmin;
+        if (isAdmin !== wasAdmin) handleResetComments();
+    }, [isAdmin]);
```
kchadha, line 49: To confirm I'm reading this right...
Does this mean that handleResetComments will run if we go from not being an admin -> being an admin, or will it also run if we go from being an admin -> not being an admin?

src/views/studio/studio-comments.jsx
```diff
@@ -98,13 +110,19 @@ StudioComments.propTypes = {
...
     handleRestoreComment: PropTypes.func,
     handleReportComment: PropTypes.func,
+    handleResetComments: PropTypes.func,
     handleLoadMoreReplies: PropTypes.func,
     postURI: PropTypes.string
 };
 
+export {
+    StudioComments
+};
```
kchadha, line 120: Is this just for use in the test?

src/views/studio/studio-comments.jsx
kchadha: Actually I think the test answers my question

kchadha (approved):
LGTM!

kchadha (changes requested):
Oops just saw the merge conflicts. LGTM _after_ those are resolved.

PR #5287, opened 2021-04-23
circleci turn on translation push cron job

BryceLTaylor:
### Resolves:

The translation push is still happening on Travis.  This is a step towards moving off travis and onto circle.

### Changes:

This turns on a cron job that runs the push translations job.  It runs at midnight UTC (8pm EDT).  It does not stop Travis from running the same job, so they should both be pushed.  This is so we can test to make sure Circle is doing it correctly before turning off the Travis cron job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,6 +114,13 @@ aliases:
             npm run test:smoke:convertReportToXunit
       - store_test_results:
           path: test/results
+  - &update-translations
+    <<: *defaults
+    steps:
+      - checkout
+      - run:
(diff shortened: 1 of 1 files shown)


chrisgarrity (approved):
LG

.circleci/config.yml
```diff
@@ -220,3 +229,19 @@ workflows:
             branches:
               only:
                 - master
+  Update-translations:
```
rschamp, line 232: The capital U is the only thing I take issue with

rschamp (approved):

PR #5306, opened 2021-04-28
Studio info styling

paulkaplan:
Implement the first round of styling for the studio "info" bar on the left with the title/image/description/follow action.

The bulk of the code change is layout refactor. I tried to make it so that each component exported a single element, using css classnames to indicate fetching/mutating state so those can be styled. The layout is implemented using css grid, since that is the easiest way to add gutters between elements, even though this is a 1d layout right now. 

The first commit also removes some unneeded status state transitions that I found while stepping through the redux dev tools to look at the loading progress.

/cc @seotts this is a pretty css-heavy PR, i'm happy to assign it elsewhere or walk through the css with you. There is some fun stuff there :)

### DIFF
--- a/src/views/studio/studio-title.jsx
+++ b/src/views/studio/studio-title.jsx
@@ -6,29 +6,28 @@ import {connect} from 'react-redux';
 import {selectStudioTitle, selectIsFetchingInfo} from '../../redux/studio';
 import {selectCanEditInfo} from '../../redux/studio-permissions';
 import {mutateStudioTitle, selectIsMutatingTitle, selectTitleMutationError} from '../../redux/studio-mutations';
+import classNames from 'classnames';
 
 const StudioTitle = ({
     titleError, isFetching, isMutating, title, canEditInfo, handleUpdate
-}) => (
-    <div>
-        <h3>Title</h3>
-        {isFetching ? (
-            <h4>Fetching...</h4>
-        ) : (canEditInfo ? (
-            <label>
-                <input
-                    disabled={isMutating}
-                    defaultValue={title}
-                    onBlur={e => e.target.value !== title &&
-                        handleUpdate(e.target.value)}
-                />
-                {titleError && <div>Error mutating title: {titleError}</div>}
-            </label>
-        ) : (
-            <div>{title}</div>
-        ))}
-    </div>
-);
+}) => {
+    const fieldClassName = classNames('studio-title', {
+        'mod-fetching': isFetching,
+        'mod-mutating': isMutating
+    });
+    return (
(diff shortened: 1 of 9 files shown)


src/views/studio/studio-title.jsx
```diff
@@ -6,29 +6,28 @@ import {connect} from 'react-redux';
...
-                {titleError && <div>Error mutating title: {titleError}</div>}
-            </label>
-        ) : (
-            <div>{title}</div>
-        ))}
-    </div>
-);
+}) => {
+    const fieldClassName = classNames('studio-title', {
+        'mod-fetching': isFetching,
```
seotts, line 15: What does the "mod" prefix mean in a bunch of these classes? On first glance, I read it as "an admin is mutating/fetching this field" which I assume is not correct.

src/views/studio/studio.scss
```diff
@@ -0,0 +1,98 @@
+@import "../../colors";
+@import "../../frameless";
+
+$radius: 8px;
+
+.studio-page {
+    background-color: #E9F1FC;
```
seotts, line 7: Is there a color variable in `_colors.scss` we could use instead?

src/views/studio/studio.scss
```diff
@@ -0,0 +1,98 @@
...
+        margin: 0 -8px; /* Outset the border horizontally */
+        padding: 5px 8px;
+        border: 2px dashed $ui-blue-25percent;
+        border-radius: $radius;
+        resize: none;
+        &:disabled { border-color: transparent; }
+    }
+
+    .studio-title {
+        font-size: 28px;
```
seotts, line 54: I think we generally try to use rem instead of px (especially for styles related to text?)

src/views/studio/studio-title.jsx
seotts: Ah, I see in the css that it is "modifying"-- spelling it out or dropping the "mod" might be more clear.

seotts (commented):
I commented a couple of nitpicks, but this is looking good to me. I haven't worked with grid layout, and I'm still wrapping my head around redux stuff, so it might make sense to finish up this review in a quick pairing session.

src/views/studio/studio.scss
mxmou: `$ui-blue-10percent` on top of `$background-color` creates the same color.

src/views/studio/studio.scss
paulkaplan: Yeah I will return to this, there will be a lot of churn in this area.

PR #5330, opened 2021-05-03
Add functionality for adding/removing projects and managing curators

paulkaplan:
This is an initial pass at the functionality for:
- Adding / removing studio projects
- Inviting, accepting invite, promoting and removing studio curators and managers

I know there is a lot here @seotts and a fair amount of the redux related stuff is subject to some churn. The difference from previous PRs is that I'm not attempting to manage all of the mutation/error state in redux, i'm letting the react components handle that by using the fact that thunks are promises. I tried a few different ways to manage mutation state for all the different actions in redux, and it was a lot of code for little benefit (there are no other consumers of an individual curators "removal in progress" state, so it isn't worth keeping in redux). I may revisit the choice to keep other mutation state in redux, since it is so easy to do in the component. 

I also added basic styling for project tiles and curator/manager tiles. The buttons are placeholders, and I haven't attempted to re-use the other thumbnail components we have since these tiles are editable. I may revisit that as well. 

This forms a solid base of functionality that should help uncover other issues. I'd love to have it in before release this week if possible.  
Screen Shot 2021-05-03 at 10 33 12 AM
Screen Shot 2021-05-03 at 10 33 19 AM

Testing / Known issues:
- Adding/Removing projects: add them by ID only (no URL parsing yet). it appears like you can add many copies of the same project to a studio, since the backend responds with 200, although there aren't actually multiple copies. We'll need to alter the server response in order to display the "project already in studio" message, maybe a "409 Conflict" or something.
- Inviting curators: puts up a system alert, will be changed to fancy in-context alert at some point.
- Accepting invitation: should add a user tile for you to the curators section as well as show the "add project" UI since you are now a curator, without requiring you to refresh the page.
- Removing manager: If you remove yourself, you should see the UI for inviting curators disappear, since you are no longer allowed to do that. You should not be able to remove the studio creator.
- Promoting a curator: There is no "confirmation" modal yet, but you should see the curator card disappear from the curators list and get added to the managers list.
- ~~Permissions: some of the action permissions do not match existing ones yet, e.g. curators only being able to remove projects they added, or anyone being able to remove their own project. This is described in the code with TODOs.~~ Nevermind, I added the correct project removal permissions.

### DIFF
--- a/src/redux/studio-permissions.js
+++ b/src/redux/studio-permissions.js
@@ -1,4 +1,4 @@
-const {selectUserId, selectIsAdmin, selectIsSocial, selectIsLoggedIn} = require('./session');
+const {selectUserId, selectIsAdmin, selectIsSocial, selectIsLoggedIn, selectUsername} = require('./session');
 
 // Fine-grain selector helpers - not exported, use the higher level selectors below
 const isCreator = state => selectUserId(state) === state.studio.owner;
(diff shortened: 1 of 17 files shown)


seotts (approved):
Per our discussion yesterday, I didn't do a close line-by-line review, but I tested locally and things are looking good. One thing I accidentally found was that on narrow browser sizes, the tiles are _very_ large -- but of course that shouldn't block release of this!
image

PR #5336, opened 2021-05-04
Run npm ci in update-translations cron on circle

BryceLTaylor:
### Resolves:

The cron job that pushes translations to transifex failed with the error: 

Error: Cannot find module 'glob'

We weren't running install (or CI in this case) in the job so it did not have access to the glob package.

### Changes:

Adds a run step that runs npm ci to the update-translations job


chrisgarrity (approved):
Yup, that would do it!

PR #5342, opened 2021-05-05
Revert "Add privacy policy and terms of use to allowlist"

benjiwheeler:
Reverts LLK/scratch-www#5220

Turns out we did need those redundant entries with the slashes after all!

Fix seems to be to either strip the trailing slash from the pathname, or reverse the order of the indexOf and use some kind of reducer to check if the pathname contains any of the goodlist entries.


benjiwheeler:
should make the revert to the release branch instead

PR #5343, opened 2021-05-05
Revert "Add privacy policy and terms of use to allowlist"

benjiwheeler:
Reverts LLK/scratch-www#5220

Turns out we did need those redundant entries with the slashes after all!

Fix seems to be to either strip the trailing slash from the pathname, or reverse the order of the indexOf and use some kind of reducer to check if the pathname contains any of the goodlist entries.

### DIFF
--- a/src/redux/session.js
+++ b/src/redux/session.js
@@ -13,10 +13,9 @@ const Types = keyMirror({
 });
 
 const banWhitelistPaths = [
-    '/accounts/banned-response',
-    '/community_guidelines',
-    '/privacy_policy',
-    '/terms_of_use'
+    '/accounts/banned-response/',
+    '/community_guidelines/',
+    '/community_guidelines'
 ];
 
 module.exports.Status = keyMirror({


Naleksuh:
@benjiwheeler Ah, the problem here is that indexOf checks if a string contains a string, but for an array, it must perfectly match one of the items, not contain.

The best way to approach this would depend on what the goal with the redirect system is. On a personal level I oppose systematically redirecting banned users to begin with, however the goals here may be different.

If a path /terms_of_use is whitelisted, would you expect /terms_of_use/aosdijfaosdijfaow to work, or not? If it is, then I'd expect the solution is to loop through each item in banWhitelistPaths and compare via indexOf on that item individually to the current path. If it isn't then it would likely be best to simply remove any trailing slashes (whether strictly for this comparison or disallowing navigating to URLs with trailing slashes entirely via the backend).

An alternative solution would be to have each item in banWhitelistPaths be a regex, which allows for both individually (allowing trailing slashes, not, allowing subdirectories, not)

PR #5346, opened 2021-05-06
Bug fixes from studio playground test on May 5 2021

paulkaplan:
Each commit resolves an issue that was uncovered during the test of the studios playground on may 5:
- Catch the error that was being returned when trying to invite a user that was already a member of the studio
- Fix the grid layout of the sidebar on firefox
- Fix the issue where going to .../curators/ (with trailing slash) would prevent the tab nav from working the first time you clicked another tab
- Correct the tab ordering
- Fix the issue where promoted or added curators would go at the beginning of the list instead of the end
- Allow the "Enter" key to submit the curator invitation / project adder forms
- Catch situations where a thumbnail is too large on the client side (file-size-wise, not dimensionally)

@seotts each commit addresses a single issue, but the code all together is spread out. I recommend using the "commit-by-commit" view to review this PR.

### DIFF
--- a/src/views/studio/lib/studio-member-actions.js
+++ b/src/views/studio/lib/studio-member-actions.js
@@ -8,13 +8,20 @@ import {selectStudioId, setRoles} from '../../../redux/studio';
 const Errors = keyMirror({
     NETWORK: null,
     SERVER: null,
-    PERMISSION: null
+    PERMISSION: null,
+    DUPLICATE: null
 });
 
 const normalizeError = (err, body, res) => {
     if (err) return Errors.NETWORK;
     if (res.statusCode === 401 || res.statusCode === 403) return Errors.PERMISSION;
     if (res.statusCode !== 200) return Errors.SERVER;
+    if (body && body.status === 'error') {
+        if (body.message.indexOf('already a curator') !== -1) {
+            return Errors.DUPLICATE;
+        }
+        return Errors.UNHANDLED;
+    }
     return null;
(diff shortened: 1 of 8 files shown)


src/views/studio/lib/studio-member-actions.js
```diff
@@ -8,13 +8,20 @@ import {selectStudioId, setRoles} from '../../../redux/studio';
...
     SERVER: null,
-    PERMISSION: null
+    PERMISSION: null,
+    DUPLICATE: null
 });
 
 const normalizeError = (err, body, res) => {
     if (err) return Errors.NETWORK;
     if (res.statusCode === 401 || res.statusCode === 403) return Errors.PERMISSION;
     if (res.statusCode !== 200) return Errors.SERVER;
```
apple502j, line 18: Errors for rate-limit (I hope it's 429?) should be handled as well, since people who try to add all followers to a studio can easily encounter that.

apple502j, line 18: 404 could be thrown for users that do not exist.

apple502j (changes requested):
Some more suggestions

src/views/studio/lib/studio-member-actions.js
paulkaplan: Good point, fixed with e2dafff

seotts:
Aside from the nitpicky comment I made, this lgtm!

PR #5359, opened 2021-05-07
Add l10n file to studios view and replace text with formatted message.

kchadha:
### Changes:

Set up translation for the Studio view. I only added the strings that I thought were going to make it into the final design, but there might be some that should be added that I was unsure about (e.g. "Loading...").

### Test Coverage:

You can build and run without running `npm run translate` and you should see some FormattedMessage ids instead of the actual copy text (e.g. "general.loadMore" or "studio.projectsHeader").

If you run `npm run translate` and restart the server, you should see the strings in English again.


paulkaplan (approved):
LGTM

PR #5367, opened 2021-05-10
Add the user projects modal for browsing projects to add to a studio

paulkaplan:
Screen Shot 2021-05-10 at 10 03 49 AM

Adds the "Browse projects" button and corresponding user projects modal. 

The favorites / shared filters already work, but the "recent" filter endpoint isn't ready yet, but I left it in there as a stub. 

There is an existing issue with being able to add a project that already exists in the studio, but this needs to be resolved on the backend (we'll talk about it next eng meeting). 

The first few commits are fixups for tests/issues i uncovered in development, the last commit is the actual user projects browser.

### DIFF
--- a/src/views/studio/lib/user-projects-actions.js
+++ b/src/views/studio/lib/user-projects-actions.js
@@ -1,0 +1,54 @@
+import keyMirror from 'keymirror';
+import api from '../../../lib/api';
+import {selectUsername} from '../../../redux/session';
+import {userProjects} from './redux-modules';
+
+const Errors = keyMirror({
+    NETWORK: null,
+    SERVER: null,
+    PERMISSION: null
+});
+
+const Filters = keyMirror({
+    SHARED: null,
+    FAVORITED: null,
+    RECENT: null
+});
+
+const Uris = {
+    [Filters.SHARED]: username => `/users/${username}/projects`,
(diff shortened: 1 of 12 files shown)


src/views/studio/lib/user-projects-actions.js
```diff
@@ -0,0 +1,54 @@
...
+
+const loadUserProjects = type => ((dispatch, getState) => {
+    const state = getState();
+    const username = selectUsername(state);
+    const projectCount = userProjects.selector(state).items.length;
+    const projectsPerPage = 20;
+    dispatch(userProjects.actions.loading());
+    api({
+        uri: Uris[type](username),
+        params: {limit: projectsPerPage, offset: projectCount}
```
apple502j, line 39: I'm sure this requires authentication for "recent"? That's how things work currently

paulkaplan: Yeah it will. We don't have that API route set up yet though, so I'll follow up when it is. (I don't want to submit too much "speculative" code that may not end up actually working).

paulkaplan:
@kchadha good point about the rows. We'll need to change the load quantity to a number that divides by each possible break point (1, 2, 3), so 18 or 24? we can update that later, we'll have to change every one of the grids.

paulkaplan:
@kchadha updated the merge conflicts and moved to formatted message

PR #5370, opened 2021-05-10
Set up studios view with 404 page

kchadha:
### Changes:

Display the 404 page when fetching a studio results in a 404.

Add info about studio availability to the studio reducer.
Hook up the studio availability state to the StudioShell to display a 404 page if we get an error while fetching the studio.

### Test Coverage:

Tested locally:
- get a 404 page when using a fake studio ID
- get the studio page when going using an existing studio ID

### DIFF
--- a/src/redux/studio.js
+++ b/src/redux/studio.js
@@ -26,7 +26,9 @@ const getInitialState = () => ({
     manager: false,
     curator: false,
     following: false,
-    invited: false
+    invited: false,
+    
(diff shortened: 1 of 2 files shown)


paulkaplan (changes requested):
@kchadha I don't think you need the extra `studioNotAvailable` state, i think you can just have the selector check `studio.infoStatus === Status.ERROR`

PR #5371, opened 2021-05-10
Add studio activity pagination

paulkaplan:
This PR moves the way studio activity items are loaded from the previous `loadMore` infinite list fetcher that didn't support pagination to the newer method already used by studio projects/curators, which is having a separate actions file that orchestrates the loading of items. This is useful for the activity feed because the pagination is not done by offset, rather by a `dateLimit` param which is the newest item you want to receive (order is newest to oldest).

The second commit also finally removes the built-in loadMore function from the infinite-list reducer, since nobody is using it anymore and it is not needed / wanted (because it is too specific).

### DIFF
--- a/src/redux/infinite-list.js
+++ b/src/redux/infinite-list.js
@@ -17,14 +17,6 @@
  */
 
 /**
- * @typedef {function} InfiniteListFetcher
- * A function to call that returns more data for the InfiniteList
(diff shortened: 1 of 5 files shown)


ericrosenbaum (approved):
LGTM!

PR #5372, opened 2021-05-10
Styled studio activity feed

ericrosenbaum:
Initial version of the styled studio activity feed, showing the five messages in the spec (added project, removed project, edited studio, promoted to curator, promoted to manager). 

Not yet included:
- Usernames with links to user profiles (these are waiting on API changes). For now, they just say "username"
- "Load more" button


src/views/studio/studio-activity.jsx
```diff
@@ -1,5 +1,6 @@
 import React, {useEffect} from 'react';
 import PropTypes from 'prop-types';
+const FormattedMessage = require('react-intl').FormattedMessage;
```
paulkaplan, line 3: Since the rest of the file uses `import`, this line should be changed to 
```
import {FormattedMessage} from 'react-intl';
```

src/views/studio/studio-activity.jsx
```diff
@@ -8,6 +9,140 @@ import {activity} from './lib/redux-modules';
...
+                iconSrc="/svgs/studio/activity-project.svg"
+                iconAlt="project activity icon"
+                imgClassName="studio-activity-icon"
+                key={item.id}
+            >
+                <FormattedMessage
+                    id="studio.activityAddProjectToStudio"
+                    values={{
+                        profileLink: (
+                            <a href={`/users/${item.actor_id}`}>
```
paulkaplan, line 31: A heads up, user links are /users/:username, not by id. I'd change this to `actor_username`, even if that isn't defined it, since it will be defined when we fix api. Same for the fixed string `username`, i'd change that to `{item.actor_username}`

paulkaplan (changes requested):
A couple small changes. Once that is done though, feel free to merge.

src/views/studio/studio-activity.jsx
ericrosenbaum: fixed!

PR #5388, opened 2021-05-12
Add error display for studio curator and project adder inputs

paulkaplan:
Include the existing `<ValidationMessage />` component when rendering the project adder and curator inviter. Show some basic error messages. Although there are some more specific messages we can include in the future, this is enough to get us started and we can come back to more later. Part of the reason I'm leaving the project one generic is because we need to resolve some issues with the backend before returning richer error messages.

responsive-errors


paulkaplan:
@ericrosenbaum as mentioned offline, I'm going to bring these in now in advance of todays test. Feel free to still leave any review comments though.

ericrosenbaum (commented):
⚠️👍

PR #5345, opened 2021-05-06
add donate banner, make close button a mode of button component

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/5318

### Changes:

* adds donate banner component and css
* modifies button component to have "close" version, to be reusable
* modifies splash/splash.jsx and splash/presentation.jsx to collect similar banner handling in one place
* adds handling for closing the banner and for redirecting user to donate page

### Issues

Closing the banner doesn't persist for long -- if you browse away from the splash screen and come back, it will reappear.

### Test Coverage:

unclear if it makes sense to add tests?

### Screenshots

image

image

image

image

image

image

PR #5391, opened 2021-05-12
fix logic for when to show intro to users

benjiwheeler:
Follow up to https://github.com/LLK/scratch-www/pull/5345 , fixing logic around when to show intro banner.

Originally, the logic was:

* if user is logged out, AND we're not showing hour of code banner, show intro banner

https://github.com/LLK/scratch-www/pull/5345 mistakenly changed the logic to:

* if we're not showing hour of code banner, show intro banner

This change changes the logic back to:

* if user is logged out, AND we're not showing hour of code banner, show intro banner


chrisgarrity (approved):
Checked locally, we should do a final check on staging.

PR #5397, opened 2021-05-13
Bug fixes from studio playground test on May 12

paulkaplan:
Fixes include:
- Correct placeholders for project/curator inputs
- Ability to add projects from URLs
- Prevention of duplicate project addition on the client side
- Prevention of empty input submission on the client side
- Prevent project browser modal from extending beyond screen height when screen is <700px tall

### DIFF
--- a/src/views/studio/lib/studio-project-actions.js
+++ b/src/views/studio/lib/studio-project-actions.js
@@ -59,8 +59,17 @@ const generateProjectListItem = (postBody, infoBody) => ({
     username: infoBody.author.username,
     avatar: infoBody.author.profile.images
 });
-                        
-const addProject = projectId => ((dispatch, getState) => new Promise((resolve, reject) => {
+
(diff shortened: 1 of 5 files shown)


src/views/studio/lib/studio-project-actions.js
```diff
@@ -59,11 +61,25 @@ const generateProjectListItem = (postBody, infoBody) => ({
...
+        // Take the last match, in case we are on localhost and there are port numbers, e.g.
+        projectId = parseInt(matches[matches.length - 1], 10);
+    }
+
     const state = getState();
     const studioId = selectStudioId(state);
     const token = selectToken(state);
+
+    // Check for existing duplicates before going to the server
+    if (projects.selector(state).items.filter(p => p.id === projectId).length !== 0) {
```
apple502j, line 80: Maybe `.find`?

seotts (approved):
Looks good to me. A quirk of the URL/id system is that any url with a number in it will work, if a project with that id exists (`gmail.com/1`). But I'm not really sure that's an issue..

PR #5399, opened 2021-05-13
Fixed problem introduced by previous fix

zoebentley:
In the last fix, we introduced a problem where "Contact Us" was not a FormattedMessage. This PR fixes that


chrisgarrity (approved):
👍

PR #5401, opened 2021-05-13
Show "students projects" filter on studio project browser

paulkaplan:
This PR connects the "students" filter up to the new api endpoints for getting projects shared by students who are in the classroom the studio is a part of (phew, that was a mouthful). @seotts this uses both the endpoints you introduced, first if the user is an educator, check if the studio is part of a classroom. Then the user-projects-modal shows a "students" filter, which requests projects from the other endpoint. 

As a bonus, i also updated the "recent" filter to have the correct uri and also send authentication along. That endpoint isn't in production yet, but i tested it locally and it was working well /cc @cwillisf. 

Screen Shot 2021-05-13 at 1 30 08 PM


seotts (approved):
Looks good to me!

PR #5405, opened 2021-05-13
Outlines for avatars

fsih:
Round and outline avatars for comments. The spec actually wants these outlines to be black and inset, but having such an outline that is also rounded is not supported on most browsers yet (see https://bugs.chromium.org/p/chromium/issues/detail?id=81556) so I put the border outside the avatar instead and instead made it the color of the dotted line around editable fields

Spec:
Screen Shot 2021-05-13 at 17 53 16

Current:
Screen Shot 2021-05-13 at 17 53 31

Proposed:
Screen Shot 2021-05-13 at 17 54 59

### DIFF
--- a/src/views/preview/comment/comment.scss
+++ b/src/views/preview/comment/comment.scss
@@ -159,6 +159,8 @@
 
     .avatar {
         margin-right: .5rem;
+        border-radius: 4px;
+        box-shadow: 0px 0px 0px 1px rgba(77, 151, 255, 0.25);
     }
 
     .comment-body {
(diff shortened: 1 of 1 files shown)


paulkaplan:
I assume this isn't also intended to address the image sizing issues? They are all supposed to be the same size

RedGuy12:
GIF profile pictures and the logged-in users' profile pictures are the only images that aren't sized correctly.

The GIF sizing issue can be fixed by adding:
```css
.comment .avatar {
  height: 60px;
  width: 60px;
}
```

PR #5418, opened 2021-05-14
Bugfixes for studio playground test May 14

paulkaplan:
7 quick fixes from the test on may 14. 

- Fix loadMore not working for user-projects-modal
- Fix awkward project thumbnail layout issues
- Fix issue where multiple overflow menus could be opened at once
- Allow curators to remove themselves
- Trim whitespace from username input
- Allow space to trigger submit on project tiles in modal because they are buttons and should act like it to the keyboard
- Make enter submit title field instead of adding newlines

Definitely recommend reviewing this in the "commit-by-commit" view.


kchadha (approved):
LGTM from looking at the code. I didn't test locally but we can do that in the next bug hunt probably?

Thanks for putting each fix into separate commits! That was really helpful for reviewing the changes!!

PR #5432, opened 2021-05-17
Update the curator invitation component to spec

paulkaplan:
This PR updates the studio-curator-invitation component to the design spec, including a "success" state (and unspec'd "error" state in case of server errors). There is a 5 second timeout on the success message.

image
image
image


src/views/studio/studio-curator-invite.jsx
```diff
@@ -6,38 +6,71 @@ import classNames from 'classnames';
...
+            <div className="studio-invitation-msg">
+                <FormattedMessage id="studio.curatorInvitation" />
+            </div>
             <button
-                className={classNames('button', {
+                className={classNames('studio-invitation-button button', {
                     'mod-mutating': submitting
                 })}
                 disabled={submitting}
                 onClick={() => {
```
kchadha, line 48: I thought we couldn't use arrow functions as JSX props...

apple502j: Technically, we can. `jsx-no-bind` eslint rule forbids that, but it's not like it breaks the code. (Bad for performance though)

And we eslint-disable'd the rule file-wide so it doesn't cause lint failure.

paulkaplan:
Yeah I’ve disabled it in much of the studios work.

I was the one who originally elevated it to an error. My thinking on it has
changed though, and once I get a bit of time to share with the team I’d
like to remove it from our eslint config.

The short version: It is bad for performance only given two things, neither
of which are true here:
1. The component is updated very frequently (eg in an animation) or very
close to the top of a deep render tree.
2. The callback is being passed down to other custom components that
implement shouldComponentUpdate/React.memo to prevent re-renders.

Both of those conditions were very frequently true when working in
scratch-gui, and the lint rule was helpful rooting out performance issues,
but are rarely the case in www.

kchadha (commented):
LGTM except for the merge conflicts!

PR #5436, opened 2021-05-17
Wait for session to be fetched before loading comments

paulkaplan:
This PR fixes an issue we found in testing where admins could not always see the deleted comments list. 

The cause is trying to fetch studio comments without waiting for the session, because the `isAdmin` flag is becoming true after the initial non-admin comment request went out, but before any comments are loaded. Since the logic for re-requesting admin comments only works if there is a change in the number of comments in the list, and the reset is being triggered before any comments are loaded, there is no change and no comments are re-requested. I was hoping to avoid this, but I think for consistency sake we need to wait until the user session is fetched to start the comment request, so we can know for sure which endpoint to use. this is how the project page works as well.

It takes a bit to repro the issue locally, just refreshing /comments while logged in as an admin, i got it about 10% of the time.


seotts (approved):
lgtm, but I have not tested locally, due to the problem with making and viewing comments in development, which has a fix in progress. But from looking at the code, it looks good to merge.

PR #5437, opened 2021-05-17
Handle more types of project add errors

paulkaplan:
Fixes an issue where all "add project" errors were being flattened into a single error message. I added text for more of the errors, which will definitely help with testing, even if we don't choose to keep all the messaging.

This uses the same `errorToMessageId(error)` pattern from the curator adder

### DIFF
--- a/src/views/studio/l10n.json
+++ b/src/views/studio/l10n.json
@@ -27,10 +27,10 @@
     "studio.projectsEmpty2": "Suggest projects you want to add in the comments!",
     "studio.browseProjects": "Browse Projects",
     "studio.projectErrors.checkUrl": "Could not find that project. Check the URL and try again.",
-    "studio.projectErrors.generic": "Could not add project",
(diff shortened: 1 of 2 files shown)


src/views/studio/l10n.json
```diff
@@ -26,7 +26,11 @@
     "studio.projectsEmpty1": "This studio has no projects yet.",
     "studio.projectsEmpty2": "Suggest projects you want to add in the comments!",
     "studio.browseProjects": "Browse Projects",
-    "studio.projectErrors.checkUrl": "Could not add project. Check the URL and try again.",
+    "studio.projectErrors.checkUrl": "Could not find that project. Check the URL and try again.",
+    "studio.projectErrors.generic": "Could not add project",
```
apple502j, line 30: Dot inconsistency?

paulkaplan: 👍 fixed in b0eae61

seotts (approved):
lgtm!

PR #5439, opened 2021-05-17
Add dynamic meta tags for studio pages

paulkaplan:
Similar to the way we do for project pages, add dynamic meta tags to adjust the title and meta info based on API responses. It is a bit hard to test that things like open graph meta tags are working because the pages cannot be accessed in production, so we'll come back to this after launch to check. 

Truncate the description to 50 words, similar to the project page.


kchadha (approved):
LGTM

PR #5440, opened 2021-05-17
Linkify studio descriptions

### LINKED ISSUE #5411: Studio description always uses textarea; no links get linkified
For those who can edit descriptions, the studio description is an editable textarea, which is expected (and is current behavior.) However, the description also uses textarea for those who can't edit. While it's disabled, it's inconsistent with current behavior.

This also **makes it so that no links are clickable.**

image

image

apple502j:
### Resolves:
Resolves #5411

### Changes:
Studio descriptions now get "linkified". This PR also adds CSS to match the behavior for editable descriptions.

Note the `height` - textarea has `rows=20`, but this does not include things like line height, so I had to take that into account.

### Test Coverage:
Manual testing.

### DIFF
--- a/src/views/studio/studio-description.jsx
+++ b/src/views/studio/studio-description.jsx
@@ -33,7 +33,7 @@ const StudioDescription = ({
     });
     return (
         <div className="studio-info-section">
-            { canEditInfo ? (
+            {canEditInfo ? (
                 <React.Fragment>
                     <textarea
                         rows="20"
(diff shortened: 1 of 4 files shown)


paulkaplan (changes requested):
This is almost there! You need to make the studio-description-text look like the `:disabled` version though: no dashed border, background color $ui-blue-10percent (see the css just above where you added css)

paulkaplan:
Thanks @apple502j . I just pushed a couple of commits to finish this up, but basically i like your idea so i applied it to the `title` as well (not the linkifying, just using a div instead of a textarea). I got rid of the `:disabled` styling since that was for the non-editable text areas, and also went back to using a single css class and changing the styling based on the tag. The motivation for that is there was some `mod-` css classes in the `fieldClassName` that i wanted to maintain, like the placeholder class while fetching.

PR #5443, opened 2021-05-18
Use circleCI build ID for saucelabs test names instead

BryceLTaylor:
### Resolves:

The tap and jest integration tests were expecting a travis build id which didn't exist since we switched to travis.  

### Changes:

Both the integration and the integration-legacy selenium-helpers.js file now set the buildID to the Circle workflow id.  The text of the name passed to saucelabs also has been updated to say CircleCI instead of Travis.

### Test Coverage:

The integration tests should still pass and on saucelabs they should report the workflow id from circle.

### DIFF
--- a/test/integration-legacy/selenium-helpers.js
+++ b/test/integration-legacy/selenium-helpers.js
@@ -5,7 +5,8 @@ require('chromedriver');
 const headless = process.env.SMOKE_HEADLESS || false;
 const remote = process.env.SMOKE_REMOTE || false;
 const ci = process.env.CI || false;
-const buildID = process.env.CIRCLE_WORKFLOW_ID;
+const usingCircle = process.env.CIRCLECI || false;
+const buildID = process.env.CIRCLE_WORKFLOW_ID || '0000';
 const {SAUCE_USERNAME, SAUCE_ACCESS_KEY} = process.env;
 const {By, Key, until} = webdriver;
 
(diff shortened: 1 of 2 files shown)


test/integration-legacy/selenium-helpers.js
```diff
@@ -5,7 +5,7 @@ require('chromedriver');
 const headless = process.env.SMOKE_HEADLESS || false;
 const remote = process.env.SMOKE_REMOTE || false;
 const ci = process.env.CI || false;
-const buildID = process.env.TRAVIS_BUILD_NUMBER;
+const buildID = process.env.CIRCLE_WORKFLOW_ID;
```
rschamp, line 8: Maybe this is a good time to use something platform-agnostic like `BUILD_ID` for this. Then do the assignment a level up. Also, maybe it should have a recognizable default (like `0000` or something)?

test/integration-legacy/selenium-helpers.js
```diff
@@ -35,7 +35,7 @@ class SeleniumHelper {
         if (remote === 'true'){
             let nameToUse;
             if (ci === 'true'){
-                nameToUse = 'travis ' + buildID + ' : ' + name;
+                nameToUse = 'circleCI ' + buildID + ' : ' + name;
```
rschamp, line 38: Or maybe we should just expect this as its own environment variable?

rschamp (approved):
This will definitely fix the problem but I have some suggestions. We don't have to do these now though.

test/integration-legacy/selenium-helpers.js
BryceLTaylor: Circle has a built in CIRCLECI environment variable set to true and Travis similarly has a TRAVIS environment variable set to true.  Since we're not expecting to use Travis anymore I'm just checking to see if we're using Circle.  If for some reason we're not it'll be set to "unknown".

It didn't seem like it was worth it to check if we were using Travis.

rschamp: Sorry I meant we could pass all of `nameToUse` in one environment variable and move all the logic about what to call it outside of these files.  But I think this is fine as is for now.

rschamp (approved):

PR #5451, opened 2021-05-19
Alert component and studio error and success alerts

paulkaplan:
This PR adds an `<Alert />` component that can be used with the `<AlertProvider />` and triggered using the functions provided by `useAlertContext()`. This uses the react context api to let us set up "alert boundaries", within which you can specify where you want the alert to display, and use the `errorAlert` and `successAlert` functions from within other components to show alerts. 

One benefit of react context is that they are nestable: any component using `useAlertContext()` will automatically be set up to use the nearest `<AlertProvider>`, even if they are nested. This is useful for the projects section, where there is one alert provider for the project list, and another (nested) inside the project browser modal. 

The first commit includes this new component, hook and an example on how to use it in the `components` view. The second commit hooks these up to be used for all the places where errors are not currently being caught:
- Errors when removing project tiles
- Errors when removing curator or manager tiles
- Errors when promoting a curator
- Success alert when promoting a curator
- Success alert when inviting a curator
- Error (shown as success) when inviting a curator who is already invited
- Error (shown as success) when adding a project that is already added
- Error when adding or removing a project in the user-projects-modal

I'm definitely looking for feedback on this use of react context, since it is new to the codebase. 

Screen Shot 2021-05-19 at 1 55 31 PM
Screen Shot 2021-05-19 at 1 58 43 PM
Screen Shot 2021-05-19 at 1 59 03 PM

Also this changes the messaging around already invited curators, so closes https://github.com/LLK/scratch-www/pull/5451


rschamp (approved):
I like it! Disclaimer, I just read the code, and did not try this out locally, so trusting you that it actually works :)

rschamp:
Also, it's nice that it still somehow fits in our split between general-use components vs view-specific ones, even though it comes with some behavior/state management.

PR #5458, opened 2021-05-20
Studio report modal style fixes

ericrosenbaum:
Fixes to the studio report modal:
- Modal is shorter, with buttons on a single row, to prevent "send" button being off the bottom of shorter screens
- Button grid is responsive for narrower screens
- Instruction text at the top of the modal is improved
- Added image to the "thank you" modal and adjusted its size


paulkaplan (approved):
LG! I pushed a commit with a one line change to make sure the tiles act more clickable by giving them a pointer cursor on hover. Also confirmed they are keyboard navigable which is neat (radio groups are navigable by arrow keys once you are focused on one of them). The button still lacks a focus state, but that isn't your fault.

PR #5461, opened 2021-05-20
Fix keyboard navigation in "add to studio" modal

ericrosenbaum:
Keyboard navigation works with interactive HTML elements, like links and buttons. Our subnavigation component expects list items as children, for styling. In many places on www, these list items are wrapped in `<a>` elements, so keyboard navigation does work (e.g. on the "explore" page). 

This change:
- changes the subnavigation styling so it can apply to list items or to buttons
- changes the studio page's "add to studio" modal subnav to use buttons instead of list items, so that keyboard nav can work

### DIFF
--- a/src/components/subnavigation/subnavigation.scss
+++ b/src/components/subnavigation/subnavigation.scss
@@ -12,7 +12,7 @@
     flex-wrap: wrap;
     
 
-    li {
+    li, button {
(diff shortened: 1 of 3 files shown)


paulkaplan (approved):
LG thanks!

PR #5466, opened 2021-05-21
Small fixes for "add to studio" modal

ericrosenbaum:
A few small fixes to the "add to studio" modal:
- The "add to studio" modal can display rows of either 3 or 2 projects. This change makes it load 24 projects per page (i.e. before the load more button shows, then each time you press it). 24 divides evenly by both 3 and 2, so we show full rows in either case.
- Style the border of projects in the "added" state
- Use check and plus icons on the project tiles
- Make the "load more" button display the full width of the modal


paulkaplan (approved):
LGTM thanks!

PR #5467, opened 2021-05-21
Prototype: muted users can't edit studios

seotts:
Some notes:
- Since we haven't done the backend piece yet, in order to test, you have to be muted when you arrive at the studio page (or get muted and then refresh.)
- The interaction with the studio info very much feels like a prototype. I'd like to get this up on staging so others can give it a try.
- The logic of that interaction can probably be pulled out into a separate component (right now, it lives in StudioTitle, StudioImage, and StudioDescription). I figured I'd wait until the design was solidified to do that.
- The messages you see should be dependent on your role in the studio; it shouldn't tell you that you will be able to do something in X minutes if you don't usually have that permission.
- My current approach to deciding whether to show the mute messages is to use new selectors specifically for that purpose, named like "selectShowProjectMuteError." The downside of that approach is the logic is a bit redundant with other permissions; if we change `selectCanAddProjects` or `selectCanRemoveProjects` we also have to change `selectShowProjectMuteError`. An alternative/additional approach could be to add even more selectors that can serve as the building blocks as all of these (e.g. `selectUsuallyHasAddProjectPermission`), but I decided to start with the simpler way.
- Two tests (currently commented out) are failing and I am not sure why -- "muted creator" for both "should show curators mute error" and "should show edit mute error"


src/redux/studio-permissions.js
```diff
@@ -1,16 +1,18 @@
...
+    selectIsLoggedIn, selectUsername, selectIsMuted} = require('./session');
 
 // Fine-grain selector helpers - not exported, use the higher level selectors below
 const isCreator = state => selectUserId(state) === state.studio.owner;
 const isCurator = state => state.studio.curator;
 const isManager = state => state.studio.manager || isCreator(state);
 
 // Action-based permissions selectors
-const selectCanEditInfo = state => selectIsAdmin(state) || isManager(state);
+const selectCanEditInfo = state => !selectIsMuted(state) && (selectIsAdmin(state) || isManager(state));
```
apple502j, line 10: Just thought this; can ST members get muted (and become unable to edit info)? That would be funny

RedGuy12: I know paddle2see accidentally banned himself once, so I assume they can be muted too.

PR #5468, opened 2021-05-21
Studios stats and icons

cwillisf:
### Changes:

Add icons and stats to studio views:

- The navigation tabs (Projects / Comments / Curators / Activity) all have icons
   - The icon for the Curators tab was already in the repository
- The Projects and Comments tabs indicate how many projects or comments are attached the studio, respectively
   - The numbers and their placeholders are hidden until loading finishes ("Projects" becomes "Projects (42)")
   - If the number is 100 or higher it's displayed as "100+" since there might be more. The API currently stops counting at 100.
- The "last updated" date and the number of followers are shown below the studio description
   - Each field has an icon attached
   - These fields, including their icons, are hidden until studio info is loaded
- The studio report button has Scratch styling and an icon

### Notes:

I had a hard time fine-tuning the positions, sizes, and spacing of the text and icons, so I'm especially interested to hear if you have suggestions on how to make it better.

The "last updated" date is formatted by `FormattedMessage`'s built-in date formatting. I could override more details to force it to match the design more closely ("23 Mar 2021" vs. "March 23, 2021"), but I'm worried about the impact on other locales. For reference, in 'en-US' the "short" date format is "3/23/2021" and the "medium" format is "March 23, 2021".

### Demo

Here's a screenshot of this PR in action:

image

Here's the current design:

image


paulkaplan (approved):
This LGTM. Think there is still some final-pass type styling to do, but definitely good enough for now. I pushed a commit to allow the counts to be wrapped in their own span so they could be styled per the spec (not bold), but other than that this LG and i'll merge it.

PR #5476, opened 2021-05-24
Bump conference URL one year

rschamp:
Resolves #5197

To test:

- /conference redirects to /conference/2021
- /conference/2020 redirects to /conference/2021
- /conference/2021 shows the same content as /conference/2020 does currently
- /conference/2019 (and below) still works as it did previously

PR #5478, opened 2021-05-24
Fix integration test broken in #5476

rschamp:
### Resolves:

One of our integration tests depends on /conference/2020, so update to the new date


BryceLTaylor (approved):
LGTM!

PR #5497, opened 2021-05-26
Prepare release branch with new changes from develop

kchadha:
Pull in latest changes from develop to stage for release


apple502j:
@kchadha I'm sure we have the chance to translate Community Guidelines - right?

yuikunyeah:
when do translation?

kchadha:
@apple502j Thanks for raising this. Unfortunately with the nature of our development cycle, sometimes this can happen where we introduce text changes into develop shortly before a release. This is definitely something we can try to be more conscious of in the future and try to communicate and schedule big text updates like these a little earlier.

PR #5500, opened 2021-05-26
Add comments not allowed box

fsih:
Add the comments not allowed box, and adjust the spacing around the comment compose box.

Screen Shot 2021-05-26 at 14 46 27


src/views/studio/l10n.json
```diff
@@ -58,6 +58,7 @@
     "studio.promote": "Promote",
 
     "studio.commentsHeader": "Comments",
+    "studio.commentsNotAllowed": "Commenting for this studio has been turned off.",
```
Naleksuh, line 66: The message for profiles is:
> Sorry, comment posting has been turned off for this user profile. 
The message for projects is:
> Sorry, comment posting has been turned off for this project.
So the studio message here doesn't make sense. Should be something like "Sorry, comment posting has been turned off for this studio."

Naleksuh (changes requested):

seotts:
This looks good to me! I tested locally, and it seems to work. Naleksuh makes a good point about the wording, but I don't think that needs to be fixed before deploy this week; might be worth checking in with Kathy about it after it's up on scratch.ly

mxmou:
This pull request increased the spacing around the compose row in project comments. Was this intentional?

PR #5503, opened 2021-05-27
Add "Promote to manager" modal

ericrosenbaum:
Adds a confirmation modal for the action of promoting a curator to a manager.


Accio1:
For the manager permission strings, it might be a good idea to also say that managers can change the description and thumbnail. I think in the past, some people haven't realized this and it has led to studio vandalism.

hiPeeps124816:
Well I do like a modernized Scratch.

kchadha (approved):
LGTM

PR #5538, opened 2021-06-02
Fix l10n string ID where it previously got missed.

kchadha:
Updates a reference to an old l10n string id that we're no longer using.


seotts (approved):
lgtm!

PR #5539, opened 2021-06-02
Use reply_count to check whether reply limit is reached

kchadha:
### Resolves:

Resolves an issue found during the bug hunt:

> In studios-playground comments, can add more than 25 replies to a comment
> 
> Steps:
> add 25 replies to a top level comment
> reload page so that replies are hidden like this:
>  image2
> click reply on the top level comment, or on any reply
> you’ll see the comments form appear normally

### Changes:

Add `totalReplyCount` prop to TopLevelComment component, and pass it in from StudioComments. Use this new prop to check whether thread has reached limit instead of checking only the 20 we load on the first try.

### Test Coverage:

Follow the repro steps described above. In the last step, instead of seeing the comments compose box appear, you should see the reply status box saying that the limit is reached.

### DIFF
--- a/src/views/preview/comment/top-level-comment.jsx
+++ b/src/views/preview/comment/top-level-comment.jsx
@@ -103,13 +103,14 @@ class TopLevelComment extends React.Component {
             replies,
             postURI,
             threadHasReplyStatus,
+            totalReplyCount,
             visibility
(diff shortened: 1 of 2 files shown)


seotts (approved):
looks good!

PR #5541, opened 2021-06-02
Fix reply status box alignment.

kchadha:
Fixes the alignment of the reply status box which appears when the comment thread limit has been reached. The status box should align with the comment bubbles.

**After this change:**
image


seotts (approved):
lgtm!

derter-menter:
So this is happening? Despite the backlash…

the wording could also be confusing. “What limit” “why limit” etc

Maybe a “learn more” button to inform the user what happened?

PR #5568, opened 2021-06-08
Hide validation message on click outside 

ericrosenbaum:
Previously, if you encountered validation error messages on the studio title, thumbnail image or description, they would be stuck on. This change makes them hide when you click outside them. The "hide" state is reset to false on each change so that they can be shown again for later errors. 

I thought it might be simpler to put this hiding behavior into the ValidationMessage component itself, but I couldn't figure out how to give it that reset behavior.


paulkaplan (approved):
Yeah I see whats strange about this. Its a bit of an impedance match between function components + hooks vs. HOCs with class components. Since your approach is the one in the documentation, I think its ok. In the future I think we may want to look for or build a "hook" version of this.

PR #5577, opened 2021-06-09
Show errors when studio edit gets a muted response from backend

seotts:
This handles errors in studios from the backend when a user is muted. This only applies to situations where the mute status is not yet in the session, such as when you have multiple tabs open.

To test: 
- You need to be running the not-yet-merged [backend change](https://github.com/LLK/scratchr2/pull/6127) 
- Go to a studio (in studios-playground) that you own/manage/curate
- Open a new tab. Get muted in that tab.
- Return to the studio tab. _Do not refresh_.
- Try to take actions that you can normally take in the studio
- For most, you should see errors like this:
image
- Toggling comments and "anyone can add projects" should fail silently.

I also included some minor l10n message renaming in this change (like "studios.mutedCurators" -> "studio.mutedCurators", for consistency.)


paulkaplan (approved):
LGTM thanks for seeing those error issues

PR #5578, opened 2021-06-09
Make the default nav button state white

fsih:
Also remove that weird tiny bit of resizing when you click on them, was that intentional?
Screen Shot 2021-06-09 at 14 12 03


paulkaplan (approved):
LG

Issue #5486, opened 2021-05-25
Slashes disappear when searching


mxmou:
When searching for a string containing a forward slash, the slash and anything after it disappears. For example, entering "scratch/project" searches for just "scratch".

### Steps to Reproduce

1. Enter "scratch/project" or any other string containing a slash into the search bar
2. Press the enter key or click the search button

### Operating System and Browser

Edge 90 on Windows

AmazingMech2418:
Can reproduce. It seems like it's likely meant as a URL sanitation thing for the search API. 

If you URL encode the "/" twice, it works. https://scratch.mit.edu/search/projects?q=scratch%252Fprojects

@BryceLTaylor I think the issue is here:
https://github.com/LLK/scratch-www/blob/32dea2d2a823756882730042b23820718a684443/src/views/search/search.jsx#L62
Here, it runs `decodeURIComponent` before it does it again here:
https://github.com/LLK/scratch-www/blob/32dea2d2a823756882730042b23820718a684443/src/views/search/search.jsx#L84

Only the second one is needed in this case. 

However, since it decodes twice, and encodes it again when searching, only one `decodeURIComponent` is needed.

Here is where it encodes again:
https://github.com/LLK/scratch-www/blob/32dea2d2a823756882730042b23820718a684443/src/views/search/search.jsx#L99

AmazingMech2418:
I would be willing to submit a PR if that would help. It's a really simple fix, just removing the first `decodeURIComponent`.

mxmou:
> I would be willing to submit a PR if that would help. It's a really simple fix, just removing the first `decodeURIComponent`.

Does #5221 fix this?

AmazingMech2418:
> > I would be willing to submit a PR if that would help. It's a really simple fix, just removing the first `decodeURIComponent`.
> 
> Does #5221 fix this?

It does not seem to. The two `decodeURIComponent`s are still there.

benjiwheeler:
@AmazingMech2418 @mxmou I agree that 5221 doesn't fix the problem. @AmazingMech2418, I'd prefer to see a modification of 5221 that would solve the problem using that approach.

BryceLTaylor:
It looks like any work done to fix this issue needs to come along with a number of automated tests to make sure we're not breaking desired behavior.

AmazingMech2418:
@benjiwheeler I'm sorry, I think I must have missed your response somehow. I think 5221 and this can both be merged without conflicts.

PR #5579, opened 2021-06-09
Remove First `decodeURIComponent` In Search Page Query Parsing

AmazingMech2418:
### Resolves:

#5486 

### Changes:

This PR simply removes the first `decodeURIComponent` from the search page. The problem in #5486 was that the query was decoded twice instead of once, and when encoding the search terms back, it only encoded it once. Removing the first `decodeURIComponent` will make it so that it is decoded once and encoded once, to balance this out and prevent failures while searching.

### Test Coverage:

I have not tested this, but this is a fairly simple logic error.

**Note:** I know this issue is not labelled `help wanted`, but it is simple to fix and I did not receive a response when I said I was willing to submit a pull request, so I'm going ahead and doing so.

### DIFF
--- a/src/views/search/search.jsx
+++ b/src/views/search/search.jsx
@@ -59,7 +59,7 @@ class Search extends React.Component {
     componentDidMount () {
         // just in case there's a URL in the wild with pluses to indicate spaces,
         // convert pluses to url-encoded spaces before decoding.
-        const query = decodeURIComponent(window.location.search.split('+').join('%20'));
+        const query = window.location.search.split('+').join('%20');
         let term = query;
 
         const stripQueryValue = function (queryTerm) {


Shluffy:
You should maybe say `resolves #5486` so it will automatically link it. resolves #5486

AmazingMech2418:
@benjiwheeler This PR has been open for a bit of time now. If you could review it, that would be great! It's a pretty simple one-line fix, so reviewing it shouldn't take too long, I don't think.

benjiwheeler:
@AmazingMech2418 the problem is that there was some reason why we decoded twice -- and it'll take a bit more time and attention to get clear on why.

AmazingMech2418:
@benjiwheeler Looking at the file history, it seems like https://github.com/LLK/scratch-www/pull/2075 was where the change was made.

AmazingMech2418:
@benjiwheeler I think a potential solution to both of these issues could be to decode when setting the value of the search box.

The line that would change is this one:
https://github.com/LLK/scratch-www/blob/f6cb3d50783e3f6f34dde821bdfef7d6d3f6768b/src/components/navigation/www/navigation.jsx#L135

AmazingMech2418:
> @benjiwheeler I think a potential solution to both of these issues could be to decode when setting the value of the search box.
> 
> The line that would change is this one:
> https://github.com/LLK/scratch-www/blob/f6cb3d50783e3f6f34dde821bdfef7d6d3f6768b/src/components/navigation/www/navigation.jsx#L135

Actually, that change might not even be needed! Given that encoding twice still displays correctly, it should work still.

AmazingMech2418:
I'm testing this PR right now to see if that issue resurfaces. I don't think it should, but it's best to check at least!

AmazingMech2418:
@benjiwheeler The test passed! This bug was fixed and #1934 didn't resurface.

PR #5585, opened 2021-06-10
"Add to studio" modal empty states and done button

ericrosenbaum:
In the "add to studio" modal, add empty state styling, and a done button at the bottom


ericrosenbaum:
@paulkaplan I noticed that the empty state shows briefly before the project tiles load- let me know if you have suggestions for how to fix that or if we should address as a follow up

PR #5587, opened 2021-06-10
Studio comments off

picklesrus:
Adds in comment status component to show when studio comments are off. 
image


src/views/studio/studio-comments.jsx
```diff
@@ -74,54 +75,67 @@ const StudioComments = ({
                 <h2><FormattedMessage id="studio.commentsHeader" /></h2>
                 {canEditCommentsAllowed && <StudioCommentsAllowed />}
             </div>
```
picklesrus, line 77: This part of the diff is much easier to read if you turn whitespace changes off.

src/redux/session.js
```diff
@@ -142,6 +142,8 @@ module.exports.selectIsSocial = state => get(state, ['session', 'session', 'perm
 module.exports.selectIsEducator = state => get(state, ['session', 'session', 'permissions', 'educator'], false);
 module.exports.selectProjectCommentsGloballyEnabled = state =>
     get(state, ['session', 'session', 'flags', 'project_comments_enabled'], false);
+module.exports.selectStudioCommentsGloballyEnabled = state =>
```
paulkaplan, line 145: @picklesrus I think that this should probably be true by default right until we hear from the session? I've been trying to make sure the path to trying to load stuff is as parallel as possible, and this change would prevent loading the comments list until the user session has been fetched, which isn't necessary.

src/views/studio/studio-comments.jsx
```diff
@@ -74,54 +75,67 @@ const StudioComments = ({
...
-            <div>
-                {shouldShowCommentComposer ?
-                    (commentsAllowed ?
-                        <ComposeComment
+            {!hasFetchedSession || studioCommentsGloballyEnabled ?
```
paulkaplan, line 78: Why is the `!hasFetchedSession` here blocking the comments from rendering? possibly related to my other comment about not blocking comment loading with session request

paulkaplan (changes requested):
A couple of questions/requests for changes. I'm happy to talk about the questions IRL.

picklesrus:
This isn't finished (e.g.tests would need to be cleaned up), but I'm curious if you like this approach better

paulkaplan:
I approve of the direction. I like the unit test on studio-comments and probably would want a unit test on the permission selectors as well.

picklesrus:
Simplified the tests in studio-comments and moved them to studio-permissions. Glad they were there because they found a bug in the refactor. Fixed that and resolved some merge conflicts. Ready for another review!

paulkaplan (approved):
LGTM

paulkaplan:
Sorry about the continuing merge conflicts, there is a lot going on with studio comments right now. Should be settled now for at least a few days, but definitely recommend merging asap

picklesrus:
I added some test cases to handle the single comment mode. If they pass on Circle, I'm merging. :)

PR #5607, opened 2021-06-15
Make the border of editable fields dashed gray for studios. Otherwise, make them match the fields on project pages

fsih:
Per feedback from design pairing.

Before
descs-before

After
descs


ericrosenbaum (approved):
LGTM!

PR #5608, opened 2021-06-15
Adjust thumbnail css

fsih:
Make thumbnails 195 * 300 and stretch to fit to match old studios.
Thumbs have a 15% black 2px border.
Transparent thumbs now have a white background.

Before:
Screen Shot 2021-06-15 at 03 47 45

After:
Screen Shot 2021-06-15 at 03 49 25

### DIFF
--- a/src/views/studio/modals/studio-report-modal.scss
+++ b/src/views/studio/modals/studio-report-modal.scss
@@ -75,9 +75,12 @@
     }
 
     .studio-report-tile-image {
-        border-radius: 0.5rem;
-        max-width: 130px;
-        max-height: 100px;
+        width: 150px;
+        height: 98px;
+        border-radius: 4px;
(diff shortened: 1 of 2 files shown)


ericrosenbaum:
I think we we should make sure to adjust the studio report modal CSS in a corresponding way so that the thumbnail appears the same there (could be in a follow-up)

ericrosenbaum:
@fsih I made a few changes:
- use "border-box" to keep the width 300px and match the layout of other fixed width components above and below (otherwise, the thumbnail element was 304px wide in total including the border)
- removed "object-fit: cover;" (looks like that doesn't do anything?)
- updated the style for the studio report modal to match the style here but half the size

ericrosenbaum (approved):
LGTM but please review my added changes before merging thanks!

PR #5609, opened 2021-06-15
Adjust responsive boundaries for grid reflow

fsih:
1-across-grid is not needed since the min width is now 300.


ericrosenbaum (approved):
L
G
T
M
!

PR #5620, opened 2021-06-16
Admins can view deleted studios

paulkaplan:
This PR allows admins to view studios that have been deleted.

1. Add the `withAdmin()` helper to API requests for studio info, curators, managers, projects and activity feed. This will add the `/admin` prefix to the URL and the `authentication: token` property if the user is an admin. Comments already do this. 
2. Move the `getInfo` call up from the studio info component to the top level component, and make it rerun if `isAdmin` changes. This is because we show the `oops` page if you get a 404 for studio info, which prevents StudioInfo from re-rendering on isAdmin changes. 

One thing to note: in order to get data rendered as fast as possible, the studio info request does not wait for the session to be fetched first. This means studio info will always be fetched as a normal user first, and then re-fetched if the session comes back and the user is an admin. If you are viewing a deleted studio, this means you'll see the "oops" 404 page flash and then once the session comes back it'll re-render. 

You'll need the corresponding api PR pulled (or merged) before being able to use this locally.

### DIFF
--- a/src/lib/admin-requests.js
+++ b/src/lib/admin-requests.js
@@ -1,0 +1,23 @@
+import {selectIsAdmin, selectToken} from '../redux/session';
+
+/**
+ * Augment an `options` object that will be used by api.js
+ * to automatically include admin authentication token and
+ * /admin url prefix if the user is an admin.
+ *
+ * @param {object} opts Object argument for api.js request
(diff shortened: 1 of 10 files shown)


kchadha (approved):
LGTM

PR #5621, opened 2021-06-16
Studio creators and managers can delete their own comments

kchadha:
### Resolves:

Updates comment deletion permissions for studios to match existing studio behavior.

### Changes:

Studio creators and managers can delete their own comments on a studio.

### Test Coverage:

Updated the studio permissions test.

Tested locally including with teacher accounts.


paulkaplan (approved):
LGTM

PR #5626, opened 2021-06-17
Studio single comment view

ericrosenbaum:
Add the ability to load a view of the studio page showing a single comment. This happens when you receive a message notifying you of someone's reply to a comment that you made on a studio. The link in the message has a comment id in the URL. The page loads only the thread containing this comment, highlights the comment, and scrolls to it. A button lets you view the full set of comments. This should match the functionality of single comment view on the project page.

### DIFF
--- a/src/redux/studio-comment-actions.js
+++ b/src/redux/studio-comment-actions.js
@@ -3,7 +3,7 @@ const eachLimit = require('async/eachLimit');
 const api = require('../lib/api');
 const log = require('../lib/log');
 
-const COMMENT_LIMIT = 20;
+const COMMENT_LIMIT = 25;
 
 const {
     addNewComment,
--- a/src/redux/studio-comment-actions.js
+++ b/src/redux/studio-comment-actions.js
@@ -3,7 +3,7 @@ const eachLimit = require('async/eachLimit');
 const api = require('../lib/api');
 const log = require('../lib/log');
 
-const COMMENT_LIMIT = 25;
+const COMMENT_LIMIT = 20;
 
 const {
     addNewComment,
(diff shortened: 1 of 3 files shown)


kchadha:
If we're changing the comment thread pagination to be the same as the thread limit, I might go back and revert this change since it caused another bug we found in a bug hunt where the reply_count is not in sync with the front-end updates:
https://github.com/LLK/scratch-www/pull/5539

src/views/studio/studio-comments.jsx
```diff
@@ -41,11 +41,27 @@ const StudioComments = ({
...
+    handleLoadMoreReplies,
+    handleLoadSingleComment
 }) => {
+    const commentHashPrefix = '#comments-';
+    const [singleCommentId, setSingleCommentId] = useState(
+        window.location.hash.indexOf(commentHashPrefix) !== -1 &&
+        parseInt(window.location.hash.replace(commentHashPrefix, ''), 10));
+
     useEffect(() => {
+        if (singleCommentId) {
```
paulkaplan, line 53: I think this `if` should be nested underneath the `comments.length===0 && hasFetchedSession`. The reason being that we only want to fetch that single comment if it hasn't already been fetched. And we only want to load comments after we've `hasFetchedSession`.

src/redux/studio-comment-actions.js
```diff
@@ -3,7 +3,7 @@ const eachLimit = require('async/eachLimit');
 const api = require('../lib/api');
 const log = require('../lib/log');
 
-const COMMENT_LIMIT = 20;
+const COMMENT_LIMIT = 25;
```
paulkaplan, line 6: I'd probably leave this as 20 for now. If we were to change this, we'd need to also change https://github.com/LLK/scratch-www/blob/develop/src/redux/comments.js#L5 which also comes in to play i think, and then we have to change the project one also.

paulkaplan (changes requested):
A couple suggestions

paulkaplan (approved):
LGTM

PR #5627, opened 2021-06-17
Fix preventing muted users from adding projects via project page

seotts:
This change was previously reverted from the deploy because it sometimes crashed the project page by checking for permissions the studio in state-- that failed since we are on the project page, not the studios page.

The only difference between this and the previous PR is that it uses `selectIsMuted(state)` instead of `selectShowProjectMuteError(state)` to decide whether to prevent the user from adding projects.


paulkaplan (approved):
LG! I wish we were able to lint/statically protect against this type of issue...

Issue #5134, opened 2021-03-14
Do not fetch replies to comments w/o any replies


Ascor8522:
### Steps to Reproduce

1. Go on the [scratch website](https://scratch.mit.edu).
2. Go on a project with many comments. Any project featured on the home page will do.
3. Notice how the (first 20) comments are loaded. (from the API, at `https://api.scratch.mit.edu/users/$USER/projects/$PROJECT_ID/comments?offset=0&limit=20`)
Some comments might no have any replies to them. (`reply_count` proprety set to `0`)
4. However, the replies for those comments are still fetched from the API.
These requests are unnecessary and can slow down the browser, making it wait for dozen of requests to complete when not needed.

Only the replies of the comments with 1 or more replies should be fetched.

I believe this line is responsible for that.
https://github.com/LLK/scratch-www/blob/ecd4fa9438fd9ba0d2e1a5741964cbec17d63791/src/redux/project-comment-actions.js#L72

An easy fix would be to filter the array of comments just recieved, based on how many replies they have.
```js
dispatch(getReplies(id, body.filter(comment => comment.reply_count > 0).map(comment => comment.id), 0, ownerUsername, isAdmin, token)); 
```

### Operating System and Browser

Any

Naleksuh:
@BryceLTaylor , I see this is still an issue after 2-3 months. Would you mind if I fixed this?

AmazingMech2418:
It might be even more efficient to send the first 20 replies per thread with the original request to prevent excessive requests, and then send the rest through further requests when the user clicks to load more replies. It also would be possible to load all replies along with the original comments, but that might lead to excessive data per request as was mentioned in a comment in #5537 .

jeffalo:
Looks like this is being resolved by #5630

paulkaplan:
FWIW the reason for the individual requests is that they are more cacheable at the CDN layer and so create less of a backend server burden (which is a bit counterintuitive, since there are more requests).

PR #5630, opened 2021-06-18
Only request replies if reply_count > 0

paulkaplan:
Instead of always requesting a top-level-comments replies, now that we have `reply_count` in the comment response, use that to decide whether to try loading the replies. 

For both studios and project comments.

/cc @kchadha I think we talked about this a while back, but now seems like a good time to make a bit of an optimization. 
Also /cc @picklesrus @benjiwheeler if this has any overlap with your work, but I don't think it will. 

This resolves https://github.com/LLK/scratch-www/issues/5134

### DIFF
--- a/src/redux/project-comment-actions.js
+++ b/src/redux/project-comment-actions.js
@@ -69,8 +69,10 @@ const getTopLevelComments = (id, offset, ownerUsername, isAdmin, token) => (disp
         }
         dispatch(setFetchStatus('comments', Status.FETCHED));
         dispatch(setComments(body));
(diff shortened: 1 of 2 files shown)


Naleksuh (approved):
Looks good. Some people suggested having replies be in one request but I personally don't think that's necessary.

paulkaplan:
Turns out there is a bug with this, which we noticed in https://github.com/LLK/scratch-www/pull/5651. I'll mark this as blocked on that one.

kchadha (approved):
Changes look good. We can merge this after the PR it's blocked on goes in.

PR #5645, opened 2021-06-21
Fix admin panel space remaining after logging out

paulkaplan:
We were incorrectly calling the `storeAdminPanelOpen` function any time the studio page was rendered, even if the user wasn't an admin. This caused two issues:
1. Non-admin users got a value of `{adminPanelToggled_studios: 'closed'}` set in localstorage, even if they weren't an admin. Nobody really noticed this, but we shouldn't be storing things in localstorage that users don't need (even if they aren't technically "false", just unapplicable).
2. Admin users who had the admin panel open and then logged out would still see the space where the admin panel would go, because `storeAdminPanelOpen` would cause the CSS class to be applied. It wasn't checking if the admin panel should even be rendered at all.

I put in tests for both of these and fixed the issue with a one line change.

### DIFF
--- a/src/views/studio/studio-admin-panel.jsx
+++ b/src/views/studio/studio-admin-panel.jsx
@@ -48,8 +48,11 @@ const StudioAdminPanel = ({studioId, showAdminPanel}) => {
     const [adminPanelOpen, setAdminPanelOpen] = useState(getAdminPanelOpen());
 
     useEffect(() => {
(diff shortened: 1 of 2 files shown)


seotts (approved):
Looks good. I'm not sure this is a problem, but it seems like the value of 'adminPanelToggled_studios' remains in local storage even if the admin logs out, since it's only changed when the user is an admin. But it no longer effects the css, so probably that's okay?

PR #5647, opened 2021-06-21
Only show reply button when comments are allowed on a studio

paulkaplan:
Fixes an issue where comments still showed the `reply` button in studios even if the studio owner had turned off comments. This is inconsistent with the project page and would cause some confusion (even though trying to submit a comment _does_ return an appropriate message).


kchadha (approved):
LGTM

PR #5649, opened 2021-06-21
Fix permissions issues on studio comments

paulkaplan:
@kchadha this is a proposal for an alternative approach to what we discussed today. Instead of making the `canReport...` a function, make the whole `Comment` component pass-in-able, defaulted to the regular Comment component. For studios, make a custom Comment component that uses ownProps to check permissions, kind of like we talked about. 

Let me know what you think about this approach. It handles both the report and the delete things we did today, including unconfirmed users.


src/views/studio/studio-comments.jsx
```diff
@@ -127,12 +121,9 @@ const StudioComments = ({
                         <TopLevelComment
                             hasThreadLimit
                             author={comment.author}
-                            canDelete={canDeleteAnyComment ||
-                                (canDeleteOwnComment && comment.author.username === username)}
                             canDeleteWithoutConfirm={canDeleteCommentWithoutConfirm}
-                            canReply={shouldShowCommentComposer}
-                            canReport={canReportComment}
                             canRestore={canRestoreComment}
+                            commentComponent={StudioComment}
```
kchadha, line 126: So basically this prop is kind of like us passing in a constructor for how the TopLevelComments component should structure its reply comments?

test/unit/components/studio-comments.test.jsx
```diff
@@ -1,7 +1,13 @@
 import React from 'react';
 import {mountWithIntl} from '../../helpers/intl-helpers.jsx';
+
 import {StudioComments} from '../../../src/views/studio/studio-comments.jsx';
 
+// Replace customized studio comment with default comment to avoid redux issues in the test
```
kchadha, line 6: This seems a little dangerous to me if we end up having project comments and studio comments diverge in more ways.

src/views/studio/studio-comments.jsx
paulkaplan: Yeah I guess you could say that. Another way that would be a bit more explicit and require less knowledge of the internals of TopLevelComment would be to use a render prop, where the consumer of TopLevelComment would provide a child that was a function called with the props that would be passed to component. That way the TopLevelComment consumer could more explicitly customize which props (as opposed to right now where we are implicitly passing a set of props that can be overridden/intercepted).

kchadha (commented):
The changes here look okay to me.

I'm not sure how to weight this against our alternative. The benefit of this PR is that we're not making any changes to the project page TopLevelComments props. 

We could have probably done something similar in our original solution by passing in a default value for the `canDelete` and `canReport` props.

test/unit/components/studio-comments.test.jsx
paulkaplan: Yeah its not ideal. Speaking of unit testing difficulties, @picklesrus one thing that is tough is unit testing components that require the redux provider in the chain. There are a couple examples of using mock-store, but I had difficulty doing that because then i couldn't manually call `setProps` on the component (it needs to be called on the top level element, but wrapping it in a Provider makes it not the top level element). I could try instead mocking the StudioComment component and wrapping that with a provider. Alternatively, i could try the shallow render method that would only render `StudioComments` and not render all the children.

paulkaplan: I'd also like to look into how to move the logic that interacts with data fetching (useEffect, isAdmin state/ref) into a custom hook to encapsulate it and then test that independently of testing the component stuff like which things get rendered.

picklesrus: I haven't looked at the code, but could you use dive to get into the component you want and set its props?

paulkaplan: I took another look at the tests. It looks like I can avoid needing to figure out how to provide a redux store by using `shallowWithIntl` instead of `mountWithIntl` because it only renders on level deep, so the comment component (nested within TopLevelComment) never needs to be rendered. This is good because it wasn't part of any of these tests anyway, so its probably good to only render as much as you need to accomplish the tests.

I also took a step to disaggregating the callback logic tests from the render tests, by putting them in different places. The callback logic tests incidentally still need to use `mountWithIntl` because annoyingly even though `mount` and `shallow` try to provide a consistent component wrapper (ie you can call setProps and update with both), `shallow` will not run calllbacks in `useEffect`.

paulkaplan: @picklesrus unfortunately i can't call `setProps` on anything that isn't the top level element, so although i can get a reference to the component using `dive` or `find`, it isn't a drop in fix because i can't call `setProps` on it...

PR #5651, opened 2021-06-22
Update the reply_count when you leave a reply

paulkaplan:
This resolves issues where things that rely on a `comment.reply_count` won't work correctly after the user has left one or more replies to a thread.

Added a couple tests for this, which also uncovered an issue where calling `addNewComment` before initializing the reply array `replies[id] = [];` would cause an error. This currently doesn't happens because we always call `setReplies()` on every top level comment, but if we stop calling setReplies, e.g. in https://github.com/LLK/scratch-www/pull/5630, replying to a top level comment that didn't have replies already wouldn't work. Glad we caught this, since it didn't come up in that PR.

### DIFF
--- a/src/redux/comments.js
+++ b/src/redux/comments.js
@@ -70,8 +70,9 @@ module.exports.commentsReducer = (state, action) => {
             return Object.assign({}, state, {
                 replies: Object.assign({}, state.replies, {
                     // Replies to comments go at the end  of the thread
-                    [action.topLevelCommentId]: (
-                        state.replies[action.topLevelCommentId] || []).concat(action.comment)
+                    [action.topLevelCommentId]:
+                        (state.replies[action.topLevelCommentId] || [])
+                            .concat(action.comment)
                 }),
                 comments: state.comments.map(comment => {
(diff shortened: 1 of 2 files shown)


src/redux/comments.js
```diff
@@ -70,7 +70,16 @@ module.exports.commentsReducer = (state, action) => {
             return Object.assign({}, state, {
                 replies: Object.assign({}, state.replies, {
                     // Replies to comments go at the end  of the thread
-                    [action.topLevelCommentId]: state.replies[action.topLevelCommentId].concat(action.comment)
+                    [action.topLevelCommentId]: (
+                        state.replies[action.topLevelCommentId] || []).concat(action.comment)
```
kchadha, line 74: I feel like the whole parenthesized expression should be on its own line for readability (assuming our linter allows that)
```suggestion
                    [action.topLevelCommentId]:
                        (state.replies[action.topLevelCommentId] || [])
                            .concat(action.comment)
```

kchadha (approved):
LGTM -- have a minor code styling suggestion, but it's ok to merge without incorporating it

PR #5653, opened 2021-06-22
Studio tab nav button styling

fsih:
- Increase the padding in and around tab nav buttons
- Change the color of the hover states

### DIFF
--- a/src/views/studio/studio.scss
+++ b/src/views/studio/studio.scss
@@ -190,15 +190,16 @@ $radius: 8px;
     border-bottom: 1px solid $active-dark-gray;
     padding-bottom: 8px;
     font-size: 14px;
+    margin: 8px;
     li {
         display: flex;
         align-items: center;
         background: white;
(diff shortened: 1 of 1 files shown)


src/views/studio/studio.scss
```diff
@@ -166,7 +166,7 @@ $radius: 8px;
         border-radius: 8px;
         background: white;
         box-sizing: border-box;
-        border: 2px solid rgba(0, 0, 0, 0.15);
+        border: 2px solid box-shadow-light-gray;
```
ericrosenbaum, line 169: looks like a typo here

ericrosenbaum (changes requested):
LGTM aside from the one type

ericrosenbaum:
I'm wondering about the "add to studio" modal, which also uses nav buttons. I assume those should be updated to match this change?

src/views/studio/studio.scss
fsih: Oops thank you!

PR #5658, opened 2021-06-23
Match Add Project placeholder & button text to old studios

seotts:
Changes the placeholder text and button text for adding projects by url, to match the text in existing studios.

image

### DIFF
--- a/src/views/studio/l10n.json
+++ b/src/views/studio/l10n.json
@@ -29,8 +29,7 @@
 
     "studio.projectsHeader": "Projects",
     "studio.addProjectsHeader": "Add Projects",
-    "studio.addProject": "Add",
-    "studio.addProjectPlaceholder": "Project URL",
+    "studio.addProject": "Add by URL",
 
     "studio.openToAll": "Anyone can add projects",
 
--- a/src/views/studio/studio-project-adder.jsx
+++ b/src/views/studio/studio-project-adder.jsx
@@ -3,7 +3,7 @@ import React, {useState} from 'react';
 import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
 import classNames from 'classnames';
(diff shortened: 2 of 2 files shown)


src/views/studio/studio-project-adder.jsx
```diff
@@ -67,7 +67,7 @@ const StudioProjectAdder = ({intl, onSubmit}) => {
                     className={classNames({'mod-form-error': error})}
                     disabled={submitting}
                     type="text"
-                    placeholder={intl.formatMessage({id: 'studio.addProjectPlaceholder'})}
+                    placeholder="http://scratch.mit.edu/projects/xxxx"
```
ericrosenbaum, line 70: maybe this detail doesn't matter since it's just a placeholder, but looks like our URLs generally use `https`

src/views/studio/l10n.json
```diff
@@ -29,8 +29,7 @@
 
     "studio.projectsHeader": "Projects",
     "studio.addProjectsHeader": "Add Projects",
-    "studio.addProject": "Add",
-    "studio.addProjectPlaceholder": "Project URL",
+    "studio.addProject": "Add by URL",
```
ericrosenbaum, line 32: I'm not sure if we need to create a new id when updating a string that has already been merged in- investigating...

src/views/studio/studio-project-adder.jsx
seotts: Good catch! Just fixed that.

src/views/studio/l10n.json
ericrosenbaum: turns out this is the correct way to do it (keep the same id and update the string).

ericrosenbaum (approved):
LGTM

src/views/studio/l10n.json
seotts: thanks for investigating!

PR #5660, opened 2021-06-23
Revert "Only request replies if reply_count > 0"

kchadha:
Reverts LLK/scratch-www#5630 due to a bug found in the bug hunt where the single comment view only shows the top-level comment for the reply that is linked, as opposed to the reply itself.


kchadha:
@paulkaplan FYI

PR #5664, opened 2021-06-24
Reduce reply requests, for real this time I promise

paulkaplan:
Revert #5660 which reverted #5630, so actually apply #5630. There was no bug with the code in this PR, it was just incompatible with our backend which was incorrectly telling the frontend that every comment loaded by id had a reply_count of 0, which caused notifications linking to project replies to not actually load the replies. 

This PR will reduce the number of requests for replies proportionally with how many comments have 0 replies, which is a win for the client side and the server side. 👑

This PR should has to be deployed _after_ the backend fix is in. 

/cc @kchadha lets keep this revert chain alive


paulkaplan:
@kchadha going to re-merge this for testing again

Issue #2729, opened 2019-01-29
main view element min height creates unwanted spacing


benjiwheeler:
To repro:
1. view a project page for someone else's project that has commenting turned off, and no comments
2. observe space at the bottom:

![image](https://user-images.githubusercontent.com/3431616/51924722-92303480-23bb-11e9-9fb1-4532e8ab9729.png)

This is because the `#view` id in `main.scss` has `min-height: 768px`. Changing this to `min-height: 680px` fixes this problem.

Related: https://github.com/LLK/scratch-www/issues/1828

R4356th:
@benjiwheeler, the space seems to have significantly decreased; is this issue still valid?

benjiwheeler:
More about this: @carljbowman and I discussed this, and we noted that scratchr2 seems to handle this better; compare the scratchr2 project 404 page. scratch.mit.edu/projects/INVALID, when you zoom out and make the page tall, the footer sticks to the bottom of the page:

![image](https://user-images.githubusercontent.com/3431616/95115823-12bc4200-0714-11eb-8164-5159ff2aca2c.png)

![image](https://user-images.githubusercontent.com/3431616/95115825-13ed6f00-0714-11eb-9f9f-401e9ed6df14.png)

 Vs. on the www page, scratch.mit.edu/projects/9999999999999999 , where the footer positions according to the `view`'s `min-height`:

![image](https://user-images.githubusercontent.com/3431616/95115886-2d8eb680-0714-11eb-9bd0-ee8654257062.png)

![image](https://user-images.githubusercontent.com/3431616/95115912-397a7880-0714-11eb-8ebb-643b307aa270.png)

benjiwheeler:
We merged a narrow fix to this case, but it could make sense to revisit the way we float the footer in the future to make it stick to the bottom of the window more aggressively.

PR #5665, opened 2021-06-24
Fix Scroll Bar on Project Fullscreen

### LINKED ISSUE #4190: In project page fullscreen mode, you can scroll the page behind the project
From the project page, if you go to full screen mode, and scroll the page (e.g. using a two finger scrolling gesture on a trackpad), the page behind the player scrolls invisibly behind. Then, when you exit full screen mode, you will be looking at a lower portion of the page, perhaps unexpectedly. You can do the scrolling gesture anywhere on the page (i.e. including on the project player). This likely has a bigger effect on projects with lots of comments, so they have long pages.

fullscreen-scroll-bug

AmazingMech2418:
### Resolves:

Fixes #4190 

### Changes:

This adds a `hidden` class with `display: none`, applies this class to the footer and comment section on fullscreen, and modifies the CSS to remove `min-height` for `#view`, along with some fixes to prevent #2729 from resurfacing.

Hiding the footer and comment section fix #4190 when added with the removal of the `min-height` that was set to 680 pixels. Other updates were made to fix the lack of padding on the 404 page, at least temporarily. 

There are issues where you still have to scroll to get to the bottom of the footer on tall pages and where there is still a padding issues on extremely small pages. I am not entirely sure, but I think these could possibly be fixed by CSS media queries.

### Test Coverage:

I tested on a local build of my fork of scratch-www and the bugs of #4190 were fixed and the only remaining bugs relating to #2729 were those mentioned (see comments in #4190 for screenshots).

### DIFF
--- a/src/components/footer/www/footer.jsx
+++ b/src/components/footer/www/footer.jsx
@@ -15,7 +15,7 @@ const getScratchWikiLink = require('../../../lib/scratch-wiki');
 require('./footer.scss');
 
 const Footer = props => (
-    <FooterBox hidden={props.isFullScreen}>
+    <FooterBox hidden={props.projectIsFullScreen}>
         <MediaQuery maxWidth={frameless.mobileIntermediate - 1}>
             <div className="lists">
                 <dl>
@@ -220,7 +220,7 @@ const Footer = props => (
 Footer.propTypes = {
     intl: intlShape.isRequired,
     scratchWikiLink: PropTypes.string,
-    isFullScreen: PropTypes.bool
+    projectIsFullScreen: PropTypes.bool
 };
 
 const mapStateToProps = (state, ownProps) => {
@@ -228,7 +228,7 @@ const mapStateToProps = (state, ownProps) => {
         scratchWikiLink: getScratchWikiLink(ownProps.intl.locale)
     };
     if (state.scratchGui) {
-        result.isFullScreen = state.scratchGui.mode.isFullScreen;
+        result.projectIsFullScreen = state.scratchGui.mode.isFullScreen;
     }
     return result;
 };
--- a/src/components/footer/www/footer.jsx
+++ b/src/components/footer/www/footer.jsx
@@ -224,13 +224,13 @@ Footer.propTypes = {
 };
 
 const mapStateToProps = (state, ownProps) => {
-    const result = {
+    const propResults = {
         scratchWikiLink: getScratchWikiLink(ownProps.intl.locale)
     };
     if (state.scratchGui) {
-        result.projectIsFullScreen = state.scratchGui.mode.isFullScreen;
+        propResults.projectIsFullScreen = state.scratchGui.mode.isFullScreen;
     }
-    return result;
+    return propResults;
 };
 
 const ConnectedFooter = connect(mapStateToProps)(Footer);
--- a/src/components/footer/www/footer.jsx
+++ b/src/components/footer/www/footer.jsx
@@ -15,7 +15,7 @@ const getScratchWikiLink = require('../../../lib/scratch-wiki');
 require('./footer.scss');
 
 const Footer = props => (
-    <FooterBox>
+    <FooterBox hidden={props.isFullScreen}>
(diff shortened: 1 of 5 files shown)


AmazingMech2418:
I just noticed a side-effect with studio pages. I'll get to work on that and see if I can fix it right now.

AmazingMech2418:
> I just noticed a side-effect with studio pages. I'll get to work on that and see if I can fix it right now.

Moving the CSS from `main.scss` to `preview.scss` seems to fix this.

mxmou:
Is it not possible to add `overflow: hidden` to the `body`?

AmazingMech2418:
> Is it not possible to add `overflow: hidden` to the `body`?

Maybe, but it would have to interface with the actual DOM, which I don't think is really wanted since this uses React. 

You'd have to use something like this:
```js
document.body.style.overflow = "hidden";
```

Another issue would be in reading when the fullscreen button is first clicked to toggle this, since the button is in scratch-gui. It might be possible to add this to the function prior to rendering, but I don't think that is really wanted either, and might have some performance issues. 

You would need something like this in the rendering function:
```js
if(isFullScreen) {
    document.body.style.overflow = "hidden";
} else {
    document.body.style.overflow = "";
}
```

Personally, I think keeping this solution pure React and Sass would be better, but this could potentially be an alternative solution

mxmou:
> > Is it not possible to add `overflow: hidden` to the `body`?
> 
> Maybe, but it would have to interface with the actual DOM, which I don't think is really wanted since this uses React.

`react-helmet` is already used in some places in this repo and it lets you change the body element declaratively from within a render function.

src/components/footer/www/footer.jsx
```diff
@@ -219,12 +219,19 @@ const Footer = props => (
...
-    scratchWikiLink: PropTypes.string
+    scratchWikiLink: PropTypes.string,
+    isFullScreen: PropTypes.bool
 };
 
-const mapStateToProps = (state, ownProps) => ({
-    scratchWikiLink: getScratchWikiLink(ownProps.intl.locale)
-});
+const mapStateToProps = (state, ownProps) => {
+    const result = {
```
seotts, line 227: Could you give this variable a more descriptive name? Maybe "props" or "propResults"?

src/components/footer/www/footer.jsx
```diff
@@ -219,12 +219,19 @@ const Footer = props => (
 
 Footer.propTypes = {
     intl: intlShape.isRequired,
-    scratchWikiLink: PropTypes.string
+    scratchWikiLink: PropTypes.string,
+    isFullScreen: PropTypes.bool
```
seotts, line 223: I wonder if a name like "projectIsFullScreen" would make more sense here, since the footer itself is not full screen

src/views/preview/preview.scss
```diff
@@ -671,3 +671,20 @@ $stage-width: 480px;
...
+}
+
+html, body, #app, .page {
+    height: 100%;
+}
+
+.footer {
+    position: absolute;
+    bottom: 0;
+}
```
seotts, line 686: I'm not sure I fully understand the purpose of these changes, or why they are in this file. Could you say more? It seems to work for me with these commented out, keeping only the change in lines 688-690.

seotts: Gotcha. I'm not seeing any issues with the 404 page when I remove those lines. Could you test it locally and confirm that?

seotts:
After taking a look at this with Ray, we'd rather a simpler solution that uses react-helmet, like @mxmou suggested. @AmazingMech2418, would you be willing to make that change?

AmazingMech2418:
@seotts I think I could! I'm a bit busy right now, so it might be a few weeks, but it seems simple enough!

benjiwheeler:
@AmazingMech2418 Just checking in! If you are free to make this change using `react-helmet`, I'd be happy to review it! Thanks for your help!

AmazingMech2418:
@benjiwheeler I'm still pretty busy, mostly with school, but I think I should be able to work on a fix with react-helmet in just a few weeks!

PR #5671, opened 2021-06-25
Studio page integration tests

BryceLTaylor:
### Resolves:

Some integration tests that link to the studio page will break when we switch to the new studio page.  Also there are no tests for the new studio page.

### Changes:

Updates some of the old tests that pointed to the old studios page.  Also, Adds a few integration tests for the new studios page.
Also updates Chromedriver.


paulkaplan (approved):
LGTM

PR #5692, opened 2021-07-06
Update manager permissions for studios.

kchadha:
### Changes:

Update the permissions for managers on studios. Only the studio creator can edit the title/description/thumbnail of the studio.

### Test Coverage:

Tested locally.

CC/ @paulkaplan


apple502j:
@kchadha I'm sure this will be noted on community posts?

Also remember that most studios are maintained by the managers, because the owner is gone. This will make maintaining such studios much, much harder. Please don't drop such a big change hours before the release. PLEASE DO NOT.

kchadha:
Here's the forum post: https://scratch.mit.edu/discuss/topic/526539/

PR #5697, opened 2021-07-06
Load full rows for projects, curators, managers

paulkaplan:
Fixes an issue where we loaded 20 per page, but we are showing rows of 3, so even when there was more to load the row would appear incomplete. I changed the max load to 24, so it will be divisible into 1,2,3 (or even 4). Aren't twelves grand?

Also change the markup/css to make the load more buttons full width, per design spec


kchadha (approved):
LGTM

PR #5711, opened 2021-07-09
check rate limit for studio creation integration test

BryceLTaylor:
### Resolves:

Ensures that the integration tests don't run into the rate limit for studio creation when run in CI.

### Changes:

adds a check to the test that creates a studio from the my-stuff page.


seotts (approved):
lgtm

PR #5719, opened 2021-07-12
Support dynamic imports to load code only when needed

paulkaplan:
This PR makes the required changes to our build system to support being able to load code asynchronously using `import('chunk').then( => `. This also allows us to use `React.lazy` and `React.Suspense` to support async react component loading.  There are some minor path resolution issues that are needed to make async bundle loading work with webpack (see webpack config and template.ejs)

The second commit includes a demo of how to use react.lazy/Suspense. I used the annual report plotly world-map, which is a multi megabyte file that is only needed if the user scrolls down far enough. 

----

Some notes about async code loading:
- The preferred way is to use `import('bundle')`, which is distinct from `import X from 'bundle'` because it is a real function that returns a promise. 
- Webpack allows you to use the `require.ensure(...)` syntax as well, if for some reason you  are stuck using requires
- React.lazy relies on the promise nature of `import('...')`. It only works when modules are exported using the es6 module syntax, in particular only works on the default export. It won't work with `require.ensure`
- Suspense renders a fallback react element while loading. The only thing to keep in mind is that Suspense does not currently work with server-side rendering. Not an issue for us right now.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,8 @@
 {
     "plugins": [
         "transform-object-rest-spread",
-        "transform-require-context"
+        "transform-require-context",
+        "syntax-dynamic-import"
     ],
     "presets": ["es2015", "react"],
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -3370,6 +3370,12 @@
         }
       }
     },
+    "babel-plugin-syntax-dynamic-import": {
+      "version": "6.18.0",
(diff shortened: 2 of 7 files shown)


rschamp (approved):
We've reverted the World Map component demo, but left the commit in place so that we can merge it separately. @chrisgarrity and I are going to work off of the dynamic import commit for some intl work.

PR #5720, opened 2021-07-12
Extract css from js into stylesheets

paulkaplan:
use the mini-css-extract-plugin to extract our styles into stylesheets and include them in the template.ejs file. 

This reduces the size of our JS and allows the browser to parallelize the loading of scripts/stylesheets. 

One thing i'm a little uncertain about is that the `htmlWebpackPlugin` data structure has css files as an array, but it appears that there is only one, named the chunk name. I'm not sure if there will _always_ be only one, or what... I think I need to do a bit more research on that.

I also included a change to the deploy process to make sure the stylesheets are uploaded with the correct mime type.


paulkaplan:
I looked more into whether we can rely on their only being a single css file. The [documentation](https://v4.webpack.js.org/plugins/mini-css-extract-plugin/) says "It creates a CSS file per JS file which contains CSS." Since our SplitChunks config will always generate only 2 initial chunks (common + view), and that won't change out from under us because we disabled auto-inject and manually require those two js chunks, I believe it is also safe to rely on their being 2 initial css files to require.

I also tested out what happens with async imports. When webpack creates the async js chunks (at build time) it also generates async css chunks corresponding with those. The async import of the JS will trigger an async loading of the css file. The plugins default way of doing this is with `document.head.appendChild(linkTag);`, which seems fine for our purposes as well.

benjiwheeler:
Filed issue for this, and added some discussion on this: https://github.com/LLK/scratch-gui/issues/7368

rschamp (approved):

PR #5722, opened 2021-07-13
test(studios): fix failing permissions tests for studios

paulkaplan:
@seotts fixed that issue where certain "creator" tests were failing mysteriously. There was a test that mutated global fixture data, causing future tests to fail. Instead of mutating, I made it make a copy and modify that using object spread syntax.

This uncovered a test that was incorrectly passing, that stated that creators couldn't add curators, which isn't true. Fixed that.


seotts (approved):
Thanks for sorting this out! Looks good.

PR #5743, opened 2021-07-19
Add studio rate limit integration test

BryceLTaylor:
### Changes:

Adds an integration test to test the rate limiting of studio creation.
It also rearranges the tests slightly and removes a wait that is no longer necessary.


seotts (approved):
I'm not super familiar with www integration tests, but the logic of it looks good to me.

Issue #5487, opened 2021-05-25
Clicking the search icon without searching anything shows an empty page


mxmou:
Clicking the magnifying glass icon without enter anything into the search bar shows an empty page (https://scratch.mit.edu/search/projects) instead of just focusing the search bar.

### Steps to Reproduce

Click the icon in the search bar without entering anything into it.

### Operating System and Browser

Edge 90 on Windows

AmazingMech2418:
Is this even really a bug? I mean, the button is meant to search for what you have typed in and if you have nothing typed in, it searches for nothing...

Shluffy:
> Is this even really a bug? I mean, the button is meant to search for what you have typed in and if you have nothing typed in, it searches for nothing...

I think it still should filter that

mxmou:
@AmazingMech2418 The behavior is confusing. One might thing that the icon needs to be clicked to search.

carljbowman:
There are two awkward / confusing behaviors which both stem from not validating if any text has been entered into the text field prior to initiating the search.

When the search field contains no text and...
1. ... clicking the search icon (magnifying glass) results in loading a search page with not results
2. ... pressing the `Return` key to initiate a search while the search field is focused  results in loading a search page with not results

In both cases we should not initiate a search unless there is text in the field.

benjiwheeler:
Sounds like the Scratch Team's intended design is (as @carljbowman said) for the search form to not submit at all (and the icon not to interpret clicks as a submit), if the search box is empty (or only whitespace).

jeffalo:
I can do this one.

jeffalo:
@benjiwheeler I made a PR: #5884.

PR #5884, opened 2021-08-05
Prevent searches for empty strings

jeffalo:
### Resolves:

Resolves: #5487

### Changes:

When the search form is submitted, don't redirect when the `q` field is an empty string. This prevents clicking the icon/pressing enter from redirecting to the blank results page.

### Test Coverage:

- Pressing the icon with something in the field correctly searches for the item
- Pressing enter with something in the field correctly searches for the item
- Pressing the icon with an empty query field does not do anything
- Pressing enter with an empty query field does not do anything

### DIFF
--- a/src/components/navigation/www/navigation.jsx
+++ b/src/components/navigation/www/navigation.jsx
@@ -84,7 +84,7 @@ class Navigation extends React.Component {
     }
 
     handleSearchSubmit (formData) {
-        if(formData.q.trim() === '') return // don't submit empty searches
+        if (formData.q.trim() === '') return; // don't submit empty searches
         
         let targetUrl = '/search/projects';
         if (formData.q) {
--- a/src/components/navigation/www/navigation.jsx
+++ b/src/components/navigation/www/navigation.jsx
@@ -84,6 +84,8 @@ class Navigation extends React.Component {
     }
 
     handleSearchSubmit (formData) {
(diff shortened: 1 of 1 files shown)


AmazingMech2418:
I think this looks good! Though, I think the error might be just a missing semicolon and a space after `if`. I'm not sure if that's it, but I know those are parts of the style from past errors I've experienced, so it might as well be worth it to fix it.

jeffalo:
> I think this looks good! Though, I think the error might be just a missing semicolon and a space after `if`. I'm not sure if that's it, but I know those are parts of the style from past errors I've experienced, so it might as well be worth it to fix it.

Thanks, dbf68ff9b4cc5518623a71030ebcd85ca14c8eb6 should fix those eslint errors

AmazingMech2418:
Let me see if I can locally build your fork to see if I can find the issue now.

AmazingMech2418:
The local builds aren't working. The only thing I can think of is possibly the void return. I'm not sure though. Returning null should work though, I think.

jeffalo:
I think the error is out of my control, it says it failed `build-staging` for "Unauthorized".

image

AmazingMech2418:
Yeah, I see that now.

FunctionalMetatable:
@jeffalo you should link this PR to the original issue (just add "resolves #5487" or something)

Issue #5693, opened 2021-07-06
Ownership transfer for studios


apple502j:
There should be a way to transfer the ownership of a studio in case someone leaves Scratch or is simply unable to be online. This is even more important because Scratch Team made a quick (and imo silly) decision to require owner status to edit studio titles or descriptions.

Alternative: studios could be owned by multiple people. Not sure how that works when it comes to deleting, tho.

leahcimto:
With the having multiple owners that sort of relates with having more ranks above manager

mxmou:
> This is even more important because Scratch Team made a quick (and imo silly) decision to require owner status to edit studio titles or descriptions.

I understand why this was done, but the manager limit and the confirmation when promoting a curator are more than enough to achieve the same goal.

miles854:
What about studios that have a need for multiple Scratchers to be able to edit the studios description and title, as well as sometimes a studios thumbnail? Also what about if a the studios owner has gone inactive and not transferred ownership to another Scratcher? This will cause lots of studios to not be updated, and puts a lot more work on the studio owner. Lots of studios used managers to help run a studio, and not being able to edit a studios description and title was a big part of why many of them needed multiple studio managers.

BryceLTaylor:
This feature has been implemented now, thanks!

PR #5920, opened 2021-08-11
Studio host transfer modal (work in progress)

ericrosenbaum:
This PR adds most of the front-end code for a feature which renames the "studio creator" role to "studio host", and allows the current studio host to transfer that host role to another studio manager. 

The feature is behind a feature flag, so it will not be available yet. No changes should be visible.

image

In addition to the backend support, there are some remaining changes to be made here before it is ready for release, including:
- Permissions changes
- More error alerts
- Studio and user notifications
- Style fixes


Naleksuh:
@ericrosenbaum Question, if this is intended to be a replacement for studio managers having permissions to edit things,  are you planning to remove "managers" entirely? Or will permissions for managers be restored at any point?

RedGuy12:
Instead of having the "change studio host" option in the owner's 3-dot menu and then choosing a host on the next screen, why not make an option in **other** manager's 3-dot menu "make host"?

LegoManiac04:
> Instead of having the "change studio host" option in the owner's 3-dot menu and then choosing a host on the next screen, why not make an option in **other** manager's 3-dot menu "make host"?

I think having the option on the owner helps emphasize the point that you, the owner, are handing permissions over to someone else. While there are the confirmation dialogs, I feel it's a lot more clear that you are no longer going to be the owner.

carljbowman:
> Instead of having the "change studio host" option in the owner's 3-dot menu and then choosing a host on the next screen, why not make an option in **other** manager's 3-dot menu "make host"?

@RedGuy12 - When making this design decision, we considered a few different options for integrating this feature and looked at an option like what you described. There is certainly something cleaner and possibly simpler about what you described. That said, given that the transfer host permissions is to someone else a larger change and decision to make, we wanted to distinguish it from other similar permission changing features e.g. promote to manager.

Similar to what @LegoManiac04 said:
> I think having the option on the owner helps emphasize the point that you, the owner, are handing permissions over to someone else. While there are the confirmation dialogs, I feel it's a lot more clear that you are no longer going to be the owner.

Dhuls-Scratch:
Shouldn't this be linked with #5693?
Also, why did you choose "studio host" over "studio owner"?

RedGuy12:
> Also, why did you choose "studio host" over "studio owner"?

probably because the host isn't always the person who created it anymore

carljbowman:
> > Also, why did you choose "studio host" over "studio owner"?
> 
> probably because the host isn't always the person who created it anymore

@Dhuls-Scratch - Yup, @RedGuy12 is correct. We wanted to choose a new name for this role that was less specific to the act of creating a studio initially and a name that is more conducive to being transferred to other Scratchers. 

Additionally, there is already some inconsistency in the name of this role: in some places it appears as "Owner" and in other places it appears as "Creator." That is something else we look to address with some of the changes to this feature.

src/views/studio/lib/studio-member-actions.js
```diff
@@ -187,6 +187,26 @@ const acceptInvitation = () => ((dispatch, getState) => new Promise((resolve, re
...
 }));
 
+const transferHost = (password, newHostName, newHostId) =>
+    ((dispatch, getState) => new Promise((resolve, reject) => {
+        const state = getState();
+        const studioId = selectStudioId(state);
+        const token = selectToken(state);
+        newHostName = newHostName.trim();
+        api({
+            uri: `/studios/${studioId}/transfer/${newHostName}?password=${password}`,
```
towerofnix, line 197: Is it always safe to embed a password right into the query parameters of a (PUT) request like this? If any part of the backend has a logging system which tracks requests by URL including non-GET, this code could inadvertently put users' passwords into a plain-text file accessible by any system admins—and possibly malicious hackers!

(This includes e.g. logging files created out of even sysadmins' control by the server providers. A mass breach of logs on those systems would leak Scratchers' passwords if PUT requests have their URLs logged.)

Disclaimer: this comment basically stolen from a comment by Chiroyce on codubee's profile, linking this line of code :P

jeffalo: Is there any reason the password can't be inside the body of the request? Using query parameters for sending data in a request is also a little weird.

src/views/studio/lib/studio-member-actions.js
```diff
@@ -187,6 +187,26 @@ const acceptInvitation = () => ((dispatch, getState) => new Promise((resolve, re
...
+            method: 'PUT',
+            authentication: token,
+            withCredentials: true,
+            useCsrf: true
+        }, (err, body, res) => {
+            const error = normalizeError(err, body, res);
+            if (error) return reject(error);
+            dispatch(setInfo({owner: newHostId}));
+            return resolve();
+        });
```
FunctionalMetatable, line 207: Does this mean that the old host will still remain as a manager?

derter-menter:
I think some more context on "you will not be able to edit the studio" would be nice to make it clearer, even if it makes the message longer and less to the point. Maybe "You will no longer be able to edit the title, thumbnail and description" instead?

src/views/studio/lib/studio-member-actions.js
colbygk: I agree, it would be better to put the password in the body of the request.

colbygk: and is already in the process of being changed

src/redux/studio-permissions.js
```diff
@@ -54,6 +54,17 @@ const selectCanRemoveManager = (state, managerId) =>
     !selectIsMuted(state) && (selectIsAdmin(state) || isManager(state)) && managerId !== state.studio.owner;
 const selectCanPromoteCurators = state => !selectIsMuted(state) && isManager(state);
 
+const selectCanTransfer = (state, managerId) => {
+    if (state.studio.classroomId !== null) return false; // Nobody can transfer class studios
```
seotts, line 58: Let's clarify this comment further, since classroomId is only set in some cases.

src/redux/studio-permissions.js
```diff
@@ -54,6 +54,17 @@ const selectCanRemoveManager = (state, managerId) =>
...
+    if (state.studio.managers > 1) { // If there is more than one manager,
+        if (managerId === state.studio.owner) { // and the selected manager is the current owner/host,
+            if (isCreator(state)) return true; // Owner/host can transfer
```
seotts, line 61: Maybe we should change `isCreator` to `isHost`?

src/views/studio/l10n.json
```diff
@@ -88,10 +89,26 @@
     "studio.managerThresholdInfo": "This studio has {numberOfManagers} managers. Studios can have a maximum of {managerLimit} managers.",
     "studio.managerThresholdRemoveManagers": "Before you can add another manager, you will need to remove managers until there are fewer than {managerLimit}.",
 
+    "studio.transfer.youAreAboutTo": "You are about to make someone else the studio host.",
+    "studio.transfer.cannotUndo": "You cannot undo this.",
+    "studio.transfer.thisMeans": "This means...",
+    "studio.transfer.noLongerEdit": "You will no longer be able to edit the studio",
```
seotts, line 95: Planning to change this string

src/views/studio/l10n.json
```diff
@@ -88,10 +89,26 @@
...
+    "studio.transfer.noLongerDelete": "You will no longer be able to delete the studio",
+    "studio.transfer.whichManager": "Which manager do you want to make the host?",
+    "studio.transfer.currentHost": "Current Host",
+    "studio.transfer.newHost": "New Host",
+    "studio.transfer.confirmWithPassword": "To confirm changing the studio host, please enter your password.",
+    "studio.transfer.forgotPassword": "Forgot password?",
+    "studio.transfer.alert.somethingWentWrong": "Something went wrong making {name} the host.",
```
seotts, line 102: Planning to change this string

src/views/studio/lib/studio-member-actions.js
seotts: @FunctionalMetatable yep!

src/views/studio/modals/transfer-host-confirmation.jsx
```diff
@@ -0,0 +1,140 @@
...
+    hostId: PropTypes.number
+};
+
+export default connect(
+    state => ({
+        hostId: state.studio.owner,
+        ...managers.selector(state)
+    }),
+    {
+        onLoadMore: loadManagers
```
seotts, line 138: Seems like we can remove this

src/views/studio/modals/transfer-host-info.jsx
```diff
@@ -0,0 +1,69 @@
...
+                <FormattedMessage id="studio.transfer.thisMeans" />
+            </span>
+            <ul>
+                <li><FormattedMessage id="studio.transfer.noLongerEdit" /></li>
+                <li><FormattedMessage id="studio.transfer.noLongerDelete" /></li>
+            </ul>
+            <div
+                className="transfer-host-button-row"
+            >
+                <button
```
seotts, line 48: We could consider using the <Button> component instead.

src/views/studio/modals/transfer-host-modal.scss
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    .transfer-host-alert-wrapper .alert-wrapper {
+        position: relative;
+        display: block;
+        margin-bottom: 2rem;
+    }
+
+    .transfer-host-alert .alert-msg {
+        font-size: 16px;
```
seotts, line 57:
```suggestion
        font-size: 1rem;
```

src/views/studio/modals/transfer-host-modal.scss
```diff
@@ -0,0 +1,176 @@
...
+    .transfer-host-name-selected {
+        color: white !important;
+    }
+
+    .transfer-selection-icon {
+        margin: auto 8px;
+    }
+
+    .transfer-host-tile-selected {
+        background: #0EBD8C;
```
seotts, line 126: This is "extensions green"-- maybe we should add to color file.

src/views/studio/modals/transfer-host-modal.scss
```diff
@@ -0,0 +1,176 @@
...
+        padding: 3rem 3rem 2rem;
+    }
+
+    .transfer-form {
+        padding: 0px 1rem 1rem;
+    }
+
+    .transfer-password-input {
+        margin-left: 2rem;
+        border: 1px solid #d9d9d9;
```
seotts, line 139: Maybe we have a dark gray that is close to this

src/views/studio/modals/transfer-host-modal.scss
```diff
@@ -0,0 +1,176 @@
...
+        border-radius: .5rem;
+        padding: 0.5rem 1rem;
+        font-size: 24px;
```
seotts, line 142:
```suggestion
        font-size: 1.5rem;
```

src/views/studio/modals/transfer-host-tile.jsx
```diff
@@ -0,0 +1,46 @@
...
+        <div className="studio-member-info">
+            <div
+                className={classNames('studio-member-name',
+                    {'transfer-host-name-selected': selected}
+                )}
+
+            >
+                {username}
+            </div>
+            {isCreator && <div className="studio-member-role"><FormattedMessage id="studio.HostRole" /></div>}
```
seotts, line 28: We might not need this anymore (or the isCreator prop)

ericrosenbaum: all set

seotts (approved):

src/views/studio/lib/studio-member-actions.js
AmazingMech2418: @colbygk 
> and is already in the process of being changed

Shouldn't this have been fixed before this was merged? Storing a password in the URL means that your password is perpetually stored in your browser history, which can be dangerous, especially if using Scratch on a shared device.

src/redux/studio-permissions.js
```diff
@@ -3,12 +3,12 @@ const {selectUserId, selectIsAdmin, selectIsSocial,
     selectHasFetchedSession, selectStudioCommentsGloballyEnabled} = require('./session');
 
 // Fine-grain selector helpers - not exported, use the higher level selectors below
-const isCreator = state => selectUserId(state) === state.studio.owner;
```
AmazingMech2418, line 6: I know this has already been merged, but shouldn't the parameter currently named as "owner" be changed to "host" to improve maintainability and readability of the code?

src/redux/studio-permissions.js
```diff
@@ -35,7 +35,7 @@ const selectCanDeleteCommentWithoutConfirm = state => selectIsAdmin(state);
 const selectCanFollowStudio = state => selectIsLoggedIn(state);
 
 // Matching existing behavior, only admin/creator is allowed to toggle comments.
-const selectCanEditCommentsAllowed = state => !selectIsMuted(state) && (selectIsAdmin(state) || isCreator(state));
+const selectCanEditCommentsAllowed = state => !selectIsMuted(state) && (selectIsAdmin(state) || isHost(state));
```
AmazingMech2418, line 38: I'm confused why a muted studio host can't close comments. If comments need to be closed due to spam or anything else of the sort, actions should be able to be taken immediately, and a mute doesn't seem to be needed in this case.

src/views/studio/l10n.json
```diff
@@ -136,5 +153,7 @@
     "studio.alertCuratorInvited": "Curator invite sent to \"{name}\"",
     "studio.alertManagerPromote": "\"{name}\" is now a manager",
     "studio.alertManagerPromoteError": "Something went wrong promoting \"{name}\"",
-    "studio.alertMemberRemoveError": "Something went wrong removing \"{name}\""
+    "studio.alertMemberRemoveError": "Something went wrong removing \"{name}\"",
+    "studio.alertTransfer": "\"{name}\" is now the host",
+    "studio.alertTransferRateLimit": "You can only change the host once a day. Try again tomorrow."
```
AmazingMech2418, line 158: I think this message might be either somewhat misleading or difficult to track on the backend.

If it means there is a 24-hour cooldown, it might be misleading since it could already be "tomorrow" and you still couldn't change the studio host. For example, if you change host at 6 PM your time and want to change the host again at noon the next day.

If it means that it is actually just the next day, verifying time zones (and making sure it isn't just a limit bypass) might be difficult to track since the backend is only in one time zone.

src/views/studio/modals/transfer-host-confirmation.jsx
```diff
@@ -0,0 +1,136 @@
...
+    handleBack,
+    handleTransfer,
+    items,
+    hostId,
+    selectedId
+}) => {
+    const currentHostUsername = items.find(item => item.id === hostId).username;
+    const currentHostImage = items.find(item => item.id === hostId).profile.images['90x90'];
+    const newHostUsername = items.find(item => item.id === selectedId).username;
+    const newHostImage = items.find(item => item.id === selectedId).profile.images['90x90'];
```
AmazingMech2418, line 25: This is an extremely minor performance enhancement, but for example in studios with a large number of managers from before the update, this might be important, but wouldn't it be better to save the host and selected user objects and then get the properties instead of running `items.find` twice?

Like this:
```js
const currentHost = items.find(item => item.id === hostId);
const newHost = items.find(item => item.id === selectedId);
const currentHostUsername = currentHost.username;
const currentHostImage = currentHost.profile.images['90x90'];
const newHostUsername = newHost.username;
const newHostImage = newHost.profile.images['90x90'];
```

src/views/studio/studio-member-tile.jsx
```diff
@@ -26,11 +30,13 @@ import removeIcon from './icons/remove-icon.svg';
 import promoteIcon from './icons/curator-icon.svg';
 
 const StudioMemberTile = ({
-    canRemove, canPromote, onRemove, onPromote, isCreator, hasReachedManagerLimit, // mapState props
+    canRemove, canPromote, onRemove, canTransferHost, onPromote, onTransferHost,
+    isCreator, hasReachedManagerLimit, // mapState props
```
AmazingMech2418, line 34: Shouldn't this be changed to `isHost` instead?

TwilightSeleneMisty:
I would prefer if manager rights were added back, but I can manage with this.

PR #5935, opened 2021-08-16
Add studio host transfer notification to the activity feed

seotts:
This adds record of studio transfer to the activity feed, with two slightly different messages depending on whether it was done by an admin or the former host.
image

This should be fine to merge at any time, including before launch of the feature. (If there are no becomehoststudio notifications, they will not be displayed.)

### DIFF
--- a/src/views/studio/l10n.json
+++ b/src/views/studio/l10n.json
@@ -87,6 +87,8 @@
     "studio.activityBecomeCurator": "{newCuratorProfileLink} accepted an invitation from {inviterProfileLink} to curate this studio",
     "studio.activityRemoveCurator": "{removerProfileLink} removed the curator {removedProfileLink}",
     "studio.activityBecomeOwner": "{promotedProfileLink} was promoted to manager by {promotorProfileLink}",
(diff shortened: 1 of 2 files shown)


src/views/studio/l10n.json
```diff
@@ -87,6 +87,8 @@
     "studio.activityBecomeCurator": "{newCuratorProfileLink} accepted an invitation from {inviterProfileLink} to curate this studio",
     "studio.activityRemoveCurator": "{removerProfileLink} removed the curator {removedProfileLink}",
     "studio.activityBecomeOwner": "{promotedProfileLink} was promoted to manager by {promotorProfileLink}",
+    "studio.activityBecomeHost": "{newHostProfileLink} was made the studio host by {actorProfileLink}",
+    "studio.activityBecomeHostAdminActor": "{newHostProfileLink} was made the studio host by a Scratch Team member.",
```
apple502j, line 91: This is the only message with a period at the end - maybe fix for consistency?

seotts: Good catch!

ericrosenbaum (approved):
looks good to me! Note that I haven't yet been able to test locally.

PR #5958, opened 2021-08-20
Increase default alert duration to 6 seconds

ericrosenbaum:
Increase the default amount of time that alerts for success and errors are shown, from 3 to 6 seconds. Currently all alerts use this default time. The aim is to help ensure that people have enough time to read the alert message before it disappears.


seotts (approved):
lgtm!

PR #5959, opened 2021-08-20
Studio host transfer modal additional work

ericrosenbaum:
We're getting there! Here are some additional updates (tested internally Aug 20th):

- Send password in the body of the request
- Show password validation error
- Add error alerts for rate limit
- Add error alert for user cannot become host
- Add unit test for permission
- Add unit test for transfer action

More updates coming...


src/views/studio/lib/studio-member-actions.js
```diff
@@ -15,6 +15,7 @@ const Errors = keyMirror({
     UNKNOWN_USERNAME: null,
     RATE_LIMIT: null,
     MANAGER_LIMIT: null,
+    CONFLICT: null,
```
seotts, line 18: I wonder if we should name this something more descriptive of the problem (like CANNOT_BE_HOST)? I chose 409 (ConflictError) as the response because it was close to what we wanted, but it is still sort of a stretch, so it feels like something more descriptive in the frontend would help clarify things.

src/views/studio/modals/transfer-host-confirmation.jsx
```diff
@@ -24,12 +31,45 @@ const TransferHostConfirmation = ({
     const newHostUsername = items.find(item => item.id === selectedId).username;
     const newHostImage = items.find(item => item.id === selectedId).profile.images['90x90'];
     const [passwordInputValue, setPasswordInputValue] = useState('');
+    const [validationError, setValidationError] = useState(null);
+    const {errorAlert, successAlert} = useAlertContext();
+
+    const errorToMessageId = error => {
+        switch (error) {
+        case Errors.RATE_LIMIT: return 'studio.alertTransferRateLimit';
+        case Errors.CONFLICT: return 'studio.transfer.alert.thisUserCannotBecomeHost';
```
apple502j, line 40: What "conflict" causes someone to be unable to become a host? Like, requesting a non-manager user?

src/views/studio/lib/studio-member-actions.js
```diff
@@ -28,6 +29,7 @@ const normalizeError = (err, body, res) => {
     if (res.statusCode === 403 && body.mute_status) return Errors.USER_MUTED;
     if (res.statusCode === 401 || res.statusCode === 403) return Errors.PERMISSION;
```
seotts, line 30: There are some 403 errors that are not related to the password for this route, and that should result in "general" error messaging, so we may need to add another kind of error here (like Errors.PASSWORD) and find it by checking for "password incorrect" in the message of the body.

src/views/studio/modals/transfer-host-confirmation.jsx
```diff
@@ -24,12 +31,45 @@ const TransferHostConfirmation = ({
...
+            .then(() => {
+                handleClose();
+                successAlert({
+                    id: 'studio.alertTransfer',
+                    values: {name: newHostUsername}
+                });
+            })
+            .catch(e => {
+                // For password errors, show validation alert without closing the modal
+                if (e === Errors.PERMISSION) {
```
seotts, line 56: If we add an Errors.PASSWORD, this should check for that

src/views/studio/lib/studio-member-actions.js
ericrosenbaum: that makes sense - I was thinking we could add that in a followup

src/views/studio/modals/transfer-host-confirmation.jsx
ericrosenbaum: to be updated in followup

src/views/studio/lib/studio-member-actions.js
ericrosenbaum: Good point - I'll update it

seotts (changes requested):
Looking great. I made a couple suggestions-- the biggest one is that we separate password errors from other permissions errors.

EDIT: Based on discussion above, we'll make that change in a follow-up. Marking this good to merge, pending the change in name for the 409 error. Thanks!

src/views/studio/modals/transfer-host-confirmation.jsx
seotts: @apple502j ha! Yeah, that's exactly the sort of thing this error would be for.

src/views/studio/lib/studio-member-actions.js
seotts: sounds good!

seotts (approved):

TwilightSeleneMisty:
l would prefer manager permissions being added back.

PR #5992, opened 2021-08-27
More transfer host modal fixes

ericrosenbaum:
- Responsive style for password validation alert
- Responsive style for transfer outcome tiles (on the confirmation step)
- Reload the manager list before closing the modal, so you immediately see the new host shown first
- Fix load more buttons

### DIFF
--- a/src/views/studio/modals/transfer-host-selection.jsx
+++ b/src/views/studio/modals/transfer-host-selection.jsx
@@ -56,7 +56,7 @@ const TransferHostSelection = ({
                                 className={classNames('button', {
                                     'mod-mutating': loading
                                 })}
-                                onClick={onLoadMore}
+                                onClick={() => onLoadMore()} // eslint-disable-line react/jsx-no-bind
                             >
                                 <FormattedMessage id="general.loadMore" />
                             </button>
--- a/src/views/studio/modals/transfer-host-selection.jsx
+++ b/src/views/studio/modals/transfer-host-selection.jsx
@@ -56,7 +56,7 @@ const TransferHostSelection = ({
                                 className={classNames('button', {
                                     'mod-mutating': loading
                                 })}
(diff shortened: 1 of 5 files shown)


src/views/studio/modals/transfer-host-selection.jsx
```diff
@@ -56,7 +56,7 @@ const TransferHostSelection = ({
                                 className={classNames('button', {
                                     'mod-mutating': loading
                                 })}
-                                onClick={onLoadMore}
+                                onClick={() => onLoadMore()} // eslint-disable-line react/jsx-no-bind
```
kchadha, line 59: It might be good to make this change when we're defining `onLoadMore` instead (e.g. line 114 of this file)

src/views/studio/studio-managers.jsx
```diff
@@ -88,7 +88,7 @@ const StudioManagers = ({
                     <h3><FormattedMessage id="studio.sectionLoadError.managersHeadline" /></h3>
                     <button
                         className="button"
-                        onClick={onLoadMore}
+                        onClick={() => onLoadMore()} // eslint-disable-line react/jsx-no-bind
```
kchadha, line 91: Same comment as above

src/views/studio/studio-managers.jsx
```diff
@@ -108,7 +108,7 @@ const StudioManagers = ({
                             className={classNames('button', {
                                 'mod-mutating': loading
                             })}
-                            onClick={onLoadMore}
+                            onClick={() => onLoadMore()} // eslint-disable-line react/jsx-no-bind
```
kchadha, line 111: Same comment as above

src/views/studio/modals/transfer-host-selection.jsx
ericrosenbaum: nice! fixed.

kchadha (approved):
LGTM

PR #6034, opened 2021-09-07
Too many password attempts error

seotts:
Adds UX for the error when a user has made too many password attempts in the studio transfer flow.

Also adds unit testing for the various error types for studio transfer.

### DIFF
--- a/test/unit/redux/studio-member-actions.test.js
+++ b/test/unit/redux/studio-member-actions.test.js
@@ -293,7 +293,7 @@ describe('inviteCurator', () => {
     });
     test('error because of rate limit', async () => {
         api.mockImplementation((opts, callback) => {
-            callback(null, null, {statusCode: 429});
+            callback(null, {}, {statusCode: 429});
         });
         await expect(store.dispatch(inviteCurator('user2')))
             .rejects.toBe(Errors.RATE_LIMIT);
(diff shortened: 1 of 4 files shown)


test/unit/redux/studio-member-actions.test.js
```diff
@@ -293,7 +293,7 @@ describe('inviteCurator', () => {
     });
     test('error because of rate limit', async () => {
         api.mockImplementation((opts, callback) => {
-            callback(null, null, {statusCode: 429});
+            callback(null, {}, {statusCode: 429});
```
kchadha, line 296: Could you explain why you needed this change?

test/unit/redux/studio-member-actions.test.js
```diff
@@ -406,26 +406,87 @@ describe('acceptInvitation', () => {
         expect(state.studio.invited).toBe(true);
         expect(state.studio.curator).toBe(false);
     });
+});
```
kchadha, line 409: Was this just incorrectly missing before? I don't see a corresponding change that adds another "describe"

kchadha (commented):
Changes mostly look good to me! I left some questions in the comments!

Thanks for adding all the tests!

test/unit/redux/studio-member-actions.test.js
seotts: That field is the body of the response, which our actual responses all have. `normalizeError()` now checks `body.message` to distinguish between `RATE_LIMIT` (message is null) and `PASSWORD_ATTEMPT_LIMIT` (message is "try again later.") If the body is null, that check would error out.

test/unit/redux/studio-member-actions.test.js
seotts: I'm not totally sure I understand what you mean by another "describe." But previously, the 'transferHost' tests were incorrectly nested under the 'promoteCurator' tests. I moved them out. (line 411)

Accio1:
I'm not sure if this is something that has been considered, but my thoughts are that the user should be signed out after too many failed password attempts. If a user is continually getting the password wrong while trying to transfer a studio, it is possible that it isn't their account and they are attempting to transfer the studio for malicious reasons. If the user is allowed to remain signed in after entering the incorrect password several times, they could go on to perform other harmful actions (such as deleting the studio or posting inappropriate comments). If it was the account owner who was signed in and they just mistyped their password a few times, then having to sign back in would be a minor inconvenience. However, it could potentially stop a person looking to perform harmful actions while signed in (this could be the case if someone forgot to sign out on a shared computer or left their computer unattended). I just thought that I would bring this up as something to consider while implementing this feature.

test/unit/redux/studio-member-actions.test.js
kchadha: Cool! That's exactly what I was asking.

kchadha (approved):
LGTM

seotts:
@Accio1 Interesting suggestion! That could be a good thing to consider doing in the future. I'll pass your idea along.

Accio1:
> @Accio1 Interesting suggestion! That could be a good thing to consider doing in the future. I'll pass your idea along.

Thanks!

PR #6061, opened 2021-09-14
Change copy of transfer rate-limit error

seotts:
Changes the error message for when a studio has been transferred too recently, to reflect that the limit is _per studio_ not _per user_.


ericrosenbaum (approved):
LGTM!

ericrosenbaum:
oh actually on second thought: maybe the string id should be updated, to reset translations?

apple502j:
@ericrosenbaum I don't think we need to change that?

PR #6063, opened 2021-09-15
Ignore querystrings when matching routes

rschamp:
This should eliminate the issue where querystrings would cause URLs to skip S3 and fall back to scratchr2.  I opted to update every relevant instance of `req.url` -> `req.url.path`, but not all of them are strictly necessary. It should just help set a pattern to follow.


colbygk (approved):
I like this, we really can only test this before production by letting it setup scratch.ly first I think, so I'm not testing it locally. Also sets us up to potentially have parameters be specified as part of route definition and automatically normalizing via fastly rules.

PR #6071, opened 2021-09-17
Add integration tests leaving comments on projects, studios, and profiles

BryceLTaylor:
### Resolves:

There are currently no tests that leave comments

### Changes:

Adds integration tests that leave a comment on:
- a project page
- a profile page
- a studio page

More comment tests will be added later.

### DIFF
--- a/test/integration/comments.test.js
+++ b/test/integration/comments.test.js
@@ -1,0 +1,124 @@
+// These tests sign in with user #4 and user #5
+
+import SeleniumHelper from './selenium-helpers.js';
+
+const {
+    findByXpath,
+    buildDriver,
+    clickXpath,
+    clickText,
+    signIn
+} = new SeleniumHelper();
+
+// Using 1 and 2 here. Hopefully this is not confusing.
+let username1 = process.env.SMOKE_USERNAME + '4';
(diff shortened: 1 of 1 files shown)


test/integration/comments.test.js
```diff
@@ -0,0 +1,124 @@
...
+        await driver.get(projectUrl);
+
+        // leave the comment
+        let commentBox = await findByXpath('//textArea[@name="compose-comment"]');
+        await commentBox.sendKeys(buildNumber);
+        await findByXpath(`//textarea[contains(text(), "${buildNumber}")]`);
+        await clickXpath('//button[@class="button compose-post"]');
+
+        // reload the page
+        await driver.sleep(5000);
```
picklesrus, line 70: You mentioned there was a config variable somewhere you might want to use for these wait times instead?
But if not, LGTM.

picklesrus (commented):
Made one comment, but lgtm.

test/integration/comments.test.js
BryceLTaylor: I think we'll only need to do that if the tests start failing on production.  For now we can put it in as is.

PR #6078, opened 2021-09-20
Use studio "host" in state, and accept "host" from api

seotts:
This switches over to using "host" (instead of "owner") terminology in the redux state for studio. It also allows www to accept a field called "host" from api. It falls back on "owner," so this can be merged anytime before the change is made in [api](https://github.com/LLK/scratch-api/pull/1669).

There should be no changes to functionality as a result of this change.

### DIFF
--- a/src/redux/studio-permissions.js
+++ b/src/redux/studio-permissions.js
@@ -3,7 +3,7 @@ const {selectUserId, selectIsAdmin, selectIsSocial,
     selectHasFetchedSession, selectStudioCommentsGloballyEnabled} = require('./session');
 
 // Fine-grain selector helpers - not exported, use the higher level selectors below
-const isHost = state => selectUserId(state) === state.studio.owner;
+const isHost = state => selectUserId(state) === state.studio.host;
 const isCurator = state => state.studio.curator;
 const isManager = state => state.studio.manager || isHost(state);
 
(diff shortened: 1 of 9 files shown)


src/redux/studio-permissions.js
```diff
@@ -61,7 +61,7 @@ const selectCanTransfer = (state, managerId) => {
     if (state.studio.classroomId !== null) return false;
     if (selectIsMuted(state)) return false; // Muted users cannot transfer studios.
     if (state.studio.managers > 1) { // If there is more than one manager,
-        if (managerId === state.studio.owner) { // and the selected manager is the current owner/host,
+        if (managerId === state.studio.host) { // and the selected manager is the current owner/host,
```
ericrosenbaum, line 64: this comment should be updated I think?

src/redux/studio-permissions.js
```diff
@@ -74,7 +74,7 @@ const selectCanRemoveProject = (state, creatorUsername, actorId) => {
 
     // Admins/managers can remove any projects
     if (isManager(state) || selectIsAdmin(state)) return true;
-    // Project owners can always remove their projects
+    // Project hosts can always remove their projects
```
ericrosenbaum, line 77: this change seems incorrect

seotts: Oh good catch!

apple502j:
Will the API still return `owner` key, or will it be removed?

ericrosenbaum (approved):
LGTM!

Naleksuh:
@seotts You replaced all instances of "owner" with "host', but what about references to a studio "creator"? Will these become host as well?

PR #6084, opened 2021-09-21
Remove use of studio-related feature flags

seotts:
This removes the use of the `new_studios_launched` and `studio_transfer_launched` feature flags, in separate commits in case we need to revert one of them. This can be released any time before these flags are removed in scratchr2. Because the studio_transfer_launched flag has not not yet been removed in api, you may see some odd behavior if you have that flag disabled in your dev environment, but everything should work fine if the flags are enabled.

scratch-www didn't make a lot of use of the feature flags, so there isn't a ton we need to test at bughunt. Things we should test:
- If you get muted, and then go to a project page and attempt to add it to a studio, you should be prevented from doing so
image
- In a studio, the host's tile says "Studio Host" and not "Studio Creator"
- If you are logged in as the host, or a mod, you should have the option to transfer host.


ericrosenbaum (approved):
LGTM!

PR #6097, opened 2021-09-24
Allow admins to promote curators to managers

seotts:
This gives admins permission to promote curators to managers, even if they are not a manager. This is a feature request by the community team.

To be consistent with other permissions, admins do not have permission to do this if they happen to be muted.

### Test Coverage:
I modified the test of this permission.

In bughunt, we should test that admins can promote curators (including themselves.) Other non-admin, non-managers should not be able to promote curators.


ericrosenbaum (approved):
LGTM!

PR #6117, opened 2021-09-29
Show "Studio Host" on manager tile

seotts:
The logic that decides whether to show "Studio Host" on the host's tile was looking for "owner." I changed it to host.


ericrosenbaum (approved):
LGTM

PR #6122, opened 2021-10-01
Add open search ability

### LINKED ISSUE #4166: Adding a <link rel="search"> tag
This is used in the _Scratch Wiki_:
```html
<link rel="search" type="application/opensearchdescription+xml" href="/w/opensearch_desc.php" title="Scratch Wiki (en)"/>
```
When the user has visited the Scratch Wiki once, and then types in "en.scratch-wiki.info" in the search bar, you see something like "Press tab to search Scratch Wiki (en)".
Why not add it to scratch.mit.edu?

funblaster22:
### Resolves:

Fixes #4166

### Changes:

The PR adds an opensearch tag to the website, which allows users to search the scratch.mit.edu website by pressing the tab button when typing into the onmibox.

### Test Coverage:

This change does not modify any javascript, so no tests fail. This is what it should look like after the change is implemented:
image


seotts:
Thanks for the contribution. I'm going to close this for now since it is related to a feature request that we haven't decided to implement. If we decide we want this feature in the future, we can reopen this PR.

PR #6126, opened 2021-10-01
Add comment and notification integration tests

BryceLTaylor:
### Changes:

Adds more automated integration tests.  Includes coverage of getting notifications and messages.

Also updates Chromedriver to version 94

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "babel-preset-es2015": "6.22.0",
     "babel-preset-react": "6.22.0",
     "bowser": "1.9.4",
-    "chromedriver": "92.0.1",
+    "chromedriver": "94.0.0",
     "classnames": "2.2.5",
     "cookie": "0.4.1",
     "copy-webpack-plugin": "4.6.0",
--- a/test/integration/comments.test.js
+++ b/test/integration/comments.test.js
@@ -29,10 +29,18 @@ let profileUrl = `${rootUrl}/users/${username2}`;
 let studioId = process.env.COMMENT_STUDIO_ID || 10005646;
 let studioUrl = `${rootUrl}/studios/${studioId}/comments`;
 
+// setup comments to leave
 let date = new Date();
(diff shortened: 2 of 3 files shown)


test/integration/comments.test.js
```diff
@@ -29,10 +29,18 @@ let profileUrl = `${rootUrl}/users/${username2}`;
...
+// setup comments to leave
 let date = new Date();
 let dateString = `${date.getFullYear()}-${date.getMonth() + 1}-${date.getDate()} ` +
 `${date.getHours()}:${date.getMinutes()}:${date.getSeconds()}`;
 let buildNumber = process.env.CIRCLE_BUILD_NUM || dateString;
+let projectComment = buildNumber + ' project';
+let profileComment = buildNumber + ' profile';
+let studioComment = buildNumber + ' studio';
+
+// let projectReply = buildNumber + ' project reply';
```
picklesrus, line 41: Is there a reason to keep these commented out lines?

package.json
```diff
@@ -76,7 +76,7 @@
     "babel-preset-es2015": "6.22.0",
     "babel-preset-react": "6.22.0",
     "bowser": "1.9.4",
-    "chromedriver": "92.0.1",
+    "chromedriver": "94.0.0",
```
picklesrus, line 79: Should this be in this PR? No strong preference, but seems unrelated to the rest of the tests and something we'd likely want to keep even if the new tests turn out to be flaky or something.

picklesrus (commented):
Couple minor comments, but lgtm.

test/integration/comments.test.js
BryceLTaylor: Oh, they are for upcoming tests.  I just didn't get that far yet.  I could take them out for this PR, though.

package.json
BryceLTaylor: I could do it in its own, but usually (and in this case in particular) it needs to be updated because you just can't run the tests without updating Chromderiver.  In the past we've just updated chromedriver in whatever PR is updating tests.  

It is in its own Commit, though, so it should be separable if it needs to be.

PR #6169, opened 2021-10-15
Annual report 2020

caseymm:
Preview of 2020 annual report


Naleksuh:
@caseymm Is 2020 Annual Report a new page? It looks like 2020 content was already inserted into the 2019 report, it mentions COVID-19 and the death of George Floyd, both of which happened in 2020. Will that content be removed from the 2019 one to be put into the 2020 one?

PR #6229, opened 2021-10-28
Annual Report Fixes

caseymm:
Fix video disappearing after fullscreen is entered and exited
Fix tweet alignment on ios

Sets a width on the parent container so that the video (which has `videoFoam=true` property) will not disappear after entering and exiting fullscreen.


kchadha (approved):
LGTM

Issue #1452, opened 2017-08-02
Update react-intl to latest version


mewtaylor:

rschamp:
When we upgraded react-intl, we hit an error due to scratch-l10n trying to import locales from react-intl that no longer exist: https://travis-ci.com/github/LLK/scratch-www/builds/219346737, this seems to be due to the import on [build-locales](https://github.com/LLK/scratch-www/blob/795d6f7bcf5bdcf7d67e3ece87e72f1e47d0262a/bin/build-locales#L37). This step is where scratch-l10n supplies a list of locales along with locale data for each. Currently it imports a lot of locale data from react-intl, but now react-intl removed that data in preference for using a polyfill. This was all used by scratch-www in order to add locale data, but this method no longer exists... in preference for using a polyfill.

We're not sure yet if react-intl needs to be in sync across scratch-www, scratch-gui, scratch-l10n, (others?...)

apple502j:
One important thing is that v3 removed `IntlShape`. That'd require rewriting every code we know...

rschamp:
With the work we've done on `scratch-l10n`, then we can upgrade scratch-www to v3 without breaking GUI. While v3 removed `intlShape` it provides `IntlShape` as an interface, which we can supposedly use instead.

Next up is polyfilling, we decided we probably want to remove the Intl and any Intl-related polyfills from the polyfill bundle we currently have, and instead use polyfill-library in a new entry, so we can selectively polyfill features, as well as provide locale data as necessary in the same file.  (Eventually we could do away with the polyfill bundle and use this entry point for the other polyfills we're using too. But to begin with we'll just use it for Intl-related stuff.)

apple502j:
Ah, it all makes sense now - the code copy-pasted from TypeScript handbook, and now we use `IntlShape` interface...

No, you can't just `sed s/intlShape/IntlShape/g`. While `intlShape` is for PropTypes, `IntlShape` is a TypeScript interface. Which probably means: we need to migrate to TypeScript (or define intlShape ourselves/just use `PropTypes.object` ignoring linters)

Another thing: we might want to consider if polyfills are actually needed. We currently polyfill everything from `Array.prototype.forEach` (es5-shim) to things like matchMedia, btoa, etc.

chrisgarrity:
[...]

PR #6243, opened 2021-11-01
Replace static plural rule and relative time polyfills with a dynamic import

rschamp:
### Resolves:

Toward https://github.com/LLK/scratch-www/issues/1452, we need to stop relying on react-intl locale data, and instead polyfill ourselves as needed.

### Changes:

- Remove relative time polyfill from the relative time library
- Remove locale data polyfill from the static polyfill file
- To replace these, add a dynamic int-polyfill file, imported when the page refreshes (which happens when the language changes)

### Test Coverage:

The tests pass. Since this is browser-specific behavior, it should be tested manually. @chrisgarrity and I confirmed that relative time translation still works on a browser that doesn't have built-in relative time format support (Safari 12). We also saw the bundle imported dynamically.  Also, on a browser that _does_ have this feature built-in, the polyfill was _not_ imported!


src/lib/intl-polyfill.js
```diff
@@ -0,0 +1,312 @@
+// this file should only be `required` in the format-time
+// when Intl.RelativeTimeFormat is not available (Safari < 14), but
+// we're not currently able to do the code splitting in www, and it
+// is always included. To reduce the amount of data that's loaded limit
+// the number of languages loaded to just the top few that are still using
```
chrisgarrity, line 5: Nit - we should correct the comment now that we've added all the languages. Eventually this will end up in scratch-l10n, so it's not urgent.

chrisgarrity (approved):
I'm merging this based on our sauce labs testing

PR #6295, opened 2021-11-10
Convert integration tests for statistics page from Tap to Jest

BryceLTaylor:
### Changes:

Converts the TAP integration tests that check the statistics page to Jest tests.
In one instance I switched from using css selectors to using xpath to match the rest of the tests
I split the second test into two separate tests so that we would have fewer assertions per test.
There are two assertions in the second test because reloading the page un-toggles the toggle, so rather than doing everything twice I just left those two assertions in the same test.  We can revisit that later if it is important.


test/integration/statistics.test.js
```diff
@@ -0,0 +1,66 @@
...
+        let chartTitle = await findByXpath('//div[contains(@class, "box0")]//h3');
+        let chartTitleText = await chartTitle.getText();
+        await expect(chartTitleText).toBe('Monthly Activity Trends');
+    });
+
+    test('New Projects label on first chart defaults to selected', async () => {
+        let toggleXpath = `(//div[@id="activity_chart"]/*[name()='svg']/*[name()='g']/*[name()='g']/*` +
+            `[name()='g'])[4]/*[name()='g']/*[name()='g']/*[name()='g']`;
+        let newProjectsToggle = await findByXpath(toggleXpath);
+        let toggleState = await containsClass(newProjectsToggle, 'nv-series');
```
rschamp, line 45: From offline discussion: Maybe this should be `nv-disabled` and we should check that it's false instead? Does the `nv-series` class mean it's enabled?

rschamp (approved):

test/integration/statistics.test.js
BryceLTaylor: fixed

PR #6311, opened 2021-11-17
fix(gh-5989): adjust comment load batch size

### LINKED ISSUE #5989: Studio comment replies should be fetched in batches of 25, not 20
Via #5929, if we fetch replies in batches of 25, we will never have to request more than one page of replies.

JoelGritter:
### Resolves:

Resolves #5989 

### Changes:

Adjust the studio reply comment load batch size to 25 instead of 20.

### Test Coverage:

Searched for test cases that interact directly with this value, didn't find any.

### DIFF
--- a/src/redux/studio-comment-actions.js
+++ b/src/redux/studio-comment-actions.js
@@ -3,7 +3,7 @@ const eachLimit = require('async/eachLimit');
 const api = require('../lib/api');
 const log = require('../lib/log');
 
-const COMMENT_LIMIT = 20;
+const COMMENT_LIMIT = 25;
 
 const {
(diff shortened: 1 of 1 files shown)


JoelGritter:
My only concern is that this value is also used to load the top-level comments. If need be we could use two constant values, i.e. `COMMENT_LIMIT` and `REPLY_LIMIT`, and set them to 20 and 25 respectively.

JoelGritter:
Yeah, the more I think about it, the more I'm concerned about this. I'll make a quick change to address this.

kchadha (approved):
LGTM!

PR #6330, opened 2021-11-23
[Master][Hotfix] HOC Banner 2021

kchadha:
### Changes:

- Updates the timing for the HOC banner. The banner will be live between 11/29/2021 00:00:00 GMT and 12/13/2021 00:00:00 GMT.
- Update the tutorials featured in the HOC Banner
- Keep the donate banner up even when the HOC banner is showing
- Fix issue with donate banner covering up regular home screen splash banner and hoc banner (use 'position: sticky')
- Fix issue with HOC banner on tablet width devices

### Test Coverage:

Tested locally and on staging and on a tablet screen using the browser dev tools.
Tested the following:
- Banner shows up and goes away at the correct times
- Tutorial links work and correct images display
- Donate Banner still appears when the HOC banner is up
- Donate banner no longer covers regular homepage splash banner


seotts:
This looks good! I did not test locally, but I did check that file names, urls, etc referred to existing things. Thanks for fixing the stickiness!

Issue #6297, opened 2021-11-12
Add app privacy policy to the website


chrisgarrity:
The app stores need to be able to link to a version of the privacy policy that corresponds to the privacy policy in the apps.

The route should be `/privacy_policy/apps`

The apps privacy policy view should be based on the website privacy_policy view.

The [copy for the apps privacy policy](https://docs.google.com/document/d/1CknRE-CTjpxnA8kXEPOWibBM91tx-pnk0HKm19zTiOw/edit?usp=sharing). _Note this document is not currently public, and therefore this is not a help wanted issue_

It should be set up so that the content can be translated (a new l10n file).

apple502j:
Website privacy policy is not translated, why translate this?

chrisgarrity:
@apple502j the policy around translating (or not) is evolving, and we've decided the new policy for the apps should be translated. The same policy is included in the android app, and will need translation.

chrisgarrity:
Resolved in #6355

apple502j:
@chrisgarrity The privacy policy is not shown on Transifex for some reason;

![image](https://user-images.githubusercontent.com/33279053/148989195-09abdbdb-223e-4d2e-bcff-a74aa2720cf6.png)

PR #6355, opened 2021-12-01
App privacy policy

seotts:
This resolves issue #6297 by creating a page for the app privacy policy. The [copy is over here](https://docs.google.com/document/d/1CknRE-CTjpxnA8kXEPOWibBM91tx-pnk0HKm19zTiOw/edit).

We will need to update the date in the policy when we know when we will deploy this.


src/views/privacypolicy-apps/l10n.json
```diff
@@ -0,0 +1,68 @@
+{
+    "privacyApps.title":"Privacy Policy",
+    "privacyApps.updated":"The Scratch Privacy Policy was last updated: [DATE]",
```
seotts, line 3: We will need to update "[DATE]" to the actual date when we know when we are going to deploy this.

src/views/privacypolicy-apps/privacypolicy-apps.jsx
```diff
@@ -0,0 +1,313 @@
+const React = require('react');
+
+const FormattedMessage = require('react-intl').FormattedMessage;
+const Page = require('../../components/page/www/page.jsx');
+const render = require('../../lib/render.jsx');
+const injectIntl = require('react-intl').injectIntl;
+const intlShape = require('react-intl').intlShape;
```
cwillisf, line 7: In order to avoid saying `require('react-intl')` multiple times I suggest something like this:
```suggestion
const {FormattedMessage, injectIntl, intlShape} = require('react-intl');
```

src/views/privacypolicy-apps/privacypolicy-apps.jsx
```diff
@@ -0,0 +1,313 @@
...
+                    </dt>
+                    <dd>
+                        <FormattedMessage
+                            id="privacyApps.deletionDescription"
+                            values={{
+                                helpEmail: (
+                                    <a href="mailto:help@scratch.mit.edu">
+                                        help@scratch.mit.edu
+                                    </a>
+                                )
```
cwillisf, line 203: Since this value is used in 3 places, and theoretically could change someday, I think it would make sense to put a constant at the top of the file to store either the literal text `help@scratch.mit.edu` or the whole `<a href=...` element.

src/views/privacypolicy-apps/privacypolicy-apps.jsx
```diff
@@ -0,0 +1,313 @@
...
+                            }}
+                        />
+                    </p>
+                    <p>
+                        Scratch Foundation<br />
+                        ATTN: Privacy Policy<br />
+                        201 South Street<br />
+                        Boston, MA, 02111
```
cwillisf, line 299: This should probably be a constant near the top as well

cwillisf (changes requested):
I added a few suggestions, including a typo or two, but overall it looks pretty good! Thank you for doing this!

hello-smile6 (commented):
Maybe run `string.replace("  "," ")` on each string?

cwillisf (approved):
👍 Looks good to me!

PR #6364, opened 2021-12-03
Remove run in band from integration tests

BryceLTaylor:
### Resolves:

The integration tests were running in series rather than in parallel so they took a long time to run.

### Changes:

removes --run-in-band when running the integration tests and runs with --maxWorkers=5 instead.  This will run 5 test suites at once rather than 1, which speeds up the tests pretty dramatically

I also increased the timeouts for the tests when running locally.  I think running more tests at once was putting strains on the bandwidth so tests would fail.


rschamp (approved):
Nice work!

PR #6370, opened 2021-12-07
Update selenium webdriver

BryceLTaylor:
### Resolves:

There is a new version of Selenium-Webdriver that we haven't updated to yet.
Also, the Join integration tests were failing sometimes.

### Changes:

Updates Selenium-Webdriver to version 4.1.0
Adds an await to the BeforeEach() function in the Join tests.  It was hard to see the effects of the Selenium update when the tests failed.

### Test Coverage:

Integration tests pass with the new version fo Selenium-Webdriver.

### DIFF
--- a/test/integration/join.test.js
+++ b/test/integration/join.test.js
@@ -24,7 +24,7 @@ describe('www-integration join flow', () => {
     afterAll(async () => await driver.quit());
 
     beforeEach(async () => {
-        driver.get(rootUrl);
+        await driver.get(rootUrl);
         await clickXpath('//a[@class="registrationLink"]');
     });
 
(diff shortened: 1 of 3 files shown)


test/integration/join.test.js
```diff
@@ -24,7 +24,7 @@ describe('www-integration join flow', () => {
     afterAll(async () => await driver.quit());
 
     beforeEach(async () => {
-        driver.get(rootUrl);
+        await driver.get(rootUrl);
```
hello-smile6, line 27: Nice! I sometimes drop promises instead of using them, then regret it.

hello-smile6 (approved):

rschamp (approved):

PR #6374, opened 2021-12-10
[Master] Add configuration for GTM

rschamp:
Conditionally outputs Google Tag Manager code in the head and body of generated html.


hello-smile6 (changes requested):
Could you stop using Google Tag Manager? Google has privacy issues.

PR #6376, opened 2021-12-10
circleci add build_pr job that runs on prs, old build doesn't run on prs

BryceLTaylor:
### Resolves:

build jobs were storing artifacts and caches when running on PRs, wasting resources.

### Changes:

Makes a new version of the build job, build_pr, that will be run on PRs only.
Stops running the old build job on PRs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -242,7 +242,7 @@ workflows:
                 - develop
   build-test-pr:
     jobs:
-      - build_pr:
+      - build-pr:
           filters:
             branches:
               ignore:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,6 +38,33 @@ aliases:
       keys:
         - v1-build-{{ .Revision }}
         - v1-build-
+  - &build_pr
+    <<: *defaults
+    resource_class: medium+
+    steps:
+      - checkout
+      - run:
+          name: "setup"
+          command: |
+            npm --production=false ci
+            mkdir ./test/results
+      - run:
+          name: "run lint tests"
+          command: |
+            npm run test:lint:ci
+      - run:
+          name: "run npm build"
+          command: |
+            WWW_VERSION=${CIRCLE_SHA1:0:5} npm run build
+      - run:
+          name: "Run unit tests"
+          command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -38,6 +38,33 @@ aliases:
       keys:
         - v1-build-{{ .Revision }}
         - v1-build-
+  - &build_pr
+    <<: *defaults
```
cwillisf, line 42: It looks like this grabs the `auth` section, mentioning `DOCKERHUB_*` variables, from the `defaults` section above. I'm not familiar with why we might need those... are they actually secret? Do they come from a context, and if so, will `build-pr` work correctly without these `DOCKERHUB_*` variables? If so, are they actually needed by other builds?

.circleci/config.yml
```diff
@@ -38,6 +38,33 @@ aliases:
...
+    resource_class: medium+
+    steps:
```
cwillisf, line 44: If you move the `store_test_results` in `build` to be before the `save_*_cache` steps, then these `build_pr` steps are identical to the middle steps of `build`. Is there a way to define them once and then include them in both places, like with the `&` / `*` / `<<:` syntax or something similar?

Another way to handle this could be conditional steps, where it would skip certain steps if running in a context where we don't want them. I do this when building the Mac & Windows app: I have a `should_sign` alias which checks the current branch to determine whether or not we want to sign the current build, and I added `condition: *should_sign` to steps which should only run for signed builds.

cwillisf (commented):
It's surprising to me how different two working CircleCI config files can be. I added some questions and maybe a suggestion, depending. That said, I think what you have is pretty good and if my questions / suggestions cause too much of a delay I think it's OK to ignore them.

.circleci/config.yml
BryceLTaylor: We'll have to do something about this if ever Circle changes its docker hub policies, but is probably fine for now.  
They announced on November 21, 2020 that they won't rate limit access to docker hub through circle, but that agreement is subject to change.
Then we should probably use the context scratch-www-all.

.circleci/config.yml
BryceLTaylor: To make this work we might want to remove restoring the git_cache.  We should do that as a next step.

PR #6445, opened 2022-01-09
Focus comment reply textarea when it appears

### LINKED ISSUE #2668: Clicking "reply" on a comment should shift focus to the form's textarea
When I click "reply" on a comment, my instinct is to start typing right away -- that's what i'm used to from other interfaces.

But I can't start typing, until I click in the textarea to focus it.

I think we should focus the textarea as soon as it appears!

mxmou:
### Resolves:

Resolves #2668

### Changes:

Adds `autoFocus` to the textarea for replying to a comment.

### Test Coverage:

Tested manually by clicking "reply" on comments on a project and in a studio. I was able to start typing a reply immediately after clicking the button.


seotts (approved):
This is awesome! Will merge.

PR #6446, opened 2022-01-09
Use font weight bold consistently for teacher registration form labels

### LINKED ISSUE #3589: Teacher registration form has inconsistent font weights
We already noted that the teacher registration form's first step has different font weights for "Username" and "Password". (See https://github.com/LLK/scratch-www/issues/3182 )

Turns out this is a bigger issue... I found inconsistent weights in more places in that form as well:

image

image

image

image

mxmou:
### Resolves:

Resolves #3589

### Changes:

Adds CSS to set the font weight of labels with help text and the "Gender" label to 500 to match the weight of other labels in the teacher registration form.

### Test Coverage:

All labels have the same weight after this change:
image
image
image

### DIFF
--- a/src/components/forms/radio-group.jsx
+++ b/src/components/forms/radio-group.jsx
@@ -11,7 +11,7 @@ require('./radio-group.scss');
 
 const RadioGroup = props => (
     <FRCRadioGroup
-        className={classNames('radio-group', props.className)}
+        rowClassName={classNames('radio-group', props.className)}
         {... props}
     />
 );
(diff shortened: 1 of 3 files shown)


seotts:
Thanks for working on this! I talked with the design team, and they'd prefer that the labels be bold instead of 500. Can you make that change?

mxmou:
@seotts Changed to use bold for all labels except checkboxes and radio buttons, which use 300 as before.

seotts:
Thanks for making that change! This looks good.

PR #4216, opened 2020-07-18
Add overflow, white-space and text-overflow CSS elements

R4356th:
### Resolves:

Partially addresses #3998.

### Changes:

Adds overflow, white-space, and text-overflow CSS elements

### Test Coverage:

Tested with Inspect Element for both project and studio search.
Studio:
Screenshot (79)
Project:
Screenshot (80)

//cc @benjiwheeler

PR #6447, opened 2022-01-09
Increase max-width of studio titles on Explore and search

### LINKED ISSUE #3998: Project and Studio search result titles are clipped
Search for "avatar" and click the Studios choice. You will see some titles clipped:

image

mxmou:
### Resolves:

Resolves #3998 (the part that was not fixed by #4216)

### Changes:

Increases the maximum width of studio titles on the Explore page and in search results to 204 pixels, which is the full width of the tile.

### Test Coverage:

With this change, studio titles are only truncated if they reach the full width of the tile:
image

### DIFF
--- a/src/components/grid/grid.scss
+++ b/src/components/grid/grid.scss
@@ -85,6 +85,12 @@
                     height: 120px;
                 }
             }
+
+            .thumbnail-info {
+                .thumbnail-title {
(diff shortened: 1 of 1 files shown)


tomlum (approved):
Looks good to me!

PR #6449, opened 2022-01-09
Left-align project instructions and notes on medium window widths

### LINKED ISSUE #4151: Notes and credits text gets centered in "narrowish" horizontal layout
The notes and credits text should get centered when the browser is narrow enough for everything to be displayed in a one-column layout:
image

But in a two-column layout, they should *not* be centered:
image

However, if you resize your browser window so that it's narrow, but not narrow enough to switch to the one-column layout, they will be centered:
image

### Steps to Reproduce

1. Go to the page of a project with some text in the description
2. Resize your browser window until it's quite narrow, but the notes and credits are still in a separate column from the project player
3. Observe that the text is now centered

Firefox + Chrome + others?

mxmou:
### Resolves:

Resolves #4151

### Changes:

Makes project instructions and Notes and Credits always left-aligned on 768px viewport or wider.

### Test Coverage:

Tested on different window widths.

### DIFF
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -482,8 +482,13 @@ $stage-width: 480px;
         white-space: pre-line;
         font-size: 1rem;
         line-height: 1.5rem;
+        text-align: left;
         flex: 1;
         overflow-wrap: break-word;
+
+        @media #{$medium-and-smaller} {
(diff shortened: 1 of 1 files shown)


seotts (approved):
Nice! Merging.

PR #6474, opened 2022-01-24
circle remove having separate jobs and caching

BryceLTaylor:
### Resolves:

We're using too much storage in circleci.

### Changes:

Moves the run steps into aliases so they can be easily called in multiple steps
Removes all caching, which leads to high storage usage
Make a new alias called Build_and_deploy which combines the existing build, deploy, and integration jobs
Replace the existing jobs needed for deploys with only build-and-deploy-staging and build-and-deploy-production

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,92 +8,74 @@ aliases:
           username: $DOCKERHUB_USERNAME
           password: $DOCKERHUB_PASSWORD
     working_directory: ~/repo
-  - &save_git_cache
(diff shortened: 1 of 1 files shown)


BryceLTaylor:
The build took about 9 minutes when I built this as a hotfix and deployed to staging.  The current workflow that deploys runs in 11-13 minutes.  That's only 1 data point but I'm hopeful about it.

rschamp (approved):

PR #6492, opened 2022-01-31
Reduce webpack memory usage

BryceLTaylor:
### Resolves:

We were running out of memory on Circle CI builds which caused the build to fail.

### Changes:

This PR decreases the parallelism used for running the minifiers to 4 rather than the default.  Webpack default uses TerserPlugin, which is what we will continue to do, but in order to not use the default setting we have to add it specifically.

This PR also updates a few packages that are used with Webpack: node-sass, sass-loader, postcss-loader, and html-webpack-plugin.  A few updates needed to be made to webpack.config, the package.json, and src/template.ejs to support those updates.

Updating html-webpack-plugin, it no longer could reference "htmlWebpackPlugin.files.chunks". We could then access the common.bundle.js and common.css directly in src/template.ejs

In order to update Autoprefixer past version 9.6.0 we had to change how we pass in the browserlist.  It is now included in the package.json and removed as an option passed in in the webpack.config


BryceLTaylor:
It's worth mentioning that the change to the TerserPlugin that reduces memory usage should only affect builds that deploy to staging or production and should not affect PRs much or at all.

BryceLTaylor:
I pulled the most significant change into this pr: https://github.com/LLK/scratch-www/pull/6494 and merged that one.  The rest of the work is helpful for updating webpack, but should be considered a separate set of work.

PR #6512, opened 2022-02-08
Update webpack dependencies

BryceLTaylor:
### Changes:

Updates dependencies required for Webpack.  The goal is to eventually update Webpack itself, but we're trying to make that transition as smooth as possible by pre-updating some of the other dependencies.


BryceLTaylor:
I'm going to move  each of these updates into their own pr

PR #6513, opened 2022-02-08
Update node-sass to version 7.0.1

BryceLTaylor:
### Changes:

Updates node-sass to version 7.0.1 (the latest version)

Higher versions of Webpack will require us to replace this package with sass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1528,6 +1528,12 @@
         }
       }
     },
+    "@gar/promisify": {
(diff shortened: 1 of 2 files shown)


BryceLTaylor:
Instead of updating node-sass we will replace it with sass (formerly dart-sass).  This will allow us to update sass-loader to the most recent version (in https://github.com/LLK/scratch-www/pull/6517)

PR #6517, opened 2022-02-09
update sass-loader to version 10.2.1

BryceLTaylor:
### Changes:

Updates sass-loader to ~~version 7.0.3.~~ (see comment blow)  

Higher versions (starting with 7.3.1) fail with node-sass version 7.0.1 but succeed with version 4.14.1.  This should be addressed in a followup PR, especially since node-sass has been replaced with sass (formerly dart-sass).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4831,35 +4831,6 @@
       "integrity": "sha1-2jCcwmPfFZlMaIypAheco8fNfH4=",
       "dev": true
     },
-    "clone-deep": {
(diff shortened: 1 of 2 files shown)


BryceLTaylor:
After discussion with @rschamp We've decided to leave node-sass at its current version and update sass-loader to its recent version that works with webpack 4.  In a followup PR we'll replace node-sass with sass (formerly dart-sass).

rschamp (approved):

PR #6518, opened 2022-02-09
Replace node sass

BryceLTaylor:
### Resolves:

node-sass has been deprecated, so updating it any further will require us to switch to using sass (dart-sass)

### Changes:

Replaces node-sass with sass (dart-sass)
updates the 2019 conference page's .scss file which had invalid sass
Update many .scss files to fix deprecated syntax.  hsl and hsla now require a % for the saturation and lightness values.  Division is now handled with the function math.div() rather than a slash. 
This removes many deprecation warnings

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4831,35 +4831,6 @@
       "integrity": "sha1-2jCcwmPfFZlMaIypAheco8fNfH4=",
       "dev": true
     },
-    "clone-deep": {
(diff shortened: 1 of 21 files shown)


BryceLTaylor:
This PR is dependent on https://github.com/LLK/scratch-www/pull/6517

rschamp (approved):

PR #6652, opened 2022-03-15
Add 2022 conference page

kchadha:
### Changes:
- Add 2022 conference page
- /conference now redirects to /conference/2022
- Also Refactor a bit of 2021 conference code/associated components so we don't have to create duplicate components

### Test Coverage:

Tested locally, comparing to 2021 conference page.


kchadha:
Moved shared banner image to a shared folder instead of `2021` folder.

PR #6687, opened 2022-03-22
adjusts navigation height and banner image

tomlum:
### Resolves:

Adjusts banner image and navigation height to accommodate smaller screens better

Before:
Screen Shot 2022-03-22 at 3 08 58 PM (2)

After:
Screen Shot 2022-03-22 at 2 56 34 PM (2)
Screen Shot 2022-03-22 at 2 56 38 PM (2)


kchadha (approved):
LGTM! Thanks for making these changes

PR #6738, opened 2022-04-15
2021 annual report (QA through map section)

caseymm:
No big changes, just some edits to clean up the alignment of the sections from the top through the map section. These sections (only) are ready to go through QA.


src/views/annual-report/2021/l10n.json
```diff
@@ -52,7 +51,8 @@
     "annualReport.2021.reachNewUsers": "New Users",
     "annualReport.2021.reachProjectsCreated": "Projects Created",
     "annualReport.2021.reachProjectCreators": "People Creating Projects",
-    "annualReport.2021.reachScratchAroundTheWorld": "Scratch is used around the world across more than 200 countries and territories",
+    "annualReport.2021.reachScratchAroundTheWorld": "Scratch is used around the world across {bold}",
+    "annualReport.2021.reachScratchAroundTheWorldBold": "more than 200 countries and territories",
```
rschamp, line 55: I am curious if this is the best way to do this, or if it will be problematic for the string to be split across two different messages.

rschamp (approved):
I am going to leave this to @seotts to merge, but it looks good to me pending one question about splitting up that l10n string.

caseymm:
Added a few more things to be QAd. Feel free to take a look down through the "Volunteer Translators" image.

src/views/annual-report/2021/l10n.json
seotts: Actually, we probably want the placeholder name to reflect the content more than the styling (something like `numberOfCountries`.) I'll make that change after I merge this PR.

seotts:
This looks good to merge, thank you!

PR #6741, opened 2022-04-21
Serve images from uploads instead of cdn2. 

picklesrus:
Cdn2 is already redirecting these requests to uploads and has been for several months now.


rschamp (approved):
Looks good!

PR #6742, opened 2022-04-25
Releases

itsHenry35:
Maybe it's a good idea to build the website through github actions
image
image


seotts:
Hi! It's cool that you thought through how to do this! We use CircleCI to automate deploy, so we won't be making this change.

PR #6745, opened 2022-04-25
Add messages for scratch week banner so they can be translated. 

picklesrus:
Actual banner changes to use these messages coming in a later PR. This is just so the messages can get translated.
Here's how it'll looks image


tomlum (approved):
LGTM!

PR #6773, opened 2022-05-05
Send fastly token to gui

tomlum:
### Resolves:

Passes the Fastly Project token from the project metadata call to GUI

Related to this follow up GUI PR: https://github.com/LLK/scratch-gui/pull/8269


apple502j:
Questions:

- When will this be deployed? Is there a migration period for tools using the API?
- Is this required for shared projects as well, or is this for unshared projects only?
- If this also applies to shared projects, can you add `Access-Control-Allow-Origin: *` header to `api.scratch.mit.edu/projects/<id>` endpoint so that external websites can continue to load shared projects?
- Is there a way to change/reset the token? (For example, does unsharing or saving cause it to change?)

susbaconhairman:
no please i still need turbowarp to see unshared projects

Dhuls-Scratch:
> no please i still need turbowarp to see unshared projects

That was something that never should have happened in the first place.

src/views/preview/project-view.jsx
```diff
@@ -259,7 +259,7 @@ class Preview extends React.Component {
         } else {
             Object.assign(opts, {
                 method: 'put',
-                url: `${this.props.projectHost}/${projectId}${qs}`
+                url: `${this.props.projectHost}/${projectId}?test="awefjiaoewfjioeawjfiaeowf`
```
GrahamSH-LLK, line 262: Should this be removed before merging?

Poly2it:
This shouldn't be implemented without an alternative way to share projects through a link but not on the website. The feature was very helpful for letting a limited section of people access a project.

Dhuls-Scratch:
> This shouldn't be implemented without an alternative way to share projects through a link but not on the website. The feature was very helpful for letting a limited section of people access a project.

But that's rejected
> Sorry - I'm going to have to reject this idea.  We depend heavily on the community to view projects and let us know when something in the project goes against the Community Guidelines.  Restricting the viewing audience to a very small number decreases the likelihood that we would get a report if the project were not okay.  It would also make it easier to do some forms of private communication, since we don't have the resources to view every project.

https://scratch.mit.edu/discuss/topic/296512/?page=2#post-3742815

Oren-Lindsey:
> - If this also applies to shared projects, can you add Access-Control-Allow-Origin: * header to api.scratch.mit.edu/projects/<id> endpoint so that external websites can continue to load shared projects?

I feel like that would probably happen because they don't want to break project embedding

Oren-Lindsey (approved):

apple502j:
Bump; we'd like to have the answer on whether this also impacts **shared projects**, and if so, whether the Scratch Team intends to add a method of viewing shared projects from other websites.

hello-smile6 (changes requested):
Needs user feedback

seotts:
Hi all! This PR is one of many pieces of this work-- you are unlikely to notice the impact of this individual PR. Eventually, we do intend to pass tokens for all requests for all types of projects. Due to the nature of this work, we won't be able to share many more details about it.

Oren-Lindsey:
> > Hi all! This PR is one of many pieces of this work-- you are unlikely to notice the impact of this individual PR. Eventually, we do intend to pass tokens for all requests for all types of projects. Due to the nature of this work, we won't be able to share many more details about it.
> 
> That's easy to understand - but please don't ignore the community at large. Being able to open projects on other sites via the API (forkphorus and Turbowarp being two examples) is a big part of the community. If you disable the ability for even shared projects to be viewed without a token, it's likely that developers will find a way to implement password-based authentication, which could potentially be a large security issue.

turbowarp would probably register a new account and just use that one's token

Chiroyce1:
> Eventually, we do intend to pass tokens for all requests for all types of projects. Due to the nature of this work, we won't be able to share many more details about it.

Yes, we do understand that implementing certain changes cannot be directly communicated with the community, but at least a hint towards what impact this will have on Scratch mods that currently utilize Scratch's project and asset server would be good to have. I think communication should be necessary between developers of Scratch and Scratch mods, this way they too can adapt to the changes that are brought in by changes to Scratch and it's project/asset system.

apple502j:
Here is the more specific details on what tool makers need:

They need **cross-origin access to projects**. There are various existing websites that access the projects API. Here are some examples:

- When **developing scratch-gui** (without scratch-www), you can no longer access projects via the hash sign URL.
- forkphorus and TurboWarp uses it to grab projects to run or mod.
- parse-sb3-blocks uses it for the scratchblocks generator.
- etc.

The most straightforward way of doing this is by adding `Access-Control-Allow-Origin: *` header to `api.scratch.mit.edu/projects/<id>` route. **This does not allow credentials to be used cross-origin**, so unshared projects are not at risk with this method. This only allows shared projects to be read by other websites, as they can today.

src/lib/storage.js
```diff
@@ -11,9 +11,18 @@ class Storage extends ScratchStorage {
...
-            projectAsset => `${PROJECT_HOST}/${projectAsset.assetId}`
+            this.getProjectGetConfig.bind(this)
         );
     }
+    setProjectToken (projectToken) {
+        this.projectToken = projectToken;
+    }
+    getProjectGetConfig (projectAsset) {
+        const path = `${PROJECT_HOST}/${projectAsset.assetId}`;
+        const qs = this.projectToken ? `?token=${this.projectToken}` : '';
```
Oren-Lindsey, line 22: why urlsearchparams? why not a header?

Oren-Lindsey:
So thanks to redguy7 on discord I have gotten more clarity on what this actually does:
The token will be required for all projects, but in the case of shared projects, a special token just for that project will be available publicly. So everything should continue to work the same (embeds, viewing when not logged in, turbowarp for shared projects)
Unshared projects will require the account token
So, basically all this does is make it impossible to view unshared projects without being on the right account and adding in an extra hoop to jump through to see shared projects

RedGuy12:
(that is just what i believe from looking at these prs; I could be very wrong. as the scratch team does not want to share more information about this, we will have to wait until it is shipped to know for sure how this works.)

src/lib/storage.js
apple502j: I believe that is how this works on Fastly.

apple502j:
This does restrict other websites from reading shared projects, unless they host a proxy server to get the token.

CST1229:
Why not just directly require auth for unshared project getting (as I think it already is for saving)?

apple502j:
I believe that is due to caching requirements. From what I can guess this token stuff is validated on Fastly CDN.

rgantzosonscratch:
When will more information on the tokens be released? I know that this is important to many developers, and I was wondering where the tokens would be stored.

Scratchfangs:
Don't do this 👎

Dhuls-Scratch:
Can y'all explain why making something private that always should've been private is a bad idea?

rgantzosonscratch:
> Can y'all explain why making something private that always should've been private is a bad idea?

I can. Scratch is a website primarily used by kids, and if developers create websites that require project access to share the projects with others, users could easily get scammed by projects and extensions that require personal data, such as passwords and session IDs.

Dhuls-Scratch:
> > Can y'all explain why making something private that always should've been private is a bad idea?
> 
> I can. Scratch is a website primarily used by kids, and if developers create websites that require project access to share the projects with others, users could easily get scammed by projects and extensions that require personal data, such as passwords and session IDs. These young kids are especially easy to trick.

They should've already been taught to not use their passwords unless they are on the site the password is for, but ok.
Plus, they should know that they can just download the project and upload it to that website.

hello-smile6:
> > > Can y'all explain why making something private that always should've been private is a bad idea?
> > 
> > 
> > I can. Scratch is a website primarily used by kids, and if developers create websites that require project access to share the projects with others, users could easily get scammed by projects and extensions that require personal data, such as passwords and session IDs. These young kids are especially easy to trick.
> 
> They should've already been taught to not use their passwords unless they are on the site the password is for, but ok. Plus, they should know that they can just download the project and upload it to that website.

Not necessarily. Some can barely open Downloads and double-click on something, let alone upload something to a website.

rgantzosonscratch:
> > > Can y'all explain why making something private that always should've been private is a bad idea?
> > 
> > 
> > I can. Scratch is a website primarily used by kids, and if developers create websites that require project access to share the projects with others, users could easily get scammed by projects and extensions that require personal data, such as passwords and session IDs. These young kids are especially easy to trick.
> 
> They should've already been taught to not use their passwords unless they are on the site the password is for, but ok. Plus, they should know that they can just download the project and upload it to that website.

No you can’t… this is a website for young kids teaching them how to use computers and how to code. You can’t expect them to know anything about the significance of session IDs when they’re just learning how to code with blocks. You can’t expect little kids to be good at this. You definitely can’t expect little kids to be good at this to the point where you give them a bunch more opportunities to get hacked.

Dhuls-Scratch:
> You can’t expect them to know anything about the significance of session IDs when they’re just learning how to code with blocks.

Fair point, but they also should at least know that their private projects are actually private.

Maybe [this project](https://scratch.mit.edu/projects/2777405/) could be featured.

hello-smile6:
> > You can’t expect them to know anything about the significance of session IDs when they’re just learning how to code with blocks.
> 
> Fair point, but they also should at least know that their private projects are actually private.
> 
> Maybe [this project](https://scratch.mit.edu/projects/2777405/) could be featured.

But they [aren't](https://turbowarp.org/1) yet. Also, avoid overestimating people. Someone could easily trick users into running a malicious bookmarklet, and the lack of a Content-Security-Policy header isn't exactly helping it.

Additionally, that says _nothing_ about bookmarklets, which are actually a bigger issue.

Dhuls-Scratch:
> Someone could easily trick users into running a malicious bookmarklet, and the lack of a Content-Security-Policy header isn't exactly helping it.
> 
> Additionally, that says _nothing_ about bookmarklets, which are actually a bigger issue.

Just make the extension policy apply to bookmarklets (in fact, they already should, as they're userscripts)
Blocking bookmarklets from working is just going to make the community madder than it is, since many use a bookmarklet for animated thumbnails

hello-smile6:
> > Someone could easily trick users into running a malicious bookmarklet, and the lack of a Content-Security-Policy header isn't exactly helping it.
> > Additionally, that says _nothing_ about bookmarklets, which are actually a bigger issue.
> 
> Just make the extension policy apply to bookmarklets (in fact, they already should, as they're userscripts) Blocking bookmarklets from working is just going to make the community madder than it is, since many use a bookmarklet for animated thumbnails

The `connect-src` CSP value just needs to be restricted, so that pages can't, e.g. change the password on a project, or post a comment from the account settings. (There's no reason to be letting scripts load from GitHub)

hello-smile6:
> Just make the extension policy apply to bookmarklets (in fact, they already should, as they're userscripts)

Policies can't prevent violations of rules. Security can.

> just going to make the community madder than it is

Wouldn't people be even more mad about losing their accounts?

RL1123:
> Yet another awful, unnecessary change from the scratch team, never fails to disappoint.

Yet another awful, unnecessary comment from an ungrateful user, never fails to disappoint.

PR #6818, opened 2022-05-23
Actually set the date in one of the tests. 

picklesrus:
It worked before because the actual date was before may 22.


tomlum (approved):
LGTM!

PR #6821, opened 2022-05-24
Fix “Creative Expression” description copy in annual report

seotts:
The copy for the Creative Expression value was a duplicate of the one for Equitable Opportunities. I fixed it to the copy on the content doc.


tomlum (approved):
LGTM!

PR #6822, opened 2022-05-24
Adds check for newly created project to not require token

tomlum:
### Resolves:

Fixes issue where newly created projects would not load without a token (even though it's not needed)


picklesrus:
LGTM

PR #6833, opened 2022-05-26
Add test for loading, copying, and remixing projects

BryceLTaylor:
### Changes:

Adds 3 integration tests that create projects

- Uploading from a file
- Creating a copy of a project owned
- Remixing an unowned project


test/integration/project-page.test.js
```diff
@@ -177,3 +180,69 @@ describe('www-integration project-page signed in', () => {
...
 });
+
+describe('www-integration project-creation signed in', () => {
+    beforeAll(async () => {
+        // expect(projectUrl).toBe(defined);
+        driver = await buildDriver('www-integration project-creation signed in');
+        await driver.get(rootUrl);
+        await driver.sleep(1000);
+        await signIn(username, password);
+        await findByXpath('//span[contains(@class, "profile-name")]');
```
seotts, line 191: what does this line do?

test/integration/project-page.test.js
```diff
@@ -177,3 +180,69 @@ describe('www-integration project-page signed in', () => {
         await expect(gfVisible).toBe(true);
     });
 });
+
+describe('www-integration project-creation signed in', () => {
```
seotts, line 184: how do you decide whether to test signed it or signed out or both?

seotts (approved):
Looks good to me! Added a couple of questions, but they are for my own education and shouldn't prevent merging.

test/integration/project-page.test.js
BryceLTaylor: This section 

- 188 loads the home page
- 189 waits 1 second (there are some race condition here and this is the easiest way around it)
- 190 signs in.  This is a function defined in selenium-helpers.  https://github.com/LLK/scratch-www/blob/develop/test/integration/selenium-helpers.js#L154 It clicks sign in in the header, types the username, then types the password.  As part of typing the password the last character is the return character.  It then waits until the profile name appears on the page.
- 191 this waits until the profile name appears on the page (in the upper right).  We likely don't need to do this because it has just done that, but I remember having trouble with timing in this area, so it might be helping some how.

test/integration/project-page.test.js
BryceLTaylor: I started with the tests being all signed out, and then added signed in tests as I went.  These last tests all involve creating and saving projects, so there needs to be a signed in user.  It's possible we could follow up with a signed out test for uploading a project, but that is already tested in the gui.  They could have different results, but it seems like that case would mostly be covered by the gui integration test and the signed in www test.

PR #6865, opened 2022-06-06
Hotfix/project token

tomlum:
### IMPORTANT

This is not really a hotfix branch, NOT intended to be merged directly to master as a hotfix.

We are only using the "hotfix/" prefix so that we can test on staging.

### Resolves:

Allows project token usage when creating new projects

### Notes:

this trunk was on a previous PR ( https://github.com/LLK/scratch-www/pull/6773 ), but we reverted it. So a bunch of these commits were previously reviewed.

### DIFF
--- a/src/views/preview/project-view.jsx
+++ b/src/views/preview/project-view.jsx
@@ -838,45 +838,46 @@ class Preview extends React.Component {
                         />
                     </Page> :
                     <React.Fragment>
-                        {showGUI &&
-                        (<IntlGUI
-                            assetHost={this.props.assetHost}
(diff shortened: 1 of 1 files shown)


seotts (commented):
Looking good. Thanks for figuring out this tricky stuff. I left a comment about adding a comment, but other than that, seems good to merge. Make sure to add a detailed test plan to the Bughunt doc -- testing is going to be super important.

A note about the PR title/description: I think a bit more detail about what this does and your approach to fixing the bugs would have helped make this a smoother review.

PR #6888, opened 2022-06-22
Update line 172 to say location, not country

derter-menter:
Not all locations on that list are countries.

### Resolves:
It says "what country you live in" when not every place is a country. 

### Changes:

Changes line 172 to a more accurate description


derter-menter:
Can someone check this?

benjiwheeler:
This is a significant change to our wording -- we can discuss this, but it needs to start as a github issue.

derter-menter:
> This is a significant change to our wording -- we can discuss this, but it needs to start as a github issue.

why?

PR #6909, opened 2022-07-09
add macOS troubleshooting info on all extension pages

cwillisf:
### Resolves:

Related to LLK/scratch-link#198

### Changes:

**This is not currently set up as a hotfix but I'd like us to consider that option. To be discussed on Monday.**

This change documents the most recent quirks of Scratch Link on macOS in the troubleshooting section of each extension landing page:
* Safari is not compatible with Scratch Link 1.4.x so the troubleshooting step recommends another browser.
* Early versions of macOS 12 are not compatible with Scratch Link at all due to what seems to be an OS bug. This seems to have been fixed in 12.3, so the troubleshooting step suggests upgrading.

Since Scratch Link will be used by at least one workshop during the conference, I'd like to publish this information before the conference.

In order to make it easier to maintain these troubleshooting steps consistently across all the extension landing pages, I made a shared `ExtensionTroubleshooting` component. The component adds shared troubleshooting steps around its children, which are assumed to be extension-specific troubleshooting steps.

In the future I plan to move existing shared troubleshooting steps into this component, but I don't think it's a good idea to do this so close to the upcoming Scratch Conference. Moving those steps will disrupt their localization strings in at least three ways:
* each page is currently its own "resource" so the strings will move from one resource to another
* some of the strings include the device name, which I plan to replace using a `formatMessage` variable, changing the string in the process
* reusing the existing string IDs won't make sense: I plan to replace `microbit.checkOSVersionTitle` with `extensions.checkOSVersionTitle` for example

For now, all existing troubleshooting steps are being treated as extension-specific in order to avoid localization disruptions. The new versions of the strings are present, though, in order to allow translators to start working on them if they'd like.

### Test Coverage:

Tested manually by viewing all the extension pages.

### DIFF
--- a/src/views/ev3/ev3.jsx
+++ b/src/views/ev3/ev3.jsx
@@ -26,6 +26,7 @@ const Step = require('../../components/steps/step.jsx');
 
 const OS_ENUM = require('../../lib/os-enum.js');
 const {isDownloaded, isFromGooglePlay} = require('../../components/install-scratch/install-util.js');
+const ExtensionTroubleshooting = require('../../components/extension-landing/extension-troubleshooting.jsx');
 
 require('../../components/extension-landing/extension-landing.scss');
 require('./ev3.scss');
@@ -279,8 +280,10 @@ class EV3 extends ExtensionLanding {
                         />
                     </Steps>
                 </ExtensionSection>
-                <ExtensionSection className="faq">
-                    <h2><FormattedMessage id="ev3.troubleshootingTitle" /></h2>
+                <ExtensionTroubleshooting
+                    deviceName={this.props.intl.formatMessage({id: 'ev3.deviceName'})}
+                    scratchLinkOnly // EV3 is Bluetooth Classic so it's only available through Scratch Link
+                >
                     {isDownloaded(this.state.OS) && (
                         <React.Fragment>
                             <h3 className="faq-title"><FormattedMessage id="ev3.checkOSVersionTitle" /></h3>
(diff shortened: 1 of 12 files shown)


src/views/ev3/ev3.jsx
```diff
@@ -279,8 +280,10 @@ class EV3 extends ExtensionLanding {
                         />
                     </Steps>
                 </ExtensionSection>
-                <ExtensionSection className="faq">
-                    <h2><FormattedMessage id="ev3.troubleshootingTitle" /></h2>
+                <ExtensionTroubleshooting
+                    deviceName={this.props.intl.formatMessage({id: 'ev3.deviceName'})}
```
chrisgarrity, line 284: I don't think this actually needs to be in a `formatMessage`. Most of the device names are brand/trademarks that we don't want people to translate.

src/views/gdxfor/gdxfor.jsx
```diff
@@ -204,8 +205,10 @@ class GdxFor extends ExtensionLanding {
                         />
                     </Steps>
                 </ExtensionSection>
-                <ExtensionSection className="faq">
-                    <h2><FormattedMessage id="gdxfor.troubleshootingTitle" /></h2>
+                <ExtensionTroubleshooting
```
chrisgarrity, line 208: This is the exception for `deviceName`, the registered trademark is `Go Direct`, which we don't actually use. So this one should keep the `formatMessage` calls.

chrisgarrity (approved):
While not actively harmful, we don't need the trademarked names to be translatable strings. It would be better to remove them from the l10n files, and just use the literal strings.

Issue #6925, opened 2022-07-13
Two people are still missing from past contributors


benjiwheeler:
Zachary Deiman and Jolie Castellucci are past contributors, and need to be added to our list!

PR #6926, opened 2022-07-13
add Jolie and Zachary to past contributors

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-www/issues/6925

### Changes:

adds Jolie and Zachary to past contributors (thanks @chrissstan for pointing this out!)


AlexisGoodfellow (approved):
Make it so

PR #7000, opened 2022-08-19
CE-176 Change "Mac OS" to "macOS" on extension pages

### LINKED ISSUE #3413: Make capitalization of "macOS" consistent across Downloads and Extensions Pages
## Background
Sometimes we write "macOS," other places we write "Mac," or "Mac OS."

**Example:** [Downloads Page](https://scratch.mit.edu/download):

Screen Shot 2019-09-26 at 7 17 49 AM

**Example:** FAQ/Troubleshooting section, Extension pages:
(Scroll to bottom of [micro:bit page](https://scratch.ly/microbit), for example)

Screen Shot 2019-09-26 at 7 21 12 AM

________
## Proposal
Apple uses "macOS" when they are talking about OS versions, and "Mac" when they talk about the hardware product.

We will soon have other OS's like ChromeOS, which we probably don't want to refer to as Chrome, since it could be confused with the browser. 

**Let's use "macOS" with lowercase capitalization everywhere**, since we're usually talking about the OS version.

mxmou:
### Resolves:

Resolves #3413

### Changes:

Changes all mentions of "Mac OS" or "MacOS" on extension pages to "macOS". The reference to "MacOS" on the download page was removed in #3569.

### Test Coverage:

Tested manually by opening each extension page.


corihudson:
@mxmou Thanks for doing this! There are a few more places that could use this change:

* https://github.com/mxmou/scratch-www/blob/2d623624d1d3c6e8cc39df15097f51b607d648ff/src/l10n.json#L433
* https://github.com/LLK/scratch-www/blob/9194e30cbf5d31f2f0b212fe11e0ac97c4e7f19c/src/views/wedo2-legacy/l10n.json#L7

after those, this should be good to merge!

mxmou:
@corihudson I made the requested changes.

corihudson (approved):
lgtm, thank you!

PR #7012, opened 2022-08-22
CE-139: update credits page

corihudson:
### Resolves:

Resolves CE-139.

### Changes:

Adds Teo and myself to the credits page and moves Eric Schilling to the past team members section.


benjiwheeler:
LGTM!

PR #7063, opened 2022-09-22
Add check on password complexity by using a backend call

colbygk:
Compliments the ability of the frontend check on username validity with a check on password weakness. This requires the deployment of changes to scratch-api before this change is deployed.

### DIFF
--- a/src/components/join-flow/username-step.jsx
+++ b/src/components/join-flow/username-step.jsx
@@ -114,11 +114,12 @@ class UsernameStep extends React.Component {
         if (!password) return null; // skip validation if password is blank; null indicates valid
         // if password is not blank, run both local and remote validations
         const localResult = validate.validatePasswordLocally(password, username);
+        if (localResult.valid === false) { // defer to local check first
+            return this.props.intl.formatMessage({id: localResult.errMsgId});
+        }
         return this.validatePasswordRemotelyWithCache(password).then(
             remoteResult => {
-                if (localResult.valid === false) { // defer to local check first
-                    return this.props.intl.formatMessage({id: localResult.errMsgId});
-                } else if (remoteResult.valid === false) {
+                if (remoteResult.valid === false) {
                     return this.props.intl.formatMessage({id: remoteResult.errMsgId});
                 }
                 return null;
--- a/src/components/join-flow/username-step.jsx
+++ b/src/components/join-flow/username-step.jsx
@@ -24,6 +24,7 @@ class UsernameStep extends React.Component {
             'handleSetUsernameRef',
             'handleValidSubmit',
             'validatePasswordIfPresent',
+            'validatePasswordRemotelyWithCache',
             'validatePasswordConfirmIfPresent',
             'validateUsernameIfPresent',
             'validateUsernameRemotelyWithCache',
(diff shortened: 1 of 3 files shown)


src/components/join-flow/username-step.jsx
```diff
@@ -92,12 +94,37 @@ class UsernameStep extends React.Component {
...
-        const localResult = validate.validatePassword(password, username);
-        if (localResult.valid) return null;
-        return this.props.intl.formatMessage({id: localResult.errMsgId});
-    }
+        // if password is not blank, run both local and remote validations
+        const localResult = validate.validatePasswordLocally(password, username);
+        return this.validatePasswordRemotelyWithCache(password).then(
+            remoteResult => {
+                if (localResult.valid === false) { // defer to local check first
+                    return this.props.intl.formatMessage({id: localResult.errMsgId});
```
towerofnix, line 120: Since this deferral happens (by definition) without regard to the remote result, is there a reason to make the remote validation request in the first place, if the local result guaranteed an overall failure anyway?

An early return seems like it should be fine, i.e. by moving this `if (localResult.valid === false)` check above the call to `validatePasswordRemotelyWithCache`.

(This suggested change would mean the surrounding function `validatePasswordIfPresent` would sometimes return sync and sometimes async, so you could do `return Promise.resolve(formatMessage(...))`, but Formik [supports sync/async validation anyway](https://formik.org/docs/guides/validation) - this PR already takes advantage of that by not making any changes to areas where `validatePasswordIfPresent` is referenced.)

src/lib/validate.js
```diff
@@ -67,14 +67,51 @@ module.exports.validatePassword = (password, username) => {
...
     }
     return {valid: true};
 };
 
+module.exports.validatePasswordRemotely = password => (
+    new Promise(resolve => {
+        api({
+            method: 'POST',
+            uri: `/accounts/checkpassword`,
+            json: {
```
benjiwheeler, line 83: Is this as secure as we can reasonably be with the password value? Will HTTPS basically protect this info? My knowledge is a little shaky here, but am I right in understanding that a password HTML element value will have a surface level of encryption applied when submitting via POST -- is that level of encryption better?

src/lib/validate.js
```diff
@@ -67,14 +67,51 @@ module.exports.validatePassword = (password, username) => {
...
+            let msg = '';
+            if (body && body.msg) msg = body.msg;
+            else if (body && body[0]) msg = body[0].msg;
+            switch (msg) {
+            case 'valid password':
+                return resolve({
+                    requestSucceeded: true,
+                    valid: true
+                });
+            case 'valid password':
```
benjiwheeler, line 103: this should be 'invalid password'

benjiwheeler:
I think we should add an integration test of validatePasswordRemotely

src/components/join-flow/username-step.jsx
benjiwheeler: Agreed -- if localResult.valid is false, we should not call validatePasswordRemotelyWithCache

src/lib/validate.js
colbygk: Yes. HTTPS is already protecting the password info during the actual sending of the password when setting up the join/change password process.

colbygk: The reason to do a POST and/or a PUT is that GETs are technically not supposed to have a body's according to the RFCs and many libraries in the world will either not support sending a body or outright respond with an error if they see one.

benjiwheeler: And putting the password in the querystring would be insecure becuase the actual URL is more visible than the POST payload?

PR #7142, opened 2022-10-25
CE-224 Updates Donation Url

tomlum:
Updates donation URL to be https://www.scratchfoundation.org/donate


benjiwheeler (approved):
LGTM! I tested the routes redirects as well.

PR #7155, opened 2022-10-30
WIP: attempt to use circleci’s browser-tools orb to force chrome version for integration tests

benjiwheeler:
DO NOT CHECK THIS IN

it's a work in progress that does NOT work as intended... 

attempting to install the Chrome version gives the error:

image
image

independently of the install attempt, attempting to even just check the Chrome version gives:

image

https://app.circleci.com/pipelines/github/LLK/scratch-www/10340/workflows/a605f150-8e66-4b35-98f1-16d5a619a46b/jobs/12755

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,8 @@
 version: 2.1
 
+orbs:
+  browser-tools: circleci/browser-tools@1.4.0
+
 aliases:
   - &defaults
     docker:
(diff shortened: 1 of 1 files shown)


cwillisf:
I used a different approach in `scratch-gui` and it seems to have worked out: LLK/scratch-gui#8635

It doesn't control the version of Chrome, but it does make sure that ChromeDriver matches Chrome.

cwillisf:
Now that I've looked at this more closely, I see that my suggestion doesn't make sense here. Sorry about that!

It seems that Chrome doesn't like to let people install older versions. I noticed that the `browser-tools` orb specifies that the `chrome-version` argument to `install-chrome` is "Only supported on centos and debian" so that might be part of the trouble you're running into. If there's no specific reason to avoid Debian, I guess that might be worth a shot.

PR #7165, opened 2022-11-04
feat: Disable cloud variables when camera is active

aoneill01:
These changes should function as-is, but the full feature depends on LLK/scratch-gui#8634.

### Resolves:

Resolves LLK/scratch-ops#565

### Changes:

- Alert the user on the project page if the cloud variables are disabled because the project uses the video sensing blocks.
- It is now possible for both this alert and the username block alert to happen at the same time. Stack alert messages if there are multiples.

### Test Coverage:

- Added new unit tests for `project-info.js` to cover the `videoSensing()` helper method used to detect whether the alert should be shown.
- Tested locally with v3 and v2 projects, using different combinations of cloud, videoSensing, and username blocks.

### DIFF
--- a/src/views/preview/project-view.jsx
+++ b/src/views/preview/project-view.jsx
@@ -379,9 +379,10 @@ class Preview extends React.Component {
                     }
 
                     if (showAlerts) {
-                        // Check for username block only if user is logged in
+                        // Check for username and video blocks only if user is logged in
                         if (this.props.isLoggedIn) {
                             newState.showUsernameBlockAlert = helpers.usernameBlock(projectData[0]);
+                            newState.showCloudDataAndVideoAlert = hasCloudData && helpers.videoSensing(projectData[0]);
                         } else { // Check for cloud vars only if user is logged out
                             newState.showCloudDataAlert = hasCloudData;
                         }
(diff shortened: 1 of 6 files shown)


src/views/preview/project-view.jsx
```diff
@@ -379,9 +379,10 @@ class Preview extends React.Component {
                     }
 
                     if (showAlerts) {
-                        // Check for username block only if user is logged in
+                        // Check for username and video blocks only if user is logged in
                         if (this.props.isLoggedIn) {
                             newState.showUsernameBlockAlert = helpers.usernameBlock(projectData[0]);
+                            newState.showCloudDataAndVideoAlert = hasCloudData && helpers.videoSensing(projectData[0]);
```
benjiwheeler, line 385: What makes this message go away -- won't it just persist as long as the project includes both cloud vars and video sensing?

benjiwheeler (commented):
A question on what makes this alert go away?

src/views/preview/project-view.jsx
benjiwheeler: Thanks, OK, that makes sense! I think I'll need to try it to see how it feels, but that's what the bug hunt is for!

benjiwheeler (approved):
LGTM -- will need to test in staging to get clearer on overall UX

PR #7205, opened 2022-11-15
CE-229 CSEd Week Banner

tomlum:
Edits copy and changes time for the splash banner for CSEd week.


src/views/splash/l10n.json
```diff
@@ -38,14 +38,16 @@
     "teacherbanner.classesButton": "My Classes",
     "teacherbanner.faqButton": "Teacher Account FAQ",
 
-    "hocbanner.title": "Get Creative with Coding!",
```
benjiwheeler, line 41: What we try to do in these situations is to leave the existing phrase and key in place, and create a new one. That makes more sense with Transifex, so that if we ever want to use the old phrase, its translations are still in place and ready to go.

benjiwheeler (changes requested):
One translations change would be good -- also, I ran the new image through a filesize reducer and will push that change to the branch

src/views/splash/l10n.json
```diff
@@ -38,14 +38,16 @@
...
+    "hocbanner.title": "Tell Your Story with Scratch!",
```
benjiwheeler, line 41:
```suggestion
    "hocbanner.title": "Get Creative with Coding!",
    "hocbanner.titleTellStory": "Tell Your Story with Scratch!",
```

benjiwheeler: And then src/views/splash/hoc/top-banner.jsx line 18 could change from `hocbanner.title` to `hocbanner.titleTellStory`

benjiwheeler:
Is the end time final now?

src/views/splash/l10n.json
benjiwheeler: I think you still need to change src/views/splash/hoc/top-banner.jsx line 18 to use `hocbanner.titleTellStory`?

benjiwheeler (approved):
LGTM!

PR #7253, opened 2022-11-28
Fix gh-7176: Fix empty h5 elements in the homepage

### LINKED ISSUE #7176: Empty <h5></h5> elements occurring on the homepage are not focusable.
### Where To Find Bug
Home page

### Discussion
The screen reader should announce the heading level and its content. There are not currently any of these elements present that a user might wish to focus on. However, in the future there might be, due to the way the <Box> component would handle a `subtitle` prop. Then, the empty heading element would not receive focus, so the screen reader would not announce it. The AT user would not know that it exists on the page.

### System Details

Paul-Clue:
### Description:

Unlabeled and unfocusable header elements should not be present. The box component accepts a prop for the text inside of the h5 element. However, no attribute is passed to the box component for the h5 element in the source code. I added a conditional that checks for the prop and renders the h5 element if a prop exists but does not render the h5 element if the prop does not exist. Resolves #7176.

### Test Cases:
I passed a prop to the box component for the h5 element and the h5 element rendered with the text from the prop.
I did not pass a prop to the box component for the h5 element and the h5 element did not render
After making the changes, I ran `npm test` and found that nothing was broken

### DIFF
--- a/src/components/box/box.jsx
+++ b/src/components/box/box.jsx
@@ -8,7 +8,7 @@ const Box = props => (
     <div className={classNames('box', props.className)}>
         <div className="box-header">
             <h4>{props.title}</h4>
-            <h5>{props.subtitle}</h5>
+            {props.subtitle ? <h5>{props.subtitle}</h5> : null}
             <p>
                 <a
                     href={props.moreHref}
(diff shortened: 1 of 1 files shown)


Paul-Clue:

PR #7261, opened 2022-11-29
Fix gh-7173: The select element in the footer needs proper aria labeling

### LINKED ISSUE #7173: The select element in the footer of every page is not labeled properly for screen reader users
### Where To Find Bug
In the footer of all pages

### Expected Behaviour
The screen reader should announce the accessible name of the element when the element is in tab-focus. 

### Actual Behaviour
The screen reader announces the following:
“English, required menu pop-up collapse, button”.

### Reproduce

- Open https://scratch.mit.edu/ in any browser on your computer
- Turn-on your screen reader of choice
- Press tab until you get to the select form-control element for language located in the footer
- Notice the screen reader gives no indication of what the select element is for

### System Details

Paul-Clue:
### Description:

AT users will have a difficult time figuring out what the select element in the footer is for because there is no proper aria label for it. Resolves #7173.

### Test Cases:
I added an arial-label attribute to the element that gives context for what the element is for to AT users. The screen reader announces the aria-label attribute.

### DIFF
--- a/src/components/languagechooser/languagechooser.jsx
+++ b/src/components/languagechooser/languagechooser.jsx
@@ -1,5 +1,7 @@
 const bindAll = require('lodash.bindall');
 const classNames = require('classnames');
+const injectIntl = require('react-intl').injectIntl;
+const intlShape = require('react-intl').intlShape;
 const PropTypes = require('prop-types');
 const React = require('react');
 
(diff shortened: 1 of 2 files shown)


Paul-Clue:

benjiwheeler (approved):
LGTM! I tested this and it worked great!

Issue #7260, opened 2022-11-29
Select element on the `/explore` page is missing an aria label


Paul-Clue:
### Where To Find Bug
`/explore`

### Expected Behaviour
The screen reader to announce the aria label given to the element.

### Actual Behaviour
The screen reader announces which option is selected and says "menu pop-up collapse item".
### Steps To Reproduce

- Open https://scratch.mit.edu/ in any browser on your computer
- Click on the Explore tab
- Tab through the page until focus lands on the select item with "Trending" as the default option
- Listen to the screen reader say what is stated above for actual behavior

### System Details
macOs Ventura, Chrome Browser, VoiceOver (AT)

PR #7262, opened 2022-11-29
Fix gh-7260: The select element on the `/explore` page needs proper aria labeling

Paul-Clue:
### Description:

AT users will have a difficult time figuring out what the select element on the `/explore` page is for because there is no proper aria label for it.

### Test Cases:
I added an arial-label attribute to the element that gives context for what the element is for to AT users. The screen reader announces the aria-label attribute.

### DIFF
--- a/src/l10n.json
+++ b/src/l10n.json
@@ -1,4 +1,5 @@
 {
+    "general.status": "Filter Projects",
     "general.accountSettings": "Account settings",
     "general.about": "About",
     "general.aboutScratch": "About Scratch",
--- a/src/views/explore/explore.jsx
+++ b/src/views/explore/explore.jsx
@@ -175,6 +175,7 @@ class Explore extends React.Component {
                         </SubNavigation>
                         <Form className="sort-mode">
                             <Select
+                                aria-label={this.props.intl.formatMessage({id: 'general.status'})}
                                 name="sort"
                                 options={[
                                     {


Paul-Clue:

corihudson:
Thanks for this! I'm noticing that this doesn't seem to work with VoiceOver on macOS 12.3 and Firefox 106 - is that a known issue?

Paul-Clue:
Hey, @corihudson, I am not sure if you are referring to the fix not working on VoiceOver and Firefox. But in case you are, run `npm run translate` then refresh your browser. It should work after that. If this is not the problem that you are referring to, please provide me with some more context.

corihudson:
@Paul-Clue Yes, I haven't been able to see the fix working in VoiceOver and Firefox, even after running `npm run translate`. Confirmed that the fix does work in Chrome, but not in Firefox.

Chrome:
Screen Shot 2022-12-14 at 10 41 24 AM

Firefox:
Screen Shot 2022-12-14 at 10 40 37 AM

Paul-Clue:
Hey, @corihudson, I am still looking into the problem that you found.

Paul-Clue:
Ok it turns out that adding a `<label>` attribute is the best, most universal solution for this, but it will take some design work to fit it in. The `aria-label` solution will have to do for now even though we know it will not work on all browsers. I will file another issue calling for adding the `<label>` element. 

However, if you don't want to accept this PR, I will update the existing issue to recommend adding an html `<label>` element to resolve the issue.

corihudson:
@Paul-Clue Thanks for the update! We'll merge this PR for now and ticket future work to improve the experience in Firefox.

corihudson (approved):

Issue #7137, opened 2022-10-24
CE-294 Make the love and favorite icons colorblind friendly


DNin01:
### The problem
Currently, colorblind people can have a hard time telling if they have loved/favorited a project because there isn't a major or consistent brightness difference between the inactive and active love/favorite buttons. See this comparison of the buttons when you haven't clicked them and when you have:
![A comparison between the love and favorite icons when you haven't clicked them and when you have clicked them](https://user-images.githubusercontent.com/106490990/197437692-397ba8de-d069-48ad-9fff-ea9ea71aee82.png)
I grayscaled this image and the loved icon was ever so slightly darker than the unloved icon and the favorited icon was a bit brighter than the unfavorited icon.
![The same image but in black and white](https://user-images.githubusercontent.com/106490990/197438258-8c5157b5-9f46-4f8c-9d7c-f90e6446454e.png)

### The solution
Make the inactive love/favorite icons an outline and fill them in when active.
![The top two icons have only a border. The next two icons are filled and colored in.](https://user-images.githubusercontent.com/106490990/197439070-6a9ad771-c02e-4825-8e12-a6b5bf981d6c.png)

edelfino11:
Hi @DNin01 @benjiwheeler I'm a computer science student, and we are currently working towards contributing to an open source GitHub for one of our classes. I would love to take this issue and work towards solving it! Is it ok if I take this issue to work on it? I'm not sure if anyone else is already assigned to this issue, so asking just in case.

DNin01:
> Is it ok if I take this issue to work on it? I'm not sure if anyone else is already assigned to this issue, so asking just in case.

@edelfino11 Don't mind if you do 🙂

benjiwheeler:
@edelfino11 Thanks so much for offering to do this change! We're checking internally with some folks working on accessibility design, and we'll get back to you!

edelfino11:
No problem! @benjiwheeler I figure it might be fixed by simply not filling the svg graphic with the gray color. Also, do you happen to know how to make the main page load properly when I run it locally? When I npm start the repo locally, it looks like this:

<img width="992" alt="Screen Shot 2022-11-27 at 2 24 51 PM" src="https://user-images.githubusercontent.com/55937173/204155683-8083eadd-47b8-473c-b260-a9b2c1efeac7.png">

So I can't check if the change I made fixes the problem. If there's a slack, I can also ask there instead! Sorry for the inconvenience!

DNin01:
[...]

PR #7264, opened 2022-12-01
CE-294 Fix(7137) Make the love and favorite icons colorblind friendly 

edelfino11:
Resolves: 

This fix solves issue [7137](https://github.com/LLK/scratch-www/issues/7137), simply changed the love and favorite svgs to the colorblind friendly svg on the stats.scss

Passing all tests: 
Screen Shot 2022-12-01 at 12 41 28 PM


DNin01 (changes requested):
It looks like the activated love button was changed to an outlined favorite icon. You can see that the second part you changed is under `.project-loves.loved {` which refers to the love icon when you loved a project, so that should stay what it was before. `background-image: url("/svgs/favorite/favorite_type-gray.svg");` should be below `.project-favorites {` instead of `.project-loves.loved {`.

Minor detail, these outlined icons are shaped slightly differently than the filled-in icons:
 
It might look more consistent if instead of changing what image is used, you modified the grayed-out icons ([`love-gray.svg`](https://github.com/LLK/scratch-www/blob/develop/static/svgs/project/love-gray.svg) and [`fav-gray.svg`](https://github.com/LLK/scratch-www/blob/develop/static/svgs/project/fav-gray.svg)) to have an outline using an SVG editor.

edelfino11:
Oh, I did not realize the subtle difference! I completely changed the SVGs and fixed the small mistake I had before, sorry about that.

DNin01 (approved):
Looks great!

benjiwheeler:
@edelfino11, this change looked great! Our design team took a look and made some tiny tweaks, and made the outline gray a little darker (and changed the remix and views icons too). Thanks for collaborating on this with us!

DNin01:
👏

Racecar-Roaster:
This is really cool and a nice feature! But I think it could be better if you make it just the thick outline so it's more colorblind friendly!

PR #7272, opened 2022-12-03
CE-273: Heading levels should increase by one.

### LINKED ISSUE #7177: CE-273 Ensure heading levels increase by one.
### Where To Find Bug
`/ideas`, `/about`, `/messages`, `/mystuff`, `/accounts`

### Expected Behaviour
The WCAG standard for decreasing heading levels (headers should decrease by one)  help screen reader users better understand a page's content. For example, every page should have only one h1 level heading that signifies what the content on the entire page is about. h2 headers should come after and represent the main sections of the main idea of the page.

### Actual Behaviour
While tabbing through the page, the screen reader user will notice that some heading levels are not in a proper order, leading to possible disorientation and confusion about the content.

### Steps To Reproduce

- Open https://scratch.mit.edu/ in any browser on your computer
- Turn-on your screen reader of choice
- Start tabbing through the page.
- Notice that the heading levels are not in the correct order

### System Details

Paul-Clue:
### Description:

I changed some header tags on the `/ideas` and `/about` pages. I put the headings in the proper order so that screen reader user can better understand the content on those pages. Resolves #7177.

### Test Cases:
No new tests have been added. All current automated tests still pass and nothing breaks in the original design of the application.

### DIFF
--- a/src/components/ttt-tile/ttt-tile.scss
+++ b/src/components/ttt-tile/ttt-tile.scss
@@ -73,6 +73,7 @@
 .ttt-tile-title {
     margin: .5rem auto;
     width: calc(100% - 3rem);
+    font-size: 1rem;
 }
 
 .ttt-tile-description {
@@ -103,8 +104,4 @@
     width: 1rem;
     height: 1rem;
     vertical-align: text-bottom;
-}
-
-.ttt-tile-title {
-    font-size: 1rem;
 }
--- a/src/components/ttt-tile/ttt-tile.scss
+++ b/src/components/ttt-tile/ttt-tile.scss
@@ -106,5 +106,5 @@
 }
 
 h3 {
-    font-size: 1em;
+    font-size: 1rem;
 }
--- a/src/components/ttt-tile/ttt-tile.scss
+++ b/src/components/ttt-tile/ttt-tile.scss
@@ -104,3 +104,7 @@
     height: 1rem;
     vertical-align: text-bottom;
 }
+
+h3 {
+    font-size: 1em;
+}
--- a/src/components/ttt-tile/ttt-tile.scss
+++ b/src/components/ttt-tile/ttt-tile.scss
@@ -105,6 +105,6 @@
     vertical-align: text-bottom;
 }
 
-h3 {
+.ttt-tile-title {
     font-size: 1rem;
 }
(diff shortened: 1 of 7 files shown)


Paul-Clue:
an @chrisgarrity review this pull request when time permits.

src/l10n.json
```diff
@@ -88,6 +88,7 @@
     "general.signIn": "Sign in",
     "general.startOver": "Start over",
     "general.statistics": "Statistics",
+    "genral.story": "Create a Story",
```
benjiwheeler, line 91: Is this string necessary? I don't see it used in the code. (Also, "general is misspelled :) )

benjiwheeler (changes requested):
This mostly looks good to me, one question about the "Create a Story" string

src/l10n.json
Paul-Clue: That string was not necessary. I removed it.

src/views/about/about.scss
```diff
@@ -47,6 +47,10 @@
         padding: 26px 40px 6px;
         width: calc(100% - 80px);
 
+        h2 {
```
benjiwheeler, line 50: Just trying to understand this -- is the h2 style used anywhere? Is this just a preemptive change for the future?

benjiwheeler (commented):
One more question

src/views/about/about.scss
Paul-Clue: That styling is for `<h2>`s that I used to replace the `<h3>`s on the page. Some how the changes that I made to the `<h3>`s did not get through. I spent some time to try to figure out why, but I could not find out why those changes did not make it upstream.

I just push those necessary changes, and they are there now.

src/components/ttt-tile/ttt-tile.scss
```diff
@@ -104,3 +104,7 @@
     height: 1rem;
     vertical-align: text-bottom;
 }
+
+h3 {
+    font-size: 1rem;
+}
```
mxmou, line 110: Won't this apply to all `h3` elements on the page, not only those in the `ttt-tile` component? It should probably be added to the `.ttt-tile-title` class instead.

Paul-Clue: Hey, @mxmou, thanks for catching that. I made the requested change.

benjiwheeler:
LGTM!

Issue #6423, opened 2022-01-04
CE-310 Title Text on EV3 Page Overlaps


leahcimto:
### Expected Behavior
The text "EV3" in the title should be next to "LEGO MINDSTORMS"

### Actual Behavior
Text "EV3" overlaps with body text in fullscreen mode.

### Steps to Reproduce
1. Go to https://scratch.mit.edu/ev3/
2. Notice that the text EV3 overlaps with the body text

### System Details
macOS 11.5.2, Firefox 95.0.2, MacBook Air

**Screenshots**

<img width="1440" alt="screenshot" src="https://user-images.githubusercontent.com/68165163/148123574-a9a08387-aee0-46cb-ba9c-29b4435d6558.png">

GrahamSH-LLK:
Seems to be a firefox only bug

mxmou:
Could be fixed by adding `width: 100%` to the `h1`.

kccuber-scratch:
This happens on Windows too
![image](https://user-images.githubusercontent.com/92891642/173246310-c289f878-8fbb-4888-89e2-116884a47c25.png)

Firefox 100, Windows 10

GrahamSH-LLK:
> Hi @GrahamSH-LLK @kccuber-scratch I'm a computer science student, and we are currently working towards contributing to an open source GitHub for one of our classes. I would love to take this issue and work towards solving it!

Hi @edelfino11, sorry for the misunderstanding. Kcuber and I are not members of the scratch team, we're actually members of the scratch community. This issue is not marked [help wanted](https://github.com/LLK/scratch-www/labels/Help%20wanted) so the scratch team _probably_ won't accept PRs for this, but you could try. If you want to try your luck I'd say ping somebody who has assigned labels to an issue recently.

edelfino11:
Hi @GrahamSH-LLK, I see! Sorry about that.

edelfino11:
Hi @chrisgarrity, I created a small PR to fix this issue! I know it is not on the help wanted list, but it's a simple fix that can impact those using Firefox. (I was let known by others in this issue to tag an actual member of the scratch team, so here we are!) Thank you for your time!

benjiwheeler:
Thanks, @edelfino11 -- we should have marked this help wanted!

benjiwheeler:
Resolved by https://github.com/LLK/scratch-www/pull/7291

PR #7291, opened 2022-12-07
CE-310 Fixes title overlap of EV3 for Firefox 

edelfino11:
Fixed the title's overlapping issue that was happening on Firefox, this fixes issue [#6423.](https://github.com/LLK/scratch-www/issues/6423) 

I tried changing this issue before, but I was working on the wrong branch. I tried to undo those commits, but I wasn't able to delete them from the issue's history. Sorry for all the confusion! 

I also added some text to one of the images, since I've seen a lot of issues regarding accessibility.  

Screen Shot 2022-12-11 at 4 51 54 PM

### DIFF
--- a/src/components/extension-landing/extension-landing.scss
+++ b/src/components/extension-landing/extension-landing.scss
@@ -106,6 +106,7 @@
                     display: flex;
                     margin-bottom: 2rem;
                     color: $ui-white;
+                    width: 100%;
 
                     img {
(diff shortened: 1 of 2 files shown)


mxmou:
I think the empty alt text was intentional because the image is decorative.

benjiwheeler (approved):
Thanks, @edelfino11!

PR #7314, opened 2022-12-19
Adds Scratch Credits (Dec 2022)

tomlum:
Adds new scratch team members as of December 2022 to the /credits page.

Also adds automatic alphabetic sorting to people.json.


src/views/credits/credits.jsx
```diff
@@ -8,6 +8,9 @@ const PeopleGrid = require('../../components/people-grid/people-grid.jsx');
 
 require('./credits.scss');
 
+// Sort people by name alphabetically
+People.sort((a, b) => (a.name.localeCompare(b.name)));
```
corihudson, line 12: this is a nice change :D

corihudson (changes requested):
Needs one quick change (I'm still listed in past contributors) but otherwise lgtm!

src/views/credits/credits.jsx
```diff
@@ -114,80 +117,22 @@ const Credits = () => (
...
-                    Lis Sylvan,
-                    Bryce Taylor,
-                    Hanako Tjia,
-                    Jaleesa Trapp,
-                    Moran Tsur,
-                    Claudia Urrea,
-                    Kathy Wu,
-                    Julia Zimmerman,
-                    Oren Zuckerman.
+                    Yusuf Ahmad, Ben Berg, Amos Blanton, Paula Bontá, Karen Brennan,
```
benjiwheeler, line 120: I think we should go back to putting names one per line -- it makes for much clearer diffs. 

My guess is that this was an unintentional change, because the same change also removed a number of names from this list, I think unintentionally. The change may have started with an out of date copy of the file.

src/views/credits/credits.jsx
benjiwheeler: I second this, great change!

PR #7348, opened 2023-01-23
feat: upgrade react-intl

aoneill01:
### Resolves:

- Resolves: [ENA-205](https://scratchfoundation.atlassian.net/browse/ENA-205)

### Changes:

React 18 is not compatible with `react-intl` version 2. As a precursor for upgrading React, this upgrades `react-intl` to version 5.

- Upgrade `react-intl`
- Replace `FormattedHTMLMessage` with `FormattedMessage`
  - Some localization messages need to be updated because of this change. So nothing will break during this transition, new ids were created for these messages. A script will be run to programmatically transform these messages and upload them to Transifex outside of this pull request. _Note that this PR will not be merged until these Transifex changes are ready._
- Use polyfill for `Intl` on old browsers
  - Change order of loading `intlPolyfill` so that it is available at the time of rendering
  - `addLocaleData` has been [removed](https://formatjs.io/docs/react-intl/upgrade-guide-3x)
- Replace `intlShape` since it was [removed](https://formatjs.io/docs/react-intl/upgrade-guide-3x)
- Replace `FormattedRelative` with [`FormattedRelativeTime`](https://formatjs.io/docs/react-intl/upgrade-guide-3x#formattedrelativetime)
  - Removed one instance that did not appear to be used

### Test Coverage:

- There should be almost no change in behavior, so the existing tests are still relevant. Some of the unit tests needed to be modified to use `react-intl` correctly.
- Manual testing of polyfilled `Intl` library was performed with Firefox version 60 on Windows.

[ENA-205]: https://scratchfoundation.atlassian.net/browse/ENA-205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/views/camp/l10n.json
```diff
@@ -3,19 +3,24 @@
...
     "camp.part3Particpate":"The <a href=\"https://scratch.mit.edu/studios/4160301/\">Final Projects Camp Cabin studio</a> will hold part 3 of this year's Scratch Camp. Here you can submit your final project, give feedback to others, and celebrate Scratch Camp! Swim on over to the studio when part 3 comes out!",
+    "camp.part3ParticpateHTML":"The <a>Final Projects Camp Cabin studio</a> will hold part 3 of this year's Scratch Camp. Here you can submit your final project, give feedback to others, and celebrate Scratch Camp! Swim on over to the studio when part 3 comes out!",
     "camp.helpfulInfo": "Helpful Information",
     "camp.infoCounselors": "The <a href=\"https://scratch.mit.edu/studios/4160300/\">Camp Counselors studio</a> offers a variety of examples for your ocean creation. You can also directly communicate with the Counselors there.",
+    "camp.infoCounselorsHTML": "The <a href=\"https://scratch.mit.edu/studios/4160300/\">Camp Counselors studio</a> offers a variety of examples for your ocean creation. You can also directly communicate with the Counselors there.",
```
corihudson, line 23: Is there a reason the other new messages remove the `href` attribute, but not this one?

corihudson (commented):
still need to try running this locally but these look like really great changes overall!

src/lib/intl.jsx
```diff
@@ -1,7 +0,0 @@
-const ReactIntl = require('react-intl');
-
-// Add locale data to react intl for all supported languages
-const localeData = require('scratch-l10n').localeData;
-ReactIntl.addLocaleData(localeData);
```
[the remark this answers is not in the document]
aoneill01: By custom, do we have some non-standard locales? Can you give me an example so I can try it out?

For older browsers, the required `Intl` functionality is polyfilled in `src/lib/intl-polyfill.js`. I have also tested that the non-updated `react-intl` in `scratch-gui` seems to play nicely. It still calls `addLocaleData` in `scratch-gui/src/reducers/locales.js`.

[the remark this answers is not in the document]
apple502j: `ja-Hira` is an example, so are `ab`, `an`, `ht`, `oc`, and `rap` I believe. Chinese is also a bit complicated, since we use `zh-cn` and `zh-tw` instead of `zh-Hans` and `zh-Hant`.

src/views/camp/l10n.json
aoneill01: I missed that one. Pushed an update.

src/lib/intl.jsx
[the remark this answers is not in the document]
corihudson: @aoneill01 I found a nice comparison by switching to Aragonés (`an` locale).

this is what's expected - note the date on the comment reading "hace 5 años":
Screen Shot 2023-01-31 at 2 28 14 PM 

this is what I see on my local machine - the date on the comment has defaulted back to English (although the formatting is still off):
Screen Shot 2023-01-31 at 2 27 51 PM

I also get a `Missing locale data for locale: "an" in Intl.NumberFormat. Using default locale: "en" as fallback.` error in the console.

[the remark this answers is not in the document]
aoneill01: @corihudson - I have tried to address these issues in the latest commits.

- Thanks for catching that the auto unit selection was only going up to "days." I considered some different options and ended up writing custom logic to select the appropriate unit of time. Created a `RelativeTime` component to encapsulate this.
- Even in browsers that natively support `Intl`, sometimes locale data is missing for some of our locales. The missing ones vary from browser to browser. Updated the polyfilling logic in `intl-polyfill.js`.
- Some of the locales were not available in the polyfill library. Locales like `ab`, `an`, `ht`, `oc`, and `rap` are swapped out for alternative locales when passed to the `IntlProvider`. Some code locations that were using the locale name from the `IntlProvider` were updated to get the true value instead.

I have compared relative times from these changes to production. For the most part it appears to match. I noted a few differences. I have discussed these with Chris G. None of these were deal breakers for her, and future updates to `Intl` and the polyfill libraries may improve the situation.

- Some relative time translations improved due to better polyfilled locale data, such as Hausa (ha) and کوردیی ناوەندی  (ckb).
- This is likely browser dependent, but I had a couple of locales that were natively supported by my browser. However, the result of the relative time translations was inferior to production. Azeri (az) and Oʻzbekcha (uz) used single-letter placeholders for the time unit.
- Sepedi (nso) and Setswana (tn) fell back to English because there was not polyfill data. Not a big loss since production just shows single-letter placeholders for time units.

src/lib/intl-polyfill.js
```diff
@@ -19,293 +21,458 @@ import {shouldPolyfill as shouldPolyfillPluralRules} from '@formatjs/intl-plural
...
         break;
     case 'kk':
-        await import('@formatjs/intl-relativetimeformat/locale-data/kk');
         await import('@formatjs/intl-pluralrules/locale-data/kk');
+        await import('@formatjs/intl-relativetimeformat/locale-data/kk');
+        await import('@formatjs/intl-numberformat/locale-data/kk');
+        await import('@formatjs/intl-datetimeformat/locale-data/kk');
+        break;
+    case 'qu':
+        await import('@formatjs/intl-pluralrules/locale-data/en');
```
corihudson, line 280: is there a reason the other imports here use `qu` but this one uses `en`?

src/lib/intl-polyfill.js
```diff
@@ -19,293 +21,458 @@ import {shouldPolyfill as shouldPolyfillPluralRules} from '@formatjs/intl-plural
...
     case 'hu':
-        await import('@formatjs/intl-relativetimeformat/locale-data/hu');
         await import('@formatjs/intl-pluralrules/locale-data/hu');
+        await import('@formatjs/intl-relativetimeformat/locale-data/hu');
+        await import('@formatjs/intl-numberformat/locale-data/hu');
+        await import('@formatjs/intl-datetimeformat/locale-data/hu');
         break;
     case 'mi':
-        await import('@formatjs/intl-relativetimeformat/locale-data/mi');
         await import('@formatjs/intl-pluralrules/locale-data/en');
```
corihudson, line 322: same here - why `en` and not `mi`?

corihudson (approved):
Couple quick notes, but this looks great to me! I noticed a slight behavior difference between this branch and prod (a comment that was posted in June 2018 is shown as 5 years old on prod but 4 years old on this branch), but I would argue that this branch's behavior is more correct.

src/lib/intl-polyfill.js
aoneill01: Good question. Despite using the latest version of `intl-pluralrules`, `intl-relativetimeformat`, `intl-numberformat`, and `intl-datetimeformat`, they don't support the same set of locales. This is a case of `qu` missing from `intl-pluralrules` but not the others. When `react-intl` can't find a locale, it looks like it falls back to `en`, so let's import `en` just to be safe.

I pretty much copied this behavior from our previous code. I hope I'm interpreting the intent correctly. The line you pointed out for `mi` existed before this PR. 

I didn't want to do this all by hand. I wrote a script to do it. Posting the important parts here in case it helps someone in the future.

```js
const {supportedLocales: pluralRulesSupportedLocales} = require('@formatjs/intl-pluralrules/supported-locales.generated');
const {supportedLocales: relativeTimeFormatSupportedLocales} = require('@formatjs/intl-relativetimeformat/supported-locales.generated');
const {supportedLocales: numberFormatLocales} = require('@formatjs/intl-numberformat/supported-locales.generated');
const {supportedLocales: dateTimeFormatLocales} = require('@formatjs/intl-datetimeformat/supported-locales.generated');

const languages = {
...
};

for (const locale of Object.keys(languages).map(l => l.toLowerCase().split('-')[0])) {
    console.log(`    case '${locale}':`);
    console.log(`        await import('@formatjs/intl-pluralrules/locale-data/${translateLocale(locale, pluralRulesSupportedLocales)}');`);
    console.log(`        await import('@formatjs/intl-relativetimeformat/locale-data/${translateLocale(locale, relativeTimeFormatSupportedLocales)}');`);
    console.log(`        await import('@formatjs/intl-numberformat/locale-data/${translateLocale(locale, numberFormatLocales)}');`);
    console.log(`        await import('@formatjs/intl-datetimeformat/locale-data/${translateLocale(locale, dateTimeFormatLocales)}');`);
    console.log('        break;');
}

function translateLocale(locale, supportedLocales) {
    return supportedLocales.includes(locale) ? locale : 'en';
}
```

corihudson: makes sense! just wanted to make sure I understood what was going on there. thanks!

PR #7353, opened 2023-01-27
fix: remove Sentry and rewrite ErrorBoundary tests

cwillisf:
### Resolves:

Half of ENA-224

### Changes:

We're no longer planning to use Sentry in this way, so this change removes all references to Sentry (the library) throughout scratch-www's code, README, and Webpack config. Note that Sentry (the organization) is still mentioned in the content of some pages.

Sentry support was tightly integrated with the ErrorBoundary component, which would report an error to Sentry and then display a crash message that includes the Sentry event ID. Now, ErrorBoundary skips sending the error to Sentry but still displays the crash message.

### Test Coverage:

The `ErrorBoundary` test sequence has been rewritten to verify that it switches from displaying children to displaying a crash message when it catches an error. Previously, the test verified that the component would call a mock Sentry interface, but did not verify user-facing behavior.

This is a little tricky to test manually. One way I've found:
1. run `npm start`
2. visit `http://localhost:8333//explore/projects/all`
3. open the dev console
4. search for `'masthead'` (with single quotes) in `explore.jsx`
5. click the `{}` button to pretty-print the file
6. find the `render()` function near `'masthead'` (for me, it's on line 197 of the pretty-printed file)
7. put a breakpoint on the `return React.createElement(` line (`React` might be a single-letter variable on minified builds)
8. reload the page so that the breakpoint hits
9. type `React=null` in the console (replace `React` with the appropriate single-letter variable if using a minified build)
10. press the "continue" button


src/components/errorboundary/errorboundary.jsx
```diff
@@ -9,39 +8,47 @@ class ErrorBoundary extends React.Component {
...
+        error = error || {
+            stack: 'Unknown stack',
+            message: 'Unknown error'
+        };
+        errorInfo = errorInfo || {
+            componentStack: 'Unknown component stack'
+        };
+
+        // only remember the first error: later errors might just be side effects of that first one
+        if (!this.state.error) {
```
ckilroy, line 31: Is there ever a case where there would be multiple distinct errors (and it would matter)?

src/components/errorboundary/errorboundary.jsx
```diff
@@ -9,39 +8,47 @@ class ErrorBoundary extends React.Component {
...
 
     handleBack () {
         window.history.back();
     }
 
     render () {
-        if (this.state.hasError) {
+        if (this.state.error) {
             return (
                 <CrashMessageComponent
```
ckilroy, line 51: I don't see edits to `CrashMessageComponent` do we need to do any updates if that component is expecting `eventId`?

ckilroy (approved):
Approved with some questions. Not sure I have full context, but LGTM generally.

src/components/errorboundary/errorboundary.jsx
cwillisf: I think it's most likely that we'd have just one error: after the first error, the `ErrorBoundary` component will stop rendering its normal children and swap in the `CrashMessageComponent` instead. After that, it seems likely that further errors would either be a problem in `CrashMessageComponent` (which would be REALLY unfortunate, but seems unlikely) or a problem in an async callback or timer or something like that, which wouldn't be caught by `ErrorBoundary` anyway. I'm not 100% confident that those are the only options, which is why the `log.error` reports every error 😅

src/components/errorboundary/errorboundary.jsx
cwillisf: Fair question! `eventId` is an optional property in `CrashMessageComponent`, and it has explicit handling for the case where it's absent. I thought about removing it, but it's currently harmless to leave it and we might want it again in the near(?) future.

PR #7358, opened 2023-01-31
Make project stats icons always opaque

benjiwheeler:
### Resolves:

follow-on to https://github.com/LLK/scratch-www/pull/7264

### Changes:

sets opacity of the new project page stats icons to default of 100% (instead of 50%), to match the intention behind their design.

### DIFF
--- a/src/views/preview/stats.scss
+++ b/src/views/preview/stats.scss
@@ -55,8 +55,6 @@
     cursor: pointer;
 
     &:before {
-        
-        opacity: .5;
         background-image: url("/svgs/project/love-gray.svg");
     }
 }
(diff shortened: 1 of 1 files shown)


src/views/preview/stats.scss
```diff
@@ -55,16 +55,13 @@
...
-        
-        opacity: .5;
         background-image: url("/svgs/project/love-gray.svg");
     }
 }
 
 .project-loves.loved {
 
     &:before {
-        opacity: 1;
```
benjiwheeler, line 67: these can be removed, since the opacity isn't .5 to start with

tomlum (approved):
LGTM! 🤘

PR #7387, opened 2023-02-22
fix: use span for localized messages

aoneill01:
### Resolves:

- Fixes [ENA-205](https://scratchfoundation.atlassian.net/browse/ENA-205)

### Changes:

Part of the `react-intl` upgrade includes the following:

> Change default textComponent in IntlProvider to React.Fragment. In order to keep the old behavior, you can explicitly set textComponent to span.

This broke some [integration tests](https://app.circleci.com/pipelines/github/LLK/scratch-www/10855/workflows/7f6b8f52-d6b4-474c-b749-00bfdc494e67/jobs/13439) that were specifically targeting the `<span>`. Used the `textComponent` property to restore original behavior. This should also ensure any css that was depending on the spans also works.

### Test Coverage:

Unit tests still pass.

[ENA-205]: https://scratchfoundation.atlassian.net/browse/ENA-205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/render.jsx
+++ b/src/lib/render.jsx
@@ -39,6 +39,7 @@ const render = (jsx, element, reducers, initialState, enhancer) => {
                 <IntlProvider
                     locale={intlLocale}
                     messages={messages}
+                    textComponent="span"
                 >
                     {jsx}
                 </IntlProvider>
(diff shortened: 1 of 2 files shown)


corihudson (approved):
lgtm! i'm curious what the motivation/benefits of using React.Fragment instead of span in the updated react-intl are, but that's definitely not a blocker for this

PR #7415, opened 2023-03-04
Change messages badge color

### LINKED ISSUE #7256: Messages badge (number of messages) can be hard to read
### The Problem

The messages badge is light orange with white text for the number of messages, which can make the text hard to read for some people with vision disabilities (and sometimes in general).

### The Solution

Here are a few other color options I came up with (the leftmost is the current one):
White text on medium-light orange, black text on medium-light orange, white text on medium-light blue, black text on sky blue, white text on medium-dark blue

### Steps to Reproduce

1. Receive a message on Scratch (new comment, love, etc.)
2. Go to any page with the messages button on the navigation bar (e.g. https://scratch.mit.edu/)
3. Look at the messages badge

DNin01:
### Resolves:

Resolves #7256

### Changes:

To improve the text contrast, I changed the background color of the messages badge from light orange to white for the navigation bar and light blue for the messages list, and also changed the text to dark gray.

### Test Coverage:

I don't have the required software to test - sorry about that. I'm sure it'll work, though!

Here's what it looks like:
The Messages page with 1 unread message with updated colors for the messages badge both on the navigation bar and the list of messages

### DIFF
--- a/src/components/navigation/www/navigation.scss
+++ b/src/components/navigation/www/navigation.scss
@@ -147,11 +147,11 @@
                 top: .5rem;
                 right: .25rem;
                 border-radius: 1rem;
-                background-color: $ui-orange;
+                background-color: $background-color;
                 padding: 0 .25rem;
                 text-indent: 0;
                 line-height: 1rem;
-                color: $type-white;
+                color: $type-dark-gray;
                 font-size: .7rem;
                 font-weight: bold;
             }
--- a/src/views/messages/messages.scss
+++ b/src/views/messages/messages.scss
@@ -51,10 +51,10 @@
 .messages-header-unread {
     margin-left: 1rem;
     border-radius: 1rem;
-    background-color: $ui-orange;
+    background-color: $ui-blue-25percent;
     padding: .25rem .5rem;
     line-height: 1rem;
-    color: $type-white;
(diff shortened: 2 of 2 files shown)


Secret-chest:
Please not. It's supposed to stand out. Keep it orange and make the text black.

lemystic:
Why not use the second example from [your issue](https://github.com/LLK/scratch-www/issues/7256) (#FFAB1A and #000000)? It stands out and also complies with every single WCAG guideline:
image

DNin01:
I want to try that because orange there pops out more, but I've run into a few problems with using orange with black text. First, [there is no `#000000` color available](https://github.com/LLK/scratch-www/blob/b154e401e3565e37e167618b55eed4abea50ab61/src/_colors.scss); and second, orange to blue has a contrast ratio of 2.09, which fails all the WCAG guidelines. While orange catches your attention a little more, white stands out more _against the background_, especially for colorblind users.

Maybe we can ignore the orange badge problem if we use accessible text on it, though, but this would work better if I could use a `#000000` text color, and this project seems to be sticking to using only colors from the palette linked above.

Secret-chest:
Maybe you could ask them to add a light orange shade?

PR #7457, opened 2023-03-16
Revert "Force Account Rename Design Check"

tomlum:
Reverts LLK/scratch-www#7455


8-6-7-5-3-0-9-1-9-8-5-0-5-1-5-2-0-6-0-7:
@tomlum Hey a couple of questions regarding this change.
1) Is this the entire new /accounts/banned-response screen or just for renames? If just for renames, how do you tell them apart?
2) Where are you getting the data from to show on the page? It looks like this is just from the entire "user" object, but that doesn't have the data that the scratchr2 page does, and has a lot of irrelevant stuff like the email address
3) There is a typo "usernamed" instead of "username". However, I would guess that the hardcoded text is temporary as Scratch usually uses translation strings?

PR #7618, opened 2023-05-25
hotfix: make banner aside use sticky position

corihudson:
### Resolves:
- Fixes an issue where the privacy policy link did not work in Chrome.

### Changes:

- Adds `position: sticky` to the `aside` element containing the banner to prevent the margin of the main page `div` from blocking the link.


aoneill01 (approved):
👍

PR #7629, opened 2023-05-28
Fix cookie expiring in session

mrsrec:
The scratchpolicyseen cookie is supposed to expire after 3 weeks, but due to a typo in the expiry time, it doesn't have a valid expiry, and falls back to expiring in Session. This has led to a lot of complaints by people about closing the popup but seeing it again later. This fixes that problem

### DIFF
--- a/src/components/privacy-banner/privacy-banner.jsx
+++ b/src/components/privacy-banner/privacy-banner.jsx
@@ -35,7 +35,7 @@ class PrivacyBanner extends React.Component {
 
     handleCloseBanner () {
         const opts = {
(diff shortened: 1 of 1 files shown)


mrsrec:
Tag @corihudson as the author of the code for popup. Is this OK?

Scratchfangs:
@corihudson Can you please merge?

aoneill01 (approved):
Thanks @mrsrec!

I tried this locally and I can see the expiration date being correctly set. I am going to merge this in now so it is included in this week's release.

PR #7636, opened 2023-06-05
feat: update cookies page with scratchtheme info

aoneill01:
### Resolves:

- [ENA-326](https://scratchfoundation.atlassian.net/browse/ENA-326)

### Changes:

- Adds entry to cookies page about the `scratchtheme` cookie
- Updates the last updated message for the cookies page
  - There is a change in how this message works. It used to be its own entry in l10n.json, including the date. It is on the translators to localize this message with the date. Now it has a placeholder for the date in the message, with `react-intl` handling the localization of the date. The translators will have to localize this new formatted message, but then any future date changes require no additional work from the translators. I believe this could be a good change for the Privacy Policy page as well the next time anyone touches it.
  - The date may not show up as June 28th if, in your timezone, it was a different date at noon Eastern. However, I feel this is more accurate anyways.

### Test Coverage:

- Manual testing
- Added an integration test for the Cookies page that verifies the last updated message


cwillisf (approved):
Looks good! I like the date solution.

PR #7637, opened 2023-06-05
Update cookies.jsx

mrsrec:
This corrects inaccurate information in the cookie page. The name of the cookie is scratchsessionsid, not scratchsessionid, and it lasts for 2 weeks, not a session

### DIFF
--- a/src/views/cookies/cookies.jsx
+++ b/src/views/cookies/cookies.jsx
@@ -70,10 +70,10 @@ const Cookies = injectIntl(() => (
                                     <td><FormattedMessage id="cookies.table.years.1" /></td>
                                 </tr>
                                 <tr>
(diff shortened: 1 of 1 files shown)


aoneill01 (commented):
@mrsrec, would you mind updating the "last updated" for this page? I think this is branched before some changes to how this is handled, so I believe you will have to:

1. Merge in `develop` or rebase this branch onto `develop`,
2. Around line 16 should look like `const updated = new Date('2023-06-28T12:00:00.000-04:00');`. Change the date portion of that string to 07-06 which should be the next release date.

mrsrec:
OK. I will open a new request with those fixed.

mrsrec:
New PR https://github.com/scratchfoundation/scratch-www/pull/7693/

PR #7650, opened 2023-06-13
hotfix: add banner for account creation outage

corihudson:
### Changes:

adds outage banner for logged out users


corihudson:
not needed anymore 🎉

PR #7698, opened 2023-06-29
Fit thumbnail instead of stretching it

lankybox02:
This one-line update would change the way studio images display - it tries to contain most of the image instead of just stretching it so that it fits. For example, when an uploaded thumbnail is wider than the ratio of the default studio thumbnail, the image won't stretch its width and make it look distorted, instead, it will just crop the edges so that the center of the image can be viewed.

### DIFF
--- a/src/views/studio/studio.scss
+++ b/src/views/studio/studio.scss
@@ -167,6 +167,7 @@ $radius: 8px;
         background: white;
         box-sizing: border-box;
         border: 2px solid $box-shadow-light-gray;
+        object-fit: cover;
     }
 
     .studio-follow-button {


Secret-chest:
`object-fit: contain` would work better, to avoid cropping the image.

lankybox02:
Nah, because then, there would be blanks in the thumbnail and it would be hard for studio owners to make thumbnails that perfectly align with the studio width and height. This is the same case for profile pictures - they get cropped.

PaperMarioFan-2022:
Good idea! I can see that fitting studio thumbs is better in some cases though.

Secret-chest:
> Nah, because then, there would be blanks in the thumbnail and it would be hard for studio owners to make thumbnails that perfectly align with the studio width and height. This is the same case for profile pictures - they get cropped.

And if they add text to the sides, it gets cropped.

lankybox02:
> > Nah, because then, there would be blanks in the thumbnail and it would be hard for studio owners to make thumbnails that perfectly align with the studio width and height. This is the same case for profile pictures - they get cropped.
> 
> And if they add text to the sides, it gets cropped.

That isn't usually good design practice. The standard is to leave a bit of space on each side, and the actual content should be on the center

Secret-chest:
> Nah, because then, there would be blanks in the thumbnail and it would be hard for studio owners to make thumbnails that perfectly align with the studio width and height. This is the same case for profile pictures - they get cropped.
>  > And if they add text to the sides, it gets cropped.
> 
> That isn't usually good design practice. The standard is to leave a bit of space on each side, and the actual content should be on the center

That's what I was saying.

lankybox02:
> > > > Nah, because then, there would be blanks in the thumbnail and it would be hard for studio owners to make thumbnails that perfectly align with the studio width and height. This is the same case for profile pictures - they get cropped.
> > > 
> > > 
> > > And if they add text to the sides, it gets cropped.
> > 
> > 
> > That isn't usually good design practice. The standard is to leave a bit of space on each side, and the actual content should be on the center
> 
> That's what I was saying.

I was talking about the actual thumbnail image, and not the way it gets displayed.

Secret-chest:
> I was talking about the actual thumbnail image, and not the way it gets displayed.

It doesn't matter. Users are not supposed to have their text cut off.

PaperMarioFan-2022:
> > I was talking about the actual thumbnail image, and not the way it gets displayed.
> 
> It doesn't matter. Users are not supposed to have their text cut off.

I agree. Thumbnails shouldn't cut off text.

PR #7738, opened 2023-07-12
LLK -> scratchfoundation

mrsrec:
Update URLs (especially root URL which does not redirect)


cwillisf (approved):
Thanks for these updates!

PR #7764, opened 2023-07-31
fix: move the 'if Scratch Link is not running' advice into the Troubleshooting section

cwillisf:
### Resolves:

ENA-348

### Changes:

- Move "If [Scratch Link] does not appear" advice from the installation steps (`install-scratch-link.jsx`) down into the troubleshooting section (`extension-troubleshooting.jsx`).
- Add a `currentOS` property to the `ExtensionTroubleshooting` component, since this info is the first OS-specific content in that component.
- Now that `ExtensionTroubleshooting` has a `currentOS` property, it can make the browser compatibility item show only for platforms which connect to Scratch Link from a browser.

### Test Coverage:

Tested locally with multiple browsers on Windows 11.

### DIFF
--- a/src/components/extension-landing/extension-troubleshooting.jsx
+++ b/src/components/extension-landing/extension-troubleshooting.jsx
@@ -4,6 +4,9 @@ const React = require('react');
 
 const ExtensionSection = require('./extension-section.jsx');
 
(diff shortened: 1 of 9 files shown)


corihudson (approved):
lgtm!

PR #7779, opened 2023-08-03
New FAR design and bug fix

tomlum:
### Resolves:

CE-445 : new design for the FAR flow
CE-448 : Bug where changing the admin message would affect the FAR message


corihudson (approved):
lgtm!

PR #7940, opened 2023-10-24
Update style and documentation for tests

cwillisf:
### Changes:

These are style & documentation changes only. These changes should not change runtime behavior of anything, including the tests. My goal here is to make the code easier to understand and debug: for example, it's easier now to see an async function called without an `await` or `then` (a relatively common mistake in our tests).

I made these changes as part of some other work, but I want to get this in on its own to separate this style "noise" from the functionality changes in the other branch.

- convert `SeleniumHelper` methods to use `async` and `await`
- document methods in `SeleniumHelper`
- enable the `scratch/es6` and `jest/recommended` rule sets for `eslint`
  - disable a couple of rules that would require even more extensive updates to our tests (effort > benefit)
- fix the resulting style issues

### Test Coverage:

Since this PR should not change any behavior, test results should not change. A few tests are failing or flaky right now, and they should continue to fail or be flaky with these changes. Fixing those tests will be that separate piece of work I mentioned above.


test/unit/components/captcha.test.jsx
```diff
@@ -33,18 +33,6 @@ describe('Captcha test', () => {
         expect(global.grecaptcha.execute).toHaveBeenCalled();
     });
 
-    test('Captcha load calls props captchaOnLoad', () => {
```
AlexisGoodfellow, line 36: Was this test supposed to be deleted entirely?

AlexisGoodfellow (approved):
Approving if complete removal of one of the tests was intentional!

test/integration-legacy/teacher-registration/teacher_registration_utils.js
```diff
@@ -9,89 +9,89 @@ module.exports.constants = {
...
+    const typePromise = typeCheckbox.click();
     return Promise.all([organizationPromise, titlePromise, typePromise]) // eslint-disable-line no-undef
-        .then(function () {
-            nextStepButton.click().then(function () {
+        .then(() => {
+            nextStepButton.click().then(() => {
                 driver.wait(seleniumWebdriver.until
                     .elementLocated(seleniumWebdriver.By.className('address-step')));
             });
         });
```
towerofnix, line 79: Does this style update also enable access to async/await? It could tidy functions like these up nicely:

```js
module.exports.fillOrganizationSlide = async function (/* ... */) {
    const organizationInput = /* ... */;
    const titleInput = /* ... */;
    const typeCheckbox = /* ... */;
    const nextStepButton = /* ... */;

    await Promise.all([
        organizationInput.sendKeys('MIT Media Lab'),
        titleInput.sendKeys('Software Developer'),
        typeCheckbox.click(),
    ]);

    await nextStepButton.click();

    driver.wait(seleniumWebdriver.until
        .elementLocated(seleniumWebdriver.By.className('address-step')));
};
```

If not, chained `.then()`s are a lot easier to read than nested ones:

```js
return Promise.all([organizationPromise, titlePromise, typePromise]) // eslint-disable-line no-undef
    .then(() => nextStepButton.click())
    .then(() => {
        driver.wait(seleniumWebdriver.until
            .elementLocated(seleniumWebdriver.By.className('address-step')));
    });
```

PR #7943, opened 2023-10-30
Gha

cwillisf:
### Changes:

Migrate to GHA

Supporting changes:
- Rename existing "Create release branch and PRs" workflow from `main.yml` to `prepare-release.yml`
- When I migrated the tests over, more integration tests started failing, presumably due to different timing on GHA vs. CCI. The bulk of this PR is made up of changes to make the integration tests more reliable:
  - Add `waitUntilDocumentReady()` method on `SeleniumHelper` which blocks until `document.readyState` is `"complete"`
  - Replace bare `driver.get(url)` calls with a new `navigate(url)` method on `SeleniumHelper` which is equivalent to `driver.get(url)` followed by `waitUntilDocumentReady()`
  - Adjust `clickXpath` method on `SeleniumHelper` to retry the click until it succeeds (or timeout). This is good for cases when the loading screen might temporarily obscure the click target, when switching UI tabs, etc.
  - Add `isSignedIn()` method to `SeleniumHelper` and use it to ensure a known initial state for tests that sign in/out
  - Test-specific changes here and there; I'll add review comments
- Remove `test:integration:remote` script since it's equivalent to `SMOKE_REMOTE=true npm run test:integration`

### Test Coverage:

These integration test changes were tested against staging by temporarily enabling integration tests without deploying: https://github.com/scratchfoundation/scratch-www/actions/runs/6695136210/job/18190015980#step:7:1


.github/workflows/ci-cd.yml
```diff
@@ -0,0 +1,145 @@
...
+
+env:
+  CXX: g++-4.8
+  FASTLY_ACTIVATE_CHANGES: true
+  FASTLY_PURGE_ALL: true
+  NODE_ENV: production
+  SKIP_CLEANUP: true
+
+jobs:
+  determine-environment:
```
cwillisf, line 21: Please share if you know of a better way to do branch-specific secrets! This seems silly...

test/health/server-health.test.js
```diff
@@ -0,0 +1,51 @@
...
+
+const rootUrl = process.env.ROOT_URL || (() => {
+    const ROOT_URL_DEFAULT = 'https://scratch.ly';
+    console.warn(`ROOT_URL not set, defaulting to ${ROOT_URL_DEFAULT}`);
+    return ROOT_URL_DEFAULT;
+})();
+
+jest.setTimeout(60000);
+
+describe('www server health check', () => {
```
cwillisf, line 17: I added this test one day when staging wasn't feeling healthy. I figured if we know the integration tests will all fail, there's no point in even starting them. Some day, we'll be able to run integration tests without deploying... when that day comes, this file can be deleted.

test/integration/comments.test.js
```diff
@@ -29,13 +30,12 @@ const studioId = process.env.COMMENT_STUDIO_ID || 10005646;
 const studioUrl = `${rootUrl}/studios/${studioId}/comments`;
 
 // setup comments to leave
+// make sure they are unique and will not be censored (avoid numbers that might look like phone numbers or other PII)
 const date = new Date();
-const dateString = `Y:${date.getFullYear()} - M:${date.getMonth() + 1} - D:${date.getDate()} ` +
-`: ${date.getHours()}:${date.getMinutes()}:${date.getSeconds()}`;
-const buildNumber = process.env.CIRCLE_BUILD_NUM || dateString;
```
cwillisf, line 35: On CCI, these build numbers were small, like `10`. On GHA, they're much larger numbers that our phone number filter blocked. It just needs to be unique, so an ISO date string with sub-second precision works well.

test/integration/footer-links.test.js
```diff
@@ -139,23 +155,14 @@ describe('www-integration footer links', () => {
...
-
-    // ==== SCRATCH FAMILY column ====
-
-    test('click Scratch Conference link', async () => {
-        await clickText('Scratch Conference');
-        const url = await driver.getCurrentUrl();
-        const pathname = (new URL(url)).pathname;
-        expect(pathname).toMatch(/^\/scratch-conference\/?$/);
-    });
-
```
cwillisf, line 155: This test failed when the conference site was having trouble. It's bad enough that we rely on the staging or production site for the integration tests; let's not rely on the conference site as well!

test/integration/join.test.js
```diff
@@ -35,22 +37,24 @@ describe('www-integration join flow', () => {
     });
 
     test('username validation message appears', async () => {
-        await clickXpath('//input[contains(@name, "username")]');
+        const clickedInput = await clickXpath('//input[contains(@name, "username")]');
+        await driver.wait(() => driver.executeScript('return document.activeElement == arguments[0]', clickedInput));
```
cwillisf, line 41: Wait for the input element to become the active element before trying to type into it

test/integration/project-page.test.js
```diff
@@ -117,14 +120,14 @@ describe('www-integration project-page signed in', () => {
...
-        await findByXpath('//span[contains(@class, "profile-name")]');
     });
 
     beforeEach(async () => {
-        await driver.get(rootUrl);
+        // The browser may or may not retain cookies between tests, depending on configuration.
+        await navigate(rootUrl);
+        if (!await isSignedIn()) {
+            await signIn(username, password);
+        }
```
cwillisf, line 130: Instead of just signing in once at the beginning, sign in _if necessary_ before each test. This covers situations like:
- tests are running in multiple browser windows and the browser is configured to discard cookies aggressively
- a problem in one test causes a sign-out that could interfere with other tests

test/integration/selenium-helpers.js
```diff
@@ -8,9 +8,9 @@ const chromedriverVersion = require('chromedriver').version;
 
 const headless = process.env.SMOKE_HEADLESS || false;
 const remote = process.env.SMOKE_REMOTE || false;
-const ci = process.env.CI || false;
-const usingCircle = process.env.CIRCLECI || false;
-const buildID = process.env.CIRCLE_BUILD_NUM || '0000';
+const ciBuildPrefix = process.env.CI ?
+    `CI #${process.env.GITHUB_RUN_ID}/${process.env.GITHUB_RUN_ATTEMPT}` :
+    ''; // no prefix if not in CI
```
cwillisf, line 13: This is the label that will appear in the Sauce Labs test browser when running remote tests.

test/integration/footer-links.test.js
```diff
@@ -28,29 +28,38 @@ describe('www-integration footer links', () => {
 
     // ==== About Scratch column ====
 
+    const pageLoadComplete = () =>
+        driver.wait(async () => {
+            return await driver.executeScript('return document.readyState;') === 'complete';
+        });
+
```
delasare: This is great!

test/integration/footer-links.test.js
```diff
@@ -28,29 +28,38 @@ describe('www-integration footer links', () => {
...
     test('click About Scratch link', async () => {
         await clickText('About Scratch');
+        await pageLoadComplete();
```
delasare: 💪

test/integration/footer-links.test.js
```diff
@@ -28,38 +29,33 @@ describe('www-integration footer links', () => {
...
 
-    const pageLoadComplete = () =>
-        driver.wait(async () => {
-            return await driver.executeScript('return document.readyState;') === 'complete';
-        });
-
     test('click About Scratch link', async () => {
         await clickText('About Scratch');
-        await pageLoadComplete();
+        await waitUntilDocumentReady();
```
delasare, line 35: even better!

.github/workflows/ci-cd.yml
delasare: I wonder if setting "Environments" in GitHub is a better approach. 
https://github.com/scratchfoundation/scratchjr-website/blob/develop/.github/workflows/deploy.yml

delasare (commented):
Looks great! I think the test changes will go along way in making in the tests less flaky. 

Wondering if GitHub Environments would work for the branch-specific secrets you are looking for.

.github/workflows/ci-cd.yml
cwillisf: Right! I (now) remember seeing that and thinking it might work for www... and then I promptly forgot, I guess. 😅
This code is doing basically the same thing -- setting `environment` -- but it's doing a lot more work to get there. I like your way better. Thanks!

cwillisf:
Sorry for the force-push noise. It turns out that GHA does some very confusing things if your `environment:` value ends with a newline, even if it's outside the quotes.

.github/workflows/ci-cd.yml
```diff
@@ -18,32 +18,26 @@ env:
...
-              ;;
-            *)
-              echo "Leaving scratch_environment unset"
-              ;;
-          esac
   build-and-test-and-maybe-deploy:
-    needs: determine-environment
     runs-on: ubuntu-latest
-    environment: ${{ needs.determine-environment.outputs.scratch_environment }}
+    environment: >-
```
delasare, line 23: 👍

delasare (approved):

PR #8036, opened 2024-01-18
Upgrade to webpack 5

cwillisf:
### Resolves:

This is a step toward ENG-394

### Changes:

Upgrade to Webpack 5 after upgrading various dependencies to versions supporting Webpack 5. Upgrades included the `babel` family, `jest`, and several webpack plugins.

Upgrading the `babel` family also meant replacing `babel-preset-es2015` with `@babel/preset-env`. I don't see a way to configure `@babel/preset-env` to match `babel-preset-es2015`. I believe this is the main source of code differences (see "Test Coverage" below).

We were using a handmade plugin to emit `version.txt` but that plugin was only compatible with Webpack 4. I replaced that with the `emit-file-webpack-plugin`, which takes an async callback for the file contents. I moved the guts of our custom plugin into that file contents callback.

I added `scratch/es6` lint rules (from `eslint-config-scratch`) so that the linter would accept that `Promise` is available. That's part of the implementation of the `emit-file-webpack-plugin` callback. That meant updating several files, mostly in `bin/`, to meet the new style rules.

Unfortunately, our `src/template.ejs` uses the `chunks` feature provided by `html-webpack-plugin@^3` but removed in `html-webpack-plugin@^4` and above. That's how we determine the CSS and JS files to refer to from the template, so I added some code to emulate the `chunks` feature. I think we could find the JS file another way, but I didn't see a good replacement for the CSS part. Anyway, the emulation is far from perfect but it's enough for our purposes.

### Test Coverage:

I've compared the build output files manually. Most JS files are changed significantly, but spot checks seem to suggest that the differences are reasonable. Other files are generally identical or rearranged in ways that should not cause problems. Of course, it's very possible I missed something with those spot checks, so... testing!

Local tests pass. Smoke tests TBD.

Before merging this, I intend to test across all the browsers listed as supported in our FAQ:

- [x] Chrome 63+
- Broken: Edge 15+
  - This is also broken on production as well (with Webpack 4). I couldn't find a combination of settings that fixes this, so I suggest treating this as a separate issue.
- [x] Edge 80+
- [x] Firefox 57+
- [x] Safari 11+
- [x] Mobile Chrome 63+
- [x] Mobile Safari 11+


cwillisf:
OK, nope, found some problems. Going back to draft... :sweat_smile:

delasare (approved):
This looks good. I also tested the branch locally and it seems to work. One of these days I'd love to dig in and see why it takes so long for the pages to build. 😆

webpack.config.js
```diff
@@ -115,7 +115,7 @@ module.exports = {
     module: {
         rules: [
             {
-                test: /\.jsx?$/,
+                test: /\.(?:js|mjs|cjs)x?$/,
```
cwillisf, line 118: This means we now transpile all JS, not just JSX.
Adding `cjs` and `mjs` aren't necessary just yet, but may help avoid surprises in the future.

PR #8053, opened 2024-02-15
fix: react-onclickouside doesn't support functional components

cwillisf:
### Resolves:

Fixes integration tests after #8036 

### Changes:

The `react-onclickoutside` package does not support functional components. Somehow it was working with the `StudioFollow` functional component anyway -- probably because of some transformations applied by webpack 4. After moving to webpack 5 with #8036, though, it stopped working.

This change moves from `react-onclickoutside` to `use-onclickoutside`, which officially supports functional components. The `use-onclickoutside` module is, as the name suggests, structured like a hook instead of a HOC, so I adjusted the code accordingly.

### Test Coverage:

Tested manually and by running integration tests against staging.


delasare (approved):
Nice, simple fix! :shipit:

PR #8366, opened 2024-05-21
Allow RADISH_URL substitution

colbygk:
Provide mechanism to specify a RADISH_URL that differentiates between different destination end points depending on if running from staging or from production or ...

### DIFF
--- a/bin/configure-fastly.js
+++ b/bin/configure-fastly.js
@@ -7,6 +7,7 @@ const routeJson = require('../src/routes.json');
 
 const FASTLY_SERVICE_ID = process.env.FASTLY_SERVICE_ID || '';
 const S3_BUCKET_NAME = process.env.S3_BUCKET_NAME || '';
+const RADISH_URL = process.env.RADISH_URL || '';
 
 const fastly = require('./lib/fastly-extended')(process.env.FASTLY_API_KEY, FASTLY_SERVICE_ID);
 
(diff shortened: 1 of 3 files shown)


bin/configure-fastly.js
```diff
@@ -18,7 +19,15 @@ const extraAppRoutes = [
     '/[^/]*.html$'
 ];
 
-const routes = routeJson.map(
+const routeJsonPreProcessed = routeJson.map(
+    (route) => {
```
cwillisf, line 23:
```suggestion
    route => {
```

bin/configure-fastly.js
```diff
@@ -18,7 +19,15 @@ const extraAppRoutes = [
...
+        if (route.redirect) {
+            route.redirect.replace('RADISH_URL',RADISH_URL);
```
cwillisf, line 25:
```suggestion
            route.redirect.replace('RADISH_URL', RADISH_URL);
```

cwillisf, line 25:
```suggestion
    route => {
        if (route.redirect) {
            route.redirect.replace('RADISH_URL', RADISH_URL);
```

colbygk: THX, I went through the process of getting it running locally, and after getting through the cairo related build trouble on node-gyp, now can run linting locally :)

PR #8419, opened 2024-06-06
fix: temporarily revert all routing changes for the new site

cwillisf:
### Changes:

Temporarily reset `routes.json` to the state it was in before any redirects related to our new website.
This should give us some breathing room to implement and test a proper fix tomorrow/later.


AlexisGoodfellow (approved):
We can (and likely should) revert this revert later.

PR #8638, opened 2024-08-22
feat: Userguiding integration

sashodada:
### Changes:

Add support for Userguiding survey integration

### Test Coverage:

Tests and lint pass, change has been verified manually

### DIFF
--- a/src/lib/user-guiding.js
+++ b/src/lib/user-guiding.js
@@ -1,0 +1,163 @@
+const api = require('./api');
+
+const USER_GUIDING_ID = process.env.USER_GUIDING_ID;
+const SCRIPT_ID = 'UserGuiding';
+
+const CONDITIONS = {condition_list: [
+    'IsLoggedIn',
+    'IsNewScratcher',
+    'NotMuted'
+]};
+
+const getUserGuidingSnippet = () => (
+    `
+        (function(g, u, i, d, e, s) {
+            g[e] = g[e] || [];
+            var f = u.getElementsByTagName(i)[0];
+            var k = u.createElement(i);
+            k.async = true;
+            k.src = 'https://static.userguiding.com/media/user-guiding-' + s + '-embedded.js';
+            f.parentNode.insertBefore(k, f);
+            if (g[d]) return;
+            var ug = g[d] = {
+                q: []
+            };
+            ug.c = function(n) {
+                return function() {
+                    ug.q.push([n, arguments])
+                };
+            };
+            var m = ['previewGuide', 'finishPreview', 'track', 'identify', 'hideChecklist', 'launchChecklist'];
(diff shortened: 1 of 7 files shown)


src/lib/user-guiding.js
```diff
@@ -0,0 +1,124 @@
...
+
+    api({
+        uri: '/user_guiding',
+        method: 'GET',
+        headers: {
+            'Content-Type': 'application/json',
+            'X-USERID': userId,
+            'X-PERMISSIONS': JSON.stringify(permissions),
+            'X-CONDITIONS': JSON.stringify(CONDITIONS),
+            'X-QUESTION-NUMBER': guideId
```
AlexisGoodfellow, line 77: Ah, looks like the survey ids aren't sequential - bummer. That's something important to note.

src/lib/user-guiding.js
```diff
@@ -0,0 +1,124 @@
...
+const attemptDisplayUserGuidingSurvey = (userId, permissions, guideId, callback) => {
+    if (!USER_GUIDING_ID || !process.env.SORTING_HAT_HOST) {
+        return;
+    }
+
+    api({
+        uri: '/user_guiding',
+        method: 'GET',
+        headers: {
+            'Content-Type': 'application/json',
```
cwillisf, line 73: During Bug Hunt, I saw `text/json` in some places. From what I've read, `application/json` is more correct, so I don't recommend changing this, but it does remind me that we may have an inconsistency somewhere in the system.

CC @AlexisGoodfellow

src/lib/user-guiding.js
```diff
@@ -0,0 +1,124 @@
...
+const onProjectLoaded = (userId, permissions) => {
+    const surveyId = sample(EDITOR_INTERACTION_SURVEY_IDS);
+
+    attemptDisplayUserGuidingSurvey(
+        userId,
+        permissions,
+        surveyId,
+        () => window.userGuiding.launchSurvey(surveyId)
+    );
+};
```
cwillisf, line 118: I like how this turned out. There's some repetition, sure, but it's a bearable amount. Most importantly, it's easy to tell what's going on.

cwillisf (approved):

src/lib/user-guiding.js
sashodada: I dug into this a bit more and found this pretty insightful answer on SO: https://stackoverflow.com/a/69518701

According to IANA's page on media types (https://www.iana.org/assignments/media-types/media-types.xhtml) the `text/json` field is not an officially registered media type, so it might be more of a leftover from the original RFC's state.

Another detail to note here is that the `api` lib doesn't populate the `body` object that's passed in the callback if the content type of the response is `text/json`.

PR #8653, opened 2024-08-27
Include vars and secrets for user guiding in CI build

AlexisGoodfellow:
_We still need to actually supply these values - should probably do that pre-merge_


AlexisGoodfellow:
Values have been supplied - just need to merge now

PR #8692, opened 2024-09-07
fix: upgrade scratch-gui to fix music extension

cwillisf:
### Resolves:

- Resolves scratchfoundation/scratch-gui#9614 in production environment

### Changes:

fix: scratchfoundation/scratch-vm@c3759e8e99fc2ae369e8d0c23b4989392fa1ef4e
supported by: scratchfoundation/scratch-webpack-configuration@2a263ff6e5f7cf895fac09d42866f3b6cf912e7f

Together, these changes ensure that the MP3 files used by the Music extension are packaged using `arraybuffer-loader` and not _also_ processed by the webpack 5 "asset modules" feature, and ensures that's consistent across repositories. It's worth noting that in webpack 4 builds, these MP3 files were not copied into the `scratch-vm/dist/` build output. This change makes that true again. They're still available in `scratch-vm/src/`, of course.

### Test Coverage:

Tested locally in both `scratch-gui` and `scratch-www`


MiroslavDionisiev:
Looks good. I was wondering for the changes does `resourceQuery: '?arrayBuffer'` allow for the `arraybuffer-loader` to be used as it matches the string in the manifest file. Also it differs with the example [here](https://www.npmjs.com/package/arraybuffer-loader). Before refactoring the webpack configuration the rule for mp3 has been using `file-loader` so I was think if we could use the new type webpack suggests in the documentation as a replacement `asset/resource`.

cwillisf:
Yeah, I think we should use asset modules in the ideal case -- probably `asset` so that webpack can automatically choose the "best" behavior. Doing so would require a more significant refactor that I didn't want to try to do in a hurry :)

Using anything other than `arraybuffer-loader` would probably involve doing some `fetch` work in the music extension's startup code. I haven't really thought that through yet, but it seems like it wouldn't be too difficult.

Also, it's worth noting that `arraybuffer-loader` hasn't been updated for about 5 years, so we should probably move away from it when we can :sweat_smile:

cwillisf:
> does `resourceQuery: '?arrayBuffer'` allow for the `arraybuffer-loader` to be used as it matches the string in the manifest file

Yep - the `?arrayBuffer` becomes the `resourceQuery` value. With the changes I made to `scratch-webpack-configuration`, theoretically we can `require` or `import` something like `myFile?arrayBuffer` "anywhere" to turn a file into an `ArrayBuffer` object. (But, as mentioned above, we should probably work to move away from this.)

> Also it differs with the example [here](https://www.npmjs.com/package/arraybuffer-loader)

Yeah, we could filter by file extension rather than resource query, but I think using `resourceQuery` is more flexible. I don't think `resourceQuery` was available when that example was written.

PR #8830, opened 2024-10-17
[UERP-51] Google Analytics events for onboarding journeys

MiroslavDionisiev:
### Jira Ticket

[UEPR-51](https://scratchfoundation.atlassian.net/browse/UEPR-51)

### Changes:

Added events:

- `editor-journey-step` - when user goes through steps during editor journey recording the picked step as parameter
- `tutorial-played` - when user plays tutorial project, sends the name of the played project as parameter
- `tutorial-remixed` - when user remixes tutorial project, sends the name of the remixed project as parameter

[UEPR-51]: https://scratchfoundation.atlassian.net/browse/UEPR-51?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/journeys/editor-journey/editor-journey.jsx
+++ b/src/components/journeys/editor-journey/editor-journey.jsx
@@ -6,6 +6,7 @@ const DriverJourney = require('../driver-journey/driver-journey.jsx');
 const {defineMessages, useIntl} = require('react-intl');
 const {useMemo, useState, useCallback} = require('react');
 const PropTypes = require('prop-types');
+const {triggerAnalyticsEvent} = require('../../../lib/onboarding.js');
 
 require('./editor-journey.scss');
 
@@ -138,6 +139,14 @@ const EditorJourney = ({onActivateDeck, setCanViewTutorialsHighlight, setShowJou
     ));
     const intl = useIntl();
 
+    const pickStep = useCallback((stepNumber, editorJourneyStep) => {
+        triggerAnalyticsEvent({
+            event: 'editor-journey-step',
+            editorJourneyStep: editorJourneyStep
+        });
+        driverObj.moveTo(stepNumber);
+    }, driverObj);
+
     const createStep = useCallback((projectId, tutorialId) => ({
         title: intl.formatMessage(messages.createStepTitle),
         showButtons: ['close'],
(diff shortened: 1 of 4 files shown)


src/lib/onboarding.js
```diff
@@ -0,0 +1,38 @@
...
+    
+    window.dataLayer.push({
+        ...eventVaribles
+    });
+};
+
+export const sendUserProperties = user => {
+    window.dataLayer = window.dataLayer || [];
+
+    const {gender, birthYear, birthMonth, ..._rest} = user;
```
sashodada, line 31: Do we need to destructure all remaining properties into `_rest` if they aren't used??

src/views/preview/project-view.jsx
```diff
@@ -255,6 +256,10 @@ class Preview extends React.Component {
                 false // Do not show cloud/username alerts again
             );
         }
+
+        if (!prevProps.user.id && this.props.user.id) {
+            sendUserProperties(this.props.user);
+        }
```
sashodada, line 262: Should these be sent over only if the user is allowed to participate in the onboarding project, i.e. is part of either the A or B group?

sashodada (approved):

src/views/preview/project-view.jsx
MiroslavDionisiev: Yeah, I have implemented the check in `sendUserProperties` in the UEPR-71 ticket as we include the toggling logic

src/components/journeys/editor-journey/editor-journey.jsx
```diff
@@ -198,6 +198,35 @@ const EditorJourney = ({onActivateDeck, setCanViewTutorialsHighlight, setShowJou
...
+                    stepName = 'pong-game-step';
+                } else if (driverObj.getActiveIndex() === 6) {
+                    stepName = 'animate-character-step';
+                } else if (driverObj.getActiveIndex() === 7) {
+                    stepName = 'make-fly-animation-step';
+                } else if (driverObj.getActiveIndex() === 8) {
+                    stepName = 'record-sound-step';
+                } else if (driverObj.getActiveIndex() === 9) {
+                    stepName = 'make-music-step';
+                }
```
sashodada, line 223: Do we have a specific reason to avoid using an array/object that contains the index-to-stepName mapping? We can have the same behavior here by implementing something like
```js
// root level
const STEP_NAMES = ['pick-genre-step', 'game-step', ..., 'make-music-step'];
...
// in place of the current code
onDestroyStarted: () => {
    const stepName = STEP_NAMES[driverObj.getActiveIndex()] || '';
...
```

PR #8848, opened 2024-10-23
Chore/remove user guiding calls

MiroslavDionisiev:
### Changes:

Removes calls to user guiding


sashodada (commented):
Should we also remove the `lib/user-guiding.js` file? I don't think it's used anywhere after removing these calls

sashodada (approved):

PR #8904, opened 2024-11-08
ci: add job to push Transifex strings

cwillisf:
### Changes:

- Add a new GHA workflow to run the `i18:push` npm script
- Fix (remove) the path to `tx-push-src` in the `tx-push-www` script

The new workflow is triggered on push to `develop`, `master`, and release branches. Back when we used Travis, we had it set up on an interval. I think this approach will still push any time we need it, but probably still less often than our Travis approach.

### DIFF
--- a/.github/workflows/tx-push.yml
+++ b/.github/workflows/tx-push.yml
@@ -1,0 +1,26 @@
+name: Transifex Push
+
+on:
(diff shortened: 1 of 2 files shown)


georgyangelov (approved):
Nice!

MiroslavDionisiev (approved):

PR #8914, opened 2024-11-12
fix: styling issues

MiroslavDionisiev:
### Changes:
- replaced the closing button content with an icon
- fixed styling issued on the project journey
- fixed a typo
- handled null values for age and month of birth

image


KManolov3 (approved):

src/components/journeys/driver-journey/driver-journey.jsx
```diff
@@ -20,6 +20,19 @@ const DriverJourney = ({configProps, driverObj}) => {
                             callback();
                         }
                         const portalData = [];
+
+                        const closeButton = document.getElementsByClassName('driver-popover-close-btn')[0];
+                        if (closeButton) {
+                            closeButton.textContent = '';
```
cwillisf, line 26: Are these element property changes necessary because `driver.js` is setting up the button in a way that we can't fully control?

src/lib/onboarding.js
```diff
@@ -71,6 +76,8 @@ export const triggerAnalyticsEvent = eventVaribles => {
 };
 
 export const sendUserProperties = (user, permissions) => {
+    window.dataLayer = window.dataLayer || [];
+
     if (!isUserEligible(user, permissions)) {
         window.dataLayer.push({
```
cwillisf, line 82: :sweat_smile:

src/components/journeys/driver-journey/driver-journey.jsx
MiroslavDionisiev: @cwillisf it is using for the cross sign the sign for multiplication and when the browsers running on mac try to render it is it not centered

PR #8980, opened 2024-12-02
bug: [UEPR-118] Player sound playback

Bogomil-Stoyanov:
**Jira ticket:**
[UEPR-118](https://scratchfoundation.atlassian.net/jira/software/c/projects/UEPR/boards/92?selectedIssue=UEPR-118)

[UEPR-118]: https://scratchfoundation.atlassian.net/browse/UEPR-118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/intro/intro.jsx
+++ b/src/components/intro/intro.jsx
@@ -76,6 +76,7 @@ class Intro extends React.Component {
                         <Video
                             className={classnames('intro-videos', !this.state.videoOpen && 'invisible')}
                             videoId="joal01i8b1"
+                            play={this.state.videoOpen}
                         />
                     </FlexRow>
                 </FlexRow>
--- a/src/components/intro/intro.jsx
+++ b/src/components/intro/intro.jsx
@@ -74,7 +74,7 @@ class Intro extends React.Component {
                         }
                         
                         <Video
-                            className={classnames('intro-videos', !this.state.videoOpen && 'invisible')}
+                            className={classnames('intro-videos')}
                             videoId="joal01i8b1"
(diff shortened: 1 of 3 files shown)


src/components/intro/intro.jsx
```diff
@@ -75,6 +72,12 @@ class Intro extends React.Component {
                                 </div>
                             )
                         }
+                        
+                        <Video
+                            className={classnames('intro-videos', !this.state.videoOpen && 'invisible')}
```
MiroslavDionisiev, line 77: Do we need the `invisible` class if we don't render the video component when `this.state.videoOpen` is false?

src/components/video/video.jsx
```diff
@@ -35,44 +40,53 @@ class Video extends React.Component {
             onReady: video => {
                 video.bind('play', () => {
                     this.setState({videoStarted: true});
-                    this.props.onVideoStart();
+                    if (this.props.onVideoStart) {
```
MiroslavDionisiev, line 43: I am wondering whether it would be better to give `onVideoStart` a default value. But I don't mind it either way

src/components/intro/intro.jsx
Bogomil-Stoyanov: I removed the style, it was left behind from another previous solution that I tried

KManolov3 (approved):

MiroslavDionisiev (approved):

PR #9060, opened 2024-12-19
fix/ [UEPR-113] Removed info box and added  a new line

Bogomil-Stoyanov:
Removing an unnecessary information box, adding a new line in the wait for approval paragraph
image
image


KManolov3 (approved):

KManolov3:
`teacherRegistration.phoneSecurityNotice` is now unused, but we'll need it anyway if we implement phone verification, so imo it's fine to stay for now.

PR #9138, opened 2025-01-13
Improve contrast for Unread Message Count in Navigation Bar.

rosics-code:
### Resolves:

Poor contrast for unread message count in navbar.

### Changes:

changes color to red

### Test Coverage:

Screenshot 2025-01-12 9 12 58 PM

### DIFF
--- a/src/components/navigation/www/navigation.scss
+++ b/src/components/navigation/www/navigation.scss
@@ -148,7 +148,7 @@
                 top: .5rem;
                 right: .25rem;
                 border-radius: 1rem;
-                background-color: $ui-orange;
+                background-color: #b94846;
                 padding: 0 .25rem;
                 text-indent: 0;
                 line-height: 1rem;


medlans:
By the way, a red is already used on the messages page

rosics-code:
> By the way, there's a different red that is already used on the messages page

I know.

rosics-code:
> By the way, there's a different red that is already used on the messages page

Fixed.

medlans:
> > By the way, there's a different red that is already used on the messages page
> 
> I know.

Yea, I just mentioned it because I think it would make more sense if they matched

rosics-code:
> > > By the way, there's a different red that is already used on the messages page
> > 
> > 
> > I know.
> 
> Yea, I just mentioned it because I think it would make more sense if they matched

I edited the files change to now request it to be changed to #c40

DDCcarr1 (approved):
Looks good.

PR #9148, opened 2025-01-15
feat: [UEPR-47] redesign starter projects page

MiroslavDionisiev:
### Resolves:
[UEPR-47](https://scratchfoundation.atlassian.net/browse/UEPR-47)

### Changes:
Redesigns "Starter Projects" page, creates new carousel component and updates `react-slick` library

[UEPR-47]: https://scratchfoundation.atlassian.net/browse/UEPR-47?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rosics-code:
Screenshots?

davidtheplatform:
> Screenshots?

[image](https://github.com/user-attachments/assets/f52bd9ac-ee05-4368-8219-642a73835984)

GoldenDuck13:
This 100% would be a great redesign. If the design looks as the image I would recommend changing the color from pink to a different color from the scratch theme.

davidtheplatform:
> This 100% would be a great redesign. If the design looks as the image I would recommend changing the color from pink to a different color from the scratch theme.

The screenshot is exactly what the website will look like, assuming they don't change anything.

src/views/starter-projects/starter-projects.scss
```diff
@@ -1,11 +1,180 @@
...
+                font-size: 1rem;
+                padding: 0.75rem 1.5rem;
+            }
+        }
+    }
+
+    @media only screen and (min-width: 1280px) {
+        .paragraph,
+        .project-sections .project-section {
+            width: 65rem;
```
KManolov3, line 89: Why is this 65rem, while below it's 64?

rosics-code: > Is it intentional that this is 65rem while below it's 64?

I would change the pink to purple

MiroslavDionisiev: Because widths of carousels in the designs are not the most accurate. For the width between 1280px and 1024px the carousel should be 1040px meaning at some point it will be wider than the screen. This is why for that range the width of the carousel is set to be the min value of the range. Similar approach is taken for the other breakpoints as well

PR #9161, opened 2025-01-21
Round the Splash Page boxes and images.

rosics-code:
### Resolves:

The boxes and images were not round before.

### Changes:

They are now round, the boxes and images.

### Test Coverage:

Works on any device supporting border radius, position, bottom, and font size.

Screenshot 2025-01-21 7 11 24 PM


GoldenDuck13:
Makes sense.

PR #9181, opened 2025-01-27
Support Newlines in comments

rosics-code:
### Resolves:

Support newlines.

### Changes:

Use` pre-wrap` in comments.

### Test Coverage:

Capture


alboxer2000:
It could be used for spam (by putting a lot of newlines), so maybe add a **Show more** button like YouTube.

PR #9292, opened 2025-03-06
feat: [UEPR-190] Connecting ToU modal UI to backend

Bogomil-Stoyanov:
Ticket:
[UEPR-190](https://scratchfoundation.atlassian.net/jira/software/c/projects/UEPR/boards/92?selectedIssue=UEPR-190)

Error UI:
image


src/components/modal/terms-of-use/modal.jsx
```diff
@@ -47,13 +65,14 @@ const TermsOfUseModal = ({
...
-            onClose={handleClose}
+            onAccept={handleAccept}
+            showErrorMessage={showError}
         />);
     }
   
     return (<TermsOfUseModalUnder16
         isOpen={isModalVisible}
-        onClose={handleClose}
+        onAccept={handleAccept}
```
MiroslavDionisiev, line 75: Why are we not passing `showError` to `TermsOfUseModalUnder16`?

MiroslavDionisiev:
There is an empty file `error-msg.jsx`

src/components/modal/terms-of-use/modal.jsx
```diff
@@ -19,16 +20,33 @@ export const TermsOfUseLink = chunks => (
     </a>
 );
 
+
```
KManolov3, line 23: nitpick: remove extra line

src/components/modal/terms-of-use/modal.jsx
```diff
@@ -47,13 +65,14 @@ const TermsOfUseModal = ({
     if (!usesParentEmail || isOver16){
         return (<TermsOfUseModalOver16
             isOpen={isModalVisible}
-            onClose={handleClose}
+            onAccept={handleAccept}
+            showErrorMessage={showError}
         />);
     }
   
     return (<TermsOfUseModalUnder16
```
KManolov3, line 73: Since we're showing the same error message in both cases, what do you think about showing it here?

src/lib/session.js
```diff
@@ -61,22 +61,7 @@ module.exports.requestSessionWithRetry = (resolve, reject, retriesLeft, totalDel
             );
         }
         console.log('session body:', body);
```
KManolov3, line 63: console.log

src/components/modal/terms-of-use/error-msg.jsx
KManolov3, line 1: Empty file

src/components/modal/terms-of-use/modal.jsx
Bogomil-Stoyanov: It needs to be in the modal itself. I can try changing it so that we have a single Modal element and only modify the content based on the age group. Initially, I took this approach because the modals behaved differently (e.g., showing the close button, being dismissible, etc.), but since we no longer have such functionality, we might as well use a single modal component and only change its contents.

PR #9312, opened 2025-03-13
CI/CD improvements: repeatable deploy, auto-deploy prepared releases

cwillisf:
I wasn't sure who would be best to review this, so I just assigned a bunch of people. :partying_face:

### Resolves:

- Resolves POD-116

### Changes:

I recommend looking at this diff with "Hide whitespace" turned on (or add `?w=1` or `&w=1` to the end of the URL).

#### Break the CI/CD workflow into multiple parts

This will allow us to re-run a deploy. The primary use case I have in mind is if we need to switch our staging environment between different things we want to test, but it also means we can rerun a deploy if anything went wrong. Also also, we can now re-run integration tests without repeating the build and deploy steps, which might be useful in case of flaky tests.

#### Trigger the CI/CD workflow on completion of the "Create release branch and PRs" workflow

By default, GHA prevents a push from one workflow triggering another workflow. That's meant to protect against infinite loops and such, but it also means that as of migrating to GHA, making release PRs stopped automatically deploying the release candidate to staging. There are several different ways to make this work; I think what I've done here is the simplest, though it has the downside that looking at `prepare-release.yml` doesn't make it clear that it will trigger a build & deploy.

#### Compress and upload artifacts differently

In the middle of this work, I used `actions/upload-artifact` and `actions/download-artifact` to communicate `node_modules` between jobs, which we were already using to store the build output. To (try to) save time and/or space, I switched to using `zstd` instead of `gzip` for both artifacts. I later switched to using `actions/cache` for `node_modules` instead, which internally uses `zstd` anyway, but I left the `zstd` change for the build output artifact as a nice-to-have.

I also added a line to tell `upload-artifact` not to try to re-compress the already-compressed file. It still wraps it in a ZIP: that's a current limitation of `actions/upload-artifact` (see actions/upload-artifact#426).

### Test Coverage:

I tested the linkage between the jobs by temporarily turning off some of the conditionals, but I have not yet tested deploying, including for a release branch. After merging this change, I'll make a pretend release branch as a test.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -36,6 +36,11 @@ jobs:
           ) ||
           ''
         }}
+    steps:
+      # apparently you're not allowed to have a job with no steps
+      - name: Report environment
+        run: |
+          echo "Environment: ${{ env.GH_ENVIRONMENT }}"
     outputs:
       environment: ${{ env.GH_ENVIRONMENT }}
       should_deploy: ${{ (env.GH_ENVIRONMENT || '') != '' }}
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -61,6 +61,7 @@ jobs:
         run: |
           cat <<EOF
           Scratch environment: ${SCRATCH_ENV}
+          Should deploy: ${SCRATCH_SHOULD_DEPLOY}
           Node version: $(node --version)
           NPM version: $(npm --version)
           github.workflow: ${{ github.workflow }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -103,23 +171,50 @@ jobs:
...
           SLACK_WEBHOOK_MODS: ${{ secrets.SLACK_WEBHOOK_MODS }}
           RADISH_URL: ${{ vars.RADISH_URL }}
+  integration-tests:
+    name: Production Integration Tests
+    runs-on: ubuntu-latest
+    needs: [set-environment, deploy]
+    environment: ${{ needs.set-environment.outputs.environment }}
+    # GHA can't reach staging
+    # TODO: Run integration tests against localhost? Not only would that let us test any and all branches,
+    # but we could also, I dunno, consider fully testing _before_ we deploy? :sweat_smile:
```
KManolov3, line 186: That'd be pretty great when we get to it :D

PR #9358, opened 2025-03-27
feat: Fixing punctuation and grammar.

DDCcarr1:
### Resolves:
* [Messages](https://scratch.mit.edu/messages/) not having punctuation.

### Changes:
Adds punctuation to many messages.

Ex: (new changes in bold)
{profileLink} favorited your project {projectLink}**.**

### Test Coverage:
None should be needed.


DDCcarr1:

DDCcarr1:
Request is ready for review!

DDCcarr1:
> I think that is intentional. Notice how many other websites don't have punctuation in the messages. The only one I can think of that doesn't is any MediaWiki wiki

If you look into the l10n.json file for the messages, you can notice some (not all) have punctuation.

AmpElectrecuted:
> Also, another note, I think it was ur request to fix “youtube” to YouTube lol

indeed

DDCcarr1:
Changed “thread” to “topic” in messages

mxmou:
> > I think that is intentional. Notice how many other websites don't have punctuation in the messages. The only one I can think of that doesn't is any MediaWiki wiki
> 
> If you look into the l10n.json file for the messages, you can notice some (not all) have punctuation.

Only those with more than one sentence have punctuation at the end.

DDCcarr1:
> > > I think that is intentional. Notice how many other websites don't have punctuation in the messages. The only one I can think of that doesn't is any MediaWiki wiki
> > 
> > 
> > If you look into the l10n.json file for the messages, you can notice some (not all) have punctuation.
> 
> Only those with more than one sentence have punctuation at the end.

That would make sense, but it’d probably make more sense if they all had punctuation…

PR #9422, opened 2025-04-18
Lower excessive redirects

colbygk:
hrefs for /users/:username without a trailing / lead to django kicking back a redirect to /users/:username/ This activity leads to scratch-www-production having an extremely high number of redirects when compared to all other requests. It would be useful to clear this noise up for SRE related operations.

I have kept edits away from `preview` routes as I believe they are no longer in use.


colbygk:

cwillisf (approved):
Wow, that's quite a few places! :sweat_smile:

PR #9444, opened 2025-04-28
[HIGHEST PRIORITY] nuke the explore page.

alctd:
see
https://scratch.mit.edu/discuss/topic/819065/
https://www.change.org/p/endorse-the-removal-or-editing-scratch-mit-edu-s-explore-and-trending-page

due to the incident that recently happened scratch could get lawsuited and possibly shut down (and this is actually pretty likely to happen since literal above-your-legs were on the explore page, and scratch literally has 5 year olds and 99% of its userbase isn't supposed to be expose to that), so killing off the explore page would be the best option.


alctd:

alctd:
wait i might have a better idea, but it might involve having to expose the NFE APIs to users

alctd:
@cwillisf @Bogomil-Stoyanov @KManolov3
what do you think
(also idk if scratch would accept contributions as large as this but this is a critical issue. scratch is for 8-to-16s, and "eggplants" are for 18+.)

Bogomil-Stoyanov:
You have not removed the explore button in the header

GvYoutubeSchool:
@alctd @Bogomil-Stoyanov @ChrisC @tmickel @southwolf 
As someone who saw this and was there during the incident (and as a person traumatized about this), you, ST, REALLY need to push this PR. If you are going to run a kids' site, remove the things that are not for kids.

alctd:
> You have not removed the explore button in the header

will fix.

alctd:
i didn't want to make the navbar look awkward with just 3 links, so i replaced the explore link with a link to the guidelines

cwillisf:
I appreciate your desire to help, but we must not give in to the trolls. Removing the Explore page would mean not just feeding the trolls but empowering them to control the design of the Scratch website and the shape of our community.

Sharing and collaboration are central to Scratch, and we're working hard to support the safety and reliability of those features. We've already made some behind-the-scenes changes in response to the recent incident, and we're working on more. We've found that detailing or highlighting these situations effectively feeds the trolls and can result in further attempts at misuse, so we make these improvements without much public fanfare.

Some folks may want to avoid the Explore page anyway. I support the right of an individual Scratcher (or parent, school, etc.) to make that decision by using a browser extension or other web filtering software.

PR #9472, opened 2025-05-08
fix: exclude new project title from remix notifications

cwillisf:
### Resolves:

- Resolves POD-214

### Changes:

Change the message for remix notifications to exclude the new/child project title. The old/parent project title is still displayed, and the new/child project link still exists; this is just a change to that link's displayed text.

### Test Coverage:

Will test manually in staging


wdickerson:
Tested on staging, LGTM
image

wdickerson (approved):

PR #9484, opened 2025-05-13
feat: disallow students from creating notifications

KManolov3:
### Resolves:
https://scratchfoundation.atlassian.net/browse/POD-220

### Changes:

Hide the 'Invite Curators' section in the Curators view in the studio view if the user is a student, even in the case where the user is the owner of the studio.

image


colbygk (approved):
I'm not the best person to take a close look at this, but in principle it appears to be good to me.

MiroslavDionisiev (approved):

PR #9505, opened 2025-05-21
fix non-English remix notifications (update scratch-l10n)

cwillisf:
### Resolves

- Resolves POD-225 "Remix notifications broken when for all languages but English"

### Changes

Update `scratch-l10n` to the latest version.

The reason this broke is that #9472 changed the placeholders in the remix message but did not change the message ID. If the message lookup had failed, the English text (with correct placeholders) would have been used. Since the message ID didn't change, looking up non-English versions of the message in `scratch-l10n` succeeded, but resulted in the old translated message with the old placeholders. Since the placeholders didn't match up, variable replacement didn't work correctly, and the message ended up looking something like this:

image

Updating `scratch-l10n` gets the new placeholders, which will cause variable replacement to work correctly. Most languages don't have a translation for this message yet, but the links will be fixed. Translations for this message can come online in future updates.

### Test Coverage:

Verified correct message rendering in staging

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -112,7 +112,7 @@
         "regenerator-runtime": "0.13.9",
         "sass": "1.87.0",
         "sass-loader": "10.5.2",
-        "scratch-l10n": "5.0.222",
+        "scratch-l10n": "5.0.247",
         "selenium-webdriver": "4.31.0",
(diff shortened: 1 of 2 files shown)


wdickerson (approved):
🛠️

PR #9506, opened 2025-05-21
Update l10n.json

LeoKids:
### Resolves:

None.

### Changes:

Updates the Windows compatible OSes to include newer OSes,as well as clarify that 10.15 and later won't work with Scratch 1.4 due to it dropping support for 32-bit apps.

### Test Coverage:

I haven't tested it yet...


LeoKids:

LeoKids:
recheck

PR #9629, opened 2025-07-10
feat: [POD-251] Search/explore admin

varunbiniwale:
### Resolves:

Work related to POD-251. Depends on https://github.com/scratchfoundation/scratch-api/pull/1891

### Changes:

* Searches now wait for session data to be ready
* If the signed in user is an admin, switch to the /admin/search/ endpoint
* Send along token in these cases

* This should have no impact on caching for regular user queries, as we're only sending the token if the user is an admin.

### Test Coverage:

Need to test:
* Admin users can search (and the new endpoint is used)
* Logged in users can still search
* Logged out users can still search
* ?

### DIFF
--- a/src/components/grid/grid.jsx
+++ b/src/components/grid/grid.jsx
@@ -1,6 +1,7 @@
 const classNames = require('classnames');
 const PropTypes = require('prop-types');
 const React = require('react');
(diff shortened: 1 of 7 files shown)


colbygk (approved):
LGTM

PR #9707, opened 2025-08-12
feat: [UEPR-290] Update face sensing configuration

adzhindzhi:
### Resolves:

[UEPR-290](https://scratchfoundation.atlassian.net/browse/UEPR-290)

### Changes:

Copy mediapipe files to the build folder to enable local serving

[UEPR-290]: https://scratchfoundation.atlassian.net/browse/UEPR-290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,7 @@
         "@formatjs/intl-numberformat": "8.15.4",
         "@formatjs/intl-pluralrules": "5.4.4",
         "@formatjs/intl-relativetimeformat": "11.4.11",
-        "@scratch/scratch-gui": "11.6.0-face-sensing",
+        "@scratch/scratch-gui": "11.6.0-face-sensing.2",
         "@types/jest": "29.5.14",
         "async": "3.2.6",
         "autoprefixer": "10.4.21",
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -243,6 +243,11 @@ module.exports = {
                     from: 'node_modules/@scratch/scratch-gui/dist/static/blocks-media',
                     to: 'static/blocks-media'
                 },
+                {
+                    context: 'node_modules/@scratch/scratch-gui/dist/',
+                    from: 'chunks/mediapipe/face_detection/face_detection_*.{js,wasm,tflite,binarypb}',
+                    noErrorOnMissing: true
```
KManolov3, line 249: I think we can omit the `noErrorOnMissing` - we do expect to find this at this path. 
Also, let's copy the whole face_detection folder - what file extensions it ends up needing can be considered an implementation detail and may be subject to change.

KManolov3 (approved):

MiroslavDionisiev (approved):

PR #9747, opened 2025-08-27
fix: use more robust css selectors and add additional property to pre…

KManolov3:
…vent tooltip overlay from breaking screen interactions

### Resolves:

https://scratchfoundation.atlassian.net/browse/UEPR-323

### Changes:

* Make sure overlay position in DOM doesn't affect whether overlay z-index is applied
* Also add visibility:hidden, in case `z-index` isn't successfully applied / doesn't work on specific browsers

### DIFF
--- a/src/views/preview/project-view.scss
+++ b/src/views/preview/project-view.scss
@@ -9,8 +9,9 @@
     pointer-events: auto !important;
 }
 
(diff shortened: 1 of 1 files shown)


KManolov3:
Pushed in scratchfoundation remote to trigger pipeline - https://github.com/scratchfoundation/scratch-www/pull/9748

PR #9748, opened 2025-08-27
fix: use more robust css selectors and add additional property to pre…

KManolov3:
…vent tooltip overlay from breaking screen interactions

### Resolves:

https://scratchfoundation.atlassian.net/browse/UEPR-323

### Changes:

* Make sure overlay position in DOM doesn't affect whether overlay z-index is applied
* Also add visibility:hidden, in case `z-index` isn't successfully applied / doesn't work on specific browsers

### DIFF
--- a/src/views/preview/project-view.scss
+++ b/src/views/preview/project-view.scss
@@ -9,8 +9,9 @@
     pointer-events: auto !important;
 }
 
-.tooltip-set-thumbnail ~ .driver-overlay {
+.driver-active:has(.tooltip-set-thumbnail) > .driver-overlay {
     z-index: -1 !important;
+    visibility: hidden;
 }
 
 .driver-popover.tooltip-set-thumbnail {


rosics-code:
> > Can the tool-tip and button be purple? It's hard to see with the blue over white.
> 
> @rosics-code The design team prefers to go with blue with some newer features, so this isn't simply a dev decision to be made. However, I believe the blue-on-white is also confirmed to match accessibility/contrast standards.

That's very interesting. Thanks for the response.

husqwc:
> > Can the tool-tip and button be purple? It's hard to see with the blue over white.
> 
> @rosics-code The design team prefers to go with blue with some newer features, so this isn't simply a dev decision to be made. However, I believe the blue-on-white is also confirmed to match accessibility/contrast standards.

Also, cheddagirl made [this post](https://scratch.mit.edu/discuss/topic/833955/?page=3#post-8731657) asking for this decision to be confirmed with the accessibility team as she hasn't heard of any "white on blue" decisions from them

PR #9799, opened 2025-09-24
Toggle community rows

cwillisf:
### Resolves:

- Resolves POD-299

### Changes:

Add a property on `SplashPresentation` called `shouldShowCommunityRows`, which controls whether or not the community-driven homepage rows should be shown.

### Test Coverage:

`SplashPresentation` was not previously under test, so I added a quick unit test for just this feature. More would be better, of course...


rosics-code:
Would us (the Scratcher) be able to toggle it? And if so, will it save?

rosics-code:
> > Would us (the Scratchers) be able to toggle it? And if so, will it save?
> 
> Good question! This is a build-time toggle so that `scratch-www` can be built for different situations.

Ah I see. Would be really cool if we could hide/show these tabs like the Scratch Forums.

PR #9850, opened 2025-10-23
Refactor StudioTabNav to use useRouteMatch

rosics-code:
### Resolves:

https://scratch.mit.edu/discuss/topic/843700/

### Changes:

Reverts the change made to studio nav since it did this
Screenshot 2025-10-22 10 25 58 PM

### Test Coverage:

Screenshot 2025-10-22 11 05 22 PM

### DIFF
--- a/src/views/studio/studio-tab-nav.jsx
+++ b/src/views/studio/studio-tab-nav.jsx
@@ -1,7 +1,7 @@
-import React, {useCallback} from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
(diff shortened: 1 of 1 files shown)


rosics-code:
> Hey @rosics-code, thanks for bringing this up. We can't revert directly here, as `useRouteMatch` is [deprecated in react-router v6 ](https://reactrouter.com/6.28.0/upgrading/v5#replace-useroutematch-with-usematch). Ideally we'd want to resolve this with minimal changes from the code currently in master. I think the likely culprit here are the changes in how the styles are passed to the NavLinks

I see. Thanks for replying! :-)

PR #9854, opened 2025-10-23
[UEPR-380] ToU: Frontend blocking

adzhindzhi:
### Resolves:

[UEPR-380](https://scratchfoundation.atlassian.net/browse/UEPR-380)

### Changes:

There are currently two approaches in discussion: Display blocking modals vs. Display blocking pages

- [The first commit](https://github.com/scratchfoundation/scratch-www/commit/26d0cb109cc777226eafe268227d9af5ddce7712) completes the modal approach implementation, showing a blocking modal when opening the editor as well (the remaining was already implemented with the non-blocking ToU modals). This solution has a couple of drawbacks - it’s easier to bypass, can conflict with other modals, and currently lacks a logout option.
- [The second commit](https://github.com/scratchfoundation/scratch-www/commit/9f3a82be652eda1e3e3644fc95b814613a3da361) reverts the modal implementation in favor of blocking pages. While this approach has some UX downsides, it is a bit more secure, harder to bypass, and supports logging out without additional work.

### Test Coverage:

- Current test cases complete successfully

**Page vs. Modal approach UI comparison:**

image
image

[UEPR-380]: https://scratchfoundation.atlassian.net/browse/UEPR-380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/views/preview/project-view.jsx
+++ b/src/views/preview/project-view.jsx
@@ -1093,6 +1093,7 @@ class Preview extends React.Component {
             this.props.parentalConsentRequired;
 
         if (!this.props.playerMode && shouldDisplayBlockingPage) {
+            // The Page components will display the blocking ToU page in this case
             return (
                 <Page />
             );
--- a/src/components/page/www/page.jsx
+++ b/src/components/page/www/page.jsx
@@ -26,6 +26,10 @@ const Page = ({
         !user.isStudent &&
         !user.acceptedTermsOfUse &&
         !isAllowedPage &&
+        // If a user has no state - we should always display the ToU modal in order to gather the state,
+        // regardless of the default jurisdiction rules
+        // If a state exists, it only makes sense to display the ToU flow when no explicit parental consent is required.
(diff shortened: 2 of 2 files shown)


src/views/preview/project-view.jsx
```diff
@@ -1080,6 +1082,25 @@ class Preview extends React.Component {
...
+            !this.props.parentalConsentRequired;
+
+        const shouldDisplayBlockingPage = this.props.userPresent &&
+            !this.props.isStudent &&
+            !this.props.acceptedTermsOfUse &&
+            this.props.parentalConsentRequired;
+
+        if (!this.props.playerMode && shouldDisplayBlockingPage) {
+            return (
+                <Page>
```
KManolov3: Isn't `ParentalConsentView` already present in the logic of `Page.jsx`? Or is this a different case?

PR #9878, opened 2025-11-02
Refactor renderHomepageRows to use dynamic rows

rosics-code:
toggle Featured Rows due to server issues

### Resolves:

It resolves things not loading on the homepage completely .

### Changes:

Uses rows.push like WTCIL and WTCIF and to toggle it.

### Test Coverage:

Before:

Screenshot 2025-11-02 10 30 51 AM

After:

Screenshot 2025-11-02 10 31 43 AM

### DIFF
--- a/src/views/splash/presentation.jsx
+++ b/src/views/splash/presentation.jsx
@@ -589,7 +589,7 @@ SplashPresentation.defaultProps = {
     sharedByFollowing: [], // "Projects by Scratchers I'm Following"
     shouldShowCommunityRows: false,
     shouldShowFeaturedProjectRow: true,
-    shouldShowFeaturedStudioRow: true
+    shouldShowFeaturedStudioRow: false
 };
 
 module.exports = injectIntl(SplashPresentation);


src/views/splash/presentation.jsx
```diff
@@ -576,7 +580,8 @@ SplashPresentation.defaultProps = {
     lovedByFollowing: [], // "Projects Loved by Scratchers I'm Following"
     news: [], // gets news posts from the scratch Tumblr
     sharedByFollowing: [], // "Projects by Scratchers I'm Following"
-    shouldShowCommunityRows: false
+    shouldShowCommunityRows: false,
+    shouldShowFeaturedRows: false
```
adzhindzhi: Hi @rosics-code, thank you for your contribution! In this change, the `shouldShowFeaturedRows` prop is always set to `false`, which ends up hiding the featured sections even when the data loads successfully. If we want to hide those sections, it should be based on whether the featured content was actually fetched, rather than disabling them by default. Also, since featured studios and featured projects are calculated separately, one of them might be available when the other isn’t, so I don't think we should hide both sections if only one of them fails. Thanks again for contributing!

rosics-code: Maybe we could only do studios for now?

PR #9894, opened 2025-11-11
feat: add static membership page

KManolov3:
### Resolves:

https://scratchfoundation.atlassian.net/browse/UEPR-404

### Changes:

Adds a static membership page and a link in the header leading to it for unlogged users

### DIFF
--- a/src/components/navigation/www/navigation.jsx
+++ b/src/components/navigation/www/navigation.jsx
@@ -124,8 +124,7 @@ class Navigation extends React.Component {
                         </a>
                     </li>
                     {
-                        this.props.user ?
-                            null :
+                        !this.props.user &&
                             (
                                 <li className="link membership">
                                     <a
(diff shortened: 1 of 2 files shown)


src/views/membership/membership.jsx
```diff
@@ -0,0 +1,167 @@
...
+                            key={section.id}
+                            className={classNames(
+                                'subsection',
+                                isEven && 'even'
+                            )}
+                        >
+                            {isEven ? image : null}
+                            {text}
+                            {isEven ? null : image}
+                        </div>
```
adzhindzhi: nitpick: I feel like the first two sections share the same structure, so I think it would be nice to extract a `<MembershipSection>` component to reduce duplication and move the rendering logic there. That said, this page probably won’t change often, so I’m fine either way.

KManolov3: That was my initial idea as well, but since this component will only be used here, and the styles are different between the sections (both for the full-size design, as well as for the responsive breakpoints), I'm not sure if the added complexity is worth it. So, given that it's already coded this way, I'll leave it as is.

PR #9901, opened 2025-11-14
[UEPR-399] Add avatar badges for Scratch Members

adzhindzhi:
### Resolves:

[UEPR-399](https://scratchfoundation.atlassian.net/browse/UEPR-399)

### Changes:

- Add flag for applying Cat Ears frame to the avatar
- Pass the flag whenever the Avatar component is used
- Adjust the styles so that displaying avatars with and without badge does not break the layout

### Test Coverage:

- All test complete successfully

[UEPR-399]: https://scratchfoundation.atlassian.net/browse/UEPR-399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/views/preview/preview.scss
+++ b/src/views/preview/preview.scss
@@ -661,7 +661,7 @@ $stage-width: 480px;
             --avatar-size: 2rem;
             --frame-size: 0.84rem;
 
-            margin-right: 8px;
+            margin-right: 0.5rem;
 
             img {
                 margin-right: 0;
--- a/src/components/grid/grid.scss
+++ b/src/components/grid/grid.scss
@@ -34,7 +34,7 @@
         .thumbnail-info {
             margin: 0 auto;
             width: $thumbnail-inner-width;
-                
+
             .creator-image {
                 float: left;
 
--- a/src/components/navigation/www/accountnav.jsx
+++ b/src/components/navigation/www/accountnav.jsx
@@ -35,7 +35,7 @@ const AccountNav = ({
             <Avatar
                 alt=""
                 src={thumbnailUrl}
-                showAvatarBadge={avatarBadge}
+                showAvatarBadge={!!avatarBadge}
             />
             <span className="profile-name">
                 {username}


src/views/preview/preview.scss
```diff
@@ -649,9 +657,23 @@ $stage-width: 480px;
...
         }
 
-        .creator-image img {
-            max-width: 2rem;
-            max-height: 2rem;
+        .creator-image .avatar-wrapper {
+            --avatar-size: 2rem;
+            --frame-size: 0.84rem;
+
+            margin-right: 8px;
```
KManolov3: Is this intentionally in pixels?

src/views/preview/project-view.jsx
```diff
@@ -1437,6 +1439,7 @@ const mapStateToProps = state => {
     const authorPresent = author && Object.keys(state.preview.projectInfo.author).length > 0;
     const authorId = authorPresent && author.id && author.id.toString();
     const authorUsername = authorPresent && author.username;
+    const authorAvatarBadge = authorPresent && author.profile.membership_avatar_badge;
```
KManolov3: Nitpick: I'm wondering whether it's confusing that we pass the badge as a number here (and in other places), but then actually use it only as a boolean. It might be cleaner to abstract away the fact that it may have different values from the frontend for now, at the places where this is returned from the backend (although the backend is okay to return it as a number, in favor of less endpoint interface changes in the future).

That said, given that it's already coded that way, I think it's fine to stay like that.

KManolov3 (approved):
Looks great! Can't wait to see it in action. 🙌

rosics-code:
This is not what Scratch is meant to be though? We shouldn't do this as it locks main-site features behind a paywall which is not Scratch's goals.

pizzawiz:
I personally don't think this is a good idea, but mrsrec is right. If this ever actually gets implemented, I'll complain then.

DawnRooney27:
This is some straight up bull. Ever since Scratch gained independence from LLK and MIT, y'all have been goin' downhill and become some money hungry, testing-AI-on-kids, ethically and morally corrupt, scoundrels. Who in God's green earth told y'all to go against the original 2003 philosophy? You should be able to donate to Scratch without setting a community on fire because you have cat ears on your profile and other users don't. I understand Scratch needs money but adding paywall features even in small bits is just gonna end badly especially considering Scratch was always meant to be free. Instead of spending your donation money on AI just so you can stay up with the trends, hire more moderators, fix bugs that have existed for years, finish the 3.0 pages, actually peep the "Suggestions" forum, and respond to "Contact Us" emails.

real-gdfsg:
No, please no
ST plz don’t add this

cwillisf (approved):
The code looks good! I have a feeling that the badge ID might eventually be a string or an object, but a number is fine until/unless, especially since this means the logic is already prepared for something that isn't just a Boolean.

leotimechannelorg:
Pretty sure there's an [extension](https://scratchaddons.com/) that does the same thing for free. And since the Scratch Team seems to be [against users using browser extensions](https://scratch.mit.edu/discuss/topic/284272/), I don't think this would be a good idea. Just my take, I think the sprite packs are cool but this is iffy and would encourage more Scratchers to use browser extensions.

JereNwa:
- https://github.com/scratchfoundation/scratch-www/issues/9924
Just want all to see this.

PR #9916, opened 2025-11-22
Made the Membership link appear on navbar when logged in

JereNwa:
### Resolves:

- https://github.com/scratchfoundation/scratch-www/issues/9910

### Changes:

When you log into Scratch, you cannot see the membership link on the nav bar. I’ve made so that it is now visible.

### Test Coverage:
[sorry, don’t have image yet.]


redspacecat:
I'm pretty sure not seeing the membership button when logged in is intentional.

PR #9918, opened 2025-11-23
feat: Membership Join promo on Splash page at the bottom before the footer

rosics-code:
### Changes:

Adds a new div with classes. Uses tumblr for the img. Uses color.$color-dark or just color.$color values.

### Test Coverage:

Screenshot 2025-11-22 11 25 59 PM

### DIFF
--- a/src/views/splash/presentation.jsx
+++ b/src/views/splash/presentation.jsx
@@ -5,7 +5,7 @@ const MediaQuery = require('react-responsive').default;
 const PropTypes = require('prop-types');
 const React = require('react');
 
-const {frameless} = require('../../lib/frameless');
+const { frameless } = require('../../lib/frameless');
 const intlShape = require('../../lib/intl-shape');
 const sessionActions = require('../../redux/session.js');
 const shuffle = require('../../lib/shuffle.js').shuffle;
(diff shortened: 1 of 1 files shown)


rosics-code:
@KManolov3 @adzhindzhi @cwillisf What do you think?

JereNwa:
I'm not a Scratch Team member, but I think it's good (probably could change the font size on the button), but I don't really feel like we need it as Scratch users. Also, the way that they've been advertising this whole thing is a little annoying already... I feel like this might contribute to making more users annoyed.

rosics-code:
> Isn’t this taking advertising the membership a bit too far? I don’t think there’s a need for this, it looks intrusive

It helps with advertising more. This is a great change.

JereNwa:
> > Isn’t this taking advertising the membership a bit too far? I don’t think there’s a need for this, it looks intrusive
> 
> It helps with advertising more. This is a great change.

Uhhhh... not sure about that...

JereNwa:
I mean come on? Cat blocks, profile decorations and profile badges, behind a paywall? I hate to be _that guy_, but I don't like where all of this is going...

rosics-code:
Best update ever!

cambegonefr:
> > This is a good way to promote membership! I support the idea of Scratch paywalling things now.
> 
> I think you need to take a break from GitHub. It looks like you’ve lost your mind.

Clock that because ain't no way you started this protest only to switch up. Attention seeking at it's finest.

PR #9923, opened 2025-11-24
feat: Add ‘membership’ to account dropdown 

rosics-code:
### Changes:

Adds ‘membership’ <a> tag to account dropdown.

### Test Coverage:

Not right now.

### DIFF
--- a/src/components/navigation/www/accountnav.jsx
+++ b/src/components/navigation/www/accountnav.jsx
@@ -52,12 +52,12 @@ const AccountNav = ({
             </li>
             <li>
                 <a href="/mystuff/">
-                    Membership
+                    <FormattedMessage id="general.myStuff" />
                 </a>
             </li>
             <li>
                 <a href="/membership/">
-                    <FormattedMessage id="general.membership" />
+                    Membership
                 </a>
             </li>
             {isEducator ? [


unconstructable13:
No support. This is literally just advertising the membership again isn’t it

medlans:
> > _"scratchfoundation.org already advertises the membership more than scratch itself, to the point where it would be better off as a .com domain instead. why?"_
> 
> I get what you're saying. The second you open up the website a giant popup appears telling you to donate as well is the button and banner the very top.
> Like, this is enough already. In fact _more_ than enough advertising.

Also, look at the Announcements forum right now
announcements

There's even more when you scroll down or go to the next 5 pages

JereNwa:
> > > _"scratchfoundation.org already advertises the membership more than scratch itself, to the point where it would be better off as a .com domain instead. why?"_
> > 
> > 
> > I get what you're saying. The second you open up the website a giant popup appears telling you to donate as well is the button and banner the very top.
> > Like, this is enough already. In fact _more_ than enough advertising.
> 
> Also, look at the Announcements forum right now announcements
> 
> There's even more when you scroll down and especially in pages 1-6

It's nothing but member badges.

unrelated but it looks so weird seeing a modern browser with old macOS.

PR #9933, opened 2025-11-28
TOU: [UEPR-426] Gather country and Birth Data when missing

adzhindzhi:
### Resolves:

[UEPR-426](https://scratchfoundation.atlassian.net/browse/UEPR-426)

### Changes:

- Update the Location Tou step to also gather Birth Data
- Handle missing country info correctly

[UEPR-426]: https://scratchfoundation.atlassian.net/browse/UEPR-426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/page/www/page.jsx
+++ b/src/components/page/www/page.jsx
@@ -57,7 +57,12 @@ const Page = ({
                     <Navigation />
                 </nav>
                 <PrivacyBanner />
-                <main id="view">
+                <main
+                    id="view"
+                    className={classNames({
+                        'blocking-view': shouldDisplayBlockingPage
+                    })}
+                >
                     {shouldDisplayTouModal && <TouModal user={user} />}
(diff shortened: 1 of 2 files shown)


mxmou:
Not related to this PR specifically, but I've noticed that on the Terms of Use branch, the background color of the splash page and other pages is slightly different (`#ffffff` instead of `#fcfcfc`) because of [this code](https://github.com/scratchfoundation/scratch-www/blob/9f553d2/src/components/tou-flow/parental-consent-required-page.scss#L4). Is that intended?

It look like this on develop:
scratch_background_fcfcfc

And on release/terms-of-use:
scratch_background_ffffff

PR #9946, opened 2025-12-04
[UEPR-423] Red Tooltip Not On Email Input

kbangelov:
### Resolves:

(https://scratchfoundation.atlassian.net/browse/UEPR-423)

### Changes:

Added the classes that add some margin top to the email field to the validation tooltip as well.

### Thoughts:

It is a little suspicious whether we really want that to be margin top of the email field or bottom of the previous component.

### DIFF
--- a/src/components/join-flow/email-step.jsx
+++ b/src/components/join-flow/email-step.jsx
@@ -230,8 +230,6 @@ class EmailStep extends React.Component {
                                 className={classNames(
                                     'join-flow-input',
                                     'join-flow-input-tall',
-                                    'join-flow-email-input',
-                                    {'join-flow-email-input-padding-sm': this.props.requiresExternalVerification},
                                     {fail: errors.email}
                                 )}
                                 error={errors.email}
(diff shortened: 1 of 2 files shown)


src/components/join-flow/email-step.jsx
```diff
@@ -240,7 +240,11 @@ class EmailStep extends React.Component {
                                 placeholder={this.props.intl.formatMessage({id: 'general.emailAddress'})}
                                 type="email"
                                 validate={this.validateEmail}
-                                validationClassName="validation-full-width-input"
+                                validationClassName={classNames(
+                                    'validation-full-width-input',
+                                    'join-flow-email-input',
+                                    {'join-flow-email-input-padding-sm': this.props.requiresExternalVerification}
+                                )}
```
adzhindzhi: While that would resolve the original issue, on smaller screens the error message appears below the input field, and this change would add a large margin between the field and the error message. For example:
image

Instead, we can use the `wrapperClassName` field on the `FormikInput` and add define the margin classes there, so that they'd be applied both for the input field and the tooltip.

src/components/join-flow/join-flow-steps.scss
```diff
@@ -41,6 +41,7 @@
 
 .validation-full-width-input {
     transform: translate(21.8125rem, 0);
+    margin-top: 0;
```
kbangelov: This will only apply for the small screen full width input, where we don't want the default margin: inherit (for margin-top anyways)

adzhindzhi (approved):
Looks good to me!

PR #10069, opened 2026-02-10
fix: revert scratch-storage to fix fetch worker errors

cwillisf:
### Changes:

Reverting to `scratch-storage@6.1.3` to fix an issue causing browsers to attempt to load `fetch-worker.*.js` from the wrong path. This doesn't seem to be causing user-facing trouble, but it's causing extra noise in our server logs.

We previously applied the same temporary fix to the editor, where the issue _was_ causing user-facing trouble. Another way to view this change is that it updates the `scratch-www` dependency on `scratch-storage` to match the version in this release of `scratch-editor`.

An upcoming release will include a new, fixed version of `scratch-storage`. The issue was present from `scratch-storage@6.1.4` through `scratch-storage@6.1.7` and was fixed in `scratch-storage@6.1.8`. I would have used 6.1.8 for this hotfix, but it hasn't been through QA yet, so I'd rather stick with this known quantity.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "react-twitter-embed": "4.0.4",
         "react-use": "17.6.0",
         "scratch-parser": "6.0.0",
-        "scratch-storage": "^6.1.6"
+        "scratch-storage": "6.1.3"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


KManolov3 (approved):
Looks good to me!

PR #10090, opened 2026-03-04
feat: add new invalid-parent-email info step as part of ToS acceptanc…

KManolov3:
…e flow

### Resolves:

https://scratchfoundation.atlassian.net/browse/UEPR-525

### Changes:

When we detect an invalid email (which was considered valid during registration) during the ToS acceptance flow, allow users to continue with recording consent, but mark the email as unverified to force re-verification (which would probably mean changing their email). Add a new info step to the modal flow.

### DIFF
--- a/src/components/tos-flow/tos-flow.jsx
+++ b/src/components/tos-flow/tos-flow.jsx
@@ -108,6 +108,11 @@ const TosFlow = ({user, onComplete, refreshSession}) => {
             () => refreshSession()
         );
     }, [user, refreshSession]);
+    
+    const handleCompleteFlow = useCallback(() => {
+        refreshSession();
+        onComplete();
+    }, [onComplete, refreshSession]);
 
     const handleOfAgeConfirmation = useCallback(() => {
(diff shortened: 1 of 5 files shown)


src/components/tos-flow/tos-flow.jsx
```diff
@@ -124,12 +133,22 @@ const TosFlow = ({user, onComplete, refreshSession}) => {
...
+                    onComplete();
+                }
+            }
+        );
     }, [user, onComplete, refreshSession]);
 
+    const handleInvalidParentEmailInfoClose = useCallback(() => {
+        refreshSession();
+        onComplete();
+    }, [onComplete, refreshSession]);
```
adzhindzhi: nitpick: Maybe we can name this something like `handleCompleteFlow()` and reuse it in `handleSubmitStep` to avoid a bit of repetition? Though that's really a nitpick, so it's fine leaving it as-is.

adzhindzhi (approved):
Looks great! 🙌

PR #10101, opened 2026-03-08
Refactor renderHomepageRows to use conditional rendering

rosics-code:
### Resolves:

They always go down and look ugly on the homepage.

### Changes:

It changes the featured studios & projects row to be on a condition if there's no projects in it, don't show it.
### Test Coverage:

Screenshot 2026-03-08 11 15 04 AM


rosics-code:
@colbygk What do you think?

PR #10122, opened 2026-03-30
[UEPR-542] Make updates to logged-out homepage, About page, and footer

kbangelov:
### Resolves:

https://scratchfoundation.atlassian.net/browse/UEPR-542

### Changes:

In logged out homepage:
- replaced old "donate" banner with new "become a member" banner.
- replaced video with new one
- replaced links in page and footer


src/views/splash/become-member/become-member-banner.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+const navigateToBecomeMemberPage = () => {
+    window.location = getBecomeMemberInfo().buttonLink;
+};
+
+// track clicks going out to the become member page from this banner
+const captureOutboundLinkToBecomeMember = () => {
+    window.dataLayer = window.dataLayer || [];
+    window.dataLayer.push({
+        event: 'donate_banner_click'
```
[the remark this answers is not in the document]
kbangelov: Yes so I was not sure if we should change that. I'm guessing it will automatically start a different statistic? Still wanted to ask.

src/components/footer/www/footer.jsx
```diff
@@ -22,7 +22,7 @@ const Footer = props => (
             <div className="lists">
                 <dl>
                     <dd>
-                        <a href="/about">
```
adzhindzhi, line 25: We should also:
1) Archive the `/about` endpoint. The code and the pages can still remain in case they are needed for future reference, but `/about `should return 404. @KManolov3 we should also double check the Fastly configuration once this gets merged (I remember we had an issue when we were removing the `/privacy_policy `and `/terms_of_service` pages)
2) Link to the Scratch Foundation homepage in scratchr2 pages as well

src/views/splash/become-member/become-member-banner.scss
adzhindzhi: Let's update this color to`colors.$ui-scratchfoundation-blue` as well, because right now the text in the "Become a member" button is still green 😅

src/components/footer/www/footer.jsx
KManolov3: Tbh I'd remove the code / pages -  we'll still have them in the git history. But not insisting.

src/views/splash/become-member/become-member-banner.jsx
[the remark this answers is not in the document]
KManolov3: We don't have such an event in Google Analytics, so let's remove the event logging logic entirely for now.

KManolov3 (approved):
Looks good to me! 🙌

adzhindzhi (approved):
Looks good! 🙌

src/components/footer/www/footer.jsx
kbangelov: I did remove them just now.

KManolov3 (approved):

PR #10137, opened 2026-04-07
accessibility: Update maintenance and trouble screens to current times

rosics-code:
### Changes:

updates background to #9966FF

And changed twitter to X


rosics-code:
@colbygk Do you think this is not needed?

PR #10139, opened 2026-04-07
fix: remove old 2023 update code

rosics-code:
### Changes:

Removes the 2023 contrast update banners since it was over 2 years ago..


ampelc:
they could be kept because they seem like a good base for other banners

PR #10142, opened 2026-04-08
[UEPR-550] Remove FAQ page from scratch-www and redirect to /contact-us

kbangelov:
### Resolves:

Part of https://scratchfoundation.atlassian.net/browse/UEPR-550

### Changes:

Deletes FAQ page and replaces links with /contact-us


src/routes.json
```diff
@@ -446,7 +433,7 @@
         "name" : "faq-redirect",
         "pattern": "^/info/faq/?$",
         "routeAlias": "/info/(cards|credits|faq|donate)/?$",
-        "redirect" : "/faq"
```
KManolov3, line 449: I think we'd have an issue updating this redirect for the same reason we can't add any new redirects. @cwillisf can confirm/deny the above.

We have two options, of which I believe both should work:
1) We leave this as-is. A request to `/info/faq` gets handled by www and rewritten to `/faq`. That gets handled by scratchr2 and redirected to `/contact-us`, which ends up being handled by www and the page served to the user
2) We delete this. In scratchr2 we add support for `/info/faq` as well (which is actually the current state of the PR there. Then `/info/faq` gets handled by scratchr2 and rewritten to `/contact-us`, which gets handled by www and the page served to the user.

src/routes.json
```diff
@@ -446,7 +433,7 @@
...
+        "redirect" : "/contact-us"
```
adzhindzhi: I think @KManolov3 already mentioned that somewhere, but we have a problem with adding new redirects. Let's do this in `scratchr2` instead by removing the FAQ page from www, letting it fallback to `scratchr2`, where we redirect to `contact-us`.

adzhindzhi: Also, do we need this redirect at all now?

src/components/modal/email-confirmation/modal.jsx
```diff
@@ -76,7 +76,7 @@ const EmailConfirmationModal = ({
                             id="emailConfirmationModal.wantMoreInfo"
                             values={
                                 {FAQLink:
-                                    (<a href="/faq#accounts">
+                                    (<a href="/contact-us">
```
adzhindzhi: Shouldn't we point to Support here as well? I think generally everywhere we used to point to `/faq`, we should now point to Support. And we should redirect to `scratch.mit.edu/contact-us` only when they try to navigate to `scratch.mit.edu/faq` directly.

src/lib/external-links.js
```diff
@@ -117,6 +117,7 @@ const externalLinks = /** @type {const} */ {
         scratchConference: 'https://www.scratchfoundation.org/scratch-conference',
         shop: 'https://shop.scratchfoundation.org/',
         shopDonate: 'https://shop.scratchfoundation.org/pages/donate',
+        support: 'https://mitscratch.freshdesk.com/en/support/home',
```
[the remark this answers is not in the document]
adzhindzhi: Yep, let's use `scratchHelpDesk.helpCenter`

src/views/ideas/ideas.jsx
```diff
@@ -412,7 +412,7 @@ const Ideas = ({
                                 <FormattedMessage
                                     id="ideas.questionsBodyHTML"
                                     values={{
-                                        faq: chunks => <a href="/info/faq">{chunks}</a>,
+                                        faq: chunks => <a href="/contact-us">{chunks}</a>,
```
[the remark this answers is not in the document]
adzhindzhi: Same comment, let's point to the Support link

src/views/parents/parents.jsx
```diff
@@ -147,13 +147,6 @@ const Landing = () => {
                         <FormattedMessage
                             id="parents.faqMoreAndAsk"
                             values={{
```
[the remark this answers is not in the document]
adzhindzhi: Let's bring this back and navigate to the Support link.

src/views/preview/unsupported-browser.jsx
```diff
@@ -35,7 +35,7 @@ const UnsupportedBrowser = () => (
                             faqLink: (
                                 <a
                                     className="faq-link"
-                                    href="/faq"
+                                    href="/contact-us"
                                 >
                                     <FormattedMessage id="general.faq" />
```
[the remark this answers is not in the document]
adzhindzhi: Same comment as above - let's navigate to Support.

src/lib/external-links.js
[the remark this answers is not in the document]
KManolov3: Actually the `helpCenter` link also includes a locale since this commit - https://github.com/scratchfoundation/scratch-www/commit/0f9cb56aec29d4ac999cbfa9f0a176c03d15d994/.

@kbangelov please pull in the newest develop changes, as I was confused by this as well for a while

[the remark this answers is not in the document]
KManolov3: Either way, I agree that we should use one link in all places. I added a question in the ticket as to whether the hardcoded locale is intentional.

src/routes.json
```diff
@@ -430,12 +423,6 @@
...
         "title": "LEGO BOOST"
     },
-    {
-        "name":"3-faq-redirect",
-        "pattern": "^/3faq/?$",
-        "routeAlias": "/3faq/?$",
-        "redirect": "/faq#scratch3"
-    },
     {
         "name" : "credits-redirect",
```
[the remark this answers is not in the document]
KManolov3: Let's keep the redirect

adzhindzhi (approved):
Looks good to me! 🙌

src/views/preview/unsupported-browser.jsx
```diff
@@ -1,3 +1,5 @@
+import externalLinks from '../../lib/external-links';
```
adzhindzhi: Actually, is this valid syntax? In Node.js, `import` can only be used in ES Modules (ESM). `CommonJS` files need to use `require()` instead.

kbangelov: Good call.

src/routes.json
rosics-code: > I think @KManolov3 already mentioned that somewhere, but we have a problem with adding new redirects. Let's do this in `scratchr2` instead by removing the FAQ page from www, letting it fallback to `scratchr2`, where we redirect to `contact-us`.

isn’t scratch.mit.edu/faq in WWW so what’s the need for scratchr2 fallback if contact-us is in WWW?

src/components/modal/email-confirmation/modal.jsx
kbangelov: Will agreed that we should point /faq to /contact-us, I might as well clear up whether that applies for /faq#accounts as well.

PR #10151, opened 2026-04-12
fix: Fix Community Guidelines off-screen navigation bug

rosics-code:
### Resolves:

The community guidelines page has been broken for months

### Changes:

Adds margin-inline (auto value) CSS property to hoepfully fix it

### Test Coverage:

Without margin-inline: auto;
Screenshot 2026-04-12 11 02 00 AM

With margin-inline: auto;
Screenshot 2026-04-12 11 02 43 AM

### DIFF
--- a/src/views/guidelines/guidelines.scss
+++ b/src/views/guidelines/guidelines.scss
@@ -46,7 +46,9 @@
         flex-direction: column;
         justify-content: center;
         gap: 4.5rem;
-        margin: 6rem 15% 0 0;
+        margin: 6rem 0 0 0;
+        padding: 0 15%;
+        box-sizing: border-box;
         padding-bottom: 6rem;
 
         border-bottom: 1px solid colors.$active-gray;
--- a/src/views/guidelines/guidelines.scss
+++ b/src/views/guidelines/guidelines.scss
@@ -48,7 +48,7 @@
         gap: 4.5rem;
         margin: 6rem 15% 0 15%;
         padding-bottom: 6rem;
-        max-width: 70%
+        max-width: 70%;
 
         border-bottom: 1px solid colors.$active-gray;
         
--- a/src/views/guidelines/guidelines.scss
+++ b/src/views/guidelines/guidelines.scss
@@ -46,9 +46,8 @@
         flex-direction: column;
         justify-content: center;
         gap: 4.5rem;
-        margin: 6rem 15% 0 15%;
+        margin: 6rem 15% 0 0;
         padding-bottom: 6rem;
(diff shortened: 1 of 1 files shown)


rosics-code:
@cwillisf @colbygk This has been a bug since January

src/views/guidelines/guidelines.scss
```diff
@@ -48,6 +48,7 @@
         gap: 4.5rem;
         margin: 6rem 15% 0 15%;
         padding-bottom: 6rem;
+        margin-inline: auto;
```
KManolov3: I think the issue stems from the `width: 100%` coming from `modal-navigation.scss` being applied to the `.navigation` class here as well. The margin-inline: auto actually overrides the horizontal margins being declared in the margin property above. 

A safe fix may be to add a `max-width: 70%` to the class here (accounting for the horizontal margins), instead of the margin-inline.

rosics-code:
Forgot a semi-colin whoops but should be all good

KManolov3 (approved):
Looks good to me! 🙌

rosics-code:
> Looks good to me! 🙌

Do I wait for a merge? I was trying to find a merge button.

KManolov3:
> > Looks good to me! 🙌
> 
> Do I wait for a merge? I was trying to find a merge button.

Yep, we'll merge this into develop when we start testing for the next release.

rosics-code:
@KManolov3 I don't think we need max width.

The margin property itself is the issue, right now its 6rem 15% 0 15% which looks like this:
Screenshot 2026-04-18 9 18 34 PM

6rem 15% 0 0:
image

src/views/guidelines/guidelines.scss
```diff
@@ -46,7 +46,7 @@
         flex-direction: column;
         justify-content: center;
         gap: 4.5rem;
-        margin: 6rem 15% 0 15%;
+        margin: 6rem 15% 0 0;
```
adzhindzhi: I don't think simply removing the margin is enough here. Even though it fixes the "off-screen" bug, it makes the boxes stick to the end of the screen for smaller screens:
Image

This is what the 15% margin was intended to do. I think what we should do instead change that 15% margin into padding by:
- Keeping only `margin-top: 6rem;`
- Adding 15% left/right padding
- Adding `box-sizing: border-box` (to make the padding not affect the whole width of the component)

rosics-code:
Is that correct?

adzhindzhi (approved):
Looks good to me! 🙌

PR #10175, opened 2026-04-22
fix: guard fetchCommunityData in componentDidMount for new projects

cwillisf:
### Resolves:

- Loading `/projects/editor/` or `/projects/editor/?tutorial=...` can sometimes result in a client-rendered "Whoops" page (no HTTP code visible)

### Changes:

Add `projectId > 0` check to `componentDidMount`, matching the existing guard in `componentDidUpdate`. With React 18's `createRoot`, the initial render is not guaranteed to be synchronous, so the `/session/` response can resolve before `componentDidMount` runs. When this race is lost on `/projects/editor/`, `componentDidMount` sees `sessionStatus=FETCHED` and calls `fetchCommunityData` with `projectId` '0', which 404s against the API and sets `projectNotAvailable=true`, showing the "Whoops" error page instead of the editor.

### Test Coverage:

Added new unit tests to kinda-sorta simulate this situation.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,3 +6,4 @@ locales/*
 **/*.min.js
 **/node_modules/*
 scratch-gui/*
+/test/generated/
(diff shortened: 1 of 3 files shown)


adzhindzhi (approved):
Looks good to me! 🙌

PR #10188, opened 2026-04-29
[UEPR-551] Display a read-only mode banner across scratch-www pages

kbangelov:
### Resolves:

Part of https://scratchfoundation.atlassian.net/browse/UEPR-551

### Changes:

Adds read-only-mode banner on said pages.


src/components/read-only-mode-banner/read-only-mode-banner.jsx
```diff
@@ -0,0 +1,59 @@
...
+        if (!bannerRef.current) return;
+        const updateHeight = () => {
+            const height = bannerRef.current ? bannerRef.current.offsetHeight : 0;
+            document.documentElement.style.setProperty('--read-only-banner-height', `${height}px`);
+        };
+        updateHeight();
+        const observer = new ResizeObserver(updateHeight);
+        observer.observe(bannerRef.current);
+        return () => observer.disconnect();
+    }, []);
```
adzhindzhi: I’m not a fan of adding extra `useEffect`s and introducing resize observers just to display a banner. That approach feels like it could introduce performance issues and adds unnecessary complexity. I wonder if switching the banner position from fixed to sticky would help us avoid all of that. It does look less "stable" than `position: fixed` while scrolling, but it seems to interact better with other banners and differently styled pages. For example, the Studio page doesn't behave as expected right now.

src/components/read-only-mode-banner/read-only-mode-banner.jsx
```diff
@@ -0,0 +1,59 @@
...
+
+    if (!READ_ONLY_MODE) return null;
+
+    return (
+        <aside
```
adzhindzhi: Is aside the semantically correct tag here?

src/components/read-only-mode-banner/read-only-mode-banner.jsx
```diff
@@ -0,0 +1,59 @@
...
+            className="read-only-mode-banner"
+            ref={bannerRef}
+            style={{zIndex}}
```
adzhindzhi: Is there a way to avoid passing the `z-index`? It's very css specific thing to pass, maybe allow for passing a `className` instead and combine it with the `read-only-mode-banner` class?

src/components/modal/join/modal.scss
```diff
@@ -18,7 +18,11 @@ being clickable, because of the standalone join view. */
     }
 }
 
+// push the modal content down to accommodate the fixed nav + read-only banner when active
 .modal-content {
+    margin-top: calc(1rem + 50px + var(--read-only-banner-height, 0px)) !important;
+    // important because otherwise gets overridden for medium and smaller screens
+
```
adzhindzhi: The way the Join modal is rendered, and the fact that we have to manually push the modal portal down, is a bit unfortunate. I wonder if we could do something like:

- Make the banner sticky and remove all the calculation logic
- Pass an `overlayClassName` to the modal that changes its position from `absolute` to `relative`, so it takes preceding DOM elements into account
- Change the nav position to relative so it’s included in the layout as well

We'd likely need to make some additional adjustments, such as tweaking paddings and making the body scrollable, but overall this approach should work.

Alternatively, we could try adding the banner directly to the modal portal, possibly by passing a custom overlay element. Something like:
```
const overlayElement = (props, contentEl) => (
    <div {...props}>
        <ReadOnlyModeBanner />
        {contentEl}
    </div>
);
```
I'm not entirely sure how well that would work, but it seems like it could give us the exact behavior we want. As opposed to my previous suggestion, which would basically render the banner as relative, and won't display over the modal when scrolling.

src/components/read-only-mode-banner/read-only-mode-banner.scss
```diff
@@ -0,0 +1,45 @@
...
+        text-decoration: underline;
+    }
+}
+
+@media only screen and (max-width: 1024px) {
+    .read-only-mode-banner .read-only-mode-banner-content {
+        width: 100%;
+        padding: 0 40px;
+        box-sizing: border-box;
+    }
```
adzhindzhi, line 56: nitpick, but I think the other banners behave a bit differently for smaller screens, and if more than one banner is displayed at a time, they seem like they are not aligned 😅

src/lib/feature-flags.js
```diff
@@ -8,5 +8,6 @@ const flagInUrl = flag => {
 };
 
 module.exports = {
-    CONTACT_US_POPUP: flagInUrl('CONTACT_US_POPUP')
+    CONTACT_US_POPUP: flagInUrl('CONTACT_US_POPUP'),
+    READ_ONLY_MODE: process.env.READ_ONLY_MODE === 'true'
```
adzhindzhi, line 12: I don't mind this, but for most flags (e.g. `MANUALLY_SAVE_THUMBNAILS`) we use `process.env` directly

src/l10n.json
```diff
@@ -603,5 +603,9 @@
     "tos.invalidParentEmailInfoStepDescriptionPart4": "Once your parent confirms the new email, your account will be confirmed, and social features will be available again.",
 
     "studentDeactivationBanner.title": "An important reminder about student accounts",
-    "studentDeactivationBanner.description": "Using a student account? Don’t lose your stuff! Student accounts are temporary. See <a>this Help Desk article</a> for details about student accounts and how to make a new account outside of your classroom."
+    "studentDeactivationBanner.description": "Using a student account? Don’t lose your stuff! Student accounts are temporary. See <a>this Help Desk article</a> for details about student accounts and how to make a new account outside of your classroom.",
+
+    "readOnlyMode.bannerText1": "<b>[PLACEHOLDER]</b> We’re moving your stuff to an upgraded version of Scratch!",
```
adzhindzhi: nitpick: Can we name this `readOnlyModeBanner.title` and the other two `readOnlyModeBanner.description1`, `readOnlyModeBanner.description2` or something like that 🙏 It just provides a bit more context as to what each text is.

src/l10n.json
```diff
@@ -603,5 +603,9 @@
...
+    "readOnlyMode.bannerText2": "Don’t worry - none of your stuff will be lost. To help everything run smoothly, we’ve paused all social actions for now. You would still be able to create and edit unshared projects.",
+    "readOnlyMode.bannerText3": "When everything is ready, you'll be taken to the upgraded version of Scratch and you'll be able to use everything like normal again. For more information, check out our {faqLink}."
```
adzhindzhi: The `faqLink` is referenced here, but isn't provided in the values when that message is used in the component.

src/components/read-only-mode-banner/read-only-mode-banner.jsx
kbangelov: Perhaps just a **div** is more appropriate. **header** (role = banner) here doesn't seem to make too much sense either.

src/l10n.json
kbangelov: The final copy is to be confirmed.

src/components/read-only-mode-banner/read-only-mode-banner.jsx
kbangelov: I resolved it but I had to use position: relative for the modal, also adding some overlay and page backround color logic. But I do believe it's cleaner and more stable now.

src/components/read-only-mode-banner/read-only-mode-banner.scss
kbangelov: All the banners I tested with seem to expand on multiple rows (become a member banner, student accounts info banner), so what kind of misalignment do you mean?

src/lib/feature-flags.js
kbangelov: I think it makes it easier to read as code since it appears in a few places.

src/components/read-only-mode-banner/read-only-mode-banner.jsx
```diff
@@ -0,0 +1,47 @@
...
+const FormattedMessage = require('react-intl').FormattedMessage;
+
+require('./read-only-mode-banner.scss');
+
+const {READ_ONLY_MODE} = require('../../lib/feature-flags');
+
+const bold = chunks => <strong>{chunks}</strong>;
+
+const ReadOnlyModeBanner = ({className}) => {
+    const bannerRef = useRef(null);
```
adzhindzhi: We no longer need this, right?

src/components/read-only-mode-banner/read-only-mode-banner.scss
```diff
@@ -0,0 +1,45 @@
...
+.read-only-mode-banner {
+    display: flex;
+    width: 100%;
+    padding: 1.5rem 0;
+    flex-direction: column;
+    align-items: center;
+    gap: 0.625rem;
+    background-color: colors.$ui-sky-blue;
+    position: sticky;
+    flex: 1;
```
adzhindzhi: You probably don't need this?

src/components/read-only-mode-banner/read-only-mode-banner.scss
```diff
@@ -0,0 +1,45 @@
+@use "../../colors";
+@use "../../frameless";
+
+.read-only-mode-banner {
+    display: flex;
+    width: 100%;
+    padding: 1.5rem 0;
+    flex-direction: column;
+    align-items: center;
+    gap: 0.625rem;
```
adzhindzhi: Do you need that gap? The `.read-only-mode-banner` class has a single div inside, so this seems unused.

src/components/read-only-mode-banner/read-only-mode-banner.scss
```diff
@@ -0,0 +1,45 @@
...
+    background-color: colors.$ui-sky-blue;
+    position: sticky;
+    flex: 1;
+    top: 50px; /* matches fixed nav height */
+    z-index: 9; /* below nav (z-index: 10) */
+
+    .read-only-mode-banner-content {
+        width: 940px;
+        max-width: 100%;
```
adzhindzhi: Does this do anything when we have a fixed `width` set?

src/components/read-only-mode-banner/read-only-mode-banner.scss
```diff
@@ -0,0 +1,45 @@
...
+        display: flex;
+        flex-direction: column;
+        gap: 10px;
```
adzhindzhi: Let's use `rems`. For the `width` as well.

src/components/read-only-mode-banner/read-only-mode-banner.scss
adzhindzhi: For example, for screens between `460px` and `767px`:

Image

Other banners (e.g. `Share banner`, `Student banner`) start centering the content, which makes this one seem a bit inconsistent when displayed next to them. I wouldn't insist on necessarily changing this, but I think it's worth mentioning.

src/views/join/join.scss
```diff
@@ -1,19 +1,40 @@
...
+.mod-join-overlay {
+    position: relative;
+    min-height: 100vh;
+
+    &::before {
+        content: "";
+        position: fixed;
+        inset: 0;
+        background-color: colors.$ui-blue;
+    }
```
adzhindzhi, line 37: Why do we need that? Can we not add the `background-color: colors.$ui-blue;` to the `.mod-join-overlay`? Or is there another reason why we need this?

src/main.scss
```diff
@@ -176,7 +176,8 @@ dl {
     display: inline-block;
 
     /* NOTE: Margin should match height in navigation.scss */
-    margin-top: 50px;
+    /* --read-only-banner-height is set dynamically by ReadOnlyModeBanner via ResizeObserver (0px when inactive) */
+    margin-top: calc(50px + var(--read-only-banner-height, 0px));
```
adzhindzhi: We no longer need this, right?

adzhindzhi (commented):
One additional general comment: The body is currently not scrollable (due to the `.overflow-hidden` class), so some of the content gets cut.

src/views/join/join.scss
kbangelov: Since we changed the position from absolute to relative, it will only cover the height of the modal itself. We add min-height and the before handling in order to cover the spaces around it. (Since the .join-page class doesn't reach the modal due to the modal computing to a child of body - on the same level in the html structure, not a child).

That's the solution I thought of, anyways.

src/components/read-only-mode-banner/read-only-mode-banner.scss
kbangelov: I made it comply with the student account banner in terms of shrinking for smaller screen width by adding the same styles.

src/components/page/www/page.jsx
```diff
@@ -58,6 +59,7 @@ const Page = ({
                 >
                     <Navigation />
                 </nav>
+                <ReadOnlyModeBanner />
                 <PrivacyBanner />
                 {shouldDisplayStudentDeactivationBanner && <StudentDeactivationBanner username={user.username} />}
```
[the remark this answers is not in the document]
kbangelov: Don't think they do in practice.

src/components/modal/base/modal.jsx
```diff
@@ -26,8 +26,9 @@ class Modal extends React.Component {
         // bodyOpenClassName prop cannot be blank string or null here; both cause
         // an error, because ReactModal does not correctly handle them.
         // If we're not setting it to a class name, we must omit the prop entirely.
-        const bodyOpenClassNameProp = this.props.useStandardSizes ?
-            {bodyOpenClassName: classNames('overflow-hidden')} : {};
+        const bodyOpenClassNameProp = this.props.bodyOpenClassName ?
+            {bodyOpenClassName: this.props.bodyOpenClassName} :
+            (this.props.useStandardSizes ? {bodyOpenClassName: classNames('overflow-hidden')} : {});
```
adzhindzhi, line 31: nitpick: I wonder if it'd make more sense to expose this as `allowVerticalOverflow`/`allowHorizontalOverflow` props, rather than passing a className, which may be too vague.

adzhindzhi (approved):
Left one minor comment, but otherwise looks good to me! 🙌

src/components/modal/base/modal.jsx
kbangelov: For this single use, yes. If we were thinking long-term, then maybe other styles that we want to apply might come up.

PR #10189, opened 2026-04-29
[UEPR-553] Front-end blocking in scratch-www

kbangelov:
### Resolves:

Part of https://scratchfoundation.atlassian.net/browse/UEPR-553

### Changes:

Disable frontend operations for PDP and Studios


src/views/preview/presentation.jsx
```diff
@@ -210,7 +212,7 @@ const PreviewPresentation = ({
                     message={<FormattedMessage id="project.share.sharedShort" />}
                 />);
             }
-        } else if (!isShared) {
+        } else if (!isShared && !READ_ONLY_MODE) {
```
adzhindzhi: I think by design the banner is displayed but the Share button is disabled

src/views/preview/presentation.jsx
```diff
@@ -591,6 +593,7 @@ const PreviewPresentation = ({
                             loveCount={loveCount}
                             loved={loved}
                             projectInfo={projectInfo}
+                            readOnlyMode={READ_ONLY_MODE}
```
adzhindzhi: You pass this flag to stats, but do you use it? Also, I think the love and favorite buttons look different when they are disabled, a bit lighter gray.

src/views/preview/presentation.jsx
```diff
@@ -601,6 +604,7 @@ const PreviewPresentation = ({
                             isAdmin={isAdmin}
                             isShared={isShared}
                             projectInfo={projectInfo}
+                            readOnlyMode={READ_ONLY_MODE}
```
adzhindzhi: Same question as above, is this actually used?

src/views/preview/project-view.jsx
```diff
@@ -754,7 +756,7 @@ class Preview extends React.Component {
         }
     }
     handleFavoriteToggle () {
-        if (!this.props.favedLoaded) return;
+        if (!this.props.favedLoaded || READ_ONLY_MODE) return;
```
adzhindzhi: You probably wouldn't need these if you apply the disabled styles to the love/favorite buttons + `pointer-events: none`.

src/views/preview/project-view.jsx
```diff
@@ -1195,10 +1197,12 @@ class Preview extends React.Component {
                                     basePath="/"
                                     canCreateCopy={this.props.canCreateCopy}
                                     canCreateNew={this.props.canCreateNew}
-                                    canEditTitle={this.props.canEditTitleInEditor}
+                                    canEditTitle={this.props.canEditTitleInEditor &&
+                                        (!READ_ONLY_MODE || !this.props.isShared)}
```
adzhindzhi: nitpick: Maybe we can extract `(!READ_ONLY_MODE || !this.props.isShared)` into `canModifyProject` or something like that to avoid repetition?

src/views/preview/project-view.jsx
```diff
@@ -1195,10 +1197,12 @@ class Preview extends React.Component {
...
                                     canRemix={this.props.canRemix}
-                                    canSave={this.props.canSave}
-                                    canShare={this.props.canShare}
+                                    canSave={this.props.canSave &&
+                                        (!READ_ONLY_MODE || !this.props.isShared)}
```
adzhindzhi: This doesn't handle pressing CTRL + 'S' in the editor, right? I wonder if we should try to disable that, or let the backend handle it and just display an error... 🤔

src/views/preview/project-view.jsx
```diff
@@ -1415,15 +1423,12 @@ const mapStateToProps = state => {
...
-    // if we don't have projectInfo, assume it's shared until we know otherwise
-    const isShared = !projectInfoPresent || state.preview.projectInfo.is_published;
-
     return {
         authorId: authorId,
         authorThumbnailUrl: thumbnailUrl(authorId),
         authorUsername: authorUsername,
         authorAvatarBadge: authorAvatarBadge,
-        canAddToStudio: isLoggedIn && isShared,
+        canAddToStudio: isLoggedIn && isShared && !READ_ONLY_MODE,
```
adzhindzhi: Hmm, this would remove the whole button, right? It's disabled by design, I don't mind it, but let's confirm with Tereza if that's okay.

adzhindzhi (commented):
I haven't reviewed everything yet, but I'm sharing my comments so far.

src/views/preview/project-view.jsx
rosics-code: Doesn't CTRL+S in Chrome (maybe some Chromium browsers too) activate a system keybind to save a page in a .mhtml?

src/views/preview/presentation.jsx
rosics-code: @adzhindzhi Doesn't the share button hide once someone shares a project?

src/views/studio/studio-curator-invite.jsx
```diff
@@ -42,7 +44,8 @@ const StudioCuratorInvite = ({showCuratorInvite, onSubmit}) => {
             </div>
             <button
                 className={classNames('studio-invitation-button button', {
-                    'mod-mutating': submitting
+                    'mod-mutating': submitting,
+                    'studio-disabled': READ_ONLY_MODE
                 })}
                 disabled={submitting}
                 onClick={() => {
```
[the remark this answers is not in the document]
adzhindzhi: Yep, that's a valid point, also `pointer-events: none` would resolve that even if the button is not disabled, but I agree that we should add `{submitting || READ_ONLY_MODE}` as we do for the other buttons.

src/views/studio/studio.scss
```diff
@@ -53,6 +53,16 @@ $radius: 8px;
     box-sizing: border-box;
 }
 
+.studio-disabled {
+    background: colors.$ui-background-disabled;
+    color: colors.$ui-text-disabled;
+    cursor: not-allowed;
```
adzhindzhi: Let's change `cursor: not-allowed` to `pointer-events: none`. Same for the other usages in this PR.

src/views/studio/studio-member-tile.jsx
```diff
@@ -59,7 +61,7 @@ const StudioMemberTile = ({
                     </div>
                 }
             </div>
-            {(canRemove || canPromote || canTransferHost) &&
+            {(canRemove || canPromote || canTransferHost) && !READ_ONLY_MODE &&
```
adzhindzhi: I think by design the menu is disabled in readonly mode, not hidden.

src/views/studio/studio-project-tile.jsx
```diff
@@ -47,7 +49,7 @@ const StudioProjectTile = ({
                         className="studio-project-username"
                     >{username}</a>
                 </div>
-                {canRemove &&
+                {canRemove && !READ_ONLY_MODE &&
```
adzhindzhi: Same comment as above, I think the menu should be disabled rather than hidden, according to the designs.

src/views/preview/presentation.jsx
adzhindzhi: Also, note that the styles differ when the project is liked/added to favorites, vs. when it's not.

adzhindzhi (commented):
One additional general comment: We should also disable the "Anyone can add projects" toggle? It shouldn't make a difference, but it would be more consistent.

src/views/preview/project-view.jsx
kbangelov: It is used only twice, I believe, I think it's easy to follow the way it is?

src/views/preview/project-view.jsx
kbangelov: It is handled in the sense that the frontend fails by displaying "Project could not save.", because of this very prop. So I think that's enough?

src/views/preview/project-view.jsx
kbangelov: I think it doesn't hurt to add that check.

src/components/overflow-menu/overflow-menu.jsx
```diff
@@ -5,41 +5,56 @@ import classNames from 'classnames';
...
             <button
                 className={classNames('overflow-menu-trigger', {
-                    'ignore-react-onclickoutside': open
+                    'ignore-react-onclickoutside': open,
+                    'is-disabled': disabled
                 })}
-                onClick={() => setOpen(!open)}
+                onClick={() => {
+                    if (!disabled) setOpen(!open);
+                }}
```
adzhindzhi: I think this check is unnecessary when we've marked the button as `disabled`.

src/components/overflow-menu/overflow-menu.jsx
```diff
@@ -5,41 +5,56 @@ import classNames from 'classnames';
...
     const [open, setOpen] = useState(false);
+    const iconSrc = disabled ? overflowIconDisabled : overflowIcon;
+
     return (
         <div className={classNames('overflow-menu-container', className)}>
             <button
                 className={classNames('overflow-menu-trigger', {
-                    'ignore-react-onclickoutside': open
+                    'ignore-react-onclickoutside': open,
+                    'is-disabled': disabled
```
adzhindzhi: Is this class used? I don't see it defined anywhere?

src/components/overflow-menu/overflow-menu.jsx
```diff
@@ -5,41 +5,56 @@ import classNames from 'classnames';
...
             </button>
-            {open && <Dropdown
-                isOpen
-                as={dropdownAs}
-                className="overflow-menu-dropdown"
-                onRequestClose={() => setOpen(false)}
-            >
-                {children}
-            </Dropdown>}
+            {open && !disabled && (
```
adzhindzhi: Isn't this check also unnecessary? How can `open` become true if the button is disabled?

src/views/preview/banner.jsx
```diff
@@ -6,16 +6,22 @@ const Button = require('../../components/forms/button.jsx');
...
         <FlexRow className="inner banner-inner">
             <span className="banner-text">
                 {message}
             </span>
             {actionMessage && onAction && (
                 <Button
-                    className="banner-button"
-                    onClick={onAction}
+                    className={classNames('banner-button', {'banner-button-disabled': !buttonActivated})}
+                    onClick={buttonActivated ? onAction : null}
```
adzhindzhi: nitpick: It would be more straightforward if the prop was actionDisabled on buttonDisabled, because you could just do:
```
{'banner-button-disabled': buttonDisabled}
disabled={buttonDisabled}
```
Also, I think it makes more sense to pass `disabled` instead of manually stopping the action.

src/views/preview/banner.jsx
```diff
@@ -28,7 +34,12 @@ Banner.propTypes = {
...
     className: PropTypes.string,
     message: PropTypes.node.isRequired,
-    onAction: PropTypes.func
+    onAction: PropTypes.func,
+    buttonActivated: PropTypes.bool
+};
+
+Banner.defaultProps = {
+    buttonActivated: true
 };
```
adzhindzhi: `defaultProps` is deprecated, if you need to explicitly pass them, you can do so in the definition as:
```
const Banner = ({
    className,
    message,
    actionMessage,
    onAction,
    buttonActivated = true
})
```

Though if we just switched it to `buttonDisabled` we wouldn't need to have a default value.

src/views/preview/banner.scss
```diff
@@ -53,3 +53,9 @@ $navigation-height: 50px;
 .banner-danger .banner-button {
     background-color: colors.$ui-red;
 }
+
+.banner-button-disabled {
+    color: colors.$ui-text-disabled;
+    background-color: colors.$ui-background-disabled;
+    pointer-events: none;
```
adzhindzhi: If the button was actually disabled (using the `disabled` prop), we wouldn't need to stop `pointer-events` manually.

src/views/preview/presentation.jsx
```diff
@@ -443,6 +446,7 @@ const PreviewPresentation = ({
                                     projectInfo={projectInfo}
                                     onFavoriteClicked={onFavoriteClicked}
                                     onLoveClicked={onLoveClicked}
+                                    readOnlyMode={READ_ONLY_MODE}
```
adzhindzhi: nitpick, but I think it'd make more sense for this prop (and the underlying classes in stats.css to be `disabled` rather than `read-only`)

src/views/preview/project-view.jsx
adzhindzhi: It's probably more of a preference, but I find this slightly hard to read because of the multiple negations and the `and`/`or` combinations. Usually, if I feel the need to add parentheses to a condition, I'd rather extract it into a named variable 😅

src/views/preview/subactions.scss
```diff
@@ -110,6 +110,16 @@
...
+.action-button.studio-button:disabled {
+    background-color: colors.$ui-background-disabled;
+    color: colors.$ui-text-disabled;
+
+    &:before {
+        // override regular icon
+        background-image: url("/svgs/project/studio-add-gray.svg") !important;
+    }
+}
+
```
adzhindzhi: Can we name the classname more specific, instead of using `!important`?

adzhindzhi: Also, a more general comment, there is a custom `filter: brightness(90%)` style that gets applied when you've clicked on the button, even when it's disabled. We can probably avoid it by adding pointer-events: none to the `.action-button.studio-button:disabled` class, or we could overwrite it manually.

src/views/studio/studio-project-tile.jsx
```diff
@@ -47,8 +48,8 @@ const StudioProjectTile = ({
                         className="studio-project-username"
                     >{username}</a>
                 </div>
-                {canRemove &&
-                    <OverflowMenu>
+                {canRemove && process.env.READ_ONLY_MODE !== 'true' &&
+                    <OverflowMenu disabled={process.env.READ_ONLY_MODE === 'true'}>
```
adzhindzhi: Will the `disabled` attribute ever be `true`? Wouldn't we just not show the menu with the `process.env.READ_ONLY_MODE !== 'true'` check?

src/components/overflow-menu/overflow-menu.jsx
```diff
@@ -39,7 +44,12 @@ const OverflowMenu = ({
...
     children: PropTypes.node,
     dropdownAs: PropTypes.string,
-    className: PropTypes.string
+    className: PropTypes.string,
+    disabled: PropTypes.bool
+};
+
+OverflowMenu.defaultProps = {
+    disabled: false
 };
```
adzhindzhi: I think I left a similar comment on one of the previous versions, but `defaultProps` is deprecated. And especially in functional components, we don't even need it. If we want to provide a default value, we can do so when destructuring the args. And in out specific use case, I don't even think we need a default value, because if disabled is not provided, that would be a falsy value and the component will treat it as not disabled.

adzhindzhi (approved):
Looks good! I left two small comments, but otherwise I think it's pretty clean! One additional general comment is that a few of the disabled buttons still have `cursor: pointer`, which may be a bit confusing from UX perspective. I think I've only seen this for the love, favorite on PDP and Edit Thumbnail on Studios, but there may be other buttons also affected.

PR #10201, opened 2026-05-03
remove: delete the "credits" view

rosics-code:
### Resolves:

https://github.com/scratchfoundation/scratch-www/issues/10200

### Changes:

Deletes the `src/views/credits` directory as it's not in the footer anymore and there's no use for it as it doesn't get updated anymore and it's a pain to update it. Also changed `src/routes.json` to never redirect to a 404 page.


rosics-code:
@KManolov3 What do you think?

PR #10226, opened 2026-05-21
Feature/sunset membership features

KManolov3:
### Resolves:

https://scratchfoundation.atlassian.net/browse/UEPR-603
https://scratchfoundation.atlassian.net/browse/UEPR-608
https://scratchfoundation.atlassian.net/browse/UEPR-604
(parts of) https://scratchfoundation.atlassian.net/browse/UEPR-607

### Changes:

Remove membership specific features. Make some member-exclusive features public


src/views/splash/donate/donate-banner.jsx
```diff
@@ -0,0 +1,65 @@
...
+};
+
+// track clicks going out to the donate page from this banner
+const captureOutboundLinkToDonate = () => {
+    window.dataLayer = window.dataLayer || [];
+    window.dataLayer.push({
+        event: 'donate_banner_click'
+    });
+    setTimeout(navigateToDonatePage, 0);
+};
```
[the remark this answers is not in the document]
KManolov3: This was similar to the preexisting state in https://github.com/scratchfoundation/scratch-www/pull/10122/changes#diff-d54ecf77c56a28a8598023f40050d136136b7ea57c0dcfcf1704d4e0d7058708

rosics-code:
Maybe delete https://scratch.mit.edu/images/membership/scratch-membership.png too? @KManolov3

src/views/membership/membership.jsx
rosics-code, line 1: Instead of nuking the membership view. Maybe keep it as is but do not redirect and/or link it. This is great for future use because we know it's being sunsetted but they are reworking on it, maybe the page can be reused again?

src/views/splash/donate/donate-banner.jsx
```diff
@@ -0,0 +1,65 @@
...
+    window.location = externalLinks.scratchFoundation.donateBanner;
+};
+
+// track clicks going out to the donate page from this banner
+const captureOutboundLinkToDonate = () => {
+    window.dataLayer = window.dataLayer || [];
+    window.dataLayer.push({
+        event: 'donate_banner_click'
+    });
+    setTimeout(navigateToDonatePage, 0);
```
adzhindzhi, line 23: Why do we need this in a `setTimeout`?

src/views/splash/donate/donate-banner.jsx
```diff
@@ -0,0 +1,65 @@
...
+                >
+                    <FormattedMessage id="general.donate" />
+                </Button>
+            </div>
+        </div>
+        <Button
+            isCloseType
+            className="donate-close-button"
+            name="closeButton"
+            type="button"
```
adzhindzhi: nitpick: We probably don't need the `type="button"`.

src/views/splash/donate/donate-banner.scss
```diff
@@ -3,36 +3,36 @@
 
 $tile-height: 244px;
 
-.become-member-banner {
+.donate-banner {
     display: flex;
     position: sticky;
     z-index: 8;
-    background-color: colors.$ui-scratchfoundation-blue;
```
adzhindzhi, line 10: nitpick: This color is now unused and we can remove it.

adzhindzhi (approved):
Looks good to me! 🙌

src/views/splash/donate/donate-banner.scss
rosics-code: I think it would be best to keep this color incase any future use.

src/views/splash/donate/donate-banner.jsx
rosics-code: > nitpick: We probably don't need the `type="button"`.

Yeah, if its already being classified as a `<Button>`. You shouldn't need a `type="button"`.

rosics-code (changes requested):
This is a good PR just need some changes.

src/views/splash/donate/donate-banner.jsx
KManolov3: I think the idea is to defer navigation so that any asynchronous event handlers that may need to react to the dataLayer push (e.g. dispatch a request to GTM) do so. This was the way it was previously implemented in https://github.com/scratchfoundation/scratch-www/pull/10122/changes#diff-03d6bb8aab1085de7fcd928206bd9d20cfd8da696fcd63449e728ab0e541a609, so I am leaving it as-is to minimize changes.

src/views/membership/membership.jsx
KManolov3: We'll still be able to reach the deleted code via the git history

rosics-code:
@KManolov3 CI/CD is screaming at you.

PR #10254, opened 2026-06-04
chore: Sunset membership-label

rosics-code:
### Resolves:

[PR](https://github.com/scratchfoundation/scratch-www/pull/10226)

### Changes:

Removes `membership-label` as membership is being sunset and we may never use these again. If we do need these again, we can look at commit history.


rosics-code:
@adzhindzhi @KManolov3 What do you think

PR #10357, opened 2026-08-27
refactor: generate Fastly route config as VCL snippets

cwillisf:
### Resolves:

No single tracking issue. This is the substantive step of the Fastly config refactor. Adding a redirect currently means adding Fastly objects that consume a shared, exhausted pool of synthetic status codes, so new footer redirects can't be added on production. (The promise-native rewrite this was stacked on, scratchfoundation/scratch-www#10356, has merged.)

### Changes:

Replace the per-route Fastly objects (one condition and header per route, plus a response object per redirect, hundreds of objects in all) with a small set of VCL snippets generated from `src/routes.js` and overwritten by name on every deploy:

- `init` `app-routes-tables`: the redirect lookup table.
- `recv` `app-routes-recv`: the redirect-table lookups (plus the explicit `tip_bar` and `projects/embed` redirects) and the app-route rewrites to the static html shells.
- `error` `app-routes-error`: turns the reused status into a 301.

Every redirect now shares one reused internal status (700) plus a `table.lookup`, so a new redirect is a new table row instead of another burned synthetic status code.

The snippets deliberately leave backend selection alone. A recv snippet weaves in ahead of the `#FASTLY recv` macro's default-backend assignment, which would overwrite any `set req.backend` the snippet made, so the S3-vs-dynamic backend split, host handling, and response TTL stay in the existing `recv-condition` and `fetch-condition` custom-VCL includes, which run after the macro. The snippets only do what survives it: `req.url` rewrites and error-driven redirects. Those includes are unchanged and stay in place.

- `bin/configure-fastly.js` renders the snippets from `src/routes.js` and validates the generated VCL (`validateVersion`) before it can activate.
- `bin/lib/fastly-extended.js` is promise-native and gains snippet and list/delete helpers. `setSnippet` updates each snippet in place (preserving its id) through the update endpoint, falling back to create when the snippet does not exist yet.
- `bin/lib/fastly-errors.js` (new): turns fastly-js's plain-object rejections into a readable one-line message instead of `[object Object]`.
- `bin/cleanup-legacy-fastly.js` (new): a one-time, dry-run-by-default script that removes the old per-route objects the snippets replace. It only deletes objects whose names carry the generated prefixes (`routes/`, `rewrites/`, `redirects/`); every other condition, header, and response object is left untouched.
- Removed three now-unused route-name helpers from `bin/lib/fastly-config-methods.js`.

Suggested rollout, per service (staging, then production):

1. Deploy the snippets (`configure-fastly`). Safe alongside the old objects during the overlap, since both mechanisms redirect identically.
2. Run `cleanup-legacy-fastly.js` (dry-run, then `CLEANUP_APPLY=true`) to remove the per-route objects.

The `recv-condition` and `fetch-condition` includes stay in place; only the per-route objects are removed.

A follow-up will add path canonicalization (single-key table plus trailing-slash normalization) and the new footer redirects. Fully retiring `recv-condition`, by moving the backend split to a native backend request-condition, is a possible separate follow-up.

### Test Coverage:

- `test/unit/lib/routes-to-vcl.test.js` (new): redirect-key expansion (optional trailing slash, alternation, throw-on-capturing-pattern) and the shape of the three generated snippets, including that the recv snippet leaves backend and host handling to `recv-condition`.
- `test/unit/lib/fastly-extended.test.js`: the promise surface, covering `getWorkingVersion` (reuse an editable draft vs clone), `setSnippet` (update-in-place, create-on-404), `validateVersion`, and the cleanup list/delete helpers.
- `test/unit/lib/fastly-errors.test.js` (new): the error formatter.
- `test/unit-legacy/test_fastly_config_methods.js` still passes; eslint clean.
- Verified on staging: deployed the snippets and diffed the full routing surface against production. Redirects, app-route views, static assets, and dynamic endpoints all route identically, aside from pre-existing environment differences.


colbygk (approved):
This looks good to me. I'd also like to see the effect it has on the outcome VCL.

PR #10358, opened 2026-08-31
refactor: collapse the redirect table to one key per source

cwillisf:
### Resolves:

No single tracking issue. Follow-up to the Fastly config refactor (scratchfoundation/scratch-www#10357), which kept a behavior-preserving "fat" redirect table with a slashed and an unslashed key for every source. This collapses it to one key per source.

### Changes:

`bin/lib/routes-to-vcl.js` only.

- The generated recv snippet now keys the redirect lookup off a slash-normalized copy of the path (`regsub(req.url.path, "(.)/+$", "\1")`), so a single slash-free table key matches both `/foo` and `/foo/`.
- `redirectSourcePaths` emits one slash-free key per source instead of both a slashed and an unslashed key, roughly halving the table. What any request resolves to is unchanged.

The request URL itself is left untouched -- there is no trailing-slash redirect. The static routes already serve either slash form, and adding a trailing slash for the paths the dynamic backend owns is left to that backend, so there is nothing to canonicalize at this layer and no redirect that could fight the backend's own trailing-slash handling. (Reducing avoidable trailing-slash 301s is handled separately, at the link layer.)

### Test Coverage:

- `test/unit/lib/routes-to-vcl.js` unit suite updated (15 tests): `redirectSourcePaths` now asserts a single slash-free key (including alternation and optional-group expansion); the table test asserts the slashed duplicate is gone; a recv test asserts the lookup is keyed off the normalized `var.path` and that the request URL is never rewritten to strip a slash. Legacy `test/unit-legacy/test_fastly_config_methods.js` passes; eslint clean.
- Simulated the generated recv against representative paths: `/jobs`, `/jobs/`, and `/jobs//` all resolve to the destination in one hop; `/ideas`, `/ideas/`, `/projects/123`, `/projects/123/` all serve 200 with no redirect; backend paths (`/discuss`, `/discuss/`) and assets pass through untouched.


colbygk (approved):
LGTM

PR #10359, opened 2026-08-31
fix: slash-terminate backend links to avoid redirects

cwillisf:
### Resolves:

No single tracking issue. Part of the Fastly redirect work: cutting down avoidable 301s.

### Changes:

The dynamic backend canonicalizes its URLs with a trailing slash -- a slash-free backend URL gets a 301 to the slashed form. So an internal link to a backend endpoint that omits the trailing slash sends the user through an extra redirect on every click. Add the missing slash to the three live ones:

- teacher banner: `/educators/classes` -> `/educators/classes/`
- parents FAQ: `/discuss` -> `/discuss/`
- admin splash: `/scratch_admin/tickets` -> `/scratch_admin/tickets/`

(The static app routes serve either slash form, so their links are left alone.)

Add a unit test to keep this from drifting back: it walks the import graph from the routed views to find live code, classifies each static internal link against the route table, and fails if a link to a dynamic-backend endpoint is missing its trailing slash. Scoping to code reachable from a routed view keeps unrouted/archived views and their private components -- which predate the convention -- out of scope, without any hardcoded exclusions.

### Test Coverage:

- New `test/unit/internal-link-conventions.test.js`: reaches 376 files from the 50 routed views; passes with the three fixes in place. Verified it has teeth (covers `teacher-banner` and the `parents` view) and correctly excludes dead code (the unrouted `terms` view and the archived 2016 conference microsite are not scanned).
- eslint clean on the changed files (only pre-existing prop-type warnings elsewhere in the touched splash file).

### DIFF
--- a/src/components/teacher-banner/teacher-banner.jsx
+++ b/src/components/teacher-banner/teacher-banner.jsx
@@ -45,7 +45,7 @@ const TeacherBanner = ({
                 {sessionStatus === sessionActions.Status.FETCHED ? (
                     user ? [
                         <a
-                            href="/educators/classes"
+                            href="/educators/classes/"
                             key="classes-button"
                         >
                             <Button>
--- a/src/views/parents/parents.jsx
+++ b/src/views/parents/parents.jsx
@@ -155,7 +155,7 @@ const Landing = () => {
                                     </a>
                                 ),
                                 discussionForums: (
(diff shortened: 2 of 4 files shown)


colbygk (approved):
Mmm tests! LGTM

PR #10365, opened 2026-09-17
fix(deps): replace scratch-storage with @scratch/scratch-storage

cwillisf:
### Resolves:

- Resolves UEPR-691 "Update www to use scratch-storage from the monorepo"

### Changes:

Update www to use scratch-storage from the monorepo :grin:

Note that while it looks like we're jumping quite a few major versions, part of that is just due to the version alignment that happened in the monorepo: `scratch-storage@6.2.1` is equivalent to `@scratch/scratch-storage@14.1.0`. The parts of the `scratch-storage` API that `scratch-www` actually uses haven't changed from v14 to v15.

### Test Coverage:

Covered by existing tests


adzhindzhi (approved):
🙌

